PR #1, opened 2021-03-25
Switch from NET Core 3.1 to NET 5.0

LanceMcCarthy:
Many folks will present the argument that .NET 5's performance improvement will change the result. I have changed the TFM of the C# project to `net5.0` (which is the moniker for .NET 5.0).

On my 11th gen i9, 64GB RAM system and the results before and after are indistinguishable. However, this is anecdotal and should be seen by any visitor to the code.

Therefore, I propose that you should merge this anyways.

Other included improvements:
- Added a gitignore so that VS generated code is not accidentally committed and pushed
- Added a README so that viewers know they landed on the correct repo


LanceMcCarthy:
Rebased my branch to line up with current.

rbergen:
I'm closing this. Next to Dave's solution that he has chosen to use .NET Core 3.1 for, we now have two separate C# solutions in place in the drag-race branch that are based on .NET 5.0.

PR #2, opened 2021-03-25
PrimePY: using array slicing

jannekotka:
Changed to use array slicing and numpy. Increases perf
by 20x

Signed-off-by: Janne Kotka <lekonna@gmail.com>

### DIFF
--- a/PrimeSievePY/PrimePY.py
+++ b/PrimeSievePY/PrimePY.py
@@ -11,6 +11,7 @@
 from sys import stdout                      # So I can print without an automatic python newline
 from math import sqrt                       # Used by the sieve
 import timeit                               # For timing the durations
+import numpy as np
 
 class prime_sieve(object):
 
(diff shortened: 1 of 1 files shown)


jannekotka:
I have also one without numpy. keep the initialization as in the original and replace the for loop with 
this.rawbits[int(factor*3/2):this.sieveSize//2:factor] = [False] * len(this.rawbits[int(factor*3/2):this.sieveSize//2:factor])

rhbvkleef:
I *think* that we can consider slicing faithful, but I am open to discussion.

oscardssmith:
IMO, it might be a good idea to have separate python and python_numpy implementations.

rbergen:
I would agree that slicing is faithful. However, we can't merge this PR until it is opened on the drag-race branch.

rbergen:

rbergen:

PR #4, opened 2021-03-25
Make C# almost twice faster

EgorBo:
A few peephole optimizations:

`x % 2 == 0` to `x % 1 == 0` - there is a PR to .NET jit for it but it's still not landed yet.

`AggressiveInlining` shouldn't be needed there with PGO-based inlining but it's not enabled by default.

Before:
```
4650 passes
```
After:
```
8732 passes
```

### DIFF
--- a/PrimeSieveCS/PrimeCS.cs
+++ b/PrimeSieveCS/PrimeCS.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 
 namespace PrimeSieveCS
 {
@@ -48,21 +49,23 @@ namespace PrimeSieveCS
                 return false;
             }
 
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
             private bool GetBit(int index)
             {
-                if (index % 2 == 0)
+                if ((index & 1) == 0)
                     return false;
-                return bitArray[index / 2];
+                return bitArray[(int)((uint)index / 2)];
             }
 
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
             private void ClearBit(int index)
             {
-                if (index % 2 == 0)
+                if ((index & 1) == 0)
                 {
                     Console.WriteLine("You are setting even bits, which is sub-optimal");
                     return;
                 }
-                bitArray[index / 2] = false;      
+                bitArray[(int)((uint)index / 2)] = false;      
             }
 
             // primeSieve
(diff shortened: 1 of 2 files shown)


PrimeSieveCS/PrimeCS.cs
```diff
@@ -48,21 +49,23 @@ public bool validateResults()
...
             private void ClearBit(int index)
             {
-                if (index % 2 == 0)
+                if ((index & 1) == 0)
                 {
                     Console.WriteLine("You are setting even bits, which is sub-optimal");
                     return;
                 }
-                bitArray[index / 2] = false;      
+                bitArray[(int)((uint)index / 2)] = false;
```
eitanhs, line 68: @EgorBo out of curiosity, why not simply do bit shift in the first place? wouldn't that be faster than this two casts?

```suggestion
                bitArray[index >> 1] = false;      
```

Kinematics:
> A few peephole optimizations:
> 
> `x % 2 == 0` to `x % 1 == 0` - there is a PR to .NET jit for it but it's still not landed yet.

What is the above optimization supposed to do?  As far as I can tell, `x % 1` will always be 0, so it will never be useful in the algorithm.

Ahhh. Looked at the code, and it's not `x % 1`, it's `x & 1`.

For comparisons (on my machine):
* Current repo: 2700 loops in 5 seconds
* Fix `factor * 3` to `factor * factor`: 2800 loops in 5 seconds
* Clean rewrite of the entire codebase (so it doesn't look like a C++ copy/paste), but still using basically the same algorithm: 3080-3090 loops in 5 seconds
* Change from `/2` to `>>1`: 3110-3120 loops in 5 seconds (1% improvement)
* * Note: The real problem in the original was the unnecessary typecasting, which is about a 4% loss (~2960 loops).
* Get rid of the check that does a `Console.Writeline` in the `ClearBit` function (since it's only there for debug purposes): 3660 loops in 5 seconds (17% improvement)
* Apply the `&1` peephole optimization: 3620-3630 loops in 5 seconds. About a 1% loss.
* Once I got rid of the `Console.Writeline` code, `AggressiveInlining` had no effect.

Assuming you were running at 10 seconds instead of 5 (because of inconsistent project values), your 4650 would be 2325. My 2700 on the current repo is 16% above that. Scaling my 3660 down by that amount would put it at 3150. That's far below the 4316 (adjusted to 5 seconds) you claim.  However I can't imagine those small changes having that large of an effect.

The aggressive inlining you added probably had the same effect as my removal of the `Console.Writeline` code, since I saw a similar benefit when I tried that initially.

This needs more investigation as to where that extra speed is coming from.

rbergen:
I'm closing this because we've merged #6. `AggressiveInlining` has been applied there, and any performance gain from turning `x % 2` to `x & 1` has been improved upon by not even evaluating the expression in Release builds.

PR #5, opened 2021-03-25
Use stopwatch instead of DateTime.UtcNow and use string interpolation

tHeCh0s3n0n3:
There is a slight performance increase when using stopwatch.


GieltjE:
Might be more than we are guessing, we have been avoiding it like the plague for even simple applications for a long, long time.

rbergen:
Closing this. We have a [drag-race C# solution](https://github.com/PlummersSoftwareLLC/Primes/tree/drag-race/PrimeCSharp/solution_3) in place that incorporates Stopwatch, string interpolation and some other differences from Dave's solution.

PR #6, opened 2021-03-25
 Adding a faithful implementation of the C# solution

tannergooding:
On my own box, C++ measures:
```
Passes: 8665, Time: 5.000000, Avg: 0.000577, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

davepl;8665;5.000000;1
```

This builds on @EgorBo's PR in https://github.com/davepl/Primes/pull/4 and gets the C# version to:
```
Passes: 7883, Time: 5.0001565, Avg: 0.0006342961436001523, Limit: 1000000, Count: 78498, Valid: True

tannergooding;7883;5.0001565;1
```

The key points are to elide bounds checks and more closely mirror what C++ in terms of types and code used to make it a more fair comparison. C# inherently does things C++ doesn't, like bounds checks.
If the version to version comparisons were to be equally safe (with C++ using things like `std::array` and doing bounds checking before accesses, eliding where it is statically determined to be "safe"), then C++ wouldn't have such a "clear lead" 😄

### DIFF
--- a/PrimeCSharp/solution_3/Dockerfile
+++ b/PrimeCSharp/solution_3/Dockerfile
@@ -1,0 +1,16 @@
+FROM mcr.microsoft.com/dotnet/sdk:5.0-focal AS build
+WORKDIR /source
+
+# copy csproj and restore as distinct layers
+COPY *.csproj /source
+RUN dotnet restore
+
+# copy and publish app and libraries
+COPY *.cs /source
+RUN dotnet publish -c release -o /app --no-restore
+
+# final stage/image
+FROM mcr.microsoft.com/dotnet/runtime:5.0-focal
+WORKDIR /app
+COPY --from=build /app .
+ENTRYPOINT ["/app/PrimeSieveCS"]
--- a/PrimeCSharp/solution_3/PrimeCS.cs
+++ b/PrimeCSharp/solution_3/PrimeCS.cs
@@ -137,6 +137,9 @@ namespace PrimeSieveCS
                 if (showResults)
                     Console.WriteLine();
                 Console.WriteLine($"Passes: {passes}, Time: {duration}, Avg: {duration / passes}, Limit: {sieveSize}, Count: {countPrimes()}, Valid: {validateResults()}");
+
+                Console.WriteLine();
+                Console.WriteLine($"tannergooding;{passes};{duration};1", passes, duration);
             }
 
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
--- a/PrimeCSharp/solution_3/PrimeCS.cs
+++ b/PrimeCSharp/solution_3/PrimeCS.cs
@@ -1,0 +1,172 @@
+﻿// ---------------------------------------------------------------------------
+// PrimeCS.cs : Dave's Garage Prime Sieve in C#
+// ---------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace PrimeSieveCS
+{
+    class PrimeCS
+    {
+        private static readonly Dictionary<ulong, ulong> myDict = new Dictionary<ulong, ulong>
+        {
+            [10] = 4,                 // Historical data for validating our results - the number of primes
+            [100] = 25,               // to be found under some limit, such as 168 primes under 1000
+            [1000] = 168,
+            [10000] = 1229,
+            [100000] = 9592,
+            [1000000] = 78498,
+            [10000000] = 664579,
+            [100000000] = 5761455,
+            [1000000000] = 50847534,
+            [10000000000] = 455052511,
+        };
+
+        class prime_sieve
+        {
+            private readonly ulong sieveSize;
+            private readonly byte[] rawbits;
+
+            [MethodImpl(MethodImplOptions.AggressiveOptimization)]
+            public prime_sieve(ulong n)
+            {
+                sieveSize = n;
+                rawbits = GC.AllocateUninitializedArray<byte>((int)((n / 8) + 1), pinned: true);
+                rawbits.AsSpan().Fill(0xFF);
+            }
+
+            public ulong countPrimes()
+            {
+                var sieveSize = this.sieveSize;
+                var rawbits = this.rawbits;
+
+                // hoist null check
+                _ = getrawbits(rawbits, 0);
+
+                ulong count = (sieveSize >= 2) ? 1UL : 0UL;
+                for (ulong i = 3; i < sieveSize; i+=2)
+                    if (GetBit(rawbits, i))
+                        count++;
+                return count;
+            }
+
+            private bool validateResults()
+            {
+                return myDict.TryGetValue(sieveSize, out ulong sieveResult) && (sieveResult == countPrimes());
+            }
+
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            private static bool GetBit(byte[] rawbits, ulong index)
+            {
+                Debug.Assert((index % 2) != 0);
+                index /= 2;
+                return (getrawbits(rawbits, index / 8U) & (1u << (int)(index % 8))) != 0;
+            }
+
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            private static void ClearBit(byte[] rawbits, ulong index)
+            {
+                Debug.Assert((index % 2) != 0);
+                index /= 2;
+                getrawbits(rawbits, index / 8) &= (byte)~(1u << (int)(index % 8));
+            }
+
+            // primeSieve
+            // 
+            // Calculate the primes up to the specified limit
+
+            [MethodImpl(MethodImplOptions.AggressiveOptimization)]
+            public void runSieve()
+            {
+                var sieveSize = this.sieveSize;
+                var rawbits = this.rawbits;
+
+                // hoist null check
+                _ = getrawbits(rawbits, 0);
+
+                ulong factor = 3;
+                ulong q = (ulong)Math.Sqrt(sieveSize);
+
+                while (factor <= q)
+                {
+                    for (ulong num = factor; num < sieveSize; num += 2)
+                    {
+                        if (GetBit(rawbits, num))
+                        {
+                            factor = num;
+                            break;
+                        }
+                    }
+
+                    // If marking factor 3, you wouldn't mark 6 (it's a mult of 2) so start with the 3rd instance of this factor's multiple.
+                    // We can then step by factor * 2 because every second one is going to be even by definition
+
+                    for (ulong num = factor * factor; num < sieveSize; num += factor * 2)
+                        ClearBit(rawbits, num);
+
+                    factor += 2;
+                }
+            }
+
+            public void printResults(bool showResults, double duration, ulong passes)
+            {
+                var sieveSize = this.sieveSize;
+                var rawbits = this.rawbits;
+
+                // hoist null check
+                _ = getrawbits(rawbits, 0);
+
+                if (showResults)
+                    Console.Write("2, ");
+
+                ulong count = (sieveSize >= 2) ? 1UL : 0UL;
+                for (ulong num = 3; num <= sieveSize; num += 2)
+                {
+                    if (GetBit(rawbits, num))
+                    {
+                        if (showResults)
+                            Console.Write(num + ", ");
+                        count++;
+                    }
+                }
+
+                if (showResults)
+                    Console.WriteLine();
+                Console.WriteLine($"Passes: {passes}, Time: {duration}, Avg: {duration / passes}, Limit: {sieveSize}, Count: {countPrimes()}, Valid: {validateResults()}");
+            }
+
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            private static ref byte getrawbits(byte[] rawbits, ulong index)
+            {
+                return ref Unsafe.Add(ref MemoryMarshal.GetArrayDataReference(rawbits), (nint)index);
+            }
+        }
+
+        static void Main(string[] args)
+        {
+            const ulong SieveSize = 1000000;
+            const long MillisecondsPerSecond = 1000;
+            const long MicrosecondsPerSecond = 1000000;
+
+            ulong passes = 0;
+            prime_sieve sieve = null;
+
+            var stopwatch = Stopwatch.StartNew();
+
+            while (stopwatch.ElapsedMilliseconds < (5 * MillisecondsPerSecond))
+            {
+                sieve = new prime_sieve(SieveSize);
+                sieve.runSieve();
+                passes++;
+            }
+            stopwatch.Stop();
+
+            if (sieve != null)
+                sieve.printResults(false, (stopwatch.Elapsed.TotalSeconds * MicrosecondsPerSecond) / SieveSize, passes);
+        }
+    }
+}
--- a/PrimeCSharp/solution_3/PrimeSieveCS.csproj
+++ b/PrimeCSharp/solution_3/PrimeSieveCS.csproj
@@ -1,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net5.0</TargetFramework>
+  </PropertyGroup>
+
+</Project>
--- a/PrimeCSharp/solution_3/PrimeSieveCS.sln
+++ b/PrimeCSharp/solution_3/PrimeSieveCS.sln
@@ -1,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.31112.23
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "PrimeSieveCS", "PrimeSieveCS.csproj", "{02AF9EB8-1B7E-42E2-89A9-73674EF99E5C}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{02AF9EB8-1B7E-42E2-89A9-73674EF99E5C}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{02AF9EB8-1B7E-42E2-89A9-73674EF99E5C}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{02AF9EB8-1B7E-42E2-89A9-73674EF99E5C}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{02AF9EB8-1B7E-42E2-89A9-73674EF99E5C}.Release|Any CPU.Build.0 = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {8BF73573-411C-4401-B939-7841524DED02}
+	EndGlobalSection
+EndGlobal
--- a/PrimeCSharp/solution_3/README.md
+++ b/PrimeCSharp/solution_3/README.md
@@ -9,4 +9,8 @@
 
 ## Output
 
-Passes: 7877, Time: 5.0005804, Avg: 0.0006348331090516694, Limit: 1000000, Count: 78498, Valid: True
+```
+Passes: 7883, Time: 5.0001565, Avg: 0.0006342961436001523, Limit: 1000000, Count: 78498, Valid: True
+
+tannergooding;7883;5.0001565;1
+```
--- a/PrimeCSharp/solution_3/README.md
+++ b/PrimeCSharp/solution_3/README.md
@@ -1,0 +1,12 @@
+# C# solution by tannergooding
+
+![Category](https://img.shields.io/badge/Category-faithful-green)
+
+## Run instructions
+
+1. Get the latest version of .NET 5 or later: https://dotnet.microsoft.com/download
+2. From the solution root, run `dotnet run -c Release`
+
+## Output
+
+Passes: 7877, Time: 5.0005804, Avg: 0.0006348331090516694, Limit: 1000000, Count: 78498, Valid: True


tannergooding:
There are likely other wins possible, such as adding `SkipLocalsInit`, avoiding allocating the `Dictionary` (since its effectively a constant) using `module initializers`, caching the array instance into a local before the loop to avoid reading it from the class over and over, etc

I've not done these yet as they are a bit more esoteric and only bring the C# numbers up to about 13000-13500. There may also be some wins possible by ensuring the method is "hot" ("tier 1") or by using aggressive optimization.

PrimeSieveCS/PrimeCS.cs
```diff
@@ -13,56 +15,59 @@ class PrimeCS
...
-            private bool GetBit(int index)
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            private bool GetBit(uint index)
             {
-                if (index % 2 == 0)
+                if ((index & 1) == 0) // index % 2
                     return false;
-                return bitArray[index / 2];
+                index /= 2;
+                return (getrawbits(index / 8U) & (1u << (int)(index % 8))) != 0;
```
ptasev: Would it be faster to change `(index % 8)` to `(index & 7)`?

saucecontrol: Nope: https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIGYACMhgYQYG8aHunGBXAJYA7DAwCyACkEiGwgCYwEASg5ce64gHYGE+YoYB6BgA4AqioBkO8nwYAeOzuEYluoQoQMApCaVKA3GrcAL40wUA==

ptasev: Hmm, based on that, I guess this change is unnecessary too on line 55:
                if (index % 2 == 0)
                if ((index & 1) == 0) // index % 2

https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIGYACMhgYQYG8aHunGBXAJYA7DAwCyACkEiGwgCYwEASg5ce64gHYGE+YoYB6BgA4AqioBkO8nwYAeOzuEYluoQoQMApCaVKA3GrcAL5BvAzSomKkUs6y7ooqnNTqGtpuHoYm5gxWEjb2jm4uGfpWmn6BKTyh1dz0DMAQEAA24nSxMnrKYcmpPAIAZk4Jnj6kDAC8kwwADEph/fXag9gtuDBVSwyLPFoMGFB8m2G1GoxNreIonaLdC3Wqj+pDOqWeVuQq03MP22kMVbrE7PHi7ZYHI4g9S1YJAA===

tannergooding: Yes, it should actually be unnecessary with the switch to use `uint` (which matches what C++ uses: `unsigned int`)

davepl:
So the index & 1 made no difference on my machine.  Neither did .Net Core 5.

The (int)((uint)index / 2) was worth something though, going from 2660 to 3000.  That's the only one that seemed to help though!

tannergooding:
> So the index & 1 made no difference on my machine. Neither did .Net Core 5.

This may have to do with exact machine setup. I tested on an AMD Ryzen 5950X and validated the changes I made improved the disassembly and were measurable via AMD uProf.

Most of the changes can also be observed in isolation via SharpLab. For example the following compares the differences between the baseline and this PR for `countPrimes`: https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABABgAJiBGAOgGEIAbRmMDASwgDtcBuAWABQZSrQBKAVy4d8MehHwAHdiygBlGFABu7MDD5CR1GpOntZNAJLTNERRu279AwYYBM5RVHMwA+rnYYLRghAG8hckjPby1sDBhydmlyAKCYNXYALwSAXnJSFyjo9lj48mAAT3iAbQBdcihsAHdgdgxccjyuCWZCqK8SuISAITaAQShGivLxyewKlwio4gBmROTIKQwABW9ZXF9gPBgACgBKJcjwwSKipIxyTeS8gsvbgDNochP7xM783h/AA85AwAAt2LgaK0MBMpvItoD2ABqZEXG63THsd7fGFw+bVdi1dGY0lRJ4YVF9THEADsjwgiLeAF83m9VusHhTdj4DgATbHvc5va5k34U/6vDGYz5Qb6/diSpHkEGpYIZbJI1EkslRbHfADiMAwowwJxOEnuZ3YZx1utuFKpbyKdIZTOl5FZrg91QAssawRA+ZYlIwTv7wUGQ4pGAB5RQcbhQsYAcxTsFwAWC1kYSSSKbOtTeA1KCWAECY5CNJraFoVXD5MAQdvIotJ+vNSUbCHIADJyFQzp0XkOAPSj9bd8gAUnIbmdpNd72wjFwMGpty7TfIo7ybg3LvpJxTxsaLTauB+De3E4AHABVIf9k5UCQqkFXjBnK9T2e321DgAhC8G5esWMRDA0MA4pUZQnhgZ4wpelrJFuzYigukSurAOL3jw2DvHIYx8nyJw4eQ/r4NAFS+tgUC4GCK40NW+IVAAInE2BiNBmgwFwehkc0SFnGg3xcFaaFnKBQjMkAA===

This shows that the diff:
* avoids needing to allocate `40 bytes` of stack space
* elides the bounds check logic that `BitArray` was introducing (L0046) reducing the number of branches per loop iteration
* reduces the number of memory reads since `sieveSize` is hoisted outside the loop

Similar diffs exist for all the other methods as well.

Notably, perf of C++ and C# could be improved by using `uint32` (`uint` and `unsigned int`) rather than `uint8` (`byte` and `unsigned char`), as that is the more natural size for x86 to operate on.
Likewise, perf of both could be improved by using `popcnt` (`BitOperations.PopCount` in .NET) so many bits (32 or 64) could be checked at once, reducing the number of loop iterations.
I hadn't opted to do these yet however as I was simply trying to improve the C# variant without deviating from the underlying algorithm used by C++.

jetersen:
I have an AMD 3950X and got it 11220 passes after moving out the dictionary. With the dictionary inside it could only do 11108 passes roughly with this PR :)
FYI there is performance to get by recreating the sln with `dotnet new sln` and `dotnet sln add *.csproj` which gives you the option to target platform x64.
The sln seems to have been created with Visual Studio 2012 compatibility.

Gigabyte0x1337:
@tannergooding Did you enable optimize code in project settings for me it doubles performance. https://github.com/davepl/Primes/pull/21

tannergooding:
> @tannergooding Did you enable optimize code in project settings for me it doubles performance.

`Optimize` is `true` by default for `Release` configurations (because you typically want perf) and `false` by default for `Debug` configurations (because you typically want to debug). You should not need to explicitly set it in your csproj.

Instead, you simply change the Configuration drop down in VS, or do `dotnet run -c Release` (rather than `dotnet run` or `dotnet run -c Debug`)

alaatm:
@tannergooding  Hmm.. How come normal array access, ie. `rawbits[index]` is about 15% faster on my machine in comparison to using `GetArrayDataReference` \ `Unsafe.Add`

tannergooding:
> Hmm.. How come normal array access, ie. rawbits[index] is about 15% faster on my machine in comparison to using GetArrayDataReference \ Unsafe.Add

Provided you are testing on .NET 5 under the `Release` configuration, there shouldn't be.

You can compare the disassembly between the two using sharplab:

#### `rawbits[index]`
https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8B6YgAjkquptrvocaYYFgAoUsgBSgEt8YAYQDKAOjC4yIMgBFsANxgBySQHFsUbAHMY3PgLLDeMRWV4A7MoIDU19p2aOnzlzXbsAAgCYAjJ4AMZB4+ogBKAK7mGPwwooIQ+AAOvAA2MFDC6fK8YDC4ANwBQSERUTGiAJJR6RCJmVDZuQXubN5kifowAPq4xorsAN7sZCPtfPLYGLoWGGS9JjBGAF66ALxk/oVso2O8E1NkwACeUwDaALpkmgDuwLwYkuvm4SkpW8OjHgDMuwI9fTAABQzMjmACUHxGQ22O1G80UyzWoK2sNGNzuDzITxg10OJxgp0BlnIAA4wWRrGQfOcUairthbvdcKIAIK4YSJbDmQFg0QAMVSKUB/gQfL5YNpIwAvi0dqcALIwDAACwgABMKkkhYqVerNYkUgB5RLRCDmZksrRaWC4XqKKopCwWLRg86Q3b7XTACAQFJkVRKgBC90B4RBFjVMAQEJhUPdO14ADMyIDgeZIwgyAAyKnk1brfzkzgRqNkACkZC88bpHgA7GRE9gUrgYJLYSXM8R1l42zs6yn0UzTh2yKSAKqXHOAnzhMgAHjnKZmYLTGfLZDJYPJAEIC22ZWx3QqlaqNVrATrT/qjSbeGaLVabXaYA6neYXW7Yx7JroPCgrGkGjBhgobhumUYxjs0J0mYyapiOOY+HmBZFuQI4VlWX5QdWqLBAAnICABEACaECzhougthg0TvmQCyWBiuBoGQ1zKjkypmJIuDhMAcC1NE+BNoREo4bCda9qMB4wSOXaVhJIyDg8w7gZ2G6TusgLHFMYIAH7TrOC5LlEK7oRuW77rKnw/H+VyREYCw8u60F0iCjZgBg0BYmQXzyWYURkAAjl5aYYGC8qTMqojCAFUAgfCiy8CsImHlhoysakuiAm5HlQPOgWQXSzkwSMiaeSFoLhPgXnZdA+QVVVi7xYidXPPgtgFcVZBFZ1oxJimAYYMBqZhsZrVbqJMHdT1sI1blTyVb503ALA2AANaLXSUk9QeE2jJwFTJoJUCrc6DbYO5nlfMxRxkSxZEpGq5hKLMR2rWQSBLhgKhkNgZD4C8swQMmXjkrgEBzBgGizNc9wcSquhfFAap+bgkPmLkZBA2QKq8JIs3ff9KTRAasS7SMnAAOq6GAXLY8qMCWKjMCJHiZ0XblABUlaHDANPhC2dGKFARxzLzZrI2a0ySFoECnR5POCwzrORomTqmuYlmdaVuXla11XnTlZBcz59V5U1iWtqb1jrLNRuVh100ATAQEhqBo2VclZNs4b1tyaJW3SuwB5AA==

#### `Unsafe.Add(ref MemoryMarshal.GetArrayDataReference(rawbits), (nint)index)`
https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8B6YgAjkquptrvocaYYFgAoUsgBSgEt8YAYQDKAOjC4yIMgBFsANxgBySQHFsUbAHMY3PgLLDeMRWV4A7MoIDU19p2aOnzlzXbsAAgCYAjJ4AMZB4+ogBKAK7mGPwwooIQ+AAOvAA2MFDC6fK8YDC4ANwBQSERUTGiAJJR6RCJmVDZuQXubN5kifowAPq4xorsAN7sZCPtfPLYGLoWGGS9JjBGAF66ALxk/oVso2O8E1NkwACeUwDaALpkmgDuwLwYkuvm4SkpW8OjHgDMuwI9fTAABQzMjmACUHxGQ22O1G80UyzWoK2sNGNzuDzITxg10OJxgp0BlnIAA4wWRrGQfOcUairthbvdcKIAIK4YSJbDmQFg0QAMVSKUB/gQfL5YNpIwAvi0dqcALIwDAACwgABMKkkhYqVerNYkUgB5RLRCDmZksrRaWC4XqKKopCwWLRg86Q3b7XTACAQFJkVRKgBC90B4RBFjVMAQEJhUPdO14ADMyIDgeZIwgyAAyKnk1brfzkzgRqNkACkZC88bpHgA7GRE9gUrgYJLYSXM8R1l42zs6ymdBh0Uy0xmyKSAKrknOAnzhMgAHgXKZmYNHpYrZLB5IAhAW2zK2O6FUrVRqtYCdWf9UaTbwzRarTa7TAHU7zC63bGPZNdB4UFYaQaMGGChuG6ZRjGOzQnSZjJqmHbZrmWIFkW5CIRWVbftB1aosEACcgIAEQAJoQPOGi6C2GDRB+ZALJYGK4GgZDXMqOTKmYki4OEwBwLU0T4E2RESrhsJ1r2oyHrBiFdpWkkjIOw4POunZkGS2brICxxTGCAB+s7zkuK5RGuGHqduB6yp8Pz/lckRGAsPLujBdIgo2YAYNAWJkF8ClmFEZAAI4+WmGBgvKkzKqIwhBVAoHwosvArKJR7YaMbGpLogIeV5UCLsFUF0q5sEjIm3lhaC4T4D5uXQPkVU1cuiWIg1zz4LYRWlWQJXdaMSYpgGGAgamYame125ibBvV9bCdX5U81X+bNwCwNgADWy10tJfWHlNoycBUyZCVA63Og22Ced5XwsUc5GseRKRquYSizCd61kEgK4YCoZDYGQ+AvLMEDJl45K4BAcwYBoszXPcnEqroXxQGqAW4ND5i5GQINkCqvCSPNv2Ayk0QGrE+0jJwADquhgFyuPKjAljozAiR4hdV35QAVJWhwwHT4QtvRihQEccz82aqNmtMkhaBA51eXzwtM+zkaJk6prmNZ3XlfllXtbVl15WQPN+Y1BUtclrbm9Y6zzSblZdbNgEwMBIZgeN1WpRTHPG7b8liTt0raz+BywMmOm6EpDJMR7swdk7M2fPW4dkBO5rYImsQsmqaqAqnir4NARyRVAuDKk2ohDSyUCaEccjQ6EMBZ7AmNAspuBgixRKrgnVlsIeQA===

Lets look at `GetBit` for example, where before:
```asm
    L0000: sub rsp, 0x28
    L0004: test dl, 1
    L0007: jne short L0010
    L0009: xor eax, eax
    L000b: add rsp, 0x28
    L000f: ret
    L0010: shr edx, 1
    L0012: mov rax, [rcx+8]
    L0016: mov ecx, edx
    L0018: shr ecx, 3
    L001b: cmp ecx, [rax+8]
    L001e: jae short L0039
    L0020: movsxd rcx, ecx
    L0023: movzx eax, byte ptr [rax+rcx+0x10]
    L0028: and edx, 7
    L002b: bt eax, edx
    L002e: setb al
    L0031: movzx eax, al
    L0034: add rsp, 0x28
    L0038: ret
    L0039: call 0x00007ff801f0bc70
    L003e: int3
```

vs

after:
```asm
prime_sieve.GetBit(UInt32)
    L0000: test dl, 1
    L0003: jne short L0008
    L0005: xor eax, eax
    L0007: ret
    L0008: shr edx, 1
    L000a: mov rax, [rcx+8]
    L000e: cmp [rax], eax
    L0010: add rax, 0x10
    L0014: mov ecx, edx
    L0016: shr ecx, 3
    L0019: movzx eax, byte ptr [rax+rcx]
    L001d: and edx, 7
    L0020: bt eax, edx
    L0023: setb al
    L0026: movzx eax, al
    L0029: ret
```

You can see that a good portion of the codegen is identical, but the "after" is now smaller, no longer has to allocate stack space, and no longer does bounds checking.

That being said, in practice, `GetBit` is being inlined and so the only really interesting function to look at is `runSieve`.

before:
```asm
  L0000: push r15
    L0002: push r14
    L0004: push rdi
    L0005: push rsi
    L0006: push rbp
    L0007: push rbx
    L0008: sub rsp, 0x28
    L000c: vzeroupper
    L000f: mov rsi, rcx
    L0012: mov edi, 3
    L0017: vxorps xmm0, xmm0, xmm0
    L001b: vcvtsi2sd xmm0, xmm0, dword ptr [rsi+0x10]
    L0020: vsqrtsd xmm0, xmm0, xmm0
    L0024: vcvttsd2si ebx, xmm0
    L0028: cmp ebx, 3
    L002b: jle L00e5
    L0031: mov ecx, edi
    L0033: mov eax, [rsi+0x10]
    L0036: cmp ecx, eax
    L0038: jge short L007c
    L003a: mov edx, ecx
    L003c: test dl, 1
    L003f: je short L0072
    L0041: shr edx, 1
    L0043: mov r8, [rsi+8]
    L0047: mov r9d, edx
    L004a: shr r9d, 3
    L004e: cmp r9d, [r8+8]
    L0052: jae L00f2
    L0058: movsxd r9, r9d
    L005b: movzx r8d, byte ptr [r8+r9+0x10]
    L0061: and edx, 7
    L0064: bt r8d, edx
    L0068: setb dl
    L006b: movzx edx, dl
    L006e: test edx, edx
    L0070: jne short L007a
    L0072: inc ecx
    L0074: cmp ecx, eax
    L0076: jl short L003a
    L0078: jmp short L007c
    L007a: mov edi, ecx
    L007c: lea ebp, [rdi+rdi*2]
    L007f: cmp ebp, eax
    L0081: jge short L00da
    L0083: lea r14d, [rdi+rdi]
    L0087: mov r15d, ebp
    L008a: test r15b, 1
    L008e: jne short L00a4
    L0090: mov rcx, 0x2489d624ca0
    L009a: mov rcx, [rcx]
    L009d: call System.Console.WriteLine(System.String)
    L00a2: jmp short L00d2
    L00a4: shr r15d, 1
    L00a7: mov rcx, [rsi+8]
    L00ab: mov eax, r15d
    L00ae: shr eax, 3
    L00b1: cmp eax, [rcx+8]
    L00b4: jae short L00f2
    L00b6: movsxd rax, eax
    L00b9: lea rax, [rcx+rax+0x10]
    L00be: mov ecx, r15d
    L00c1: and ecx, 7
    L00c4: mov edx, 1
    L00c9: shl edx, cl
    L00cb: not edx
    L00cd: movzx edx, dl
    L00d0: and [rax], dl
    L00d2: add ebp, r14d
    L00d5: cmp ebp, [rsi+0x10]
    L00d8: jl short L0087
    L00da: add edi, 2
    L00dd: cmp edi, ebx
    L00df: jl L0031
    L00e5: add rsp, 0x28
    L00e9: pop rbx
    L00ea: pop rbp
    L00eb: pop rsi
    L00ec: pop rdi
    L00ed: pop r14
    L00ef: pop r15
    L00f1: ret
    L00f2: call 0x00007ff801f0bc70
    L00f7: int3
```

after:
```asm
    L0000: push r15
    L0002: push r14
    L0004: push rdi
    L0005: push rsi
    L0006: push rbp
    L0007: push rbx
    L0008: sub rsp, 0x28
    L000c: vzeroupper
    L000f: mov rsi, rcx
    L0012: mov edi, 3
    L0017: vxorps xmm0, xmm0, xmm0
    L001b: vcvtsi2sd xmm0, xmm0, dword ptr [rsi+0x10]
    L0020: vsqrtsd xmm0, xmm0, xmm0
    L0024: vcvttsd2si ebx, xmm0
    L0028: cmp ebx, 3
    L002b: jle L00dc
    L0031: mov ecx, edi
    L0033: mov eax, [rsi+0x10]
    L0036: cmp ecx, eax
    L0038: jge short L0075
    L003a: mov edx, ecx
    L003c: test dl, 1
    L003f: je short L006b
    L0041: shr edx, 1
    L0043: mov r8, [rsi+8]
    L0047: cmp [r8], r8d
    L004a: add r8, 0x10
    L004e: mov r9d, edx
    L0051: shr r9d, 3
    L0055: movzx r8d, byte ptr [r8+r9]
    L005a: and edx, 7
    L005d: bt r8d, edx
    L0061: setb dl
    L0064: movzx edx, dl
    L0067: test edx, edx
    L0069: jne short L0073
    L006b: inc ecx
    L006d: cmp ecx, eax
    L006f: jl short L003a
    L0071: jmp short L0075
    L0073: mov edi, ecx
    L0075: lea ebp, [rdi+rdi*2]
    L0078: cmp ebp, eax
    L007a: jge short L00d1
    L007c: lea r14d, [rdi+rdi]
    L0080: mov r15d, ebp
    L0083: test r15b, 1
    L0087: jne short L009d
    L0089: mov rcx, 0x2489d624ca0
    L0093: mov rcx, [rcx]
    L0096: call System.Console.WriteLine(System.String)
    L009b: jmp short L00c9
    L009d: shr r15d, 1
    L00a0: mov rax, [rsi+8]
    L00a4: cmp [rax], eax
    L00a6: add rax, 0x10
    L00aa: mov ecx, r15d
    L00ad: shr ecx, 3
    L00b0: mov edx, ecx
    L00b2: add rax, rdx
    L00b5: mov ecx, r15d
    L00b8: and ecx, 7
    L00bb: mov edx, 1
    L00c0: shl edx, cl
    L00c2: not edx
    L00c4: movzx edx, dl
    L00c7: and [rax], dl
    L00c9: add ebp, r14d
    L00cc: cmp ebp, [rsi+0x10]
    L00cf: jl short L0080
    L00d1: add edi, 2
    L00d4: cmp edi, ebx
    L00d6: jl L0031
    L00dc: add rsp, 0x28
    L00e0: pop rbx
    L00e1: pop rbp
    L00e2: pop rsi
    L00e3: pop rdi
    L00e4: pop r14
    L00e6: pop r15
    L00e8: ret
```

Where again, the overall code is largely the same, but now without bounds checking and so it should therefore be faster for `dotnet run -c Release`

tannergooding:
Here is a SxS diff of the assembly differences between `rawbits[index]` and `Unsafe.Add(ref MemoryMarshal.GetArrayDataReference(rawbits), (nint)index)` in `runSieve`, for easier comparison:

image

alaatm:
@tannergooding Yea, I can see the smaller code gen and that's why I'm baffled! I literally cloned your fork and ran it a couple of times, then changed `getrawbits(...)` to `rawbits[...]` then ran it a couple of times. All runs in release mode.

Using `Unsafe.Add(ref MemoryMarshal.GetArrayDataReference(rawbits), (nint)index)`
`Passes: 7493, Time: 10.0008581, Avg: 0.0013346934605631923, Limit: 1000000, Count: 78498, Valid: True`

Using `rawbits[index]`:
`Passes: 9443, Time: 10.0002272, Avg: 0.001059009552049137, Limit: 1000000, Count: 78498, Valid: True`

I used Egro's Disasm extension to compare the code gen diff for `runSieve`: https://www.diffchecker.com/76Al1Rxc

It's clear that the one with `Unsafe.Add(ref MemoryMarshal.GetArrayDataReference(rawbits), (nint)index)` is generating less code and no bound checks but I am not sure why the results aren't reflecting that!

PrimeSieveCS/PrimeCS.cs
```diff
@@ -101,37 +106,44 @@ public void printResults(bool showResults, double duration, int passes)
...
         static void Main(string[] args)
         {
-            var tStart = DateTime.UtcNow;
             var passes = 0;
             prime_sieve sieve = null;
 
-            while ((DateTime.UtcNow - tStart).TotalSeconds < 10)
+            var stopwatch = Stopwatch.StartNew();
+
+            while (stopwatch.ElapsedTicks < (10 * TimeSpan.TicksPerSecond))
```
mike-barber: Weirdly, this gives me about 100ms on my box - running under WSL2.
However, this works correctly: `while (stopwatch.ElapsedTicks < (10 * Stopwatch.Frequency))`
In my case at least, StopWatch.Frequency != TimeSpan.TicksPerSecond

tannergooding: Good catch. I forgot that `ElapsedTicks` isn't normalized to .NET ticks but stays as timer ticks. The other functions are normalized to timer ticks and I've fixed this accordingly.

tannergooding:
@alaatm, it may be something subtle with your particular hardware.

You could try to profile this in the VS Performance Profiler (Debug->Performance Profiler, then check CPU Usage). If you have an Intel based machine you can also use VTune (https://devblogs.microsoft.com/dotnet/net-core-performance-profiling-with-intel-vtune-amplifier-2018/ gives a basic overview of how it works) and with an AMD Machine you have uProf (https://developer.amd.com/amd-uprof/) which provides similar functionality, but doesn't work as nicely with .NET Core today.

This might tell you if there is some subtle thing, like a false dependency chain, that would cause the newer codegen to run the same or slower.

If there is something obvious, then I may be able to do something to help in the codegen.

tannergooding:
Latest changes bring codegen more inline with C++.

C#
```asm
    L0000: push r15
    L0002: push r14
    L0004: push r12
    L0006: push rdi
    L0007: push rsi
    L0008: push rbp
    L0009: push rbx
    L000a: sub rsp, 0x20
    L000e: vzeroupper
    L0011: mov esi, [rcx+0x10]
    L0014: mov rdi, [rcx+8]
    L0018: cmp [rdi], edi
    L001a: mov ebx, 3
    L001f: mov ecx, esi
    L0021: vxorps xmm0, xmm0, xmm0
    L0025: vcvtsi2sd xmm0, xmm0, rcx
    L002a: vsqrtsd xmm0, xmm0, xmm0
    L002e: vcvttsd2si rbp, xmm0
    L0033: cmp ebp, 3
    L0036: jbe L00d7
    L003c: mov ecx, ebx
    L003e: cmp ecx, esi
    L0040: jae short L0074
    L0042: mov eax, ecx
    L0044: test al, 1
    L0046: je short L006a
    L0048: shr eax, 1
    L004a: lea rdx, [rdi+0x10]
    L004e: mov r8d, eax
    L0051: shr r8d, 3
    L0055: movzx edx, byte ptr [rdx+r8]
    L005a: and eax, 7
    L005d: bt edx, eax
    L0060: setb al
    L0063: movzx eax, al
    L0066: test eax, eax
    L0068: jne short L0072
    L006a: inc ecx
    L006c: cmp ecx, esi
    L006e: jb short L0042
    L0070: jmp short L0074
    L0072: mov ebx, ecx
    L0074: lea r14d, [rbx+rbx*2]
    L0078: cmp r14d, esi
    L007b: jae short L00cc
    L007d: lea r15d, [rbx+rbx]
    L0081: mov r12d, r14d
    L0084: test r12b, 1
    L0088: jne short L009e
    L008a: mov rcx, 0x2489d6296b8
    L0094: mov rcx, [rcx]
    L0097: call System.Console.WriteLine(System.String)
    L009c: jmp short L00c4
    L009e: shr r12d, 1
    L00a1: lea rax, [rdi+0x10]
    L00a5: mov ecx, r12d
    L00a8: shr ecx, 3
    L00ab: mov edx, ecx
    L00ad: add rax, rdx
    L00b0: mov ecx, r12d
    L00b3: and ecx, 7
    L00b6: mov edx, 1
    L00bb: shl edx, cl
    L00bd: not edx
    L00bf: movzx edx, dl
    L00c2: and [rax], dl
    L00c4: add r14d, r15d
    L00c7: cmp r14d, esi
    L00ca: jb short L0081
    L00cc: add ebx, 2
    L00cf: cmp ebx, ebp
    L00d1: jb L003c
    L00d7: add rsp, 0x20
    L00db: pop rbx
    L00dc: pop rbp
    L00dd: pop rsi
    L00de: pop rdi
    L00df: pop r12
    L00e1: pop r14
    L00e3: pop r15
    L00e5: ret
```

as compared to C++:
```asm
        mov     QWORD PTR [rsp+32], rsi
        push    rdi
        push    r14
        push    r15
        sub     rsp, 32                             ; 00000020H
        mov     edi, DWORD PTR [rcx]
        mov     r14, rcx
        vxorps  xmm1, xmm1, xmm1
        vcvtsi2sd xmm1, xmm1, edi
        vxorpd  xmm0, xmm0, xmm0
        vucomisd xmm0, xmm1
        mov     esi, 3
        ja      SHORT $LN35@runSieve
        vsqrtsd xmm0, xmm1, xmm1
        jmp     SHORT $LN36@runSieve
$LN35@runSieve:
        vmovaps xmm0, xmm1
        call    sqrt
$LN36@runSieve:
        vcvttsd2si r15d, xmm0
        cmp     r15d, esi
        jle     $LN3@runSieve
        mov     QWORD PTR [rsp+64], rbx
        mov     r9d, edi
        mov     QWORD PTR [rsp+72], rbp
        mov     QWORD PTR [rsp+80], r12
        mov     r12d, 1
        npad    2
$LL2@runSieve:
        mov     r8d, esi
        cmp     esi, r9d
        jge     SHORT $LN34@runSieve
$LL6@runSieve:
        test    r8b, r12b
        je      SHORT $LN33@runSieve
        mov     ecx, r8d
        shr     ecx, 1
        mov     eax, ecx
        shr     rcx, 3
        and     eax, 7
        shlx    edx, r12d, eax
        mov     rax, QWORD PTR [r14+8]
        movzx   ecx, BYTE PTR [rcx+rax]
        test    dl, cl
        jne     SHORT $LN23@runSieve
$LN33@runSieve:
        inc     r8d
        cmp     r8d, r9d
        jl      SHORT $LL6@runSieve
        jmp     SHORT $LN34@runSieve
$LN23@runSieve:
        mov     esi, r8d
$LN34@runSieve:
        lea     ebx, DWORD PTR [rsi+rsi*2]
        mov     r9d, edi
        cmp     ebx, edi
        jge     SHORT $LN8@runSieve
        lea     ebp, DWORD PTR [rsi+rsi]
$LL9@runSieve:
        test    bl, r12b
        jne     SHORT $LN18@runSieve
        lea     rcx, OFFSET FLAT:`string'
        call    printf
        jmp     SHORT $LN7@runSieve
$LN18@runSieve:
        mov     rcx, QWORD PTR [r14+8]
        mov     r8d, ebx
        shr     r8d, 1
        mov     edx, r8d
        and     r8d, 7
        shr     rdx, 3
        movzx   eax, BYTE PTR [rdx+rcx]
        btr     eax, r8d
        mov     BYTE PTR [rdx+rcx], al
$LN7@runSieve:
        mov     edi, DWORD PTR [r14]
        add     ebx, ebp
        cmp     ebx, edi
        jl      SHORT $LL9@runSieve
        mov     r9d, edi
$LN8@runSieve:
        add     esi, 2
        cmp     esi, r15d
        jl      $LL2@runSieve
        mov     r12, QWORD PTR [rsp+80]
        mov     rbp, QWORD PTR [rsp+72]
        mov     rbx, QWORD PTR [rsp+64]
$LN3@runSieve:
        mov     rsi, QWORD PTR [rsp+88]
        add     rsp, 32                             ; 00000020H
        pop     r15
        pop     r14
        pop     rdi
        ret     0
```

tannergooding:
Switching to use a raw pointer like C++ does (which is safe since the array is allocated as pinned), would bring it to: 
`Passes: 14549, Time: 10.0000641, Avg: 0.0006873368685132999, Limit: 1000000, Count: 78498, Valid: True`

The main benefit here is being able to remove a secondary indirection required to get the reference to element 0. Manually hoisting the lookup to a `ref var` wasn't providing the same benefit. I didn't push either of these changes.

```asm
    L0000: push r15
    L0002: push r14
    L0004: push r12
    L0006: push rdi
    L0007: push rsi
    L0008: push rbp
    L0009: push rbx
    L000a: sub rsp, 0x20
    L000e: vzeroupper
    L0011: mov esi, [rcx+0x10]
    L0014: mov rdi, [rcx+8]
    L0018: mov ebx, 3
    L001d: mov ecx, esi
    L001f: vxorps xmm0, xmm0, xmm0
    L0023: vcvtsi2sd xmm0, xmm0, rcx
    L0028: vsqrtsd xmm0, xmm0, xmm0
    L002c: vcvttsd2si rbp, xmm0
    L0031: cmp ebp, 3
    L0034: jbe L00ca
    L003a: mov ecx, ebx
    L003c: cmp ecx, esi
    L003e: jae short L006b
    L0040: mov eax, ecx
    L0042: test al, 1
    L0044: je short L0061
    L0046: shr eax, 1
    L0048: mov edx, eax
    L004a: shr edx, 3
    L004d: movzx edx, byte ptr [rdx+rdi]
    L0051: and eax, 7
    L0054: bt edx, eax
    L0057: setb al
    L005a: movzx eax, al
    L005d: test eax, eax
    L005f: jne short L0069
    L0061: inc ecx
    L0063: cmp ecx, esi
    L0065: jb short L0040
    L0067: jmp short L006b
    L0069: mov ebx, ecx
    L006b: lea r14d, [rbx+rbx*2]
    L006f: cmp r14d, esi
    L0072: jae short L00bf
    L0074: lea r15d, [rbx+rbx]
    L0078: mov r12d, r14d
    L007b: test r12b, 1
    L007f: jne short L0095
    L0081: mov rcx, 0x2489d6296b8
    L008b: mov rcx, [rcx]
    L008e: call System.Console.WriteLine(System.String)
    L0093: jmp short L00b7
    L0095: shr r12d, 1
    L0098: mov ecx, r12d
    L009b: shr ecx, 3
    L009e: mov eax, ecx
    L00a0: add rax, rdi
    L00a3: mov ecx, r12d
    L00a6: and ecx, 7
    L00a9: mov edx, 1
    L00ae: shl edx, cl
    L00b0: not edx
    L00b2: movzx edx, dl
    L00b5: and [rax], dl
    L00b7: add r14d, r15d
    L00ba: cmp r14d, esi
    L00bd: jb short L0078
    L00bf: add ebx, 2
    L00c2: cmp ebx, ebp
    L00c4: jb L003a
    L00ca: add rsp, 0x20
    L00ce: pop rbx
    L00cf: pop rbp
    L00d0: pop rsi
    L00d1: pop rdi
    L00d2: pop r12
    L00d4: pop r14
    L00d6: pop r15
    L00d8: ret
```

tannergooding:
Given these, the two core parts of the logic are, cpp:
```asm
$LL2@runSieve:
        mov     r8d, esi
        cmp     esi, r9d
        jge     SHORT $LN34@runSieve
$LL6@runSieve:
        test    r8b, r12b
        je      SHORT $LN33@runSieve
        mov     ecx, r8d
        shr     ecx, 1
        mov     eax, ecx
        shr     rcx, 3
        and     eax, 7
        shlx    edx, r12d, eax
        mov     rax, QWORD PTR [r14+8]
        movzx   ecx, BYTE PTR [rcx+rax]
        test    dl, cl
        jne     SHORT $LN23@runSieve
```

vs C#:
```asm
    L003a: mov ecx, ebx
    L003c: cmp ecx, esi
    L003e: jae short L006b

    L0040: mov eax, ecx
    L0042: test al, 1
    L0044: je short L0061
    L0046: shr eax, 1
    L0048: mov edx, eax
    L004a: shr edx, 3
    L004d: movzx edx, byte ptr [rdx+rdi]
    L0051: and eax, 7
    L0054: bt edx, eax
    L0057: setb al
    L005a: movzx eax, al
    L005d: test eax, eax
    L005f: jne short L0069
```

---------------------
cpp:
```asm
$LN33@runSieve:
        inc     r8d
        cmp     r8d, r9d
        jl      SHORT $LL6@runSieve
        jmp     SHORT $LN34@runSieve
$LN23@runSieve:
        mov     esi, r8d
$LN34@runSieve:
        lea     ebx, DWORD PTR [rsi+rsi*2]
        mov     r9d, edi
        cmp     ebx, edi
        jge     SHORT $LN8@runSieve
        lea     ebp, DWORD PTR [rsi+rsi]
$LL9@runSieve:
        test    bl, r12b
        jne     SHORT $LN18@runSieve
        lea     rcx, OFFSET FLAT:`string'
        call    printf
        jmp     SHORT $LN7@runSieve
$LN18@runSieve:
        mov     rcx, QWORD PTR [r14+8]
        mov     r8d, ebx
        shr     r8d, 1
        mov     edx, r8d
        and     r8d, 7
        shr     rdx, 3
        movzx   eax, BYTE PTR [rdx+rcx]
        btr     eax, r8d
        mov     BYTE PTR [rdx+rcx], al
$LN7@runSieve:
        mov     edi, DWORD PTR [r14]
        add     ebx, ebp
        cmp     ebx, edi
        jl      SHORT $LL9@runSieve
        mov     r9d, edi
```

vs C#:
```asm
    L0061: inc ecx
    L0063: cmp ecx, esi
    L0065: jb short L0040
    L0067: jmp short L006b

    L0069: mov ebx, ecx

    L006b: lea r14d, [rbx+rbx*2]
    L006f: cmp r14d, esi
    L0072: jae short L00bf
    L0074: lea r15d, [rbx+rbx]

    L0078: mov r12d, r14d
    L007b: test r12b, 1
    L007f: jne short L0095
    L0081: mov rcx, 0x2489d6296b8
    L008b: mov rcx, [rcx]
    L008e: call System.Console.WriteLine(System.String)
    L0093: jmp short L00b7

    L0095: shr r12d, 1
    L0098: mov ecx, r12d
    L009b: shr ecx, 3
    L009e: mov eax, ecx
    L00a0: add rax, rdi
    L00a3: mov ecx, r12d
    L00a6: and ecx, 7
    L00a9: mov edx, 1
    L00ae: shl edx, cl
    L00b0: not edx
    L00b2: movzx edx, dl
    L00b5: and [rax], dl

    L00b7: add r14d, r15d
    L00ba: cmp r14d, esi
    L00bd: jb short L0078
```

The core is that in the L0095-L00B5 block, .NET is generating much worse codegen and is doing a bit of register shuffling rather than generating what is effectively `btr`

tannergooding:
Added some notes to https://github.com/dotnet/runtime/issues/27382#issuecomment-808771570, which tracks having a way to do things like `btr`, `bts`, and `btc` in .NET

rhbvkleef:
Please remove the gitignore in the root. If you really want a gitignore, place it lower in the tree.

Kinematics:
I did a full rewrite of the C# project, set up so that I could pass commandline parameters to select the algorithm to use (along with sieve size, length of run, etc).

I currently have four versions: The standard version that's basically copied from Dave's, using a bitarray (though with a few small fixes that make it about 35% faster than the current repo version); a version that uses a bool array; a version that inverts the boolean logic of the bool array so that I can skip re-initializing the array values to true (after the runtime already initialized them to false); and the rawbits version provided here. 

I also ran a version of the rawbits that used array access instead of `Unsafe.Add`, because of @alaatm's comment.  It's listed as `RawBitsSieve[]`.

Rather than providing loops, I hooked in BenchmarkDotNet to do the performance profiling.  This is what I'm seeing:

```
|            Method | SieveSize |       Mean |   Error |  StdDev | Ratio |
|------------------ |---------- |-----------:|--------:|--------:|------:|
|     StandardSieve |   1000000 | 1,369.2 us | 4.08 us | 3.61 us |  1.00 |
|         BoolSieve |   1000000 | 1,040.0 us | 6.23 us | 5.82 us |  0.76 |
| InvertedBoolSieve |   1000000 |   782.5 us | 3.85 us | 3.01 us |  0.57 |
|      RawBitsSieve |   1000000 |   960.0 us | 3.14 us | 2.62 us |  0.70 |
|    RawBitsSieve[] |   1000000 |   870.2 us | 2.81 us | 2.49 us |
```

NB: An unmodified version of the C++ program runs at about 600 us (@tannergooding's OP's value corresponds to 601 us).  A target of 750 us would mean that the C++ build would be 25% faster than the C# build.

This is on Windows 10-64, with an AMD 3700X processor.  Given that I'm seeing the same thing @alaatm is seeing, with the array access being faster than the `Unsafe.Add`, I'm wondering if @tannergooding is running on an Intel processor? It's one of several comments around here that suggest far greater speedups from certain changes than I'm seeing, and the processor is the only guess I have at the moment.

Also also: I have looked at changes like `index % 2` being changed to `index & 1`, and while the latter has several fewer instructions (identical surrounding code, but with the % taking 5 instructions compared to the &'s 1 instruction), the `index & 1` design still ended up about 1% slower than the %.  I can only guess that there's better branch prediction for the % than for the &, or similar runtime optimization issues.

333fred:
> I'm wondering if @tannergooding is running on an Intel processor?

If there is one thing I can tell you about tanner, it's that he's _definitely_ not running on an Intel processor.

tannergooding:
> I'm wondering if @tannergooding is running on an Intel processor?

As per a couple posts above, I tested on an AMD Ryzen 5950X.

> The index & 1 design still ended up about 1% slower than the %. I can only guess that there's better branch prediction for the % than for the &, or similar runtime optimization issues.

`index % 2` being converted to `index & 1` is implicitly done by the C++ compiler and the .NET JIT for unsigned inputs, but not for signed. Any noise here is likely caused by other differences in the JIT output, such as method or loop alignment. For additional context, you might be interested in reading https://devblogs.microsoft.com/dotnet/loop-alignment-in-net-6/. The summary however is that prior to .NET 6, when viewing results across multiple distinct runs or even running a benchmark 1st rather than second, the codegen would be identical but the alignment of the address at which the code was emitted would differ. This tends to cause downstream perf problems and in the worst case can make a given run look much worse in some cases since the code will be emitted in just such a way as to not play well with the decoder.

tannergooding:
I resolved the merge conflict.

> Please remove the gitignore in the root. If you really want a gitignore, place it lower in the tree.

Placing the `.gitignore` in the root is standard practice, just like `.gitattributes`, and is even where GitHub will place it for newly created repos.
In many scenarios, it is required in the root due to being where IDEs and other artifacts are generated by default.

mike-barber:
Hi @tannergooding I know you've put a fair bit of work into this and it's a great C# example. 

Rolf and Tudor put together a contributing guide defining things the branch to target (drag-race) and where to put the solution in the folder structure: https://github.com/davepl/Primes/blob/main/CONTRIBUTING.md

If it's helpful, PR #129 was to update my solution accordingly.

tannergooding:
Thanks, I'll take a look later today and will update the PR as appropriate

tannergooding:
This should be updated now to follow what's laid out in Contributing.md

rbergen:
It looks to me like the only real "requirement" that's missing is the line of output in drag-race format?

tannergooding:
> It looks to me like the only real "requirement" that's missing is the line of output in drag-race format?

I must have missed this step in the contributing doocs. Could you clarify what's expected here?

The readme currently lists `Passes: 7877, Time: 5.0005804, Avg: 0.0006348331090516694, Limit: 1000000, Count: 78498, Valid: True` under the output, which matches what CPP outputs as far as I can tell.

rbergen:
It does match wat CPP outputs, except CPP (in the drag-race branch) outputs one extra line of output, which is the one I was referring to. It's described in CONTRIBUTING.md under "[Rules](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md#rules)" as `<name>;<iterations>;<total_time>;<num_threads>`. The guidelines then describe the various `<placeholders>` in that expression. 

As I'm sure you've already guessed, the long and short of it is that this is the output line that the automatic benchmark suite looks for (using a regex), to retrieve the data it needs for the report it generates.

tannergooding:
Thanks for the clarification, I had missed those two lines.

I've updated accordingly and the output is now:
```
Passes: 7883, Time: 5.0001565, Avg: 0.0006342961436001523, Limit: 1000000, Count: 78498, Valid: True

tannergooding;7883;5.0001565;1
```

rbergen (approved):
LGTM

marghidanu (approved):

PR #7, opened 2021-03-25
Python: Use python's count function and normal time.

wgaylord:
The time module works exactly like C#'s time stuff, looks a lot simpler. 

Python has a built-in count function for arrays.

time.time() returns the current time in seconds as a float.

### DIFF
--- a/PrimeSievePY/PrimePY.py
+++ b/PrimeSievePY/PrimePY.py
@@ -7,10 +7,12 @@
 # call the runSieve method to do the calculation. printResults will dump the count to check validity.
 #
 # Updated 3/22/2021 for Dave's Garage episode comparing C++, C#, and Python
+#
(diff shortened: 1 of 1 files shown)


wgaylord:
On my system using count instead of summing managed to pull out about 10 more passes. (I have a  i7-6700K in my laptop.)

dasper:
I am getting about the same results on my machine running Python 3.8.5.

wgaylord:
Interesting, I am also using Python 3.8, I wonder what could have caused it to increase by 10 passes when I did this initially, as if it truly made it faster then it shouldn't matter who runs it.

rbergen:
I think that, from a perspective of sieve run performance, these changes are basically cosmetic. For that reason, I'm closing this.

Issue #3, opened 2021-03-25
Not an issue - Python with Numba, faster than C#


cristi-neagu:
Hello,

This is not really an issue. Wanted to share a version of the Python code using Numba. But because this is the first time i used Numba and i don't really know how to use jitclass, i had to pull the code apart so it's pretty ugly. As such, i didn't want to submit it as a pull request. Hopefully someone can make a prettier version.

On my machine this ran faster than the C# version, with 4855 iterations vs 4150.

```
from sys import stdout
from math import sqrt                       # Used by the sieve
import timeit                               # For timing the durations
import numpy as np
from numba import njit

@njit
def primeSieve(limit=1000000):
    rawbits = np.full(int((limit+1)/2), np.bool(True))

    factor = 3
    q = sqrt(limit)
    num = 0

    while factor < q:
        for num in range (factor, limit):
            if num % 2 == 0:
                continue

            if rawbits[int(num/2)]:
                factor = num
                break

        # If marking factor 3, you wouldn't mark 6 (it's a mult of 2) so start with the 3rd instance of this factor's multiple.
        # We can then step by factor * 2 because every second one is going to be even by definition

        for num in range (factor * 3, limit, factor * 2):
            if num % 2 == 0:
                continue
            rawbits[int(num/2)] = False

        factor += 2 # No need to check evens, so skip to next odd (factor = 3, 5, 7, 9...)
    return rawbits

def printResults(showResults, rawbits, duration, passes, sieveSize):
    if showResults: # Since we auto-filter evens, we have to special case the number 2 which is prime
        stdout.write("2, ")

    count = 1
    for num in range (3, sieveSize): # Count (and optionally dump) the primes that were found below the limit
        if num % 2 == 0:
            continue
        if rawbits[int(num/2)]:
            if showResults:
                stdout.write(str(num) +", ")
            count+=1

    countPrimes = sum(1 for b in rawbits if b)
    assert count == countPrimes
    stdout.write("\n")

    primeCounts = { 10 : 1,                 # Historical data for validating our results - the number of primes
                    100 : 25,               # to be found under some limit, such as 168 primes under 1000
                    1000 : 168,
                    10000 : 1229,
                    100000 : 9592,
                    1000000 : 78498,
                    10000000 : 664579,
                    100000000 : 5761455
                  }

    if sieveSize in primeCounts:      # the data, and (b) our count matches. Since it will return
        validateResults = (primeCounts[sieveSize] == countPrimes)
[...]

PR #9, opened 2021-03-25
Made PrimePY a LOT faster

rhbvkleef:
You probably dont want to merge this as it significantly changes the spirit of the piece of code, but I wanted to show that Python can be VERY in some cases.

### DIFF
--- a/PrimeSievePY/PrimePY.py
+++ b/PrimeSievePY/PrimePY.py
@@ -67,12 +67,13 @@ class PrimeSieve:
     def sieve(sieve_size):
         factor = 3
         q = sqrt(sieve_size)
-        rawbits = np.full(sieve_size, True, b1)
+        rawbits = np.full(sieve_size // 2, True, b1)
 
         while factor < q:
-            for i in range(factor, sieve_size, 2):
-                if rawbits[i]:
-                    factor = i
+            for num in range(factor, sieve_size, 2):
+                if rawbits[num // 2]:
+                    factor = num
+                    break
 
             for num in range(factor * 3, sieve_size, factor * 2):
                 rawbits[num // 2] = False
@@ -132,7 +133,7 @@ if __name__ == '__main__':
 # Using time.time as opposed to timeit: 65 passes
 # Adding a better counting method:      69 passes
 # Finding next prime with increment 2:  85 passes
-# More efficient division by two:         114 passes
+# More efficient division by two:       114 passes
 # Inline clear_bit:                     209 passes
 # Naive JIT:                            2020 passes
-# Proper JIT with numpy:                25939 passes
+# Proper JIT with numpy:                10301 passes
--- a/PrimeSievePY/PrimePY.py
+++ b/PrimeSievePY/PrimePY.py
@@ -8,128 +8,131 @@
 #
 # Updated 3/22/2021 for Dave's Garage episode comparing C++, C#, and Python
 
-from sys import stdout                      # So I can print without an automatic python newline
 from math import sqrt                       # Used by the sieve
-import timeit                               # For timing the durations
+import time                                 # For timing durations
+from typing import ClassVar, Dict     # For specifying types of variables
+import itertools                            # For chaining iterators
+from numba import njit, b1, int32
+import numpy as np
 
-class prime_sieve(object):
 
-    rawbits = None   # Storage for sieve - since we filter evens, just half as many bits
-    sieveSize = 0    # Upper limit, highest prime we'll consider
+class PrimeSieve:
+    rawbits: np.ndarray
+    """
+    Contents of the sieve.
+    We only store odd numbers, as they are the only interesting ones.
+    """
 
-    primeCounts = { 10 : 1,                 # Historical data for validating our results - the number of primes
-                    100 : 25,               # to be found under some limit, such as 168 primes under 1000
-                    1000 : 168,
-                    10000 : 1229,
-                    100000 : 9592,
-                    1000000 : 78498,
-                    10000000 : 664579,
-                    100000000 : 5761455
-                  }
+    sieve_size: int
+    """
+    Highest prime we'll consider
+    """
 
-    def __init__(this, limit):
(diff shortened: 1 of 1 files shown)


cristi-neagu:
This closes #3 by being almost twice as fast. Nice job!

cristi-neagu:
Btw, would have been nice to post the numbers you were getting with C# and C++ as well, for comparison. All machines are different, so we don't know what "25939 passes" actually means.

rhbvkleef:
This is tested on Arch Linux with kernel 5.11.6, 32GB DDR4 RAM, i7-9750H.

| Language | Iterations |
| ------------- | ------------  |
| C++ (GCC) | 11556 |
| C++ (Clang) | 11267 |
| C# (Mono) | 1521 |
| C# (DotNet Core) | 3317 |
| Python | 10345 |

### C++
- GCC 10.2.0
- `g++ -O3 PrimeCPP.cpp`
- Clang 11.1.0
- `clang++ -O3 PrimeCPP.cpp`

### C#
- Mono 6.12.0
- `csc -o+ -debug- PrimeCS.cs`
- DotNet Core 5.0.104
- `dotnet run --configuration=release`

### Python
- CPython 3.9.2
- `python PrimePY.py`

pepper-lees:
This is quite impressive. But here's a question. It's using multiple libraries instead of using native Python. Do those additional libraries use native code instead of pure python? If they are, it feels a bit like cheating, and perhaps there should be two versions - one with pure Python and built-in libraries, and the other which takes advantage of extra libraries for better performance.

rhbvkleef:
> If they are, it feels a bit like cheating, and perhaps there should be two versions

Both Numpy and Numba use native code (although none of Numpy´s native code is actually executed). Numba contains quite a significant amount of native code, and actually produces machine-code at runtime. Is it cheating? Maybe, but not because the libraries use C code. The whole runtime is written in C, so using it as a qualifier is either meaningless or incredibly ambiguous. I think the benchmark for that should be whether we use specific-purpose native code, and that is NOT the case.

I do, however, se a case for keeping two separate solutions, one using the pure STL (in which case we can expect performance that is on par with the dotnet performance) and one where we use Numba´s JIT (or another Python JIT for that matter).

pepper-lees:
Fair reasoning!
I'm still quite unfamiliar with python at the moment, so I apologize if my question seemed accusatory. I see now that the biggest improvement was by using a JIT compiler, correct? Does that mean that Python does not do JIT compiling by default?

cristi-neagu:
Well, think of Numba kinda like it used to be when people were writing C or C++ but the compilers weren't super optimized like today, and if they had something performance critical they would insert some assembly code to make it as fast as possible. It's sort of like that, except code written for Numba is much, much close to Python than assembly is to C. It's basically just Python with types.

So in this respect i wouldn't consider using Numba cheating. It's still Python. It's like choosing a different compiler for your C code cause it's a bit better optimized.

rhbvkleef:
> Fair reasoning!

Thanks! I was afraid it would come across as a bit rant-ey. I´m happy my point got across despite that.

> I apologize if my question seemed accusatory

Don´t worry about it! It didn´t. I just wanted to make sure that I explained my position properly, and wanted to explain why I see it differently.

> I see now that the biggest improvement was by using a JIT compiler, correct?

Yes, that´s where the largest speed gain is.

> Does that mean that Python does not do JIT compiling by default?

CPython doesn´t by default. There are interpreters (like Pypy and Jython) that do, but they are oftentimes slower than using Numba. They do provide a more "Pythonic" experience: Numba restricts the amount of Python we can write, so that it will be able to optimize it better.

knowlen:
> This is quite impressive. But here's a question. It's using multiple libraries instead of using native Python. Do those additional libraries use native code instead of pure python? If they are, it feels a bit like cheating, and perhaps there should be two versions - one with pure Python and built-in libraries, and the other which takes advantage of extra libraries for better performance.

Python built-ins are implemented in C. The source code for Python can be found [here](https://github.com/python/cpython). Many Numpy ops just link to the Python implementation (eg; [numpy.sum()](https://github.com/numpy/numpy/blob/v1.20.0/numpy/core/fromnumeric.py#L2111-L2248) is just sum()). Numpy still gets called through the interpreter though; which means everything has to happen in memory + you cannot optimize around multiple / overlapping functions. For example, 
```
import numpy as np
def f(n):
    return np.sum(np.arrange(n)

print(f(10000000))
```
will literally create a 100000000 dimensional array in memory without any consideration to how sum will use it. This is pretty expensive, but if you decorate with numba, 
```
import numpy as np
from numba import njit

@njit
def f(n):
    return np.sum(np.arrange(n)

print(f(10000000))
```
f() gets JIT compiled and no array generation occurs. 

 
The Numba JIT compilation is _potentially*_ cheating in the sense that Numba is doing some extra manual translation work in the frontend. It's _sort of_ comparable to clang (which is also just an LLVM frontend), but I think the translation from c/++ -> LLVM IR in clang is more deterministic / automatic. The Numba team literally re-implements builtins / Numpy functions in their own language and then maps the Python code "something they know how to compile" before translating the Numba IR to LLVM IR. So there is probably a side effect here where Numba implicitly re-writes poor Python implementations to be optimal before lowering to LLVM IR --in a way that is more rigid than clang. I could be wrong though. 


vs. 

  
Generally, the name of the game for compiling Python is to either translate Python -> C (which can then be compiled by clang/gcc) or translating Python -> LLVM IR (which can be compiled directly by LLVM backend). There's dozens of mature libraries / tools for JIT and AOT compilation, but the methodology is essentially bimodal. Usually the former for AOT and the latter for JIT. 

While it is possible make Python as fast as C languages in 2021 using JIT compilers, I concede that the comparison is somewhat beyond the scope of Dan's video. The C languages are designed to be compiled, and Python is designed to be interpreted. It would be equivalent to benchmark native Python vs. C++ using a 3rd party C interpreter.

rbergen:
I'm wondering how this relates to #2?

rbergen:

PR #10, opened 2021-03-25
Added .gitignore and optimized C++ code to use multiple bitArray objects.

maw136:


davepl:
It's a cool solution, but I'm trying not to make any implementation too custom!  Could you check this in as PrimesCPP-BitArray-Alternate.cpp and then we can keep it without altering the main code?

rhbvkleef:
It would also be nice if you don't reformat your code in this way. It makes the diff unnecessarily big. Also, please remove the gitignore in the root. If you really want a gitignore, place it lower in the hierarchy.

rhbvkleef:
Additionally, I think @davepl hints that this is not faithful anymore. I still need to come up with a reasonable criterion for that.

rbergen:
We will not be able to merge this. Aside from the fact that it's opened on "original" (formerly "main"), the target implementations have changed to an extent that the vast majority of the proposed changes have become moot.

PR #13, opened 2021-03-26
Use vector and unique_ptr.

mookid:
1. Using vector in prime_sieve allows to make the default destructor work out the resource management.
2. Using unique_ptr has similar upsides.

I also added cmath so that the code builds on g++ / clang++.


rbergen:
The changes proposed by this PR have been made moot by subsequent changes to the target solution.

PR #14, opened 2021-03-26
Fixed pylint errors to conform to Python naming conventions, etc.

mckoss:
The code is now more "Pythonic" in naming conventions.  I did the minimal pass of running pylint (and ignored
all the line length reports, but fixed almost all the rest).


rbergen:
I think these changes can be considered cosmetic. That being the case, I'm closing this.

PR #15, opened 2021-03-26
Optimized python 25 passes -> 4,848 passes

dougrizo:
Changes include:

1. Adding numpy library and using arrays instead of lists. This optimization results in about a 100% speed improvement.
2. The real improvement comes from numba, which builds off of numpy and jit compiles our primary worker function. This optimization results in about a 19,000% speed improvement.
3. Also refactored to a more function-based approach.
4. Added ability to pass arguments (with default values) via command line.


dougrizo:
Closing as #9 seems equally or more optimal - and closer to the original source

davepl:
Cool, I’ll check it out!  I only want to actually include things that are part of a standard python install, to be fair.  But I’m curious no matter what!

Cheers,
Dave

dougrizo:
Makes sense! Would love to see a sequel to this comparison with additional packages / frameworks included.

Btw - great videos! Keep them coming.

PR #16, opened 2021-03-26
python has bytearrays and bitwise operators

samamiller:
Also fixed number of primes under 10 :)

### DIFF
--- a/PrimeSievePY/PrimePY.py
+++ b/PrimeSievePY/PrimePY.py
@@ -1,24 +1,29 @@
+"""
+Python Prime Sieve
 
-# Python Prime Sieve
-#
-# MyFirstPython Program (tm) Dave Plummer 8/9/2018
-#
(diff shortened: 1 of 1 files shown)


rbergen:
@samamiller We now have a `bytearray`-based Python implementation in place in the drag-race branch ([PrimePython/solution_2](https://github.com/PlummersSoftwareLLC/Primes/tree/drag-race/PrimePython/solution_2)). You could open a PR on the drag-race branch to have your solution added as the third Python solution. However:

- Please consider if your solution is sufficiently different from the 2 existing drag-race Python solutions. We will do the same when reviewing and merging.
- Your solution needs to conform to the [contributing guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md). 

Concerning the number of primes under 10: there are 4, being 2, 3, 5 and 7.

rbergen:
Closing this due to alternative solutions being in place.

PR #17, opened 2021-03-26
Faster, uses less memory, and fixes a few bugs.

realXCV:
Some of the changes could probably be applied to the Python and C# versions too.

### DIFF
--- a/PrimeCPP/PrimeCPP.cpp
+++ b/PrimeCPP/PrimeCPP.cpp
@@ -3,148 +3,172 @@
 // ---------------------------------------------------------------------------
 
 #include <chrono>
-#include <ctime>
-#include <iostream>
-#include <bitset>
+#include <cstdlib>
 #include <map>
+
 
 class prime_sieve
 {
-  private:
+private:
 
-      int sieveSize = 0;
-      unsigned char * rawbits = nullptr;
-      const std::map<const int, const int> myDict = 
-      {
-            { 10 , 1 },                 // Historical data for validating our results - the number of primes
-            { 100 , 25 },               // to be found under some limit, such as 168 primes under 1000
-            { 1000 , 168 },
-            { 10000 , 1229 },
-            { 100000 , 9592 },
-            { 1000000 , 78498 },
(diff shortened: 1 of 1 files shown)


bplessis:
Hi,

FYI you should not re-indent a code in the same commit as any other changes, doesn't allow for code review.
Well in fact you shouldn't re-indent someone else code at all, expect when really necessary

davepl:
All code should be tab size 8, indent size 4.  Any code which isn’t should be reindented before being checked in.

Thanks!

realXCV:
> Hi,
> 
> FYI you should not re-indent a code in the same commit as any other changes, doesn't allow for code review.
> Well in fact you shouldn't re-indent someone else code at all, expect when really necessary

I think what you meant is: you should not let Visual Studio re-indent code.

Spaces (even useless ones at the end of a line) are back.

realXCV:
> Hi,
> 
> FYI you should not re-indent a code in the same commit as any other changes, doesn't allow for code review.
> Well in fact you shouldn't re-indent someone else code at all, expect when really necessary

I think what you meant is: you should not let Visual Studio re-indent code.

rhbvkleef:
I have tagged this as `stl-faithful` as it does use classes. One of the criteria I have for `stl-faithful` is that a new sieve class is constructed each time, and this does not do that. It would be great if that is changed.

rbergen:
Acknowledging fully that this PR was opened before the current contributing guidelines were in place, there is a couple of reasons why we can't merge this, in this form:

- It basically selects one of two implementations based on compiler-set `#define`s. The two implementations differ significantly.
- It doesn't comply with the [contributing guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md). Part of this is that the PR isn't opened on the drag-race branch.

If you still want to submit this solution, I think it would need to be a separate C++ solution that only implements the `USE_RAWBITS` approach. As already indicated, it would need to conform to the current contributing guidelines.

rbergen:
Closing this due to alternative C++ solutions being available.

PR #23, opened 2021-03-27
Golang implementation

dasper:
Yet another not-very-idiomatic port of the benchmark but in Golang. One large difference is just using a slice of bool values rather than bit manipulation for storage. This uses more memory but for the advantage of a simple implementation and comparable benchmarks on my AMD Ryzen 3900X. 

Here are the fastest of 5 runs.
- dotnet core 3.1
    - `Passes: 2084, Time: 10.0008253, Avg: 0.004798860508637237, Limit: 1000000, Count: 78498, Valid: True`
- Go 1.15.3
    - `Passes: 7268, Time: 10.0019661s, Avg: 1.376164ms, Limit: 1000000, Count: 78498, Valid: true`

A struct has been created for `primeBool` and the calls are methods of that struct to enable switching between bool and sieve implementations easily in the future.

## Quick start

Install Go at https://golang.org/doc/install

Navigate to the  `PrimeGo` directory, 
#### To run with a temporary binary:
- run: `go run prime.go`
#### To build a binary:
- build: `go build prime.go`


dasper:
I will further mention if you want just raw speed and unsafe execution, you can get about 25% - 30% more performance by disabling the garbage collector with `debug.SetGCPercent(-1)` however this is extremely frowned upon for production code and generally a bad solution for any concurrent language. Plus I feel this is a more honest representation of the language itself as the majority of developers will not be implementing their own memory management system with syscall.Mmap (nor would I recommend it to the majority of developers)

mike-barber:
Thanks for this -- been meaning to play around with Go for some time, and a program I already know is a good place to start! It's very readable too :)

One thing to be aware of regarding the choice to use a slice of bool values instead of flipping bits: it has a massive performance impact. To compare apples to apples, you'd probably need a similar bit-vector approach. To see what I mean, here are the results for the Rust and C# implementations using both approaches.

C#
```
bit vector    Passes: 5045, Time: 10.0015904, Avg: 0.0019824757978196233, Limit: 1000000, Count: 78498, Valid: True
byte vector   Passes: 10894, Time: 10.0008661, Avg: 0.0009180159812740958, Limit: 1000000, Count: 78498, Valid: True
```
C# does debug build by default; `dotnet build -c Release` will give you as fast one :)

Rust
```
bit vector     Passes: 21170, Time: 10.000431, Avg: 0.00047238692, Limit: 1000000, Count: 78498, Valid: true
byte vector    Passes: 33495, Time: 10.000135, Avg: 0.00029855606, Limit: 1000000, Count: 78498, Valid: true
```

dasper:
Thanks! Perhaps I should do a PR with a more apples to apples comparison but I noticed the performance was affable and I felt a clean and readable example for the language was better so it did not seem too abstract for someone to start learning it if they wanted to.

In the end I am content not "winning the benchmark" as long as the code is something I feel is deployable to production. I did have a version with an array hard coded into the application and utilized that so the memory usage did not fall to the heap stack or get slowed down by the garbage collector but it did not feel true to the exercise to help people learn a new language.

The structure of this should be simple enough to make a new struct and interface the public method calls. This way main() can us a flag/arg to switch between the two.
```
type primeBase struct {
	size int
	sqrt int
}

type primeBool struct {
	primeBase
	isPrime []bool
}

type primeByte struct {
	primeBase
	isPrime []byte
}
```

Course it probably will have to wait for the next weekend.

mike-barber:
I agree -- what you're saying makes a lot of sense! I definitely struggled with that decision with the Rust implementation -- manipulating bits directly makes it quite different from idiomatic Rust. 

Specializing for different storage as you're suggesting would make cover both cases nicely, but possibly make it more complicated for new people to understand. Although Go seems pretty readable, so it might work quite well :) I might try the same thing in Rust, but it'll definitely look quite intimidating to new users.

I'm guessing it'll get even more interesting when generics are available for Go? I've heard they're in the works, but haven't been following it too closely.

ciplogic:
What about GCC Go...?

rbergen:
@dasper If you would like to still submit your Golang solution, you'll have to close this PR and open a new one on the drag-race branch. We can't merge this one just on the basis that it's opened on original (formerly main).

rbergen:
Closing this due to another Go solution being in place.

PR #24, opened 2021-03-27
add a numpy version

ch0001:
On my computer 

- cpp-version : ~9000 passes in 10s
- numpy-version: ~4900 passes in 10s

"No,no,no ...  numpy is an external extension using C and FORTRAN. It's not really Python and shouldn't count" 

Well, yes, but one feature of Python is, that you can easily extend it  with C/C++. And every pythonista would use numpy for problems like that.


ch0001:
Sorry, just realized there was already a pull request for a numpy version.

PR #26, opened 2021-03-27
Close to original, but pythonic

ZettZet:
I`ve try to save main idea, but rewrite it in more pythonic way + pep8


rbergen:
I think these changes can be considered to be cosmetic. That being the case, I'm closing this.

PR #27, opened 2021-03-28
More pythonic code

alfest:
Made it more pythonic.


rbergen:
I think these changes can be considered to be cosmetic. That being the case, I'm closing this.

PR #28, opened 2021-03-28
Make C++ fast

Thalhammer:
So there was this comment in youtube:
"probably C++ calculated the result at compile time and printed the result"
And since all the input parameters are known at compile time (this is common about all languages so Imho its only a small bit of cheating) I went ahead and rewrote the C++ implementation a bit to take advantage of this fact.  So here we go:
```
g++-10 -O2 main.cpp -o main -std=c++2a -fconstexpr-ops-limit=100000000 && ./main
Passes: 4854305, Time: 10.000001, Avg: 0.000002, Limit: 1000000, Count: 78498, Valid: 1
```

Note: One could probably rewrite this in a way that keeps the array inside the primesieve class without making anything static, but I am bad at constexpr/consteval stuff, so this is how I managed to do it. If anyone knows how to do it right feel free to correct it.
Note2: This is more of a joke than a serious improvement, I also did some changes without bending the rules and reached about 20k (vs the 17k baseline on my cpu).

### DIFF
--- a/PrimeCPP/PrimeCPP.cpp
+++ b/PrimeCPP/PrimeCPP.cpp
@@ -7,144 +7,125 @@
 #include <iostream>
 #include <bitset>
 #include <map>
+#include <cstring>
+#include <cmath>
+#include <array>
 
+const std::map<const int, const int> myDict =
+    {
+        {10, 1},   // Historical data for validating our results - the number of primes
+        {100, 25}, // to be found under some limit, such as 168 primes under 1000
+        {1000, 168},
+        {10000, 1229},
+        {100000, 9592},
+        {1000000, 78498},
+        {10000000, 664579},
+        {100000000, 5761455}};
+
+template <size_t N>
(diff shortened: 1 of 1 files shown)


davepl:
Good stuff, but it turns out that vector<bool> is as fast or faster than the approach you and I took to manually set and clear bits, so it must be doing about the same thing, and it's much simpler.  That wasn't always true, but now compilers can specialize vector<bool> so that it doesn't waste memory!

Please test it against the current code and let me know if you can get any improvement over it now!

Thalhammer:
Vector wont work for this case, as vector is not constexpr. I dont think you actually read the source/description. This does not do any "real" work at runtime, all the primes are generated at compile time by the compiler.
I can surely try it against the current code but I highly doubt using vector<bool> (which should actually be slower than manually setting bits by definition) can come anywhere close to a 20000% speed improvement. ;)

PR #30, opened 2021-03-28
C++, improved performance

bterwijn:
Hello Dave,

Nice youtube video: https://www.youtube.com/watch?v=D3h62rgewZM

I made some suggestions that improve the style and performance of your C++ code that you might like. See the commit logs.

C++ has modernized a lot even though code written in the old style still compiles. The new guidelines now also say use algorithms:

  https://www.youtube.com/watch?v=bFSnXNIsK4A

instead of raw for loops. But your for loops don't iterate over each item so therefore we should use the new "ranges" from C++ 2020 to go this route, but that is too new for me:

  https://www.youtube.com/watch?v=d_E-VLyUnzc

For this small program raw for loops are easiest, but for a large project the higher level of abstraction make algorithms/range the better option although there is a learning curve involved. They don't make it easier, but the code gets more expressive (so the experts say).

Also, we now use 1 byte for each bool:
 
  sizeof(bool) == 1 byte

I tried to use 1 bit for each bool to improve caching, but that only slowed things down (see 1_bit_per_bool_branch).

Best regards,
Bas

### DIFF
--- a/PrimeCPP/PrimeCPP.cpp
+++ b/PrimeCPP/PrimeCPP.cpp
@@ -20,7 +20,6 @@ class prime_sieve
 {
   private:
 
-      long sieveSize = 0;
       vector<bool> Bits;
       const std::map<const long, const int> myDict = 
       {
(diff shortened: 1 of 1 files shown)


dasper:
Validation does not appear to pass on my machine.
`Passes: 28889, Time: 10.000000, Avg: 0.000346, Limit: 1000000, Count1: 500000, Count2: 500000, Valid: 0`

bterwijn:
Oops, I now fixed the bug and made some other modifications.

dasper:
passes on my machine now

bterwijn:
Even faster now.

realXCV:
It crashes if I set the size to 1000000000.

bterwijn:
Yes, it now uses memory on the stack instead of allocating it dynamically on the heap. That means you have to have enough stack memory reserved, increase your stack memory. On say Ubuntu use:
$ ulimit -s unlimited

hohounk:
Pretty sure not using stack was a deliberate decision to make the comparisons fair between languages.

bterwijn:
In C++ you have the option to choose either which is a language decision to help optimize some things. Blame Python for not having the option and only doing run-time memory allocation which has overhead. I'm not sure about C#.

hohounk:
When you need to reconfigure your OS to make it not crash you're not comparing the same things any more and it isn't a language feature you just toggle on.

bterwijn:
It depends how you look at it, but I would argue it is as language feature. If you prefer use compiler flags to adjust the stack size if you don't want to adjust your OS. Languages differ in many ways, if you only use the common demoninator, are you really comparing the languages?

Spongman:
err, 

| One potential optimization involves coalescing vector elements such that each element occupies a single bit instead of sizeof(bool) bytes.

https://en.cppreference.com/w/cpp/container/vector_bool

clang, gcc, & msvc all ship with `std::vector<bool>` implementation that use this optimization.

rbergen:
I'm closing this. Dave has asked us not to make changes that change the performance of his implementations. It is still possible to open a new PR (on the drag-race branch and conforming to the contributing guidelines) to add a new C++ solution based on this PR's premises.

Do note that the number of bits per prime (candidate) has now been marked as a relevant characteristic of implementations. Put simply, for an implementation to be considered a "fair" contestant, it needs to use 1 bit to store prime candidate flags.

PR #31, opened 2021-03-29
Adds Dockerfile and docker-compose to the projects

ProFL:
Running all projects in docker containers makes the benchmarks easier to run on any platform/PC :)


rbergen:
Closing this. We have adopted Docker as the means to run the benchmarks, across the board.

PR #35, opened 2021-03-29
Huge Improvement on Python

callowaysutton:
435,000% improvement on #15 's 19,000% improvement on the original version by using the Python 3 built-in LRU Caching decorator. Can be used in Python 2.7 by installing the functools library as well.

This should be competitive with even the C++ version :D


aberent (commented):
Did you intend to replace PrimePY.py with PrimeCPP.cpp? That appears to be what this pull request is doing.

callowaysutton:
> Did you intend to replace PrimePY.py with PrimeCPP.cpp? That appears to be what this pull request is doing.

Yeah I have no idea how that happened...

rbergen:
Closing this as an effective duplicate of #9 (and possibly #2).

PR #36, opened 2021-03-30
Python speedups

aberent:
Using a bit array rather than a list is the major speedup here. As a result this now runs faster on my Linux machine when using PyPy3 than the c++ version using g++ -O2.

### DIFF
--- a/PrimeSievePY/PrimePY.py
+++ b/PrimeSievePY/PrimePY.py
@@ -11,6 +11,7 @@
 from sys import stdout                      # So I can print without an automatic python newline
 from math import sqrt                       # Used by the sieve
 import timeit                               # For timing the durations
+from array import array
 
 class prime_sieve(object):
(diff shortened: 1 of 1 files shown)


ch0001:
Your current code doesn't validate. I think the bit array is not correct initialized

       39 -      this.rawbits = array('I', [255] * ((this.sieveSize+1)//DIVISOR+1))
       39 +      this.rawbits = array('I', [2**ITEMSIZE-1] * ((this.sieveSize+1)//DIVISOR+1))

With pypy3 I have now ~5300 passes (vs. ~9000 for the cpp version).  Thx, for telling me about pypy :-)

aberent:
Thanks, now fixed, and my results are now similar to yours (Python slower than C++ but C++ is less than twice the speed of Python).

rbergen:

PR #37, opened 2021-03-30
pedantic BS, not with a mind to optimizations

mgdotdev:
Thought it would be fun to play around with this, wanted to first push minor touchups. Also idk if this would be against the spirit of the comparison, but I think most of this could be done in NumPy (which would provide you a more performant experience w/o much more difficulty) and if we really had the need for speed I would push the computation down into a static form using something like Cython or even in C via a C extension. Interested in seeing any of these approaches? Would be happy to take a swing at it.


mgdotdev:
It looks like a bunch of people have already done the NumPy approach, so no need to repeat there :P

PR #38, opened 2021-03-31
Native CPython implementation, 1737 passes / 5 sec @ bound = 1M

mgdotdev:
Given all of the focus around increasing Python's execution speed using 3rd party libraries, I thought I'd take a different approach and use the C/Python API to enhance the execution performance. The results are provided in the README.md. I'm providing the source code for the extension instead of a pre-built binary, so python will need access to a compiler for install.


rbergen:
We can't merge this PR in its current form, as it doesn't conform to the contributing guidelines. If a PR is opened on the drag-race branch that does conform, we're happy to consider it.

mgdotdev:
> We can't merge this PR in its current form, as it doesn't conform to the contributing guidelines. If a PR is opened on the drag-race branch that does conform, we're happy to consider it.

well that's understandable considering that CONTRIBUTING.md didn't exist at the date of this PR :P I'll take a look at it though

rbergen:
> well that's understandable considering that CONTRIBUTING.md didn't exist at the date of this PR :P 

Obviously no argument there. Then again, it doesn't change where we stand right now :)

> I'll take a look at it though

That's highly appreciated. If any questions arise while doing so, don't hesitate to let us know.

PR #39, opened 2021-03-31
Some tiny changes that result in a significant speedup

Omnifarious:
First, I make some tiny changes to your version to clean up some things that really stuck out to me.

Then I write my own version with an eye towards making it perform really well.

My own version takes less than half the time of the original, which is what I claimed I could do in my YouTube comment.


rbergen:
We already have a number of earlier PRs that move to change to integer division. Concerning your own implementation, you can open a PR for that on the drag-race branch, that conforms to the contributing guidelines.

rbergen:

PR #40, opened 2021-04-01
Pure Python, but 65 times faster

ssovest:
This is an optimized implementation in pure Python (no third-party modules, no C extensions). Runs about 65 times faster than the original on my Win64 machine (130 times faster on 100_000 and 1_000_000 limits).

And, as the main reason of tinkering with other people's code and reading random pull-requests is to learn someting new, here's the details:
1. GetBit and ClearBit were removed, their work moved to the runSieve method. The reason is how Python calls user-defined methods: being a dynamic language, it has to actually [search](https://docs.python.org/3/reference/datamodel.html#object.__getattribute__) instance attribute by its name and then [transform](https://docs.python.org/3/reference/datamodel.html#index-35) the retrieved function to a bound method.
2. The loop in which bits were being cleared was replaced with extended slice assignment. In situations where you need to assign some value to every n-th element of a sequence this way is considerably faster.
3. ~Doubled the size of the sequence containing bits~ got rid of index divisions. Python's numbers can be arbitrary large, but at a cost: the minimum space in memory that a number takes is 24 bytes (not bits!) and many operations on numbers are expensive, including bitwise operations.
4. List of bools was replaced with a [bytearray](https://docs.python.org/3/library/stdtypes.html#bytearray). The reason is CPython list implementation. Under the hood it's a dynamic C array, but as list elements may differ in size, it can't hold values themselves and holds pointers to them, which consumes additional memory and slows down iteration. Bytearray, on the other hand, just holds bytes and is more appropriate for the purpose of this program.

Also:
 - "Pythonized" the code
 - Fixed a couple of bugs
 - Added command line arguments support for setting different limits

Interestingly, this implementation works faster on CPython than on pypy, at least on my machine.

### DIFF
--- a/PrimeSievePY/PrimePY.py
+++ b/PrimeSievePY/PrimePY.py
@@ -68,9 +68,11 @@ class PrimeSieve:
 
         if self._size > 1:
             yield 2  # Since we auto-filter evens, we have to special case the number 2 which is prime
-        for num, is_prime in enumerate(self._bits):
-            if is_prime and num:
+        if self._size > 2:
+            num = 1
+            while num > 0:
                 yield num * 2 + 1
+                num = self._bits.find(1, num + 1)
 
     def print_results(self, show_results, duration, passes):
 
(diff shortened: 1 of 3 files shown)


wentz89:
This Code Looks pretty good!
It would be really cool to put the main loop as it is in a cython-file then compile it and see how much Performance you can gain.

pepper-lees:
Dang, this is VERY impressive.

PR #43, opened 2021-04-02
Add a table of scores and changed CPP duration

gfelisberto:


rbergen:
I'm closing this. We're using 5 seconds as the "standard" run time and only grandfathering some very early solutions that run for 10. Adding the table with results doesn't improve the readability of the README; we'll reporting on results to the benchmark tool :)

PR #44, opened 2021-04-04
Multithreaded C Primes Implementation

jsrobson10:
This was written completely in C and uses arrays to find a range of primes. It uses multi-threading so it will utilize more cores to calculate more prime numbers. It also has error checking to check if the results are valid from pre calculated values.


davepl:
To be clear, does this produce each batch of 1000000 any faster or does it do more in parallel?  ie: is each run faster in seconds individually than before?

rbergen:
Closing due to lack of response from contributor.

PR #45, opened 2021-04-04
add implementation using C++20 std::bitset instead of std::vector<bool>

wolframw:
C++20 introduces `std::bitset`, which is a bitfield that does not pretend to be a vector. Other differences from `std::vector<bool>` include that its data resides on the stack and that its size must be given at compile time.

The results are interesting: On my Linux machine (gcc 10.2.0, Intel Core i5-8265U), using the bitset is slightly but consistently and measurably faster. On my Windows machine on the other hand (cl 19.28.29912, Visual Studio Community 2019 16.9.1, AMD Ryzen 5 3600), the vector seriously outpaces the bitset. Here are some stats (10 runs each):
```
Linux std::vector<bool>: min=6525 max=6638 avg=6603.8
Linux std::bitset:       min=6650 max=6745 avg=6694.9

Windows std::vector<bool>: min=8275 max=8303 avg=8291.4
Windows std::bitset:       min=7544 max=7577 avg=7561.7
```

Unfortunately, I don't have access to a Mac at the moment.

### DIFF
--- a/PrimeCPP20-bitset/PrimeCPP20-bitset.cpp
+++ b/PrimeCPP20-bitset/PrimeCPP20-bitset.cpp
@@ -1,0 +1,136 @@
+// ---------------------------------------------------------------------------
+// PrimeCPP20-bitset.cpp : Dave's Garage Prime Sieve in C++ (using C++20 std::bitset)
+// ---------------------------------------------------------------------------
+
+#include <chrono>
+#include <ctime>
+#include <iostream>
+#include <bitset>
+#include <map>
+#include <cstring>
+#include <cmath>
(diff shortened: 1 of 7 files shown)


davepl:
My only concern with bitset as I remember it is that you need to statically declare the size, which I'd like to avoid, even though it works!  All of the other languages keep it dynamic (the upper limit) and I'd like to remain that way, but it's cool to have this optional impl!

Since the change is available here, if I close the topic, does it keep the code available forever or do I need to create a branch that contains it?

micwoj92:
The code will be available "forever", until you delete this repo.

PR #46, opened 2021-04-04
Add a PHP prime sieve

DennisdeBest:
definitely not a language for this sort of stuff but why not for the fun of it ;)

Edit : After a few changes it seems to be going quite a bit faster than the python version in this repo

### DIFF
--- a/PrimePHP/PrimePHP.php
+++ b/PrimePHP/PrimePHP.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /**
  * PHP Prime Sieve
  * Class PrimeSieve
@@ -27,7 +29,7 @@ class PrimeSieve
     )
     {
         $this->sieveSize = $sieveSize;
-        $rawbitSize = (int)($this->sieveSize + 1) / 2;
+        $rawbitSize = (int)(($this->sieveSize + 1) / 2);
         $this->rawbits = array_fill(0, $rawbitSize, true);
     }
 
@@ -118,7 +120,7 @@ printf(
 
 function getTimeDiffInMs(float $tStart): int
 {
-    return (microtime(true) - $tStart) * 1000;
+    return (int)(microtime(true) - $tStart) * 1000;
 }
 
 function validateResult($sieveSize): ?int
--- a/PrimePHP/PrimePHP.php
+++ b/PrimePHP/PrimePHP.php
@@ -49,16 +49,15 @@ class PrimeSieve
         $q = sqrt($this->sieveSize);
 
         while ($factor < $q) {
-            foreach (range($factor, $this->sieveSize) as $num) {
-                if ($this->getBit($num)) {
-                    $factor = $num;
+            for ($i = $factor; $i <= $this->sieveSize; $i++) {
+                if ($this->getBit($i)) {
+                    $factor = $i;
                     break;
                 }
             }
 
-
             for ($i = $factor * 3; $i <= $this->sieveSize; $i += $factor * 2) {
-                $this->clearBit($num);
+                $this->clearBit($i);
             }
 
             $factor += 2;
--- a/PrimePHP/PrimePHP.php
+++ b/PrimePHP/PrimePHP.php
@@ -85,8 +85,9 @@ $passes = 0;                            //Init passes
 $sieveSize = 1000000;                   //Set sieve size
 $printResults = false;                  //Print the prime numbers that are found
 $rawbitCount = null;                    //Init a rawbitCount to validate the result
+$runTime = 10;                          //The amount of seconds the script should be running for
 
-while (getTimeDiffInMs($tStart) < 10000) {
+while (getTimeDiffInMs($tStart) < $runTime * 1000) {
     $sieve = new PrimeSieve($sieveSize);
     $sieve->runSieve();
     $rawbitCount = $sieve->getRawbitCount();
--- a/PrimePHP/PrimePHP.php
+++ b/PrimePHP/PrimePHP.php
@@ -24,7 +24,8 @@ class PrimeSieve
         private int $sieveSize = 1000000,
     )
     {
-        $this->rawbits = array_fill(0, (int)($this->sieveSize + 1) / 2, true);
+        $rawbitCount = (int)($this->sieveSize + 1) / 2;
+        $this->rawbits = array_fill(0, $rawbitCount, true);
     }
 
     private function validateResult(): ?int
@@ -103,11 +104,16 @@ class PrimeSieve
 $tStart = hrtime(true);
 $passes = 0;
 
-while ((hrtime(true) - $tStart) / 1e+6 < 10000) {
+while (getTimeDiffInMs($tStart) < 10000) {
     $sieve = new PrimeSieve(1000000);
     $sieve->runSieve();
-    $passes += 1;
+    $passes++;
 
-    $sieve->printResults((hrtime(true) - $tStart) / 1e+6, $passes, false);
+    $sieve->printResults(getTimeDiffInMs($tStart), $passes, false);
     echo "\n";
+}
+
+function getTimeDiffInMs(int $tStart): int
+{
+    return (hrtime(true) - $tStart) / 1e+6;
 }
--- a/PrimePHP/PrimePHP.php
+++ b/PrimePHP/PrimePHP.php
@@ -9,7 +9,7 @@ class PrimeSieve
 {
     private array $rawbits;
 
-    private array $primeCounts = [
+    public static array $primeCounts = [
         10 => 1,
         100 => 25,
         1000 => 168,
@@ -24,13 +24,8 @@ class PrimeSieve
         private int $sieveSize = 1000000,
     )
     {
-        $rawbitCount = (int)($this->sieveSize + 1) / 2;
-        $this->rawbits = array_fill(0, $rawbitCount, true);
-    }
-
-    private function validateResult(): ?int
-    {
-        return $this->primeCounts[$this->sieveSize];
+        $rawbitSize = (int)($this->sieveSize + 1) / 2;
+        $this->rawbits = array_fill(0, $rawbitSize, true);
     }
 
     private function getBit(int $index): bool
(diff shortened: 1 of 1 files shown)


PrimePHP/PrimePHP.php
```diff
@@ -0,0 +1,119 @@
...
+
+$tStart = hrtime(true);
+$passes = 0;
+
+while (getTimeDiffInMs($tStart) < 10000) {
+    $sieve = new PrimeSieve(1000000);
+    $sieve->runSieve();
+    $passes++;
+
+    $sieve->printResults(getTimeDiffInMs($tStart), $passes, false);
```
HugoSantiagoBecerraAdan: Shouldn't be this call outside of the loop? You're calling it in every single iteration but it's only needed after the loop ends.

PrimePHP/PrimePHP.php
```diff
@@ -0,0 +1,119 @@
...
+        foreach (range(3, $this->sieveSize) as $num) {
+            if ($this->getBit($num)) {
+                if ($showResults) {
+                    echo $num . ", ";
+                }
+                $count++;
+            }
+        }
+
+        echo sprintf("Passes: %d, Time: %dms, Avg: %dms, Limit: %d, Count: %d, Valid: %s",
```
HugoSantiagoBecerraAdan: You can use `printf` directly instead of `echo` and `sprintf`:
```php
printf("Passes: %d, Time: %lf, Avg: %lf, Limit: %d, Count: %d, Valid: %d\n",
```
You can copy the format argument from the C++ version, it will work the same.

DennisdeBest: indeed ! I'll clean it up some more, thanks !

HugoSantiagoBecerraAdan:
From my own research and the different versions I've tested the use of `foreach` with `range` it's way slower than a raw `for` loop. The `range` allocates a whole array with the number of values defined and that allocation takes time when the array is big. 

Changing property `$myDict` to a constant, and using `SplFixedArray` instead of a raw array will also boost performance.
```
// array version
Passes: 248, Time: 10.034645, Avg: 0.040462, Limit: 1000000, Count: 78498, Valid: 1
// SplFixedArray version
Passes: 330, Time: 10.007499, Avg: 0.030326, Limit: 1000000, Count: 78498, Valid: 1
```
I think `microtime` is good enough for this benchmark and there is no need for `hrtime`.

PrimePHP/PrimePHP.php
HugoSantiagoBecerraAdan: Here is my own version if you want to check it out: https://github.com/HugoSantiagoBecerraAdan/Primes/blob/main/PrimePHP/PrimePHP.php

It includes the last changes from the CPP version, like reducing the execution time to 5 seconds instead of 10, remove `getBit` and `setBit` methods, etc.

I've found there's no fast way to set the whole list of initial values to `true`, like in the original CPP version, by using `SplFixedArray` but `array_fill` has a huge weight in the execution time. So I decided to invert the checks and value asingments of the bit list to keep the main functionality intact without compromising the performance.

Hope you can find it useful. ;)

DennisdeBest:
Indeed the for loop made a big difference : 
```php
///foreach(range(...))
Passes: 5, Time: 11649ms, Avg: 2329ms, Limit: 1000000, Count: 78498, Valid: True%        
//for                                                                                                                                      
Passes: 186, Time: 10032ms, Avg: 53ms, Limit: 1000000, Count: 78498, Valid: True%   
```

I tried he SPLFixedAray before aswell but this seems to not work correctly on PHP8, or I am missing something.

Thanks !

GXTX:
No real reason to have this PHP8 only. A simple change in the class constructor makes this PHP7 compatible. Also consider playing around with DateTime class that's built in.

PrimePHP/PrimePHP.php
```diff
@@ -0,0 +1,124 @@
...
+
+    public function __construct(
+        private int $sieveSize = 1000000,
+    )
+    {
+        $rawbitSize = (int)($this->sieveSize + 1) / 2;
+        $this->rawbits = array_fill(0, $rawbitSize, true);
+    }
+
+    private function getBit(int $index): bool
```
AJenbo, line 36: Type checking is done runtime, so despite being a fan of typed PHP for this you would get better performance by removing it and possibly using old-school PHPDoc instead.
```suggestion
    /**
     * @param int $index
     * @return bool
     */
    private function getBit($index)
```
Doing this to all of the code improves the performance by about 12%
Passes: 67, Time: 10010ms, Avg: 149ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 75, Time: 10062ms, Avg: 134ms, Limit: 1000000, Count: 78498, Valid: True

AJenbo: Also worth noting the python version comes in at 63 so this seams to be in the valid range of languages :P

DennisdeBest:
> No real reason to have this PHP8 only. A simple change in the class constructor makes this PHP7 compatible. Also consider playing around with DateTime class that's built in.

This got me curious so I added some docker config in a branch : https://github.com/DennisdeBest/Primes/tree/docker-environment

And I ran it on different php-cli-alpine containers, the results are interesting, there is a jump in performance between php 7.1 and 7.2

### 8

Passes: 95, Time: 10052ms, Avg: 105ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 107, Time: 10094ms, Avg: 94ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 102, Time: 10027ms, Avg: 98ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 104, Time: 10014ms, Avg: 96ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 107, Time: 10065ms, Avg: 94ms, Limit: 1000000, Count: 78498, Valid: True

### 7.4

Passes: 102, Time: 10015ms, Avg: 98ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 104, Time: 10123ms, Avg: 97ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 97, Time: 10000ms, Avg: 103ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 98, Time: 10037ms, Avg: 102ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 99, Time: 10068ms, Avg: 101ms, Limit: 1000000, Count: 78498, Valid: True

### 7.3

Passes: 93, Time: 10028ms, Avg: 107ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 94, Time: 10035ms, Avg: 106ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 93, Time: 10108ms, Avg: 108ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 93, Time: 10100ms, Avg: 108ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 92, Time: 10003ms, Avg: 108ms, Limit: 1000000, Count: 78498, Valid: True

### 7.2

Passes: 93, Time: 10066ms, Avg: 108ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 90, Time: 10059ms, Avg: 111ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 95, Time: 10093ms, Avg: 106ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 95, Time: 10070ms, Avg: 106ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 95, Time: 10013ms, Avg: 105ms, Limit: 1000000, Count: 78498, Valid: True

### 7.1

Passes: 47, Time: 10092ms, Avg: 214ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 49, Time: 10113ms, Avg: 206ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 50, Time: 10094ms, Avg: 201ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 48, Time: 10168ms, Avg: 211ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 48, Time: 10084ms, Avg: 210ms, Limit: 1000000, Count: 78498, Valid: True

### 7.0

Passes: 43, Time: 10039ms, Avg: 233ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 43, Time: 10044ms, Avg: 233ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 43, Time: 10134ms, Avg: 235ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 44, Time: 10248ms, Avg: 232ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 43, Time: 10104ms, Avg: 234ms, Limit: 1000000, Count: 78498, Valid: True

PrimePHP/PrimePHP.php
DennisdeBest: Hi, I do not see any real difference on my end.

## PHP (8.0.3)
### All types
```shell
Passes: 184, Time: 10015ms, Avg: 54ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 187, Time: 10052ms, Avg: 53ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 186, Time: 10017ms, Avg: 53ms, Limit: 1000000, Count: 78498, Valid: True
```

### No types (parameters / return)

```shell
Passes: 123, Time: 10027ms, Avg: 81ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 191, Time: 10003ms, Avg: 52ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 188, Time: 10004ms, Avg: 53ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 183, Time: 10008ms, Avg: 54ms, Limit: 1000000, Count: 78498, Valid: True
```

### No types only getBit and clearBit parameters

```shell
Passes: 187, Time: 10012ms, Avg: 53ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 184, Time: 10057ms, Avg: 54ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 186, Time: 10031ms, Avg: 53ms, Limit: 1000000, Count: 78498, Valid: True
```

maybe this is where JIT comes in ? I have enabled it on my cli

```shell
php -i | grep -E '^opcache\.(enable_cli|jit|jit_buffer_size) '
opcache.enable_cli => On => On
opcache.jit => tracing => tracing
opcache.jit_buffer_size => 50M => 50M
```


And indeed now that it is getting a lot better thanks to the comments it is running quite a bit faster than Python on my computer

## Python (3.6)

```shell
Passes: 40, Time: 10.209030065016123, Avg: 0.25522575162540306, Limit: 1000000, Count: 78498, Valid: True
Passes: 40, Time: 10.075028802995803, Avg: 0.25187572007489506, Limit: 1000000, Count: 78498, Valid: True
Passes: 40, Time: 10.057532230013749, Avg: 0.2514383057503437, Limit: 1000000, Count: 78498, Valid: True
```

AJenbo: I also removed casting when I removed the type, maybe that is what makes the difference?

Types
```
php8.0 PrimePHP.php
Passes: 170, Time: 10035ms, Avg: 59ms, Limit: 1000000, Count: 78498, Valid: True
```
No types and no casting
```
php8.0 PrimePHP.php
Passes: 186, Time: 10011ms, Avg: 53ms, Limit: 1000000, Count: 78498, Valid: True
```
No types and no casting + JIT
```
php8.0 -dopcache.enable_cli=1 -dopcache.jit_buffer_size=100M PrimePHP.php
Passes: 366, Time: 10003ms, Avg: 27ms, Limit: 1000000, Count: 78498, Valid: True
```

AJenbo: No types, explicit casting:
`Passes: 173, Time: 10022ms, Avg: 57ms, Limit: 1000000, Count: 78498, Valid: True`
No types, implicit casting:
`Passes: 179, Time: 10002ms, Avg: 55ms, Limit: 1000000, Count: 78498, Valid: True`

HugoSantiagoBecerraAdan: There's no real benefit on performance here by removing type of arguments and returns, so it's better to use it wherever you can. Result may vary so should run each benchmark few times.

I suggest to use `strict_types` so automatic type casting on arguments and returns will be avoided at all.
```
declare(strict_types=1);
```

DennisdeBest: I added the strict_types and indeed I do not really see any performance difference on my end.

HugoSantiagoBecerraAdan:
What trouble did you find using `SplFixedArray`? Tell me what happend and I will help you.

`SplFixedArray` has a huge improvement on the performance.

```
php8.0 -dopcache.enable_cli=1 -dopcache.enable=1 -dopcache.jit_buffer_size=100M -dopcache.jit=1255 -dxdebug.mode=off PrimePHP.php

// Your version with array
Passes: 202, Time: 10038ms, Avg: 49ms, Limit: 1000000, Count: 78498, Valid: True
// My implementation with SplFixedArray
Passes: 407, Time: 10.023325, Avg: 0.024627, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1
```

See my implementation here, with all changes from CPP version: https://github.com/HugoSantiagoBecerraAdan/Primes/blob/main/PrimePHP/PrimePHP.php

DennisdeBest:
@HugoSantiagoBecerraAdan actually the only problem was my countPrimes function, I modified it a bit looking at your ode and everything works.

I now initialize the rawbit array like this 

```php
        $this->rawbits = SplFixedArray::fromArray(array_fill(0, $this->rawbitSize, true));
```

but I do not see any performance imrovements weirdly enough so there must be some more stuff to change on my side

HugoSantiagoBecerraAdan:
@DennisdeBest `array_fill` has a has a huge penalty in the execution time.

To avoid that problem, I decided not to use `array_fill` on `SplFixedArray` creation. `SplFixedArray` creates all entries needed for the list using `null` as default value.

```php
// This creates the whole list with null values
$this->Bits = new SplFixedArray($this->sieveSize);
```

Then I inverted the checks and value asingments of the bit list to keep the main functionality intact without compromising the performance.

```php
// Then invert the checks, this applies in several places
if (!$this->Bits[$num]) {
```

```php
// Also invert value asignment to keep main functionality intact
$this->Bits[$num] = true;
```

This is by far the best result:
```
// SplFixedArray
Passes: 196, Time: 5.004823, Avg: 0.025535, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

// SplFixedArray with array_fill
Passes: 114, Time: 5.028326, Avg: 0.044108, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

// Array with array_fill
Passes: 154, Time: 5.015931, Avg: 0.032571, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1
```

Notice `Time` is 5 secs. instead of 10.

davepl:

PR #47, opened 2021-04-04
Ziglang implementation of Prime, staying somewhat true to the original implementation.

devblok:
Since someone already has a submssion with my day-job language (Golang), here goes one for Ziglang! It's a young, pre-release language, based on LLVM (for now at least).

My findings vs what I expected:
GNU Compiler PrimeCPP:
```
Passes: 9858, Time: 5.000000, Avg: 0.000507, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1
```
LLVM-11 Clang PrimeCPP:
```
Passes: 10417, Time: 5.000000, Avg: 0.000480, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1
```
My Ziglang implementation (LLVM):
```
Passes: 8927, Time: 5.00039, Avg: 0.00056, Limit: 1000000
```

The CPU is `AMD Ryzen R7 3700X`.

Well, wasn't as great as I hoped, but then again, I bet there's a lot of language specific stuff I could use to bump this. In the interest of keeping this more or less true to original I didn't. I'll try those on my own another day and see what I can squeeze out. EDIT: Found a silly thing, performance is much better now. See comment.

As for how to run this:
* Download the package for you at https://ziglang.org/
* Install, yada-yada
* Navigate to the `PrimeZig` directory
* Run `zig build -Drelease-fast=true`, to get the optimized build
* You will find the binary at `./zig-cache/bin/`

### DIFF
--- a/PrimeZig/build.zig
+++ b/PrimeZig/build.zig
@@ -1,0 +1,27 @@
+const Builder = @import("std").build.Builder;
+
+pub fn build(b: *Builder) void {
+    // Standard target options allows the person running `zig build` to choose
+    // what target to build for. Here we do not override the defaults, which
+    // means any target is allowed, and the default is native. Other options
+    // for restricting supported target set are available.
+    const target = b.standardTargetOptions(.{});
+
+    // Standard release options allow the person running `zig build` to select
+    // between Debug, ReleaseSafe, ReleaseFast, and ReleaseSmall.
+    const mode = b.standardReleaseOptions();
+
+    const exe = b.addExecutable("PrimeZig", "src/main.zig");
+    exe.setTarget(target);
+    exe.setBuildMode(mode);
+    exe.install();
+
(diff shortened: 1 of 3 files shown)


devblok:
Okay, yeah. I poked around some more and found, peculiarly, that replacing bool with u8 (unsigned byte), improves performance dramatically.

Now I got:
```
Passes: 14304, Time: 5.00012, Avg: 0.00035, Limit: 1000000
```

Don't know why, maybe the compiler does not like optimizing bool for some reason.

davepl:

ManDeJan:
> Don't know why, maybe the compiler does not like optimizing bool for some reason.

This is probably a memory alignment issue, I forked your repository and added alignments to the field data and the run function, which makes both the `bool` and `u8` functions perform similar (~14500 passes on my 3700x) :-)

PR #48, opened 2021-04-04
Fast Inverse Square Root

byte-surgeon:
You can significantly improve performance using the Fast Inverse Square Root algorithm.


davepl:
If I remember my magical constants correctly, this is an ESTIMATE, is it not?  As in it might be wrong for certain numbers, which would be OK when doing graphics but bad in a prime sieve.  I could be wrong, let me know if I am!

byte-surgeon:
The estimate is far accurate enough for this purpose considering we only need the integer part of the number. Apart from a stupid bug in my code where I forgot to return the reciprocal of the inverse square root, it yields the correct results.

PR #49, opened 2021-04-04
Add benchmark action (CI)

flostellbrink:
Runs benchmark on every commit and pull request to main.
Posts a comment on those commits like the one below.

The hardware these actions will run can of course change between runs, so the results might not be directly comparable.
Still this makes sure that the implementations actually work, and gives us some clue of how they perform.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,75 @@
+name: Benchmark
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  python:
+    runs-on: ubuntu-latest
+
+    outputs:
+      result: ${{ steps.benchmark.outputs.result }}
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+      - name: Benchmark
+        id: benchmark
+        working-directory: PrimeSievePY
+        run: echo ::set-output name=result::$(python3 PrimePY.py)
+
+  cpp:
+    runs-on: ubuntu-latest
+
+    outputs:
+      result: ${{ steps.benchmark.outputs.result }}
+
+    steps:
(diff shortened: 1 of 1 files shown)


rhbvkleef:
Boy GH Actions are slow 😳. I want to modularize the actions, so that each language can define their own action. Do you see any posibility of doing something like that? I think it'll make it a lot neater, looking at how many languages there already are.

flostellbrink:
From my understanding separate workflows have a hard time communicating.

At the moment the jobs for all languages run in parallel and then post a single message to the commit.
This can be scaled to many languages and implementations all running in parallel.
image


Granted, this workflow file will get pretty big, but at ~20 lines of code per implementation it won't get that bad.

rhbvkleef:
> From my understanding separate workflows have a hard time communicating.
> 
> At the moment the jobs for all languages run in parallel and then post a single message to the commit.
> This can be scaled to many languages and implementations all running in parallel.

This is not what I meant. What I was interested about is whether we can define separate `yml` files, one for each language/implementation, and share the basics, like the publishing of results between them.

marghidanu:
Yes, we can. And we can also make the execution sequential if we need.

rhbvkleef:
> Yes, we can.

Awesome!

> And we can also make the execution sequential if we need.

That might make a lot of sense, but only if we can force runners to only run one job at a time.

flostellbrink:
Okay cool. I have no idea how to make that work.
You wanna take it from here @marghidanu?

marghidanu:
Sure, the idea is to support all future solutions on different platforms. We can create a build matrix with supported solutions (submitted solutions), platforms, etc. so we can have all combinations. We can set the strategy for the build matrix with max-parallel jobs set to 1. We will also collect all output from these jobs in the matrix and allow the jobs to continue on fail. Each job will build and run the docker image expecting a standard output, which we can collect in the later step and generate a report directly in GitHub.

Hope I'm not missing anything.

marghidanu:
Please also take a look at #84

flostellbrink:
Closing in favor of #84

PR #50, opened 2021-04-04
Lua Roblox

seanbouk:
Lua code running in Roblox. Can but run your local machine or Roblox's server.


sudocode1:
this seems mildly pointless and difficult to execute, you would have to download the entire game to just run this little script. - this should be in raw Lua, rather than Roblox supported Lua in my opinion.

seanbouk:
> this seems mildly pointless and difficult to execute, you would have to download the entire game to just run this little script. - this should be in raw Lua, rather than Roblox supported Lua in my opinion.

We won't know if it's pointless until we can compare both versions!

It's also no harder to download the whole game than a whole Lua interpreter, and a Lua interpreter won't let you run the code on Roblox server.

micwoj92:
Doesn't roblox need account? I think that makes it a no-no.

seanbouk:
> Doesn't roblox need account? I think that makes it a no-no.

Perhaps. I don't remember reading that requirement.

micwoj92:
What I meant is that it makes it much harder to run and compare to other implementations, what If someone doesn't want to register account just to run this simple benchmark?

seanbouk:
> What I meant is that it makes it much harder to run and compare to other implementations, what If someone doesn't want to register account just to run this simple benchmark?

They're be out of luck. That wouldn't be this version. 

It would be a quick job to convert this code to plain Lua (just an event to remove and a function to unwrap), and it'd be interesting to see how they compare. I doubt they'd be the same, but that's the point of the drag race, no?

This code also allows you to choose to run it on a Roblox server. That means a plain Lua conversion wouldn't be the same. That's why I marked it as Lua Roblox.

It's unusual to be able to do things on someone's server without making an account. No different to what you did to be able to comment here.

MansenC:
I mean, this is 100% a meme and I appreciate a good meme. But at least including the source code would be an option and not pushing the plain binary. Include a way of building it and the source code (and perhaps the plain lua code) so we can see what's going on. Now let's try and make this sieve run doom

seanbouk:
> I mean, this is 100% a meme and I appreciate a good meme. But at least including the source code would be an option and not pushing the plain binary. Include a way of building it and the source code (and perhaps the plain lua code) so we can see what's going on. Now let's try and make this sieve run doom

I'm not sure you and I share the same definition of a meme, but you make a good point about pushing the binary (it actually is the source code in this case, but granted it's also a binary file).

I'll run it against against a lua interpreter, fix what breaks, and put a new PR up with source code in a .lua file and an updated README

seanbouk:
…actually I added it to this PR

ghost:
> this seems mildly pointless and difficult to execute, you would have to download the entire game to just run this little script. - this should be in raw Lua, rather than Roblox supported Lua in my opinion.

Roblox in the title is misleading as this can run in a non-Roblox environment Lua completely fine.

If you don't have Lua installed and you want to test this, the easiest way to do so is by using [replit.com](https://replit.com/languages/lua) or any online sandbox.

rhbvkleef:
Do we need a separate language tag for Roblox? I think we should just categorize this with Lua.

foxt:
> Doesn't roblox need account? I think that makes it a no-no.

@micwoj92 Yes, Roblox does require an account, even just to run a script sadly

----

It's also important to note that Roblox Lua is **not** standard Lua. It is very similar in many ways, but Roblox have their own custom Lua engine named [Luau](https://roblox.github.io/luau/performance). There is a Roblox CLI, which does not require an account, as it is an internal (read: unless you're willing to get a job at Roblox, and then quickly lose it, we're not getting our hands on it)

marghidanu:
If you can build it in a Docker image and read the Robolx credentials from environment variables I don't see any problems with this. It would build normally as any other implementation. The only difference would be that I have to pass the Robox credentials when executing this implementation in Docker. Let me know if that's possible.

foxt:
From what I know, Docker runs everything in a Linux container, where the problem being that Roblox having no public CLI tools, all .lua files are ran in the GUI Roblox Studio application which is Windows and macOS only. Wine is possible however.

seanbouk:
> > Doesn't roblox need account? I think that makes it a no-no.
> 
> @micwoj92 Yes, Roblox does require an account, even just to run a script sadly
> 
> It's also important to note that Roblox Lua is **not** standard Lua. It is very similar in many ways, but Roblox have their own custom Lua engine named [Luau](https://roblox.github.io/luau/performance). There is a Roblox CLI, which does not require an account, as it is an internal (read: unless you're willing to get a job at Roblox, and then quickly lose it, we're not getting our hands on it)

I think Roblox Lua can be thought of as a subset or continuation of vanilla Lua. This sieve runs FOUR times as fast on my machine. These reasons are why I've kept both in the PR. They're too closely related to separate, but too similar to dismiss one version of (in my opinion)

rbergen:
We can, and I personally would love to, include a "plain Lua" solution. If you would like to contribute that, the solution would need to comply with the [contribution guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md). Amongst others, that means that the PR needs to be opened on the drag-race branch, not original (formerly main).

Concerning Roblox Lua, there are two problems with including that in the (Docker-executed) benchmark setup: 
* The need for a Roblox account.
* The fact that the rblx file is a form of binary executable; we don't merge binary versions of solutions.

That said, it is perfectly fine to include documentation on how to build and run the solution in the Roblox Lua context, to support those people who individually want to do so.

foxt:
> The fact that the rblx file is a form of binary executable; we don't merge binary versions of solutions. 

Just for completeness, while `rblx` files are binary files, they are not binaries or executables (more like a zip file), and contain an entire Roblox game. `rbxlx` files are stored as XML. You can also just click the `Run Script` button in a blank project.

However I do agree that Luau would be quite hard to get to work in the context of this project, as the CLI code runner is private and propriatery to Roblox staff, and the Luau programming language is [closed-source](https://luau-lang.org/why#open-source).

There are [some](https://luau-lang.org/compatibility) differences from standard Lua, but it'd be interesting to see any performance differences from Luau to Lua.

rbergen:
I should indeed not have made a claim about the nature of the binary that rblx is. At the same time, it doesn't matter for the key reason we don't merge binaries, being that it is practically impossible to establish what's in them (particularly when proprietary tooling is involved). That, in turn, opens up the project to a range of potential issues, legal concerns being one of them.

I understand that the argument for including the rblx file is being made. In fact, I would also be interested in seeing comparisons between the two Lua "flavors". However, if there is no way to represent its contents in a human-readable form, then I'm afraid we can't include it. 

As I said, I would still very much like to include "plain" Lua, regardless. :)

foxt:
> However, if there is no way to represent its contents in a human-readable form

Roblox Studio also has the ability to load/save it's legacy XML based format, which while not clean, is definately human readable, which when cleaned results in a human readable file. (There are also no IO features in Roblox except HTTP, so malware is not possible)

image

rbergen:
XML is clean enough. That could be added to the Lua solution if that were to be submitted for inclusion on the drag-race branch.

foxt:
Infact, you don't even need to include a RBXL file. You can just open a blank Studio project, and go to `Model`, and `Run Script`
image

(Note that there is an expected `arg` variable in the script that isn't defined in Roblox, and therefore will not run.)

rbergen:
Great! In that case I hope we'll soon have a PR that we can merge to add Lua to the list of prime sieve languages. :)

rbergen:
Closing after merge of #251

PR #51, opened 2021-04-04
Add Swift version

j-f1:
On my machine, PrimeCPP gets ~7500 passes while PrimeSieveSwift gets ~2100 passes.


davepl:

PR #52, opened 2021-04-04
Add java implementation of prime sieve

MansenC:
I created a java implementation of the sieve. This is a port of the original C# version with additional optimizations:

- I'm using a boolean array instead of the equivalent of BitArray for Java (BitSet)
- The getBit and setBit methods have been optimized

You can execute this sieve with an installed java version by just calling the run-sieve.sh file from (hopefully) Windows, Mac and Linux if you have a java version (8+ works for sure, probably works with 6+) installed.

Benchmarks using my Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz and 10 seconds time:
- Running in shell I get ~3740 passes
- Running in eclipse I get ~10800 passes (Does anyone have a clue why that is using the git bash)
- Compares to C++ running in shell for 5 seconds ~7200 passes

Any comments/reviews/suggestions?


davepl (commented):
Are all the support files needed?  The shell scripts and so on?  Less is more if possible!

davepl:

PR #55, opened 2021-04-04
Add Common Lisp Implementation

mikehw:
This adds a prime sieve using common lisp, and is set up to count primes under 1000000 for 10 seconds to measure performance.

I testing using sbcl. Installed sbcl using apk, and then ran by calling `sbcl --script PrimeSieve.lisp`


davepl:

PR #57, opened 2021-04-04
Delphi. 4,809 passes

JackTrapper:
Delphi 10.3 implementation. 4,809 passes in 10 seconds.


DennisdeBest:
>nobody caught the number of primes under ten.

Indeed I think nobody noticed, that bit felt like copy pasting some lorem ipsum

davepl:

PR #59, opened 2021-04-05
Create prime sieve implementation in Bash

Nitepone:
This is a pure bash implementation with exception of usage of "awk" and "time" external commands. It uses bash builtin arithmetic blocks for all critical calculations.

The result output on my machine with a 3950x:
`Passes: 1, Time: 10408, Avg(ms): 10408, Limit 1000000, Count: 78498, Valid: 1`

Yes, it can't complete a single iteration with a 1 million limit in the 10 second runtime.

### DIFF
--- a/PrimeBASH/PrimeBASH.sh
+++ b/PrimeBASH/PrimeBASH.sh
@@ -1,0 +1,157 @@
+#! /bin/bash
+#
+# A simple prime sieve based on https://github.com/davepl/Primes
+# Written in bash.
+#
+# PrimeSH.sh
+# Copyright (C) 2021 nitepone <admin@night.horse>
+#
+# Distributed under terms of the MIT license.
+#
+
+
+declare -A primeCounts=( \
+	[10]=1
+	[100]=25
+	[1000]=168
+	[10000]=1229
+	[100000]=9592
+	[1000000]=78498
+	[10000000]=664579
+	[100000000]=576145
+)
+
+sieveSize=0
+bitArray=()
+RUNTIME_SEC=10
+
+
+
+function pr {
+	printf "\\e[36m[PrimeBash]\\e[0m "
+	printf "%s\\n" "$@"
+}
+
+function pr_err {
+	printf "\\e[36m[PrimeBash]\\e[31m " >&2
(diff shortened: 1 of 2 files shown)


ciplogic:
Still faster than C64... Awesome work!

micwoj92:
@Nitepone Have you thought of using mawk instead of awk? It could be faster, but actually I am not sure what awk implementation do the runners (running ubuntu) use by default, it might be using mawk already.

Nitepone:
I have updated the output to meet the original standard:
`Passes: 2, Time: 20.340799132, Avg: 10.170399566, Limit: 1000000, Count: 78498, Valid: True`

Additionally, my implementation is slightly faster now due to changes in how the bit array is allocated.


@micwoj92 I decided to remove the awk usage entirely. I can get by using bash builtin math to find the integer square roots. 

@rhbvkleef I have tried to make changes so that this is faithful to the original style. What in particular caused this PR to be marked as "unfaithful"?

minecraftchest1:
I wish I looked here before doing the same thing.

rbergen:
@Nitepone We are happy to work towards including your solution in the drag-race; we don't yet have a Bash implementation in the repo. I for one would **love** to have one :) 

However, to be able to do so, it would need to comply with the [contributing guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md). That also means that the PR needs to be opened on the drag-race branch of this repo, not original (formerly main).

With regards to faithful and unfaithful, a faithful implementation basically has to wrap the full sieve state in an object/class instance (or whatever a language's equivalent of that is), that is created from scratch for each run. We recognize that not all languages have such a construct. The "unfaithful" moniker should not be interpreted as criticism to the author of a solution, but as a characteristic of the implementation. The contributing guidelines I pointed to earlier elaborate on this and other characteristics we now consider relevant.
Concerning that relevance: I don't think it matters much in this case, due to the overall performance of the implementation. This stuff generally starts to matter towards the top of the performance leaderboard. :)

rbergen:
Closing after merge of #251

PR #61, opened 2021-04-05
I modified your code to compile with FPC 

jaybinks:
https://github.com/jaybinks/Primes/tree/main/PrimeSievePascal


jaybinks:
@davepl this version simply requires the freepascal compiler (https://www.freepascal.org/) and should compile without any issue using run.sh

ciplogic:
Why yourr PR adds primes executable to Cpp folder?

May you add a Readme file that documents at least on your machine the performance ratio against C++? And Delphi?

ciplogic:
I also tested with latest FPC/Lazarus on Windows and it beat C++/64 bit, really!

My number of iterations (on a Zen2 4800H laptop):

C++ | 15019
-- | --
FPC | 19207

I added "inline" keyword as the last Delphi PR includes it too:
```@pascal
		function GetBit(Index: Integer): Boolean; inline; // <-- added here
		procedure ClearBit(Index: Integer);inline;    //<-- added here    
```
But still, cool stuff!

rhbvkleef:
Could you remove the executable file from the diff?

davepl:
I sure can, and I know it’s a terrible idea, but I’ve had a ton of requests from folks who want to run it but that don’t want to or don’t know how to compile it.  

 

What’s the shortest path safe alternative? A simple .zip of the exe?  

 

Cheers,

Dave

rhbvkleef:
> I sure can, and I know it’s a terrible idea, but I’ve had a ton of requests from folks who want to run it but that don’t want to or don’t know how to compile it.
>
> What’s the shortest path safe alternative? A simple .zip of the exe?
>
> Cheers,
>
> Dave

Ah, I see. I hadn't noticed you committed the binary. I think we could probably create a release on the `main` branch, and add downloadable binaries there for the people that want to give running it a go. I just don't really think publishing executables in git is a good idea, as it's not really meant for it. It would cause a lot of merge conflicts.

marghidanu:
@davepl you can create a tag and upload the binary as an artifact.

rbergen:
Closing this as we have a separate Pascal solution in place that compiles using FPC.

PR #65, opened 2021-04-05
Added C, and I did a very small "improvement" to CPP

alexte:
The small change in the C++ code is: search for next prime factor after using it. 
I think this is cleaner, as the first prime in already known to be 3.

I always thought of C code to look cleaner than C++, so I could not resist adding C.

It's actually about 8% faster than C++, at least on my maschine.
I hope I got it right.

### DIFF
--- a/PrimeC/PrimeC.c
+++ b/PrimeC/PrimeC.c
@@ -1,0 +1,120 @@
+// ---------------------------------------------------------------------------
+// PrimeC.c : based on Dave's Garage Prime Sieve in C++
+// ---------------------------------------------------------------------------
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <math.h>
+#include <time.h>
+
+#define FALSE 0
+#define TRUE 1
+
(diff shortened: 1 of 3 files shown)


rbergen:
@alexte We currently have 2 C solutions in place in the drag-race branch, one of which actually contains a range of different implementations. If you want to still add yours, please make it comply with the [contributing guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md). That does mean that this PR would need to be opened on the "drag-race" branch, not "original" (formerly "main").

Because of the fact that this would constitute another solution next to the ones that are already in place, we would consider if this solution differs from the others in a meaningful way, before we decide to merge.

Also, please separate the change to the C++ solution from your C submission. To be honest, I wouldn't open a new PR for it if I were you. This is because a) the C++ change is basically cosmetic (meaning that the performance impact is negligible) and b) Dave has asked us not to modify his implementations.

rbergen:
Closing this for reasons indicated.

PR #67, opened 2021-04-05
Update PHP implementation with changes from PrimeCPP

HugoSantiagoBecerraAdan:
Provides a closer aproach to the changes from CPP version introduced in commit [be15c62 - Rewrote timing code to be simpler and to not reset the object every pass](https://github.com/davepl/Primes/commit/be15c6282549f5ce8f30404f8f787f3494aaf832#diff-9f5d08557424a3b31a92099cba99c326bb0ec2caa67b70f53a0d5a099dac5dcc)

But a little difference with the CPP version has to be taken in consideration, due to the penalty found when initializing to `true` the elements in the list, even with function `array_fill` takes a significant amount of time.

`SplFixedArray` provides a faster aproach to handle large lists with a fixed number of elements in comparison to raw arrays, but in order to mantain good permformance, I decided not to initialize the value of the elements created inside the object, and left them as `null`.

```php
// Creates the whole list of elements with 'null' as value
$this->Bits = new SplFixedArray($this->sieveSize);
```

Then I inverted the checks and value asingments of the bit list to keep the main functionality intact without compromising the performance.

```php
// Invert the checks, this applies in several places
if (!$this->Bits[$num]) {
```

```php
// Also invert value asignment to keep main functionality intact
$this->Bits[$num] = true;
```

In the end, this is by far the best result I've found:
```
// SplFixedArray using 'null' as default value for all elements
Passes: 196, Time: 5.004823, Avg: 0.025535, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

// SplFixedArray and array_fill using 'true' as default value for all elements
Passes: 114, Time: 5.028326, Avg: 0.044108, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

// Array with array_fill using 'true' as default value for all elements
Passes: 154, Time: 5.015931, Avg: 0.032571, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1
```


photodude:
Would be interesting to see this expanded to show the performance evolution of PHP from PHP 4 through PHP 8 with some of the milestone minor branches in PHP 5 and PHP 7 (5.2, 5.3, 5.4, 5.6, 7.0, 7.2, 7.3, 7.4) along with HACK running on HHVM. As some say PHP 7.3+ is not your grandparent's PHP.
PHP sadly often gets a bad rap due to people's experience with PHP 4 and early PHP 5. Showing the performance evolution would be interesting.

aidabeorn:
> Would be interesting to see this expanded to show the performance evolution of PHP from PHP 4 through PHP 8 with some of the milestone minor branches in PHP 5 and PHP 7 (5.2, 5.3, 5.4, 5.6, 7.0, 7.2, 7.3, 7.4) along with HACK running on HHVM. As some say PHP 7.3+ is not your grandparent's PHP.

That would require a new implementation since SplFixedArray didn't show until 5.3, probably better to have that as a separate solution anyways

rbergen:
Closing this after merge of #307

PR #68, opened 2021-04-05
Minify the java implementation

MansenC:
@davepl Regarding #52 I have minified the implementation, removing gradle and the IDE setup completely, leaving only the source code in PrimeSieveJava.java, a gitignore, the readme for steps and additional information and a small script for execution behind. I agree, less is more and better.
I wasn't able to do this on the old PR since it was already merged.


rbergen:
@MansenC If you still want to apply this (which I think would be a good idea :) ) then please open the PR on the drag-race branch, where we can merge changes. Note that: 

- In that branch, the solution directory is PrimeJava/solution_1. 
- The Dockerfile I've added already "circumvents" Gradle, and performs a plain javac compile. That said, it would obviously have to be changed if the name or location of PrimeSieveJava.java change.

rbergen:
Closing this. #325 implements most of the changes in this PR.

PR #69, opened 2021-04-05
Some optimizations for Crystal  

marghidanu:
* ~~Dropped the `Array(Bool)` in favor of `BitArray` for the `bits` variable~~
* ~~Reversed functionality of `bits`, conditions on it evaluate the opposite way. This way I don't have to reinitialize the `BitArray` (by default it initializes to false).~~
* Switched from class to struct.
* More idiomatic code.

P.S.: If I reuse the instance on each pass the performance improves significantly. @davepl do you consider this a possible optimization?


ciplogic:
I am not having a final say on the matter, but reusing the instance of the array looks like a no-no, as GC is a part of Crystal and if it takes too long, so be it.

marghidanu:
Correct. Actually, I'm re-running the benchmark as we speak, I will commit the new results as part of this PR.

marghidanu:
Ok, the results are now correct.

PR #70, opened 2021-04-05
Added Dart version of algorithm

eagerestwolf:
I created a Dart version of the algorithm. I do know there is one issue already, and that is the initial count is always one; and I know why, but I'm not sure how to fix it. The reason is in the C++ algorithm, a boolean can be treated as a numerical value (for example `int count = (primeSieve >= 2)`). However, this is not the case with Dart, and I'm not entirely sure how to fix it, so `Count1` always displays 1 because my solution to that problem was to do `int count = (primeSieve >=2) ? 1 : 0`. Other than that the numbers seem to work out correctly.

Furthermore, I am very unimpressed by the performance of Dart. Granted, I'm not a "dart developer", so my code is probably sub optimal. Regardless, I was always under the impression that Dart was considered a competitor to C++ or C# for it's performance (considering it's an IL-based, interpreted language), but it's really more comparable to Python or NodeJS in my opinion (which makes sense, it is a web programming language). I may go ahead and do a NodeJS port today as well and see how they compare. For reference, the algorithm does ~1050 passes in 5 seconds running using the Dart CLI, and ~1330 passes in 5 seconds compiling using the Dart CLI and running the resulting binary on an M1 MacBook Pro.


rbergen:
@eagerestwolf We're happy to include this solution in the drag race, but for that it would need to comply with the [contribution guidelines](https://github.com/davepl/Primes/blob/drag-race/CONTRIBUTING.md). I've taken a look at the contents of your solution, and it looks like you're already pretty close to that. 
However, it would also be necessary to open the PR on the `drag-race` branch, not `main`.

eagerestwolf:
10/4. My apologies, these guidelines weren't in place at the time I submitted. I am working on a revised version now. First to address some dart quirks; and to comply with the guidelines. I should have it done this evening, and then I'll amend my PR.

eagerestwolf:
Ok, I goofed on the git side of things, so I had to create a new PR (#180), but everything is all good now. Sorry for the hassles.

PR #72, opened 2021-04-05
added Octave implementation

fspigel:
Hello! I saw that MATLAB was missing from your list of languages! Since this is essentially a mathematical problem, and MATLAB is such a popular and well-liked language, I wanted to correct that error. However, I am too poor to have access to MATLAB, so i did the next best thing and wrote it in Octave, which can interpret MALAB code, even though the inverse isn't true. 

I could blindly convert the code to what I think MATLAB would probably be able to interpret (by changing all the end statements like `endfor` and `endfunction` to just `end` and changing the increments from `var++` to `var+=1`, but I'm not 100% sure those are the only differences, and I wouldn't be able to test the code, so here we are.

I enjoy your content and I'm pretty sure you're a wizard. In the meantime, see you next time.

### DIFF
--- a/PrimeSieveOctave/PrimeSieveOctave.m
+++ b/PrimeSieveOctave/PrimeSieveOctave.m
@@ -1,0 +1,51 @@
+%%  MATLAB Prime Sieve
+%%
+%%  Fran Spigel, 5. 4. 2021.
+%%
+%%  This is the main function. Change the 'limit' variable in order to count
+%%  prime numbers up to that number.
+
+% A small library of pre-computed prime counts for powers of 10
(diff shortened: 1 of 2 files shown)


rbergen:
@fspigel We are happy to work towards including your solution in the drag-race. However, to be able to do so, it would need to comply with the [contribution guidelines](https://github.com/davepl/Primes/blob/drag-race/CONTRIBUTING.md). That also means that the PR needs to be opened on the `drag-race` branch of this repo, not `main`.

rbergen:
Closing after merge of #251

PR #73, opened 2021-04-05
touched up python implementation

fspigel:
Some of your python code needed touching up, here's the changes I made

* changed all the `this` instances to `self` (I didn't even know python would successfully interpret `this` as a keyword
* changed `sum(1 for b in this.rawbits if b)` to just be `sum(self.rawbits)` which is more pythonic (python can add booleans together by treating them as ones and zeros 
* sped up the calculation by about 30% by removing the parts of GetBit and ClearBit where you check if the input is even first. I get that this is a safety feature, but I think it's ok to leave it out in such a small codebase, considering how much CPU time was spent checking it every single time. If you insist on guarding against any kind of exception, I would suggest just making the `rawbits` array full length instead, and getting rid of the two functions entirely.

### DIFF
--- a/PrimeSievePY/PrimePY.py
+++ b/PrimeSievePY/PrimePY.py
@@ -27,15 +27,15 @@ class prime_sieve(object):
                     100000000 : 5761455
                   }
 
-    def __init__(this, limit):
-        this.sieveSize = limit
-        this.rawbits = [True] * (int((this.sieveSize+1)/2))
+    def __init__(self, limit):
+        self.sieveSize = limit
+        self.rawbits = [True] * (int((self.sieveSize+1)/2))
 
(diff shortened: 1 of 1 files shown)


dasKeks28:
`self` isn't even a real keyword but just a convention. You could also name it `foobar` or whatever you like. Calling `object.foo()` just passes the `object` as first argument into `foo(self)` (or `foo(im_the_self_variable)`).

+1 for replacing the `sum`, also noticed that in the last video. :smile:

fspigel:
You know, thinking about it now, that seems very obvious, haha. My linter was complaining about it and I'd never seen it done any other way, so I just went ahead and changed it. Cheers!

rbergen:
I think these changes are basically cosmetic. That being the case, I'm closing this.

PR #75, opened 2021-04-06
Update C# code to closer match the C++ changes.

iJohnnis:
I've updated the C# code to closer match the updated C++ operation. Given that C# cannot initialize an array without setting its values to false (without using unsafe code anyway) I suggest using the reverse array to keep track of the non-prime numbers. This code is around 3.7 times faster than the original without taking any shortcuts. Please, consider it as a fairer representation of the C# language.

### DIFF
--- a/PrimeSieveCS/PrimeCS.cs
+++ b/PrimeSieveCS/PrimeCS.cs
@@ -1,5 +1,5 @@
-﻿// ---------------------------------------------------------------------------
-// PrimeCS.cs : Dave's Garage Prime Sieve in C++
+// ---------------------------------------------------------------------------
+// PrimeCS.cs : Dave's Garage Prime Sieve in C#
 // ---------------------------------------------------------------------------
 
 using System;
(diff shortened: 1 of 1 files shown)


Kinematics:
I've implemented this in my own project (where I'm consolidating the different versions posted), and while it's indeed much faster, I'm not seeing a 3.7x improvement. What I get is a 75% improvement (3660 to 6350).

Hmm. Actually, I see you do direct access to the array, instead of using the `GetBit()` and `SetBit()` functions.  And, strictly speaking, the code checks in those functions aren't needed if the original loop calls are structured correctly. Let me re-run the benchmarks, with this variation.

```
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19042
AMD Ryzen 7 3700X, 1 CPU, 16 logical and 8 physical cores
.NET Core SDK=5.0.201
  [Host]     : .NET Core 5.0.4 (CoreCLR 5.0.421.11614, CoreFX 5.0.421.11614), X64 RyuJIT
  DefaultJob : .NET Core 5.0.4 (CoreCLR 5.0.421.11614, CoreFX 5.0.421.11614), X64 RyuJIT


|             Method | SieveSize |       Mean |    Error |   StdDev | Ratio |
|------------------- |---------- |-----------:|---------:|---------:|------:|
|      StandardSieve |   1000000 | 1,362.0 us | 12.52 us | 11.71 us |  1.00 |
|          BoolSieve |   1000000 |   984.0 us |  7.40 us |  6.56 us |  0.72 |
|  InvertedBoolSieve |   1000000 |   797.6 us |  7.74 us |  7.24 us |  0.59 |
| DirectInvertedBool1 |  1000000 |   576.0 us |  5.23 us |  4.08 us |  0.42 |
| DirectInvertedBool2 |  1000000 |   697.9 us |  4.21 us |  3.73 us |  0.51 |
|       RawBitsSieve |   1000000 |   828.4 us |  3.17 us |  2.81 us |  0.61 |
```

That's actually a pretty impressive speedup, though the benchmark seems to bounce between the 600 us and the 700 us values.  The 700 us value is closer to the loop count I get (around 7200). Also, the Standard version in the list already fixes some issues with the original code, so ends up doing a lot more loops (roughly 3660 vs 2800), which could factor into your comparison (and seem more reasonable).

Running the DIB just to get a loop count gives me 7220, a substantial improvement over the ~6200 for the normal inverted bool and the rawbits implementation.

It still gets hit by the memory overhead issues when multithreading, though, only getting about 29000 vs the 39000 for the rawbits and ref struct versions.

iJohnnis:
@Kinematics Let me clarify, 3.7 ~= (new version's iteration) / (old version's iteration). It was actually more like 3.685...(more digits)... and it was the result of a single run, not extensive testing. I just wanted to put a very rough estimation, to give an idea to the reader about how significant the difference is.

The memory hit it is to be expected. But given that the C++ version uses a boolean vector, I do not see the problem. Obviously, if you had to use something like that (although I do not see why you would use a prime-sieve) and it was not just a simple test, the memory would be an important issue.

rbergen:
@iJohnnis We will not be able to merge this PR in its current form. Dave has asked us to not make further changes to the "original" branch (formerly "main"), and also not to apply performance-altering changes to his implementations. You could consider opening a PR on the "drag-race" branch to add this as a new solution, keeping the [contributing guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md) in mind. As you will understand, in that case, when reviewing your solution and deciding to merge or not, we would consider that we do have 3 existing C# solutions in place. One of those is actually a broad range of different implementations.

Concerning the memory use: it does matter, in part because the vector<bool> in C++ is actually a special case. Although it pretends to be a "vector of bools", each "bool" is mapped to one bit in practice. This is (indeed) a deliberate inconsistency in the C++ standard library's implementation of the "vector<>" template. It has been a source of heated debate for 2 decades within the C++ community, but that is what it is.

rbergen:
Closing this due to other C# solutions being in place.

PR #76, opened 2021-04-06
Added a C11 implementation

asdfjkluiop:


rbergen:
@asdfjkluiop We currently have 2 C solutions in place in the drag-race branch, one of which actually contains a range of implementations of which some use multiple threads. If you want to still add yours, please make it comply with the contributing guidelines. That does mean that this PR would need to be opened on the "drag-race" branch, not "original" (formerly "main").

Because of the fact that this would constitute another solution next to the ones that are already in place, we would consider if this solution differs from the others in a meaningful way, before we decide to merge.

rbergen:
Closing this due to other C solutions being in place.

PR #78, opened 2021-04-06
Add Scheme implementation

mihapiha:
This is implementation of original algorithm with most iterations transformed into recursion which is more Schemey. It can be run with Guile and Chez implementations. Both are available on Debian and only Chez on Windows.

### DIFF
--- a/PrimeSieveScheme/prime.scm
+++ b/PrimeSieveScheme/prime.scm
@@ -9,68 +9,73 @@
     (100000000 . 5761455)))
 
 
-(define (find-first vector start stop)
-  (let finder ((i start))
-    (cond
-     ((> i stop) start)
-     ((even? i) (finder (+ 1 i)))
-     ((vector-ref vector (quotient i 2)) i)
-     (else (finder (+ 1 i))))))
+(define (count-primes bit-array)
+  (do ((i 0 (+ i 1))
+       (count 0))
+      ((= i (vector-length bit-array)) count)
+    (when (vector-ref bit-array i)
+      (set! count (+ 1 count)))))
+  
 
-
-(define (update vector start stop)
-  (do ((i (* start 3) (+ i (* start 2))))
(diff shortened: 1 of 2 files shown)


mihapiha:
I've now remodeled the recursion into loops and named all the variables as they are named in original C# implementation so I think it's completely faithful now.

rhbvkleef:
I´m sorry for not giving you the reasons I tagged this as `unfaithful`. You see, the ruleset hasn´t been completely defined as of yet. The primary reason I used for tagging this as such was that no class is used in order to contain some state and logic. Of course the difficulty with that is that Scheme isn´t an OOP language, so for languages such as Scheme, we still need to think about how and whether to expand the `faithful` criteria.

mihapiha:
Would a good criteria be that no buffers are reused between iterations in addition to closely following the alghoritm? In the case of Python those are rawbits and primeCounts.

rbergen:
@mihapiha The interpretation of "faithful" has basically settled on what you are saying as far as buffers are concerned, in the sense that C struct-like implementations are also accepted. Generally speaking, they contain at least the sieve size and the prime array. I'm certainly no expect on Scheme, but the contents of [this page](http://people.cs.aau.dk/~normark/prog3-03/html/notes/oop-scheme_themes-classes-objects-sec.html) for example, do make it seem like such a setup could be implemented in Scheme, as well.

Concerning staying close to the original algorithm, issue #177 was recently opened about this. At the moment, the drag-race basically is setup to make comparisons in two dimensions (language and algorithm), but we may well end up separating those category-wise, soon. 

In any case, for this solution to be included in the comparison, it would need to adhere to the [contribution guidelines](https://github.com/davepl/Primes/blob/drag-race/CONTRIBUTING.md). That also means that the PR would need to be opened on the `drag-race` branch, not `main`.

rbergen:
Closing after merge of #452

PR #81, opened 2021-04-06
Fix integer rounding bug in CPP version

fatho:
The number of microseconds returned is an integer, and since the divisor was also an integer, the result would always be exactly 5, even if it took slightly longer than that.


rbergen:
I've opened #220 on the drag-race branch to fix this.

PR #82, opened 2021-04-06
Add Haskell implementation

fatho:
Just as the title says, adds a Haskell implementation of the sieve.
The benchmark results from my machine are in the readme.


fatho:
@rhbvkleef The meaning of `stl-unfaithful` is "lack of bitsets" I assume? In any case, I added a variants using a bit set to this implementation as well.

rbergen:
@fatho We are still happy to work towards including your solution in the drag-race; we don't yet have a Haskell implementation in the repo. However, to be able to do so, it would need to comply with the [contribution guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md). That also means that the PR needs to be opened on the drag-race branch of this repo, not original (formerly main).

fatho:
@rbergen I finally found the time to (hopefully) make this PR conform with the updated rule set.

rbergen:
@fatho Thank you for taking the time for this. I'll review the solution when I have the time to do so.

rbergen (approved):
LGTM. I'm happy that we can finally add this. :)

marghidanu (approved):

PR #83, opened 2021-04-06
Fortran Version of the PrimeSieve

johandweber:
This pull request contains a Fortran version of the PrimeSieve, along with a Shell scipt that compiles the program using gfrotran, the Fortran compiler of the GNU Compiler collection that is available along with gcc and g++ for Linux, Max and Windows.

I found a performance very similar to the C++-Implementation.

Because Fortran does not contain a bitfield data type, I used 64bit integers to store the bits.

### DIFF
--- a/PrimeFortran/PrimesFortran.f03
+++ b/PrimeFortran/PrimesFortran.f03
@@ -1,0 +1,180 @@
+program primes_fortran
+  use iso_fortran_env
+  implicit none
+
+  ! All bits are set to 1 for an integer value of -1 as in Fortran there are only signed numbers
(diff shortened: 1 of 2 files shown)


rbergen:
@johandweber We can include this solution in the drag race, but for that to happen:

- The solution must be setup in accordance with the [contribution guidelines](https://github.com/davepl/Primes/blob/drag-race/CONTRIBUTING.md).
- As indicated in the guidelines, the PR needs to be opened on the `drag-race` branch, instead of `main`.

marghidanu:
Wow! Super cool, was waiting for something like this.

PR #84, opened 2021-04-06
Setting CI base for multiple builds via Docker

marghidanu:
I've set up a build matrix using the implementation name(just one for now)  and the os for the runner (just Ubuntu for now). As the next steps, we need all to adhere to the same output, run duration, and the number of iterations to run (that can be controlled from the CI). Additionally, I've added one last step in the job to gather all the data and in the future, we can add a script to aggregate results and generate a report.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -29,13 +29,11 @@ jobs:
           string: ${{ matrix.solution }}
         
       - name: "Build container image"
-        continue-on-error: true
         run: |
           cd "${{ matrix.solution }}"
           docker build -t "${{ steps.container_name.outputs.lowercase }}" .
       
       - name: "Run tests"
-        continue-on-error: true
         run: |
           docker run --rm "${{ steps.container_name.outputs.lowercase }}" | tee "${{ matrix.solution }}-${{ matrix.os }}.txt"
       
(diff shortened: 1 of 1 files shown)


marghidanu:
@rhbvkleef any chance of getting this merged?

rhbvkleef:
Yeah, merging. Let´s build it from here.

rhbvkleef:
Do I really need to place a workflow on master for us to get any workflows running, or is this a property that can be changed in the settings?

micwoj92:
Change first lines to

```
on:
  push:
    branches: [ drag-race ]
  pull_request:
    branches: [ drag-race ]
```

and it should work

marghidanu:
@rhbvkleef I assumed that the drag race will actually be run on master and with each merge we can re-run the CI automatically. You can change the branch name as described above. All output will be written to files, we need to write a script that goes over the files and generates some stats. Unfortunately today I was busy with work stuff no time to do it.

rhbvkleef:
> I assumed that the drag race will actually be run on master

I see. The thing is that Dave indicated that he wanted his own branch for his own experimentation and code. As such, I proposed to create a separate branch for all user submissions, and that is what I´ve currently implemented.

> You can change the branch name as described above.

Great! Thank you both!

> Unfortunately today I was busy with work stuff no time to do it.

Understood. When, and if you do, might I propose using output properties like `::set-output name=result::`? There´s an excellent idea in the initial CI action created by @flostellbrink in #49.

marghidanu:
@rhbvkleef The reason I went with storing information in the file is that I can't iterate over the outputs programmatically in a script. The outputs in that form make sense only if you want to use them as configuration within the CI. That's not what we want; we want to gather the results and process them in some script.

PR #86, opened 2021-04-06
C version with optimizations

danielspaangberg:
Only store odd primes. Minimize operations by only calculating indices for odd primes. Runs 25% faster on my i5-9400.


danielspaangberg:
This one does 2753 passes on my pi4

ednl:
Ha! Your C version is twice as fast again as mine! Well done :) I got 3140 from yours, best out of 20 on a Pi 4 overclocked to 1.8 GHz (same as the Pi 400). Mine got a little faster too with your native and unroll compiler flags. I'll delete my pull request.

wentz89:
I got a speed up of your implementation from around 10400 passes up to 11900 passes with this simple trick (insert commercial link here :P). You can get rid of some if statements, which enables the processor/os to improve branch prediction.

```
void cleararray(uint8_t *a, unsigned int numbits) {
  size_t numints=numbits>>3;
  numints += 1 * (numbits & 0x7U);
  for (int i=0; i<numints; i++)
    a[i]=0U;
}
```

I think this was the performance gainer, but i did the same with the other if-statements. However even better would be to get rid of the if-statement in findprimes. But i dont know how to handle the break, for the rest you could do something like:

```
for (int i=factorh; i<maxintsh; i++) {
        factor= ((i<<1)+1) * (!primeget(a,i)) + factor * (primeget(a,i));
        // break // dont know how to handle the break without if;
 }
```
**Edit:**
I'm onto something, but it is not working correctly (and beside looking pretty ugly but we want speed not beaty). I could need some help to fix it. Here is the code
```
void findprimes(uint8_t *a, int maxints) {
  int maxintsh=maxints>>1;
  int factor=3;
  int q=(int)sqrt(maxints);
  void *add, *st, *en, *mid;
  st = &&start;
  en = &&end;
  mid = &&middle;
  while (factor<=q) {
    int factorh=factor>>1;
    int i = factorh - 1;
start:
    i++;
    uint8_t n_pg = (primeget(a,i) == 0);
    uint8_t stop = (i == (maxints-1));
    factor = ((i<<1)+1) * (n_pg) + factor * (!n_pg);
    add = (void*)(uintptr_t)((int)(uintptr_t)en * (n_pg) + (int)(uintptr_t)st * (!n_pg && !stop) + (int)(uintptr_t)mid * (stop && !n_pg));
    goto *add;
middle:
    ;
    int f2h=(factor*factor)>>1;
    for (int i=f2h; i<maxintsh; i+=factor)
      primeset(a,i);
    factor+=2;
  }
end:
  ;
}
```
I replaced the for-loop and the if-statement with this weird goto-construct. The start-label is the beginning of the for-loop, the middle-lable right after the for-loop and the end-label replaces the break-statement. It first calculates the factor and then the adress to jump-to. But there is something wrong maybe with my logic or the address calculation. I get a lot more passes but always 500000 from count primes. 

**Edit2:**
Fixed! But im not smarter then gcc -Ofast ...Still ~11900 passes
```
void findprimes(uint8_t *a, int maxints) {
  int maxintsh=maxints>>1;
  int factor=3;
  int q=(int)sqrt(maxints);
  void *add, *mid, *sta;
  mid = &&middle;
  sta = &&start;
  uint8_t n_pg, stop;
  int factorh, i, f2h;
  while (factor<=q) {
    factorh=factor>>1;
    i = factorh - 1;
    f2h;
start:
    i++;
    n_pg = (!primeget(a,i));
    stop = (i == (maxints-1));
    factor = ((i<<1)+1) * (n_pg) + factor * (!n_pg);
    add = (void*)((int)mid * (n_pg) + (int)sta * (!n_pg && !stop) + (int)mid * (stop && !n_pg));
    goto *add;
middle:
    f2h=(factor*factor)>>1;
    for (i=f2h; i<maxintsh; i+=factor)
      primeset(a,i);
    factor+=2;
  }
}
```

danielspaangberg:
Hi!
Nice! I think the problem was the return of primeget is either 0 or a non zero value (1-7), so you need it all to be either 0 or 1 for your algo. The break is a goto anyway, but if I get your solution you try to avoid one goto right? Your solution would compile into a indirect branch, (yes, confirmed looked at the assembly: jmp *%r12 on my intel). I had to change from the (int) casts in your add= line to (intprt_t). But I can still not get your solution to be faster on my i5-9400. I also tried it on  my pi4 (32 bit) and it has about the same runtime. The modification in cleararray cannot change the speed much. That if only executes once per pass. It can also simply be replaced by +1 always, it does not hurt. On what platform do you see the improvements? Anyway, Mike Koss, published a C-version that beats this solution, by including even more special cases, so I think we should try to look at his solution, and see if it can be improved instead!

danielspaangberg:
I'll remove the pull request - look at the version by @mckoss  instead. #104

PR #87, opened 2021-04-06
PrimePowershell

justusiv:
Basics of a powershell script to find primes. I did not include the timer.

### DIFF
--- a/PrimePowershell/PrimePowershell.ps1
+++ b/PrimePowershell/PrimePowershell.ps1
@@ -1,0 +1,69 @@
+﻿function getBit {
+    param([Parameter(Mandatory=$True)][int]$getBitIndex)
+    if ($getBitIndex % 2 -eq 0){
+        return $false
+    }else{
+        return $bitarray[$getBitIndex / 2 + .5]
+    }
+}
+
+function clearBit() {
+    param([Parameter(Mandatory=$True)][int]$clearBitIndex)
+    if (( $clearBitIndex % 2 -eq 0 )){
+		Write-Host "You are setting even bits, which is sub-optimal"
+		return $false
+        }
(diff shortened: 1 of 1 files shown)


pepper-lees:
Couldn't you implement a timer with this? https://docs.microsoft.com/en-us/dotnet/api/system.timespan?view=netframework-4.7.2

```powershell
$StartTime = Get-Date
# do stuff

  # inside your loop
  $Interval = Get-Date - $StartTime 
  # compare Interval.TotalSeconds
```

justusiv:
> Couldn't you implement a timer with this? https://docs.microsoft.com/en-us/dotnet/api/system.timespan?view=netframework-4.7.2
> 
> ```powershell
> $StartTime = Get-Date
> # do stuff
> 
>   # inside your loop
>   $Interval = Get-Date - $StartTime 
>   # compare Interval.TotalSeconds
> ```

Yeah i typically do that for timing things. This code didn't really fit into the category of how many times in 5 seconds so I scrapped it all together. I am no expert just some guy in a basement so i'm sure there are many language optimizations that could help out. I would love to see it. I did leave a tiny bit of debug code in which slows it down i'm sure but not enough to matter.

rbergen:
We now have two PowerShell solutions in the drag-race branch, both of which include a timer. We could add another PowerShell solution, but it would have to comply with the [contributing guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md) and differ from the existing solutions in a meaningful way. This means that the solution does need to be timed, and the PR would need to be opened on the "drag-race" branch, not on "original" (formerly "main").

rbergen:
Closing this due to alternative PowerShell solutions being in place.

PR #88, opened 2021-04-06
C version, only odd factors and primes, RPi4 = 1461 passes

ednl:
About twice as fast because only odd numbers are stored and only odd factors sieved. Best result was 1461 passes on a RPi 4 with RaspiOS 32-bit. All 32-bit integers, so UbuntuPi64 wasn't faster (it was actually a little slower).


ednl:
The other C version I saw in the pull requests is twice as fast *again*, so I'm closing this one here.

PR #90, opened 2021-04-06
Added F# implementation

dmannock:
Ported from the C++ version. Consistent .NET build targets with the C# version


rbergen:
@dmannock We are still happy to work towards including your solution in the drag-race. However, to be able to do so, it would need to comply with the [contribution guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md). That also means that the PR needs to be opened on the drag-race branch of this repo, not original (formerly main).

Do note that another F# solution has been added to the repo after you opened this PR. One reason I would like to include your solution as well, is that I think it is much more interesting than the one we have. As I am the author of that other solution, I **am** allowed to say that. :)

dmannock:
> @dmannock We are still happy to work towards including your solution in the drag-race. However, to be able to do so, it would need to comply with the [contribution guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md). That also means that the PR needs to be opened on the drag-race branch of this repo, not original (formerly main).
> 
> Do note that another F# solution has been added to the repo after you opened this PR. One reason I would like to include your solution as well, is that I think it is much more interesting than the one we have. As I am the author of that other solution, I **am** allowed to say that. :)

Thanks for the kick, that explains why these PR's are no longer accepted. I've converted to drag-race format and opened in a new PR #235. Hopefully, all is well & I've not missed anything.

Ha, can never have too many. What's the situation with additional perf tweaks? These versions & by the looks of it yours also are  'out-of-the-box' whatever the fsharp compiler magic gives you :)

rbergen:
> Thanks for the kick, that explains why these PR's are no longer accepted.

It's taken us a while to figure out how to a) deal with the large number of languages and implementations (we're pushing towards 40 and 100, respectively), b) cater to the unexpectedly creative approaches some came up with, c) structure things to allow automatic and distributed benchmark run execution, while d) not getting in the way of Dave's wishes and his work on his videos.

I think we've now reached a place where we can incrementally develop from, and that's why I'm now starting to gently ask people if they're willing to adapt their submissions to what we've settled down on. 

Thank you for doing the work on yours. I will review it first thing tomorrow morning.

> What's the situation with additional perf tweaks? These versions & by the looks of it yours also are 'out-of-the-box' whatever the fsharp compiler magic gives you :)

Additional perf tweaks are always welcome. :)) To be honest, I wouldn't know where to start; I do most of my development in C#. We do have a C# solution now that contains work of two Microsoft developers with some optimizations I still never knew were possible. Maybe we'll at some point get similar help on F#, too?

rbergen:
Closing due to replacement by #235.

PR #93, opened 2021-04-07
Readme and minor changes to Julia Implementation

dcbi:
As requested, I included a Readme with how to run the code.

Small changes were made to the code: (1) replaced type conversion after floor function with native integer floor division function and (2) added two type declarations in the definition of the prime_sieve struct.

Performance is much better than before.
Passes: 2639, Time: 5.0, Avg: 0.0018946570670708603, Limit: 1000000, Count: 78498, Valid: true

### DIFF
--- a/PrimeSieveJulia/PrimeSieveJulia.jl
+++ b/PrimeSieveJulia/PrimeSieveJulia.jl
@@ -8,10 +8,10 @@ const primeCounts = Dict( 10 => 1,
                           100000000 => 5761455 )
 
 struct prime_sieve
-	sieveSize::Int64
-	rawbits::BitArray{1}
+	sieveSize::Int
+	rawbits::BitVector
 
-	prime_sieve(limit::Int64) = new(  limit, trues( fld(limit+1,2) )  )
+	prime_sieve(limit::Int64) = new(  limit, trues( div(limit+1,2) )  )
 end
 
 function validateResults(sieve::prime_sieve)
@@ -23,19 +23,19 @@ function validateResults(sieve::prime_sieve)
 end
 
 function runSieve!(sieve::prime_sieve)
-	factor::Int64 = 3
-	q::Float64 = sqrt(sieve.sieveSize)
+	factor = 3
+	q = sqrt(sieve.sieveSize)
 
 	while factor <= q
 		for num in factor:sieve.sieveSize
-			if num % 2 == 0 ? false : sieve.rawbits[fld(num+1,2)]
+			if num % 2 == 0 ? false : sieve.rawbits[div(num+1,2)]
 				factor = num
 				break
 			end
 		end
 
 		for num in factor*3:factor*2:sieve.sieveSize
-			sieve.rawbits[fld(num+1,2)] = false
+			sieve.rawbits[div(num+1,2)] = false
 		end
 
 		factor += 2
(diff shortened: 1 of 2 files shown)


PrimeSieveJulia/PrimeSieveJulia.jl
```diff
@@ -23,19 +23,19 @@ function validateResults(sieve::prime_sieve)
 end
 
 function runSieve!(sieve::prime_sieve)
-	factor = 3
-	q = sqrt(sieve.sieveSize)
+	factor::Int64 = 3
```
oscardssmith: This type annotation isn't necessary

PrimeSieveJulia/PrimeSieveJulia.jl
```diff
@@ -23,19 +23,19 @@ function validateResults(sieve::prime_sieve)
...
+	q::Float64 = sqrt(sieve.sieveSize)
```
oscardssmith: ditto

PrimeSieveJulia/PrimeSieveJulia.jl
```diff
@@ -8,10 +8,10 @@ const primeCounts = Dict( 10 => 1,
                           100000000 => 5761455 )
 
 struct prime_sieve
-	sieveSize::Int
-	rawbits::BitArray
+	sieveSize::Int64
+	rawbits::BitArray{1}
 
-	prime_sieve(limit::Int) = new(  limit, trues( floor(Int,(limit+1)/2) )  )
+	prime_sieve(limit::Int64) = new(  limit, trues( fld(limit+1,2) )  )
```
oscardssmith: For positive numbers `div` is the same as `fld` and probably a little more clear.

PrimeSieveJulia/PrimeSieveJulia.jl
```diff
@@ -8,10 +8,10 @@ const primeCounts = Dict( 10 => 1,
                           100000000 => 5761455 )
 
 struct prime_sieve
-	sieveSize::Int
-	rawbits::BitArray
+	sieveSize::Int64
+	rawbits::BitArray{1}
```
oscardssmith: Maybe use `BitVector`? IMO, it's clearer.

PrimeSieveJulia/PrimeSieveJulia.jl
```diff
@@ -8,10 +8,10 @@ const primeCounts = Dict( 10 => 1,
                           100000000 => 5761455 )
 
 struct prime_sieve
-	sieveSize::Int
-	rawbits::BitArray
+	sieveSize::Int64
```
oscardssmith: `Int` is probably better so that it is `Int32` on 32 bit systems

PrimeSieveJulia/PrimeSieveJulia.jl
dcbi: Thanks for all the great suggestions!

PrimeSieveJulia/PrimeSieveJulia.jl
dcbi: Again, thanks for the suggestions! After I included them, the performance went up again a little!

oscardssmith (approved):

rbergen:
I realize this is an additional request on top of one that was made earlier, but if you still want to make these changes, then please open the PR on the drag-race branch.

rbergen:
Closing this after merge of #401; changes in this PR were not added as suggestions to #401. A new PR can of course still be opened  to further improve Julia/solution_1.

PR #94, opened 2021-04-07
Optimisations for the Python code

timsavage:
Took a run at speeding up the Python code (using Python 3.8.5, Ubuntu 20.04, AMD Ryzen 9 3900X):

Original code
Passes: 65, Time: 10.12448023702018, Avg: 0.15576123441569506

Remove dict lookups
Passes: 71, Time: 10.124506268068217, Avg: 0.1425986798319467

Inline GetBit/ClearBit
Passes: 81, Time: 10.08509167295415, Avg: 0.12450730460437222

Switch to integer only division
Passes: 145, Time: 10.047150881029665, Avg: 0.06929069573123907

### DIFF
--- a/PrimeSievePY/PrimePY.py
+++ b/PrimeSievePY/PrimePY.py
@@ -71,18 +71,21 @@ class prime_sieve(object):
 
         factor = 3
         q = sqrt(this.sieveSize)
+        sieveSize = this.sieveSize
+        GetBit = this.GetBit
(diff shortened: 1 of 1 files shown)


timsavage:
Remains entirely with the Python Standard Library, eg does not include Numpy

rbergen:
Closing this. We already have a number of earlier PRs that make the switch to integer division.

PR #95, opened 2021-04-07
Update run.sh

richardgroves:
Added the suffix _PAR to the source file in the build line


rbergen:
The target file has changed in a way that voids the proposed change.

PR #96, opened 2021-04-07
PrimeRust

jmatthewpeters:
So I was playing around with rust and saw your video and it seemed like a really good first step.  This is rough and has a bunch of mostly style warnings. But when run with release optimizations it turned in 12,347 passes on my M1 MacBook Air.  I'm sure people can make it better but it does work and follows the CPP original pretty closely minus I didn't use the ClearBit function.


old-mibmo:
Consider running rustfmt (`cargo fmt`) and making a commit with the changes.

rbergen:
@jmatthewpeters We now have 4 Rust implementations in place in the drag-race branch. If you want to still add yours, please make it comply with the contributing guidelines. That does mean that this PR would need to be opened on the "drag-race" branch, not "original" (formerly "main").

Because of the fact that this would constitute the 5th Rust implementation, we would consider if this implementation differs from the other 4 in a meaningful way, before we decide to merge.

rbergen:
Closing due to 4 alternative Rust solutions being in place.

PR #97, opened 2021-04-07
add Kotlin version

xvolks:
This commit adds a Kotlin version compilable for JVM and native on macOS, Linux and Windows.


rbergen:
@xvolks We are still happy to work towards including your solution in the drag-race; we don't yet have a Kotlin implementation in the repo. However, to be able to do so, it would need to comply with the [contribution guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md). That also means that the PR needs to be opened on the drag-race branch of this repo, not original (formerly main).

rbergen:
I'm closing this after the merge of #309. If you feel anything in this PR should be applied to that solution, you can open a PR that aims to improve the current PrimeKotlin/solution_1.

PR #101, opened 2021-04-08
Added cython implementation

fspigel:
Cython is kind of a bridge between Python and C. It lets you strongly type your python code, and then does its best to compile it into C. In this case, it's a roughly 50% speed upgrade over standard python. To be perfectly honest, I was expecting a bigger gain, so if someone can point out some inefficiencies in my code, I'd be interested to know.


rbergen:
@fspigel We are still happy to work towards including your solution in the drag-race; we don't yet have a Cython implementation in the repo. However, to be able to do so, it would need to comply with the contributing guidelines. That also means that the PR needs to be opened on the drag-race branch of this repo, not original (formerly main).

rbergen:
Closing this after merge of #259.

PR #102, opened 2021-04-08
Adjust all projects to run the same amount of time

Kinematics:
Adjust sources to all use the same duration for the test. Half the languages had 5 seconds, and half 10.  I set everything to 5 seconds based on it being the time used in the C++ project.


Kinematics:
Closing this, as apparently it's on the wrong branch.

PR #104, opened 2021-04-08
A variety of prime sieve variants written in C.

mckoss:
I explored several algorithm options in C - not relying on any libraries so I could
see directly how the compiler would optimize by code.


danielspaangberg:
Hi! Very nice work! Impressive! On my machines I get even higher speed if I compile for the specific CPU by using -march=native -mtune=native. And for some of your algorithms unrolling (-funroll-all-loops) also gives higher speed on my i5-9400 (on the rpi the differences are small and not always beneficial). For 8 of 30 I get:
With -Ofast: 11092 passes
With -Ofast -march=native -mtune=native: 14579 passes.

danielspaangberg:
Also, I got a couple of warning with -Wall (Mostly %d / %ld mixups and a warning for an unused lookup table (isModCandidate).

danielspaangberg:
I incorporated what I did in my previous attempt in your fastest version (my changes are half bit map and predivide in the hot loop). The fastest code as you stated seems to be with 32 bits, so this is what I include below. I tested with uint64_t, uint16_t and uint8_t as well). This change makes it 10% faster on my i5-9400 and 15% faster on my RPI4 with -Ofast -march=native -mtune=native, 17% faster on the RPI with -Ofast -march=native -mtune=native -funroll-all-loops. C shall win over them all! :-)
`#include <stdint.h>
uint32_t inline primeget32(uint32_t *a, unsigned int prime) {
  unsigned int i=prime>>6;
  unsigned int bp=(prime>>1) & 0x1FU;
  return a[i]&((uint32_t)1<<bp);
}

void inline primeset32(uint32_t *a, unsigned int prime) {
  unsigned int i=prime>>5;
  unsigned int bp=prime & 0x1FU;
  a[i]|=((uint32_t)1<<bp);
}

#define indexOf32(n) n / 32
#define allocOf32(n) indexOf32(n) + 1
int countPrimes8of30_32_bit_odd(int maxNumber, int fNeedCount) {
  int maxNumberh=maxNumber>>1;
  // Starts off zero-initialized.
   uint32_t *buffer = calloc(allocOf32(maxNumberh), sizeof(uint32_t));
   unsigned int maxFactor = sqrt(maxNumber) + 1;
   // Only numbers congruent to candidates mod 30 can be prime.
   unsigned int candidates[8] = {1, 7, 11, 13, 17, 19, 23, 29};

   // Build a stepping map.
   unsigned int steps[8];
   for (int i = 0; i < 8; i++) {
      steps[i] = (candidates[(i + 1) % 8] - candidates[i] + 30) % 30;
   }

   // We get 2, 3 and 5 for "free" since we ignore their multiples.
   int count = 3;
   unsigned int p;

   // First step is from 7 to 11 (steps[1]).
   unsigned int step = 1;

   // Look for next prime
   for (p = 7; p <= maxFactor; p += steps[step], step = (step + 1) % 8) {
      // A 1 bit means it's composite - keep searching.
      if (primeget32(buffer,p))
        continue;
     
      count++;
      // printf("%d, ", p);
      // The following loop is the hotspot for this algorithm.
      // No need to start less than p^2 since all those
      // multiples have already been marked.
      for (unsigned int m = (p*p)>>1; m < maxNumberh; m += p)
 	 primeset32(buffer,m);
   }

   // Count all the remaining primes above sqrt(maxNumber)
   if (fNeedCount) {
      for (; p < maxNumber; p += steps[step], step = (step + 1) % 8) {
	 if (primeget32(buffer,p))
	   continue;
         count++;
         // printf("%d, ", p);
      }
   }

   free(buffer);

   return count;
}
`
[sieve32.zip](https://github.com/davepl/Primes/files/6286274/sieve32.zip)

mckoss:
@danielspaangberg Those compiler flags are definitely providing some boost.  My best 8 of 32 is now at 17,313 passes (0.289 ms) - compared to about 0.387 ms with just -Ofast.  Thanks for letting me know about them.

I was a bit confused by why the index calculate in primeget32 was dividing by 64, while primeset32 was dividing by 32.  I see now that you did a nice factoring of getting the /2 out of the loop index.

Is there a speed savings of only storing bits for the odd values (1/2 bitmap) once you take your trick of halving the indices of the bit-setting pass into account?

I assume there is no (speed) benefit to using inline functions vs the macros I use in my code for these calculations?

ednl:
This is great. However, with the wheels it is now a completely different algorithm. I admit that I did it too by skipping even multiples, which is a wheel of just 2 :) So, it is no longer a fair comparison between languages or compilers.

danielspaangberg:
Nice! There are some speed differences due to cache, since half the amount of memory is used. In my original code I tested larger sizes than the first million too, where there was more apparent differences.

I think the compiler should be smart enough with your macros, it's just me who like to see which bits I am manipulating directly, so that was the code I was using in my version. I just spent 30 minutes looking at the assembler listing differences for the generated codes, and in the inner loop the differences are quite small. There is storage size differences (orq vs. orl etc, and the number of shifted bits). One wierd thing is that for your code gcc (8.3) compiles the outer loop test to bit test (btq), while my code compiles to a shift followed by a test (shlx/testl). Yet I get 5% faster for my 64 bit code and 10% for the 32 bit code. Maybe this is only cache (L1D is 32k, and the data is 122kB vs 61kB). I changed to primes within first 2 million, and this changes the differences to 15% and 20%, so it seems like cache.
Full code tested:
[sieve_storage.zip](https://github.com/davepl/Primes/files/6286745/sieve_storage.zip)

Also @ednl , I agree that the algorithm has changed. But I think the nice thing with @mckoss implementation is that there are several different algorithms to compare to.

mckoss:
@ednl - Yeah, this is no longer focused on language differences but algorithmic (and optimization) improvements.

mckoss:
I hand tuned some more and got quite a bit of speed improvements in the fastest algorithm.  The code is now filled with (neat) tricks and special cases to squeeze more performance out.  Whee! @danielspaangberg

danielspaangberg:
Now it starts to be complex, but fast. I like it. I think that this is one advantage of a low level language like C, where you actually see what's going on and how to squeeze out the last bit of performance. I tried your code on both my intel and arm (RPI and a chromebook duet) and they are indeed substantially faster. On my 64 bit system RPI (1.9GHz) I get a speedup of 45% compared to the old code (now doing 6410 passes). On one of my RPIs I only have a 32 bit system and then using a 32 bit word is faster than 64 bit (by 20%). Again, nice!

ednl:
Are the speedups in sync with the tables at https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes#Computational_analysis ? (What exactly the columns represent takes some careful reading...)

mckoss:
@ednl Thanks for pointing out that wikipedia page - I hadn't seen that (I guess you can say I "reinvented the wheel" :-) ).

The last optimizations had more to do with eliminating redundant calculation of bit mask patterns by pre-calculating them and then applying them in the right order in the inner loop for the composite marking loop.

Kinematics:
Cool stuff. I was pondering a mod 6 approach, and then saw someone had already done it. Very nice performance.

_However_ ... I found an improvement when I was writing my own version (in C#) that I think could be applied to yours as well.

This basically matches your version:
```
            int q = (int)Math.Sqrt(SieveSize);

            for (int factor = 5, inc = 2; factor <= q; factor += inc, inc = 6 - inc)
            {
                if (GetBit(ref bits, factor))
                {
                    for (int num = factor * factor; num <= SieveSize; num += factor * 2)
                    {
                        ClearBit(ref bits, num);
                    }
                }
            }
```

However your inner loop is making extra checks it doesn't need to. You can iterate the inner loop in the same fashion as the outer loop, as long as you start with the proper increment value. (For example, when testing 5, you go: 25, 35, 55, 65, 85, etc, because 45, 75, etc can be discarded as multiples of 3.)

```
            int q = (int)Math.Sqrt(SieveSize);

            for (int factor = 5, inc = 2; factor <= q; factor += inc, inc = 6 - inc)
            {
                if (GetBit(ref bits, factor))
                {
                    for (int num = factor * factor, nInc = inc; num <= SieveSize; num += factor * nInc, nInc = 6 - nInc)
                    {
                        ClearBit(ref bits, num);
                    }
                }
            }
```

I see about a 20% speed improvement with this change, which is fairly substantial. Using `factor * 2` in the loop resulted in calling the `ClearBit` function 644402 times (with a sieve size of 1,000,000).  Using the mod 6 increment in the loop resulted in calling the `ClearBit` function 429626 times.  A 33% reduction in what is basically the most expensive part of the algorithm.

I'm happy with a C# implementation hitting almost 10,000 loops in 5 seconds (hit 9850), but the C version will continue to run circles around it. Hope this is useful to you.

Your 8 of 30 version is too convoluted for me to guess whether the same improvement can be made in there.

mckoss:
@danielspaangberg - and finally added a parallel (multi-threaded) version here too. See the README for the summary on performance.

danielspaangberg:
Looks good. Your byte implementation actually runs each sieve faster in parallel. Having looked at the other parallel implementation, it just runs a separate sieve in each thread (embarrassingly parallel style). So if it is just the numbers that matters, that could be done with the bit versions as well?

danielspaangberg:
I wrote an OpenMP embarrassingly parallel version of your previous code. Compile with -Ofast -fopenmp -march=native -mtune=native -funroll-all-loops 
I had to change the time measurement to measure wall clock instead of CPU time too.
Gets 131564 passes for 8 of 30 on 6 threads on my i5-9400.
[sieve_par.zip](https://github.com/davepl/Primes/files/6310442/sieve_par.zip)

mckoss:
> Looks good. Your byte implementation actually runs each sieve faster in parallel. Having looked at the other parallel implementation, it just runs a separate sieve in each thread (embarrassingly parallel style). So if it is just the numbers that matters, that could be done with the bit versions as well?

I don't know how to do the bit mask version without creating locks for each word before setting the bit. Otherwise competing threads could accidentally erase the bits set by parallel threads.

danielspaangberg:
@mckoss - no sure. I guess it could be done with locks, but it'll likely perform slow as ... But I just looked at Dave's latest vid as well, and he seems to be just running the threads completely separately. Look at the CPP_PAR version. There's no threads updating the same memory, so no need to creating locks. Hence embarrassingly parallel.

mckoss:
> @mckoss - no sure. I guess it could be done with locks, but it'll likely perform slow as ... But I just looked at Dave's latest vid as well, and he seems to be just running the threads completely separately. Look at the CPP_PAR version. There's no threads updating the same memory, so no need to creating locks. Hence embarrassingly parallel.

Yeah - I wouldn't really count that as very interesting (hence, the embarrassing part).  I guess it measures how much concurrent independent work a processor/os can do - and I suppose there is some unshared resources like processor cache that will interfere between processors.  I thought it much more interesting to see how quickly a single problem could be solved by using parallelism.

Kinematics:
I got a parallel version to work (finally) by breaking up each base prime's (below sqrt(sieveSize)) ClearBit run into segments of the overall sieve, rather than parallelizing all the primes.  That way there would never be any memory conflicts.

On the 1,000,000 target, it peaked at about 3-4 threads, but never came close to the linear version (it ran about 70% of the linear version's speed).

On 10,000,000, however, it peaked at 32 threads at 15% better than the linear version, having surpassed the linear version at about 5 threads.

On the 100,000,000 target, it flattened out at about 16 threads (staying roughly the same up to 96 threads, though with a soft peak at around 32 threads), and was 50% faster than the linear version.

Against 1,000,000,000, the linear version completed 8 passes in 22.5 seconds (2818 ms per pass), while the parallel version (32 threads) completed 17 passes in 20 seconds (1178 ms per pass) — 2.4x faster.

danielspaangberg:
Tried a modification as well with only doing the masking in parallel. 
On 100,000,000 I get almost 50% on thread parallel 1 of 2 on 6 threads compared to single thread 8 of 30, making sure I only touch bits separated by 8 bits (i.e. bytes or larger). But I am not sure that is still correct, since I seem to recall that one has to stay one cache line apart to guarantee atomicity? (I did try to test without the bit distance >=8, and the code ran >200 passes, and I have not been able to detect a bit clash, but that is just stats I guess). Anyway,
[sieve_par.zip](https://github.com/davepl/Primes/files/6313567/sieve_par.zip)

Calculate primes up to 100000000.
Timer resolution: 1000000 ticks per second.
Word size: 64 bits.

      Byte-map - 1 of 2 tested:    12 passes completed in 5 seconds (416.667 ms per pass).
       Bit-map - 1 of 2 tested:    47 passes completed in 5 seconds (106.383 ms per pass).
       Bit-map - 2 of 6 tested:    53 passes completed in 5 seconds (94.340 ms per pass).
      Bit-map - 8 of 30 tested:    79 passes completed in 5 seconds (63.291 ms per pass).
                   1/3 Bit-map:    46 passes completed in 5 seconds (108.696 ms per pass).
Thread parallel Bit-map - 1 of 2 tested:   116 passes completed in 5 seconds (43.103 ms per pass).
Embarrassingly Parallel Byte-map - 1 of 2 tested:    18 passes completed in 5 seconds (277.778 ms per pass). (6 Threads)
Embarrassingly Parallel Bit-map - 1 of 2 tested:    50 passes completed in 5 seconds (100.000 ms per pass). (6 Threads)
Embarrassingly Parallel Bit-map - 2 of 6 tested:    49 passes completed in 5 seconds (102.041 ms per pass). (6 Threads)
Embarrassingly Parallel Bit-map - 8 of 30 tested:    48 passes completed in 5 seconds (104.167 ms per pass). (6 Threads)
Embarrassingly Parallel 1/3 Bit-map:    85 passes completed in 5 seconds (58.824 ms per pass). (6 Threads)

danielspaangberg:
And, yes even the embarrssingly parallel are slower. Running out of cache...

Kinematics:
So, I was trying to write the [8 of 30] algorithm in C#, looking at what applies and doesn't apply in that language.  While doing so, I came across an oddity.

I initially wrote the for loop as:

    for (int num = factor * factor, nInc = steps[iStep]; num <= SieveSize; num += factor * nInc, iStep = (iStep + 1) % 8, nInc = steps[iStep])

And that ran at 7800 loops.  However I noticed @mckoss using a shorter version:

    for (int num = factor * factor; num <= SieveSize; num += factor * steps[iStep], iStep = (iStep + 1) % 8)

And switched to that for simplicity.  However the result then runs at only 7400 loops (~5% loss).

I'm guessing that when it's broken out as a separate variable, the update from the array can be tossed out as a separate processor action while it does the multiply on the known items, whereas when it's consolidated the array access is inline with the multiply, delaying the processor instruction. Or something vaguely along those lines.

This was written for all the for loops, but only the inner for loop is run enough for it to really matter.

I don't know whether this is specific to C#, but it's something for you to check on in your C version.

danielspaangberg:
@Kinematics can you point to your code? I tried to find it. Do you have a pull request or is your code somewhere else?

danielspaangberg:
@mckoss I added thread parallel code for each sieve too. I tried atomic ops when needed, but that was indeed slower (not as slow as I expected though). To do the full parallelization of the bit maps requires some trickery for the bits to ensure two threads do not update the same word. There is still quite a bit(!) of room for improvement, since I did not manage to parallelize your fast masking loop. I did manage to gain a little speed even on primes up to 1 million (27133 passes, or about 16% increase when using 6 threads, ), but the changes for larger sieves are more apparent. For 10 million I get 3608 passes compared to fastest single thread (168% increase in speed). This can be compared to the embarrassingly (uninteresting) parallel run of 6859 passes. The most interesting tested results here, the rest are in the attached zip.
Calculate primes up to 10000000.
Timer resolution: 1000000 ticks per second.
Word size: 64 bits.

                                              Byte-map - 1 of 2 tested:    282 passes completed in 5 seconds (17.730 ms per pass).
                                               Bit-map - 1 of 2 tested:    654 passes completed in 5 seconds (7.645 ms per pass).
                                               Bit-map - 2 of 6 tested:    814 passes completed in 5 seconds (6.143 ms per pass).
                                              Bit-map - 8 of 30 tested:   1346 passes completed in 5 seconds (3.715 ms per pass).
                                                           1/3 Bit-map:    553 passes completed in 5 seconds (9.042 ms per pass).
                                 Bit-map no fast mask - 8 of 30 tested:   1015 passes completed in 5 seconds (4.926 ms per pass).
 Thread parallel Bit-map no fast mask no word overlap - 8 of 30 tested:   3608 passes completed in 5 seconds (1.386 ms per pass).
                 Thread parallel Bit-map no fast mask - 8 of 30 tested:   1692 passes completed in 5 seconds (2.955 ms per pass).
       Thread parallel atomic or Bit-map no fast mask - 8 of 30 tested:    371 passes completed in 5 seconds (13.477 ms per pass).
               Thread parallel Bit-map no byte overlap - 1 of 2 tested:   2054 passes completed in 5 seconds (2.434 ms per pass).
                               Thread parallel Bit-map - 1 of 2 tested:   1106 passes completed in 5 seconds (4.521 ms per pass).
                     Thread parallel atomic or Bit-map - 1 of 2 tested:    891 passes completed in 5 seconds (5.612 ms per pass).
                      Embarrassingly Parallel Byte-map - 1 of 2 tested:    185 passes completed in 5 seconds (27.027 ms per pass). (6 Threads)
                       Embarrassingly Parallel Bit-map - 1 of 2 tested:   3546 passes completed in 5 seconds (1.410 ms per pass). (6 Threads)
                       Embarrassingly Parallel Bit-map - 2 of 6 tested:   4451 passes completed in 5 seconds (1.123 ms per pass). (6 Threads)
                      Embarrassingly Parallel Bit-map - 8 of 30 tested:   6859 passes completed in 5 seconds (0.729 ms per pass). (6 Threads)
                                   Embarrassingly Parallel 1/3 Bit-map:   3138 passes completed in 5 seconds (1.593 ms per pass). (6 Threads)
[parallel_sieves_with_results.zip](https://github.com/davepl/Primes/files/6317762/parallel_sieves_with_results.zip)

Kinematics:
> @Kinematics can you point to your code? I tried to find it. Do you have a pull request or is your code somewhere else?

OK, I pushed my code and put it in pull request #135, or [my repo](https://github.com/Kinematics/Primes/tree/drag-race-kinematics/PrimeCSharp/stl-faithful/Kinematics)

danielspaangberg:
@Kinematics The C algorithms might have changed since you checked last time. The inner loop no longer needs to refer to the steps array.

Kinematics:
> @Kinematics The C algorithms might have changed since you checked last time. The inner loop no longer needs to refer to the steps array.

Ah, OK, right. I did the first implementation of the 8 of 30 algorithm straight, the same way as the 2 of 6 algorithm. I'm in the middle of implementing the masked version, but was trying to make sure I understood it, to make sure I knew it was working correctly. The earlier comment was on the straight version, and of course doesn't apply to the masked version.

Would still be curious about whether that performance hit applies in C as well, though.

danielspaangberg:
> > @Kinematics The C algorithms might have changed since you checked last time. The inner loop no longer needs to refer to the steps array.
> 
> Ah, OK, right. I did the first implementation of the 8 of 30 algorithm straight, the same way as the 2 of 6 algorithm. I'm in the middle of implementing the masked version, but was trying to make sure I understood it, to make sure I knew it was working correctly. The earlier comment was on the straight version, and of course doesn't apply to the masked version.
> 
> Would still be curious about whether that performance hit applies in C as well, though.

I reimplemented that inner loop with your two variants. And the one where you have the nInc in the loop initializer runs 10%-15% faster. The load is earlier, so maybe the CPU can overlap the iterations better that way. I looked at the generated assembly and without -funroll-all-loops it somewhat made sense. With -funroll-all-loops, not so much. I haven't used vtune or performance counters in linux in 7-8 years, so it would take me way to long to understand this difference right now. The newer implementations run 1.7 times, and 2.6 times faster than your fast version though.

Kinematics:
So I'm trying to understand the 8 of 30 algorithm, so that I can implement it properly. And while writing out all the details of the issues I was having,  I finally figured out the problem.

The issue was that the masking method would never mark 7 or 11 as prime.  Instead, because of the looparound, they would always be marked as composite.

However somehow you ended up with the correct count, while I was always 2 short.  I finally figured out why: You don't keep the buffer as part of the class, and count the found primes after the fact.  Instead, you count the primes as you process them.  Even if the masking fails to mark 7 and 11 correctly, you've already counted them as part of the basic loop.

(Note: 13 and higher don't have the same problem, because their first composite value is always at an offset of at least 1, so it doesn't overwrite the original number.)

This is written as a function that returns an int, while I've been writing mine as self-contained classes that has one function to run the sieve calculation, and other functions to count and extract the primes from the buffer that's part of the class.

While I can see that it might make more sense to write it that way, the drag race design goes for the stored version.  So I'll just have to account for those in the counting and generating portions of the program.

danielspaangberg:
@mckoss and @Kinematics - Now I thought, why stop at 8 of 30? So I wrote a generator for the candidates. Now there is also 48 of 210, 480 of 2310 and 5760 of 30030. I haven't finished with all parallelization and masking speeding up, so I do not have a final number for how much faster it might get, but wanted to share the code in this state. The 5760 of 30030 version runs almost 50% faster than the 8 of 30 version with the same inner loop. (It ought to be 40% faster, so not sure why it is running faster 30030*8/5760/30). 
[higher_sieve.zip](https://github.com/davepl/Primes/files/6328868/higher_sieve.zip)

danielspaangberg:
Also, @Kinematics Replacing your inner loop with this is much faster (but still not as fast as just setting everything):
     ` int istep=step;
      int ninc=steps[istep];
      for (int num=p*p; num<=maxNumber; ) {
	primesetw(buffer,num>>1); // Or whatever you use to set the bits in the buffer
	num+=p*ninc;
	if (++istep==8)
	  istep=0;
	ninc=steps[istep];
      }`

mckoss:
> @mckoss and @Kinematics - Now I thought, why stop at 8 of 30? So I wrote a generator for the candidates. Now there is also 48 of 210, 480 of 2310 and 5760 of 30030. I haven't finished with all parallelization and masking speeding up, so I do not have a final number for how much faster it might get, but wanted to share the code in this state. The 5760 of 30030 version runs almost 50% faster than the 8 of 30 version with the same inner loop. (It ought to be 40% faster, so not sure why it is running faster 30030*8/5760/30). 
> [higher_sieve.zip](https://github.com/davepl/Primes/files/6328868/higher_sieve.zip)
>
wow.  Each prime, p, added to the exclusion set reduces the work needed by 1/p.  So I expect diminishing returns.    But the gains compound.  So I'd expect your p=13 version to be 28% faster than the p=5 version.

Kinematics:
> Also, @Kinematics Replacing your inner loop with this is much faster (but still not as fast as just setting everything):
> ` int istep=step; int ninc=steps[istep]; for (int num=p*p; num<=maxNumber; ) { primesetw(buffer,num>>1); // Or whatever you use to set the bits in the buffer num+=p*ninc; if (++istep==8) istep=0; ninc=steps[istep]; }`

Wow. Went from 7800 to 12000.  Looks like it's just swapping this line:
```
iStep = (iStep + 1) % 8;
```
to this:
```
if (++iStep == 8)
    iStep = 0;
```
that gives a 50% speedup.   Though I guess division always has been expensive.

Also, it's now finally faster than the mod 6 version (9800).

danielspaangberg:
So, interestingly when testing fewer prime candidates, code that takes care of only writing bits that will be read starts to get quicker. So for the primes below 1 million, the fastest is 5760 of 30030, doing 28714 passes / 5 seconds, compared to 24000 passes / for the code which quickly sets the bits. Setting all bits in parallel on 4 threads gets the 5760 of 30030 code to 31084 passes. The fastest 8 of 30 code is at 26900 passes (for thread parallel on 4 threads), so the increase on up to a million is not really that impressive, as I initially thought, The improvements are larger for 10, 100 and 1000 million though. 
[sieve.zip](https://github.com/davepl/Primes/files/6329605/sieve.zip)

danielspaangberg:
> > Also, @Kinematics Replacing your inner loop with this is much faster (but still not as fast as just setting everything):
> > ` int istep=step; int ninc=steps[istep]; for (int num=p*p; num<=maxNumber; ) { primesetw(buffer,num>>1); // Or whatever you use to set the bits in the buffer num+=p*ninc; if (++istep==8) istep=0; ninc=steps[istep]; }`
> 
> Wow. Went from 7800 to 12000. Looks like it's just swapping this line:
> 
> ```
> iStep = (iStep + 1) % 8;
> ```
> 
> to this:
> 
> ```
> if (++iStep == 8)
>     iStep = 0;
> ```
> 
> that gives a 50% speedup. Though I guess division always has been expensive.
> 
> Also, it's now finally faster than the mod 6 version (9800).

I did not expect it to be that much quicker, and I have not analyzed the differences in detail. It should be possible to convert the % 8 to just a simple mask. a%8 = a & 7, so I am not sure what is going on, but speed is always nice.

Kinematics:
> I did not expect it to be that much quicker, and I have not analyzed the differences in detail. It should be possible to convert the % 8 to just a simple mask. a%8 = a & 7, so I am not sure what is going on, but speed is always nice.

A quick check has the &7 version running at 12200, while the =0 is at 12500.  Minuscule difference, but slightly worse, so I'll keep the current version.

danielspaangberg:
@mckoss For fun - run the code on the web. I made a javascript version and also compiled the C to webassembly. On firefox this runs at about 60% of native speed. So about 17000 iterations in 5 seconds. [https://www.papdiger.se/primesieve/primesieve.html](https://www.papdiger.se/primesieve/primesieve.html)
[html5_sieve.zip](https://github.com/davepl/Primes/files/6330614/html5_sieve.zip)
BTW, did you consider making your code into a drag-race contribution?

mckoss:
> @mckoss For fun - run the code on the web. I made a javascript version and also compiled the C to webassembly. On firefox this runs at about 60% of native speed. So about 17000 iterations in 5 seconds. https://www.papdiger.se/primesieve/primesieve.html
> [html5_sieve.zip](https://github.com/davepl/Primes/files/6330614/html5_sieve.zip)
> BTW, did you consider making your code into a drag-race contribution?

I tried the Emscripten thing too.  Seemed to be about 1/2 speed to native.

I hadn't noticed the contribution guidelines for the drag race entry.  Sure, I'll do that with the 8 of 30 algorithm.

mckoss:
> > > Also, @Kinematics Replacing your inner loop with this is much faster (but still not as fast as just setting everything):
> > > ` int istep=step; int ninc=steps[istep]; for (int num=p*p; num<=maxNumber; ) { primesetw(buffer,num>>1); // Or whatever you use to set the bits in the buffer num+=p*ninc; if (++istep==8) istep=0; ninc=steps[istep]; }`
> > 
> > 
> > Wow. Went from 7800 to 12000. Looks like it's just swapping this line:
> > ```
> > iStep = (iStep + 1) % 8;
> > ```
> > 
> > 
> > to this:
> > ```
> > if (++iStep == 8)
> >     iStep = 0;
> > ```
> > 
> > 
> > that gives a 50% speedup. Though I guess division always has been expensive.
> > Also, it's now finally faster than the mod 6 version (9800).
> 
> I did not expect it to be that much quicker, and I have not analyzed the differences in detail. It should be possible to convert the % 8 to just a simple mask. a%8 = a & 7, so I am not sure what is going on, but speed is always nice.

I wonder if the %8 is not optimized because it's and int value instead of unsigned?

danielspaangberg:
> > > > Also, @Kinematics Replacing your inner loop with this is much faster (but still not as fast as just setting everything):
> > > > ` int istep=step; int ninc=steps[istep]; for (int num=p*p; num<=maxNumber; ) { primesetw(buffer,num>>1); // Or whatever you use to set the bits in the buffer num+=p*ninc; if (++istep==8) istep=0; ninc=steps[istep]; }`
> > > 
> > > 
> > > Wow. Went from 7800 to 12000. Looks like it's just swapping this line:
> > > ```
> > > iStep = (iStep + 1) % 8;
> > > ```
> > > 
> > > 
> > > to this:
> > > ```
> > > if (++iStep == 8)
> > >     iStep = 0;
> > > ```
> > > 
> > > 
> > > that gives a 50% speedup. Though I guess division always has been expensive.
> > > Also, it's now finally faster than the mod 6 version (9800).
> > 
> > I did not expect it to be that much quicker, and I have not analyzed the differences in detail. It should be possible to convert the % 8 to just a simple mask. a%8 = a & 7, so I am not sure what is going on, but speed is always nice.
> 
> I wonder if the %8 is not optimized because it's and int value instead of unsigned?

Not sure. I thought it was, but I needed modulus of 48, which is not as easy.

mckoss:
> @mckoss For fun - run the code on the web. I made a javascript version and also compiled the C to webassembly. On firefox this runs at about 60% of native speed. So about 17000 iterations in 5 seconds. https://www.papdiger.se/primesieve/primesieve.html
> [html5_sieve.zip](https://github.com/davepl/Primes/files/6330614/html5_sieve.zip)
> BTW, did you consider making your code into a drag-race contribution?

https://github.com/davepl/Primes/pull/152

rbergen:
@mckoss In light of #152, would it be fair to assume that you will not convert this PR to one on drag-race? In that case, I would like to close this as we will not be able to merge it in "original".

rbergen:
Closing this in line with my previous comment.

mckoss:
> Closing this in line with my previous comment.

That's fine.  I moved the best version to the drag-race pr.

PR #106, opened 2021-04-08
Added Rust as a language

StappsWorld:
My attempt at the benchmark written in Rust. Probably not the best way to write it, but I had fun :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 PrimeCrystal/primes
+PrimeRust/target
+PrimeRust/Cargo.lock
--- a/PrimeRust/.gitignore
+++ b/PrimeRust/.gitignore
@@ -1,0 +1,2 @@
+./target
+Cargo.lock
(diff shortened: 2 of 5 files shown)


micwoj92:
I think this is already like 3rd rust implementation lol
also should be targeted at drag-race branch

rbergen:
@StappsWorld We now have 4 Rust implementations in place in the drag-race branch. If you want to still add yours, please make it comply with the [contributing guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md). As @micwoj92 already mentioned, that also means that this PR would need to be opened on the "drag-race" branch, not "original" (formerly "main").

Because of the fact that this would constitute the 5th Rust implementation, we would consider if this implementation differs from the other 4 in a meaningful way before we decide to merge.

rbergen:
Closing this due to alternative Rust solutions being in place.

PR #107, opened 2021-04-09
add Rustlang drag-race version

joshallen64:
I saw there was 3 or so Rust implementations, but I thought I'd throw mine in here to see how it compares to the others.


rbergen:
@joshallen64 We're still happy to consider this solution, but some work will be required to make it possible to add it.

If you still want this solution to be included in the drag race then please:
- Read the [contributing guidelines](https://github.com/davepl/Primes/blob/drag-race/CONTRIBUTING.md) and structure your solution accordingly.
- Consider that there is an open PR that targets the PrimeRust/solution_3 directory. To prevent a merge conflict later, please use PrimeRust/solution_4, for yours.

Due to the large amount of changes to the `drag-race` branch since you opened this PR, it is probably best to base your next commit on the current version of this branch.

joshallen64:
@rbergen Thanks, it looks like a lot has changed!  I'll update it to the new contribution guidelines and retarget for solution_4.

joshallen64:
@rbergen  I believe I have updated everything to conform with the contribution guidelines.  Let me know if there are any issues.

PrimeRust/solution_4/PrimeRust.rs
```diff
@@ -0,0 +1,156 @@
...
+                break
+            }
+        }
+
+        if show_results {
+            println!("");
+        }
+        println!("joshallen64;{};{:?};1",
+                passes,
+                duration
```
rbergen: This prints the duration including an "s" at the end, My guess is that you'll have to explicitly extract a numerical version of the duration (second + fraction) out of the Duration struct.

joshallen64: I corrected the output to remove the "s".

joshallen64:
@rbergen I had to fix the docker file for the correct rust 1.51.  It broke when I updated it for the correct output format.  The method to get a float format for seconds in the duration class didn’t exist in rust 1.31.

rbergen (approved):

PR #112, opened 2021-04-09
Added .NET 5.0 project with the new top-level statements (and more)

syntax-tm:
Hi Dave!

Wanted to say that I am a fan of your channel and despite being a developer myself, I don't usually like watching stuff like that for entertainment. But, your channel has been the one exception to that and your ability to have a fluid conversation covering technical and interesting information makes it feel like I'm at work listening to my co-workers stories and that's (by far) where I've learned the most is from other people. One last thing before I ramble about the actual changes, your "just in it for the subs and likes" is fantastic. I have no idea how people can shamelessly do the things they do nowadays and most YouTubers make me cringe but your just wanting to do it for personal reasons makes a huge difference. I'm sure you already know that, but in a world with sponsors ruining every video and adblockers basically being a requirement it's really refreshing. I haven't seen that attitude on YouTube in a long, long time. Thanks for everything. 👍

Added new .NET 5.0 test project using the new Top-level Statements feature from C# 9.0 as well as other optimizations. Changed the `printResults` in `PrimeSieveCS/PrimeCS.cs` file to match the new one (since it doesn't affect performance). Added conditional compilation to the `ClearBit` function to not perform the check unless you're targeting Debug. My logic was when running the test you should already have the algorithm optimized and the constant validation is unnecessary but, you can still target Debug to leave it in. Last but not least the `Test-CSProjects.ps1` script is just a quick way to run multiple passes of all .NET projects for an easy comparison. It will search for all `*.csproj` files so theoretically you could make multiple `*.csproj` files with different runtime, configuration, optimization, etc. Since Powershell is cross-platform it can still be used on other operating systems.

I don't expect this to be merged, mostly was just curious to see if .NET 5 was that much better and was trying to make everything as optimized as possible for my own enjoyment. It's a fair comparison, especially with the conditional compilation, but there's a lot of little things like moving the starting `DateTime` declaration to right before starting the first run (as opposed to letting time pass declaring other globally scoped variables). Also changed it to not run the same calculations every time for the array size and stuff like that. Having a static `Init` function in that class to initialize the static members so they can only be set that one time makes a difference when you're running millions of iterations. It's not how I'd normally set it up, but for being a single file test it's fine (or so I am telling myself).

Feel free to take whatever from this you want, or none of it, or don't even read all of this. I just wanted to mess around with it personally so there's no expectations of anything on my end.

Here's the output from running the '`Test-CSProjects.ps1`' file:

```
[Running] powershell -ExecutionPolicy ByPass -File "c:\projects\Primes\Test-CSProjects.ps1"
Running 'PrimeSieveCS.csproj'.....
Running 'PrimeSieveCS_net5.0.csproj'.....

project             run passes time        avg        limit   count valid
-------             --- ------ ----        ---        -----   ----- -----
PrimeSieveCS          1 1456   10.00763260 0.00687337 1000000 78498 True 
PrimeSieveCS          2 1432   10.00324410 0.00698551 1000000 78498 True 
PrimeSieveCS          3 1510   10.00325290 0.00662467 1000000 78498 True 
PrimeSieveCS          4 1526   10.00189520 0.00655432 1000000 78498 True 
PrimeSieveCS          5 1580   10.00400600 0.00633165 1000000 78498 True 
PrimeSieveCS_net5.0   1 5824   10.00045270 0.00171711 1000000 78498 True 
PrimeSieveCS_net5.0   2 5238   10.00112630 0.00190934 1000000 78498 True 
PrimeSieveCS_net5.0   3 5449   10.00094680 0.00183537 1000000 78498 True 
PrimeSieveCS_net5.0   4 5773   10.00148790 0.00173246 1000000 78498 True 
PrimeSieveCS_net5.0   5 5678   10.00083980 0.00176133 1000000 78498 True 

[Done] exited with code=0 in 122.615 seconds
```

Thanks,
Trey

### DIFF
--- a/PrimeSieveCS/PrimeCS.cs
+++ b/PrimeSieveCS/PrimeCS.cs
@@ -112,7 +112,8 @@ namespace PrimeSieveCS
                 }
                 if (showResults)
                     Console.WriteLine("");
-                Console.WriteLine("Passes: " + passes + ", Time: " + duration + ", Avg: " + (duration / passes) + ", Limit: " + this.sieveSize + ", Count: " + count + ", Valid: " + validateResults());
+                Console.WriteLine($"Passes: {passes}, Time: {duration:n8}, Avg: {(duration / passes):n8}, Limit: {this.sieveSize}, Count: {count}, Valid: {validateResults()}");
+
             }
         }
 
--- a/PrimeSieveCS_net5.0/PrimeCS.cs
+++ b/PrimeSieveCS_net5.0/PrimeCS.cs
@@ -1,0 +1,151 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+
+prime_sieve.Init();
+
+prime_sieve sieve = null;
+double elapsed;
+var passes = 0;
+var tStart = DateTime.UtcNow;
+
+do
+{
+    sieve = new prime_sieve();
+    sieve.runSieve();
+    passes++;
+
+    elapsed = (DateTime.UtcNow - tStart).TotalSeconds;
+} while (elapsed < prime_sieve.DURATION);
+
+sieve?.printResults(elapsed, passes);
+
+class prime_sieve
+{
+    public const int SIEVE_SIZE = 1000000;
+    public const int DURATION = 10;
+    public static int BitArraySize;
+    public static int SieveSizeSqrt;
+
+    public static void Init()
+    {
+        BitArraySize = (int) ((SIEVE_SIZE + 1) / 2);
+        SieveSizeSqrt = (int) Math.Sqrt(SIEVE_SIZE);
+    }
+
+    private readonly BitArray bitArray;
+    private readonly Dictionary<int, int> myDict = new Dictionary<int, int> 
+    { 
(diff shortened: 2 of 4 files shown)


syntax-tm:
Also, if you're looking for a new video idea, it would be cool to see what, if any, design patterns you actually like or find the most useful.

I don't like the concept of "this is how you do this" as the single solution for all languages/environments/etc. but, I find design patterns and things like the SOLIDD principles beneficial just in a general sense. They're things that most people think of and probably already do, or eventually figure out, but having them written down makes it easier to keep in mind. Important things to keep in mind but like anything else they're just a tool for developers.

rbergen:
@syntax-tm We're happy to work towards adding your C# .NET 5.0 solution to the collection; I think it would be interesting to see what the .NET 5.0/C# 9.0 features do to the performance of the sieve. However, to be able to do that, the solution would have to be modified (mainly in structure) to conform to the [contributing guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md). That also means that this PR would need to be opened on the "drag-race" branch, not "original" (formerly "main").

rbergen:
Closing this for reasons indicated.

PR #113, opened 2021-04-09
Update PrimeCPP.cpp

Diconica:
Sorry my system crashed yesterday up submitting this not seeing any record of it the submission today.
It should net a 15 to 20% improvement in performance. runSeive was doing several things repeatedly that didn't need to be done.
Corrected the order of initialization of variables in the class so that it gets rid of warnings at compile time.
Converted most values to uint32_t this will eliminate other warnings and conversions that aren't needed.
I would have used uint64_t except there C++ doesn't seem to handle it with %I64d which is why I dropped back to uint32_t with the exception being on line 24 the map.

### DIFF
--- a/PrimeCPP/PrimeCPP.cpp
+++ b/PrimeCPP/PrimeCPP.cpp
@@ -10,6 +10,7 @@
 #include <cstring>
 #include <cmath>
 #include <vector>
+#include <cstdio>
 
 using namespace std;
 using namespace std::chrono;
@@ -18,9 +19,9 @@ class prime_sieve
 {
   private:
 
-      long sieveSize = 0;
+      uint32_t sieveSize = 0;
       vector<bool> Bits;
-      const std::map<const long long, const int> resultsDictionary = 
+      const std::map<const uint64_t, const uint32_t> resultsDictionary =
       {
             {          10LL, 4         },               // Historical data for validating our results - the number of primes
             {         100LL, 25        },               // to be found under some limit, such as 168 primes under 1000
@@ -35,6 +36,7 @@ class prime_sieve
 
       };
 
+
       bool validateResults()
       {
           auto result = resultsDictionary.find(sieveSize);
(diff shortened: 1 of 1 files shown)


rbergen:
@Diconica We can move to adding your C++ solution to the set of solutions, but not in the way this PR aims for. The modifications you propose are such a significant change to the original implementation that it is warranted to make a new solution out of it that we add, next to the one it is based on.
To be able to do add your solution in that wat, it would have to be modified (mainly in structure) to conform to the [contributing guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md). That also means that this PR would need to be opened on the "drag-race" branch, not "original" (formerly "main").

rbergen:
Closing this for lack of response from the contributor. As indicated in my previous comment, a new PR that conforms to the contributing guidelines can still be opened on the drag-race branch to add this as a separate C++ solution.

Diconica:
Just gained access to this today.
Honestly, not sure what you mean by structure.
Basically all I did was fix a number of issues in the one posted on github
at the time.
You had issue that caused compiler problems such as in the constructor the
variables were out of order. (That actually matters)
The original program has to many casts in variables. Which is why I
switched everything to uni32_t and uin64_t. (Those casts cost time)
Then the inefficient sieve loop that did more checks than it needed and
recounts rather than start where you already made it to.
If that's what you meant by structure then there isn't anything left to
change and submitting anything is a waste of time.

Or maybe you are talking about the hidden sections of code github does by
collapsing it with up and down arrows.

To make things clear that first forloop inside the Sieve is entirely
unnecessary. Step through it by hand you will see what I mean.
What it amounts to is num =factor; factor = num; That's how bad of
programming that is.
So all I did as use num = factor; then surround the second loop by the if
check for the bits[num].

I submitted it so you it could be fixed and you not look so num = factor;
factor = num;
You get my point. That's a pretty embarrassing bit of coding.

PR #116, opened 2021-04-09
Start from square of factor

Guiorgy:
Half of the implemented languages start from the square of `factor` when a new prime is found, which is a valid optimization. However, the other half did not, which is unfair for them when comparing performance.
Comparison in C# (.Net Core 3.1 x64):
Start at `factor*3`: `Passes: 5267, Time: 10.0007961, Avg: 0.0018987651604328842, Limit: 1000000, Count: 78498, Valid: True`
Start at `factor*factor`: `Passes: 5539, Time: 10.0009852, Avg: 0.0018055578985376422, Limit: 1000000, Count: 78498, Valid: True`

### DIFF
--- a/PrimePHP/PrimePHP.php
+++ b/PrimePHP/PrimePHP.php
@@ -61,7 +61,7 @@ class PrimeSieve
                 }
             }
 
-            for ($i = $factor * 3; $i <= $this->sieveSize; $i += $factor * 2) {
+            for ($i = $factor * $factor; $i <= $this->sieveSize; $i += $factor * 2) {
                 $this->clearBit($i);
             }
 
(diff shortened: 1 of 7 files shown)


Walrusbonzo:
I'm no expert with Prime Sieves, but this also stuck out at me. I found the C# implementation to be 5% faster like you have. Good spot.

Guiorgy:
@Walrusbonzo The funny thing is, as I already pointed out, I looked at the implementations in other languages, and half of them did have this optimization

Walrusbonzo:
👍 yep, it's an odd one.

rbergen:
I'm closing this. I believe the vast majority if implementations now do include the squared factor optimization. If the change is still desired for those that don't yet, a PR can be opened on the drag-race branch to that effect.

PR #118, opened 2021-04-10
Add naive Scala implementation

rom1dep:
It compiles both for the JVM and to Native


micwoj92:
naive :(

rom1dep:
> naive :(

Just a straight port of the python version, I haven't tried to optimize anything for raw performance :)

micwoj92:
> > naive :(
> 
> Just a straight port of the python version, I haven't tried to optimize anything for raw performance :)

I see you still haven't seen this typo xp

rbergen:
@rom1dep  We're still happy to consider this solution, but some work will be required to make it possible to add it.

If you still want this solution to be included in the drag race then please read the [contributing guidelines](https://github.com/davepl/Primes/blob/drag-race/CONTRIBUTING.md) and structure your solution accordingly.

Due to the large amount of changes to the `drag-race` branch since you opened this PR, it is probably best to base your next commit on the current version of this branch.

rom1dep:
@rbergen I think I went through the requested changes, although a bit in a haste (haven't actually tried to run the Dockerfile).

Let me know if anything more is required.

rbergen:
I think we're nearly there. As far as I can tell, there is one thing missing: the solution_&lt;n&gt; folder between PrimeScala and your sources, as described here: https://github.com/davepl/Primes/blob/drag-race/CONTRIBUTING.md#folderdirectory. Once that's in place I will take a closer look at the source code. I can already confirm that the Dockerfile works, now.

rom1dep:
@rbergen updated the directory structure. Regarding behavior, I updated the code to run for 5s, as indicated in `CONTRIBUTING.md`, but then I saw that other implementations (Python, Java, …) were not updated (they run 10s), so I'm not sure what the "reference implementation" is anymore :)

rbergen:
Thank you, I'll do a full review in my evening.

Concerning the duration of the sieve runs: that is a fair question to ask. New submissions should, in principle, take the 5 seconds as the guideline, but it is true that we have indeed not retroactively enforced that guideline in "older" solutions. What we will do is execute all benchmark runs sequentially to prevent them from interfering with each other, and repeat the run for each implementation, a number of times. We will then base our reporting on the _average number of passes per second_. That's also why 10 seconds per run really has to be the upper limit; the process will otherwise quickly start taking too much time to complete. 

In practice, for now, we won't be too strict on the 5 seconds, as long as the duration does not go beyond 10. 
We may revise this stance once we have the automated benchmark runs on stable runners, and its reporting, setup. The difference between now and then is that we then have actual benchmark data to look at and base a (revision of) strategy on.

PrimeScala/solution_1/sieve/src/Sieve.scala
```diff
@@ -0,0 +1,94 @@
...
+      if (getBit(num)) {
+        if (showResults)
+          print(f"$num, ")
+        count += 1
+      }
+    }
+
+    if (showResults) println()
+
+    println(s"rom1dep;$passes;$duration;1")
```
rbergen, line 73: It seems your code currently outputs the duration in milliseconds. Please change this to output in seconds, with any milliseconds after the decimal point (so 5027 would become 5.027, for example).

rom1dep: this should be fixed with 09427b6 :)

rbergen: It is indeed.

rbergen (approved):

rbergen:
I think this is ready for merge.

@rhbvkleef or @marghidanu please review and merge if you agree.

marghidanu:
LGTM

Issue #110, opened 2021-04-09
drag-race branch paths 


marghidanu:
Honestly, I find the manual classification of the implementations overengineered. Do we need to reflect this in the structure of the project, a markdown table to mark them as faithful/unfaithful should be enough IMHO.

rhbvkleef:
I thought about this, and I agree the directory structure might need to change. The reason I initially went for the current structure, separating each category, is because a single person might make multiple contributions in different categories. We might decide to drop it down to one category: `stl-faithful`, and only keep those solutions.

A thing I´ve been contemplating is how to deal with parallelized solutions. I think that in the end, we would still need to have some categories. I agree it might be sensible to significantly reduced them. This is what I am thinking of:

* `faithful`
* `parallel`
* `unfaithful` (this would bucket together everything of `stl-unfaithful`, `dependencies-faithful`, `dependencies-unfaithful`, `native`, and any unfaithful parallel implementations)

I think we would still need to keep all these solutions in separate subdirectories, because it is conceivable that people might want to take part in all of these categories.

Kinematics:
Does there even need to be directory structure branching?  Why not just use a proper class-per-file approach, and have commandline switches to indicate which class to target? (Going along with the suggested commandline switches for single-threaded vs multi-threaded runs, and potentially linear vs parallelized versions.)

micwoj92:
I think best one would be per language directories and then inside there whether its faitful or not.

rhbvkleef:
> I think best one would be per language directories and then inside there whether its faitful or not.

@micwoj92, so you propose two categories?

> Does there even need to be directory structure branching? Why not just use a proper class-per-file approach, and have commandline switches to indicate which class to target? (Going along with the suggested commandline switches for single-threaded vs multi-threaded runs, and potentially linear vs parallelized versions.)

@Kinematics, I think we do need some separation, because doing a command-line switch on whether or not to be faithful might not be possible with some languages, as the unfaithful solution might have additional dependencies. Althoug... perhaps those dependencies could just be installed and not used.

micwoj92:
So now since there are also multithreaded options I am not 100% sure. But I still think it would be best if it looked something like this,
directory layout in root folder per language implementation, so let's say python, haskell, c++, autoit3 etc. And in there just faithful single thread, then multi thread and third folder for whatever different implementation someone might think of. I think that per language separation at the top level is best.

OR
[...]

PR #120, opened 2021-04-11
Draft for contribution guidelines

rhbvkleef:


micwoj92:
Maybe also move readme and contributing to .github folder? Main folder is already cluttered enough.

CONTRIBUTING.md
```diff
@@ -0,0 +1,59 @@
...
+Finally, submit a Pull request targetting the branch `drag-race`, and
+place at least the name of the language in the title.
+
+## Categories
+
+### Faithful
+
+Directory name: `faithful`.
+
+* Your solution uses no external dependencies.
```
mike-barber: This is unrealistic in many cases, because it's unclear what external dependencies are. 

For instance, Java, C#, Scala have heavy runtimes with tons of "standard" features. 

In contrast, if we take something like JS, C++, or Rust: these are "composable" because they're meant to build and use only the things they need, so standard practice in these languages is to pull in dependencies to do specific things. I'd argue that many of these dependencies are inseparable from the language as it's the way the language was designed to work -- composable vs. batteries-included.

In short, not sure how to define what these external dependencies are: it's quite subjective.

CONTRIBUTING.md
```diff
@@ -0,0 +1,59 @@
...
+* Your solution uses a class to encapsulate the sieve, or an equivalent
+   feature in your language.
+* Your solution does not use multi-threading or multi-processing.
+* Your solution conforms to the base [rules](#Rules).
+
+### Parallel faithful
+
+Directory name: `parallel`.
```
mike-barber: I think that having a separate directory for `parallel` is going to create quite a bit of duplication, and we might be able to simplify our lives a bit.

I've been playing around with a Rust implementation, and it's easiest to have a single implementation that can run in either single-threaded mode or parallel mode. I know this will apply to C# too.

The most ergonomic way to deal with this would be to record both sets of results via the output mentioned below, but perhaps modify the spec to comma separated (spaces inserted for clarity only)

This also allows us to support different variants within the same codebase. For instance, the Rust implementation allows you to run the sieve using either `bits` or `bytes` for storage.
```
label,                sieve_size,    threads,   total_iterations,  total_time,         num_primes
single-thread-bits,   1000000        1,         6789,              5.012000,           78498
single-thread-bytes,  1000000        1,         8123,              5.000023,           78498
multi-thread-bits,    1000000        8,         25001,             5.000120,           78498
multi-thread-bytes,   1000000        8,         35001,             5.000012,           78498
``` 
Last little detail: total_time should be in seconds, to 6-decimal places. It's too easy to stuff up results and report seconds rounded down to an integer.

This might have an impact on #110, so worth chatting about it there.

CONTRIBUTING.md
```diff
@@ -0,0 +1,59 @@
...
+   feature in your language.
+* Your solution conforms to the base [rules](#Rules).
+
+### Unfaithful
+
+Directory name: `unfaithful`.
+
+* Your solution conforms to the base [rules](#Rules).
+
+## Rules
```
mike-barber, line 65: I think we should add the required sieve size(s) and run time(s) we're aiming for: the "standard" at the moment appears to be a sieve_size = 1,000,000 and runtime = 5s.

It might be worth requiring multiple sizes, but that's open to discussion :)

CONTRIBUTING.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Faithful
+
+Directory name: `faithful`.
+
+* Your solution uses no external dependencies.
+* Your solution uses a class to encapsulate the sieve, or an equivalent
+   feature in your language.
+* Your solution does not use multi-threading or multi-processing.
+* Your solution conforms to the base [rules](#Rules).
```
mike-barber, line 50: I'd also add one here: you don't re-use allocations, and follow the same general pattern as the reference C++ implementation. Otherwise it's very easy to just take the sieve class and "reset" it, which results in a fairly significant performance difference compared to all the other solutions that are doing the same thing as the C++ implementation.

mike-barber:
> Maybe also move readme and contributing to .github folder? Main folder is already cluttered enough.

I'd definitely put these in the root folder. It's the standard place for them. The .github folder is only for GitHub automated admin stuff, like actions, permissions, etc.

mike-barber:
@rhbvkleef thanks for this -- I reckon it's a very good start!

CONTRIBUTING.md
Guiorgy: Actually, about half of the implementations seem to be targeting 10 seconds instead (lisp, py swift, delphi, java, php and c#). Maybe we should adress that?

Edit: #122
Edit 2: Seems that originally 10 seconds was intended, but then he only changed c++ to be 5 seconds 🤔

micwoj92:
> > Maybe also move readme and contributing to .github folder? Main folder is already cluttered enough.
> 
> I'd definitely put these in the root folder. It's the standard place for them. The .github folder is only for GitHub automated admin stuff, like actions, permissions, etc.

.github folder supports things like contributing.md, code_of_conduct.md, readme.md etc 

and this project is hosted on GitHub So I don't see why not. But please do not say that it is "only for GitHub automated admin stuff" Because it is not only for that.

mike-barber:
Sorry @micwoj92 I should probably explain what I mean a bit better; didn't mean to sound dismissive of your suggestion!

My feeling is that general repo stuff, like `README.md` and `CONTRIBUTING.md` stuff should live in the root, since that's where most people expect them to be. I've honestly never encountered them placed anywhere else, but that's just my specific experience: I only checked 15 or so open-source projects I follow, and that definitely seems to be the case in those.

My understanding for the `.github` folder is that it's for the more github-specific stuff: actions workflows, things like specifying CODEOWNERS, templates for pull requests, issues, etc. At least, that's how we use it at work.

I'm not categorically against putting `README` and `CONTRIBUTING` files in the .github folder too, but I think it would be somewhat non-standard. At least in my experience; let's see what others think too. Hope that helps!

micwoj92:
> Sorry @micwoj92 I should probably explain what I mean a bit better; didn't mean to sound dismissive of your suggestion!
> 
> My feeling is that general repo stuff, like `README.md` and `CONTRIBUTING.md` stuff should live in the root, since that's where most people expect them to be. I've honestly never encountered them placed anywhere else, but that's just my specific experience: I only checked 15 or so open-source projects I follow, and that definitely seems to be the case in those.
> 
> My understanding for the `.github` folder is that it's for the more github-specific stuff: actions workflows, things like specifying CODEOWNERS, templates for pull requests, issues, etc. At least, that's how we use it at work.
> 
> I'm not categorically against putting `README` and `CONTRIBUTING` files in the .github folder too, but I think it would be somewhat non-standard. At least in my experience; let's see what others think too. Hope that helps!

One of projects where I contribute most to, [mastercomfig](https://github.com/mastercomfig/mastercomfig)  has these in .github folder and I think it works pretty good there. 

I think it depends whether you want to only use GitHub for code hosting or other source code forges. I don't know if for example GitLab recognizes readme in .github folder or not.

CONTRIBUTING.md
realXCV: Yes if you're creating a reference implementation for a new language. But if you're trying to make an existing one faster,  it shouldn't matter if you're re-using or not the same instance.

Guiorgy: > it shouldn't matter if you're re-using or not the same instance

It does, otherwise it's not a fair comparison. If not, then why should that matter:
> Your solution uses a class to encapsulate the sieve, or an equivalent

Afterall, not using a class will probably result in petter ferformance

CONTRIBUTING.md
mike-barber: Agreed - it was originally 10s. I changed mine from 10s to 5s following the C++ version. There is one extra benefit to using 5s: CI will be twice as fast :) I'm anticipating a lot of projects to build+test.

CONTRIBUTING.md
realXCV: It's not about using a class or not. It's about needing to reallocate memory or not. 

If you want a fair comparison then every implementations should  be identicals. And in that case, the separation should be: "Same as original" and "Everything else". One would be used to compare between OSes, compilers or languages. The other would be about comparing with others implementations for the same language.

Guiorgy: I might've misunderstood you. So, if I understand you correctly, you think it should be allowed to submit implementations that are based on others for a fair comparison, as well as alternative implementations that are faster, but can't be dirrectly compared to others in a (relatively) fair manner?
If so then, first of all, allowing the reuse of buffers wouldn't even improve performance. The only reason we reallocate now is because we are running the sieve many times. If you want maximum performance, you wouldn't care how many times it would run, because in a real scenario you would only run it onece, and in such case current implementations wouldn't reallocate, now would they. Reusing buffers would only serve to misslead, as in a real scenario, how much faster can one implementation allocate memory compared to the other is actually relevant, as you are only running the algorithm once.
And secondly, I think this repository, at least initially, was meant to be one way to compare relative performance across different languages, and not a source of the best and fastest way to implement the sieve in language X. Doing both would require a lot more moderation, and at least as of now, I think they already have enough on the plate. Perhaps this could be considered in the future imo, or someone else can create a repo dedicated to that

realXCV: > I might've misunderstood you. So, if I understand you correctly, you think it should be allowed to submit implementations that are based on others for a fair comparison, as well as alternative implementations that are faster, but can't be dirrectly compared to others in a (relatively) fair manner?

That's already happening. Some of the pull requests are about making the Python, C# or C++ version faster. Some others are about adding versions for different languages.
> The only reason we reallocate now is because we are running the sieve many times.

Are you sure it's not because in the original implementation, the buffer cannot be reset?

Guiorgy: Again, I don't see the point in reusing the buffer, other than to misslead. In a real application, you wouldn't run the sieve multiple times, but just once.

Imagine a hypotetical situation:
Language A performs the sieve about as fast as language B, but the allocation is inefficient in A and thus is 3 times slower than B. If you reuse the buffer and run it many times, that difference will basically be unnoticeable, maybe a couple percentage slower. But in a real scenario, because you only run once, that will result in A being 3 times as slow! A big difference from a benchmark that reuses the buffer.

Now, I know that the abow is a big exaggeration, but the point stands. Language A may not be able to create a simple array of bits, and creating a list object may have overhead, while not a lot, still enough to make a difference in performance. We should include and reflect that in the score imo!

mike-barber: It's kind of what you're both saying already. 
1. We *do* care about how long it takes to run a single iteration
2. A single iteration has to allocate
3. It's unreliable to measure the time for a single pass, so we run as many as we can in 5 seconds and count them; then `time per iteration = 5s / num passes`. 
4. In order for each pass to be **independent**, you need to do the same work. This includes the allocation, and not caching results etc. i.e. don't make subsequent passes faster by reusing stuff previous passes did.

Ideally we'd use benchmarking tools for this measurement, like BenchmarkDotNet, Criterion, JMH, etc. but there's nothing universal across all languages. I think the above is a decent enough compromise.

Guiorgy: @mike-barber yes, that bout right

CONTRIBUTING.md
rhbvkleef: The main reason I added this rule is to prevent (for example) adding Numpy to speed up your implementation. Next to that, I don't think calculating primes requires any dependencies, at least in most languages. I agree it's a bit unfair, but I can't see a better discrimination currently.

CONTRIBUTING.md
rhbvkleef: I think we sould expand this rule:

* Your solution uses a class to encapsulate the sieve, or an equivalent feature in your language.

and add: "This class must contain the full state of the sieve. Each iteration should re-create a new instance of this class."

CONTRIBUTING.md
rhbvkleef: Yes, I'd forgotten to add these requirements. I'll add time=5s + size=1,000,000.

CONTRIBUTING.md
mike-barber: That's a fair point!

I'm wondering how to permit things to make ancillary things (like command-line parsing) easier, without impacting the sieve. Might work by qualifying with something like "no external dependencies used to calculate the sieve." What do you think?

rhbvkleef:
The directory structure was updated in the last commit. I will fix this on the `drag-race` branch later.

CONTRIBUTING.md
rhbvkleef: I had a discussion with @marghidanu about this, and we agree that this categorization using directories is problematic. As such, we decided to go for a single sub-level below the languages, just numbering each solution. The first CI implementation likely will not support outputting multiple test-results for each implementation, but the output format has been adjusted to allow for that in the future.

CONTRIBUTING.md
rhbvkleef: I think that's better. Yeah. I'll change that.

CONTRIBUTING.md
```diff
@@ -0,0 +1,86 @@
...
+
+*Show the output you got on your machine(s) here, in code-blocks*
+```
+
+*TODO: Add the different category badges here*
+
+Now look on the `drag-race` branch, and see what the highest numbered
+implementation is for the language you chose, and place your solution
+in the following folder:
+`Prime<Language>/<solution>_<highest+1>/` where highest is the number of the
```
rhbvkleef: Oops, I made a mistake here. `<solution>` should be `solution`.

CONTRIBUTING.md
```diff
@@ -0,0 +1,86 @@
...
+* Your solution uses the sieve of Erastosthenes.
+* Your benchmarked code returns either a list of primes
+   or the `is_prime` array, containing the result
+   of the sieve.
+* Your solution runs for at least 5 seconds, and stops as quickly as possible
+   after that.
+* Your solution calculates all the primes up to 1,000,000.
+* The test code outputs the following text to standard output:
+   ```
+   <name>;<iterations>;<total_time>
```
mike-barber: Would it make sense to include threads? something like this:
```
<name>;<iterations>;<total_time>;<threads>
```
where threads = 1 for single-threaded solutions.

rhbvkleef: Agreed! I'll change it.

README.md
```diff
@@ -4,3 +4,8 @@ Source code to [Dave's Garage](https://www.youtube.com/c/DavesGarage/featured) v
 benchmarking the same prime number sieve in Python, C#, and C++.
 
 [![Software Drag Racing | Dave's Garage](https://img.youtube.com/vi/D3h62rgewZM/0.jpg)](https://youtu.be/D3h62rgewZM)
+
+## Contributing
+
+If you want to add your own solution, please read [CONTRIBUTING.md](CONTRIBUTING.md).
```
rhbvkleef, line 11: I should add a link to the `drag-race` branch, if people want to just look at community contributions.

CONTRIBUTING.md
Guiorgy: Wait, would single-threaded solutions also have to print the number of threads?

README.md
Guiorgy: Does it mean that people should no longer target the main branch? And what is going to happen to the current implementation on it?

CONTRIBUTING.md
mike-barber: I think they should, yes. It'll make it easier to do reporting.

Guiorgy: Then there's a need to update current implementations, I guess

mike-barber:
I've opened a draft PR #129 for my Rust solution; I'll try to keep it in line with what we decide here so that it can serve as a concrete example.

PR #122, opened 2021-04-12
Normalize runtime to 5 seconds

Guiorgy:
Some implementations (namely in lisp, py swift, delphi, java, php and c#) run the benchmark for 10 seconds, while the rest run for 5 seconds.

### DIFF
--- a/PrimePHP/PrimePHP.php
+++ b/PrimePHP/PrimePHP.php
@@ -85,12 +85,12 @@ class PrimeSieve
 }
 
 //Entry
-$tStart = microtime(true);       //Init time
+$tStart = microtime(true);              //Init time
 $passes = 0;                            //Init passes
 $sieveSize = 1000000;                   //Set sieve size
 $printResults = false;                  //Print the prime numbers that are found
 $rawbitCount = null;                    //Init a rawbitCount to validate the result
-$runTime = 10;                          //The amount of seconds the script should be running for
+$runTime = 5;                           //The amount of seconds the script should be running for
 
 while (getTimeDiffInMs($tStart) < $runTime * 1000) {
     $sieve = new PrimeSieve($sieveSize);
--- a/PrimeSieveCS/PrimeCS.cs
+++ b/PrimeSieveCS/PrimeCS.cs
@@ -122,7 +122,7 @@ namespace PrimeSieveCS
             var passes = 0;
             prime_sieve sieve = null;
 
-            while ((DateTime.UtcNow - tStart).TotalSeconds < 10)
+            while ((DateTime.UtcNow - tStart).TotalSeconds < 5)
             {
                 sieve = new prime_sieve(1000000);
                 sieve.runSieve();
--- a/PrimeSieveDelphi/PrimePas.dpr
+++ b/PrimeSieveDelphi/PrimePas.dpr
@@ -229,7 +229,7 @@ begin
 	passes := 0;
 
 	sieve := nil;
-	while TTimeSpan.Subtract(Now, dtStart).TotalSeconds < 10 do
+	while TTimeSpan.Subtract(Now, dtStart).TotalSeconds < 5 do
 	begin
 		if Assigned(sieve) then
 			sieve.Free;
--- a/PrimeSieveJava/PrimeSieveJava/src/main/java/PrimeSieveJava.java
+++ b/PrimeSieveJava/PrimeSieveJava/src/main/java/PrimeSieveJava.java
@@ -114,7 +114,7 @@ public class PrimeSieveJava
 		int passes = 0;
 		PrimeSieveJava sieve = null;
 		
-		while ((System.currentTimeMillis() - start) < 10000)
+		while ((System.currentTimeMillis() - start) < 5000)
 		{
 			sieve = new PrimeSieveJava(1000000);
 			sieve.runSieve();
--- a/PrimeSieveLisp/PrimeSieve.lisp
+++ b/PrimeSieveLisp/PrimeSieve.lisp
@@ -71,7 +71,7 @@
     (get-universal-time))
 (loop 
     (if 
-        (<= (- (get-universal-time) start) 10)
+        (<= (- (get-universal-time) start) 5)
         (progn
             (prime-sieve 1000000)
             (run-sieve)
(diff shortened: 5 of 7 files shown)


Guiorgy:
Actually, fun fact, looking through git blame it seems that originally it was meant to be 10 seconds:
```c++
while (std::chrono::duration_cast<std::chrono::seconds>(std::chrono::steady_clock::now() - tStart).count() < 10)
```
The above is the initial `C++` code, and below is the current (and original/unchanged) code for `C#`:
```cs
while ((DateTime.UtcNow - tStart).TotalSeconds < 10)
```
And `Python`:
```py
while (timeit.default_timer() - tStart < 10):
```

So, maybe we should set 10 seconds as the standard?

marghidanu:
This looks good! Not sure about the duration yet.

Guiorgy:
Well, in the [original video](https://www.youtube.com/watch?v=l1j-aF_wyzU) and the [video](https://www.youtube.com/watch?v=h4IDTblOHYY) he just released where he compares multithreaded performance, he used 5 second runtime, so I'll consider that as the standard for now

enebo:
If you care about peak performance of languages which JIT then you should run a warmup or rehearsal period and then measure from that point on.  10s for many work loads will be short enough where you will measure the program while it is still optimizing.  For some work loads this matter and others not.  It depends on what you are measuring.

marghidanu:
I usually run several iterations of 5s. The first iteration always has the smallest number of passes.

Guiorgy:
> I usually run several iterations of 5s. The first iteration always has the smallest number of passes.

Most likely, this is because of the JIT compilation. Running many times and taking the best result is usually valid, but not in this case imo. Why? Because if you want to answer the question "in the real world how much faster can I expect language A to run than language B?", then it is undeniable that compiled languages will have faster runtime than JIT languages. And while you can say that, tasks that are run often may be slower on the first run, but after that the speeds will normalize, so that's more important. But the problem with our case is that, in the real world you would only run the sieve once and cache the result. Therefore, I'd argue that we most care about the performance comparison of **The First Run** of the sieve. In this sense, the current benchmark is a little flawed imo, but still better than cherry picking results.
I believe that for the sieve, the overhead of a languages preparing (compile in case of JIT) and/or optimizing should reflect in the comparisons.

marghidanu:
What JIT compilation? What are you talking about? I’m simply saying that we should execute the application 5 times consecutively and retrieve the results.

Guiorgy:
> What JIT compilation? What are you talking about? I’m simply saying that we should execute the application 5 times consecutively and retrieve the results.

An what I am saying is, that you would never really run the sieve multiple time, you would run it once and store the results for reuse, so the unoptimized first run performance is the important one

marghidanu:
Usually, multiple runs are in place to remove any biases. I'm simply suggesting that a repeated experiment might yield more plausible results.

enebo:
"An what I am saying is, that you would never really run the sieve multiple time, you would run it once and store the results for reuse, so the unoptimized first run performance is the important one"

This is a made up microbenchmark.  If we were to find primes we would not use this algorithm and we would certainly run it a lot longer (to find a lot more primes) because we could just memoize the first million primes right?  why calculate them more than once ever?  In that sense who cares how long it takes to figure out the first million primes.

As I said in a previous depends on what you want to measure.  If you want to know how fast these technologies are at peak you either will run with rehearsal or run it long enough that warmup effects of the JIT become a rounding error.

For Java it appears the warmup is fairly quick.  The first 5 seconds (on my machine at least) is nearly at peak perf already.

Guiorgy:
@enebo I agree. What I was saying is that, yes we could use the peek performance, but personally I think that the time a language needs to warm up is also an important metric that shouldn't be entirely discarded, and should be reflected in the scores, especially since you would only run a prime sieve once. Yes, we are only running it to a million, but you could easily take the code and scale it to however big you want. But again, that is just my opinion.

enebo:
@Guiorgy I am not invested in this repo but I think so long as people include warmup in the scores it will be complained about.  No doubt if you don't include warmup some will also complain.  So if I wanted to show both effects I would measure both.  One for peak values and another showing iterations to show how slow the warmup curve is.

Of course the other problem is we all run on different hardware and OSes.  None of these numbers really tell a story for all people either.  Benchmarking is hard. :) 

A lot of work I know but having witnessed a number of these benchmark repos I can see some problems which will occur.

Guiorgy:
@enebo, indeed, there's never a perfect solution, just good enough. A compromise misght be a good choice

ciplogic:
So the compromise is likely better to be 10 seconds, this would mean that even VMs written to compile themselves (i.e. GraalVM) would have a chance to not lose more than 10 percent (i.e. a second) to do this operation. 

I remember a Java talk where just using one lambda could escalate into around 100 ms startup cost, and probably are few places like this on other runtimes with no ahead of time compilation, so up to 1 second could happen on some runtimes. Especially as some runtimes load lazily on disk so the IO cost could also be a factor. 

Obviously: why not 50 seconds could someone say, and I think that people can consider 1 second as a cost that they can pay for a runtime to start, maybe 2. And if 2 seconds is the cost of starting up a runtime, this means 40% of a penalty of VM like machines, but out of 10 seconds is basically showing peak performance penalized by up-to 20%, which is more reasonable. For 50 seconds would mean that a runtime to be significantly penalized to have a startup phase of 10 seconds which misses the point of Dave's video to compare runtimes for few seconds of peak performance.

rbergen:
For the time being, we've decided to stick to the 5 second runtime for all (new) solutions, except for some initial solutions with a 10 second runtime that we've grandfathered. 

Once we're in a place where the community can start posting their results, we can see what makes sense then based on the discussions they lead to. I think a first step would be to perform multiple runs in sequence and average the results (in line with what @marghidanu suggested). One thing we could also do is allowing those languages/platforms to which warm-up impact is considered an unfair disadvantage, to add a "warm-up cycle" of a second or so prior to the actual benchmark run. With the implementation characteristic tags we've recently introduced, we could easily make visible in reporting which implementations use this and which don't. As things stand, none do.

To be honest, I think disk I/O is something we don't need to worry about in the case of this particular benchmark algorithm.

rbergen:
Closing this for reasons indicated in my last comment.

PR #124, opened 2021-04-12
Change code style of creating and running threads.

goeiecool9999:
Same code, different style.


rbergen:
This is a cosmetic change in an implementation and branch that the original author has specifically asked us not to change. As such, we can't merge this.

goeiecool9999:
> This is a cosmetic change in an implementation and branch that the original author has specifically asked us not to change. As such, we can't merge this.

Sorry. I was unaware. I'll close the PR.

PR #126, opened 2021-04-13
PrimeCPP_PAR : Let each thread run multiple sieves until the time is up

aas1226:
Let each thread run multiple sieves until the time is up, instead of recreating a new thread each time

Results On AMD 3700X running in wsl2 ubuntu 20.04
5s 6304 -> 8195 
10s 13312 -> 16067
30s 38816 -> 48838

### DIFF
--- a/PrimeCPP_PAR/PrimeCPP_PAR.cpp
+++ b/PrimeCPP_PAR/PrimeCPP_PAR.cpp
@@ -11,6 +11,7 @@
 #include <cmath>
 #include <vector>
 #include <thread>
+#include <mutex>
 #include <memory>
 
 using namespace std;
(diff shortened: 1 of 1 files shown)


marghidanu:
Not sure I understand what the numbers above mean. Can you elaborate on what you're trying to show?

aas1226:
> Not sure I understand what the numbers above mean. Can you elaborate on what you're trying to show?

For a 5 seconds run before the changes the sieves ran 6304 after the changes the sieves ran 8195, so 30% improvement
For 10 seconds run, before 13312 after 16067, 20% improvement
For 30 seconds run, before 38816 after 48838, 25% improvement

aas1226:
The original program started n new thread waited for all of them to complete the started n other threads until the time is up
There is multiple problems with this approach,
1. If a thread completed 20% before the other, we just waist the time waiting for the other threads to complete
2. the second problem is the overhead of recreating the threads each time
This too problems are addressed by this commit

rbergen:
@aas1226  We will not be able to merge this PR in its current form. For one, it's opened on the `original` branch (formerly `main`), for another Dave has asked us not to make changes to his solutions that change their performance. 

You could turn your approach into a separate solution for inclusion next to the ones we have. However, to be able to do so, it would need to comply with the [contribution guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md). As mentioned, the PR would then need to be opened on the `drag-race` branch.

rbergen:
Closing this for reasons indicated. As also stated, a new PR can still be opened to add this as a separate solution to the drag-race branch.

PR #127, opened 2021-04-13
PrimeSieveAda: Improved implementation size, readability, and preformance

michael-hardeman:
I also included some improvements to the project file that set optimization levels for speed and create missing directories.


michael-hardeman:
@BoopBeepBoopBeep Please review these changes to your original code 🎉

BoopBeepBoopBeep (approved):
👍

MichaelAllenHardeman:
Something is wrong. 

I really shouldn't be nearly doubling up on Dave's C++ performance on a weaker processor with Ada. The Ada impl should be similar or slightly (2%-4%) slower than optimized C/C++ with -O3 and runtime type checks off. I think we are missing something important; there is probably more room for performance to be gained for both of us.

BoopBeepBoopBeep:
The main goal of my original implementation was to make an almost exact
copy of the C++ version in Ada.
The doubling of the performance is due to not setting the Num variable in
the Factor <= Q loop (line 44 and 45) in the original version.
I originally did not compare the C++ and Ada version because they would be
similar and if they were not, would it probably be me who made the mistake
and me who would be unable to fix it.
Knowing the open-source community, mistakes are quickly fixed and updated
so thank you for that.
Also, I would like to thank you for updating the readme to a more detailed
version and fixing the spelling mistake :).

On my computer the C++ implementation using the MSVC C++ compiler gets to
around 6200 passes, whereas the fixed Ada implementation reaches around
4800.
The GNAT toolchain is based on/around the GCC compiler and does support
compiling C/C++ code.
Using this compiler with C++ gives around 5600 passes, resulting in the Ada
implementation still being 15% slower.

If I have to take a guess on where this difference comes from, I would
argue that the C++ implementation uses a vector instead of an array.
Although the first version did not really use the char* as an array but
more like a long list of single bits it did complete less passes.
This version comes close to the number of passes that the Ada
implementation reaches on my computer.
The first version of the C++ code had 2 minor mistakes which impacted
performance: factor * 3 on line 85 instead of factor*factor and factor < q
instead of factor <= q on line 75
Around 5000 passes are made when using the first version of the C++ code
with fixes and the GNAT toolchain on my computer, similar to the Ada
implementation.

I am aware that this all sounds like cherry picking the best scenario for
the Ada implementation, but I would find it nice that with this knowledge
you, me, or somebody else be able to close the 'gap' between the faster C++
version and the Ada implementation.

Edit : Removed email reply text

michael-hardeman:
@BoopBeepBoopBeep Your comment is quite insightful, I can see the reason for the performance discrepancy is from the issues:

> factor * 3 on line 85 instead of factor*factor and factor < q
> instead of factor <= q on line 75

I tried all afternoon today but  was unable to make the Ada implementation go faster. I did get a lead just as I was posting this comment, however. I investigating the `Bit_Array` type. Printing `Bits.all'Size` yields `7_999_984` since it's range is `3 .. 1_000_00` it's length is `999_998`. `7_999_984 / 8 = 999_998` this leads me to believe it's a byte array instead of a true bit array.

michael-hardeman:
Changing the `bit_array` into a `byte_array` and doing the math to convert `bit_index` -> `byte_index` and mask give 50% of the performance.

Strangely precomputing the bit/byte and bit/mask values in big arrays is far slower giving only 25% of the performance.

PR #129, opened 2021-04-13
Rust: refactor for contributing

mike-barber:

This is a refactor of my *existing* Rust solution to bring it in line with the contributing guide. It should be ready to merge now.

Initially created this while discussing rules in #120, and hopefully a it's useful example for others to get an idea of the structure -- Dockerfile, outputs, etc.

Changes:
- moved to correct folder
- updated readme
- using correct output format on stderr
- deleted the second implementation called `simple-bits`
- just simpler overall: just uses the same method to run single or parallel versions.
- updated docker


Guiorgy:
In line **179**:
```
let start = factor * 3 / 2;
let skip = factor;
```
I havent used Rust, but if I understand correctly, you start flagging as non prime starting from `factor * 3`, but you should be able todo so from the square of factor. All non primes below that will already be flagged previousely

mike-barber:
Thanks @Guiorgy! You don't need to know much Rust to spot an algorithm improvement :) 

I'll have to have a think about this with a pen and paper. I see the C++ version has been updated to do this too, so I might have missed it... potentially a little bit more performance on the table :)

Guiorgy:
@mike-barber,
you can check like I did, or be smarter and just google. I found later that [Wikipedia ](https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes) has it all:
> The optimization of starting at the square of each prime and only culling for primes less than the square root

Edit: you can aslo just make the change and run and validate it

PrimeRust/solution_1/README.md
```diff
@@ -0,0 +1,137 @@
...
+## Output
+
+Tested on a Ryzen 3900X, Rust 1.51, running on WSL2.
+
+This is as reported on `stdout`:
+```
+byte-storage;16849;5.0003237724;1
+bit-storage;11171;5.0002226830;1
+byte-storage;157714;5.0008201599;24
+bit-storage;126534;5.0010280609;24
```
rhbvkleef: So, note that this output won't initially be supported, we expect, so ideally, we can pass command-line flags to select each type of output, and so have multiple runs, each with just one line of output.

rhbvkleef:
That's one gorgeous example of a readme ;)

mike-barber:
> @mike-barber,
> you can check like I did, or be smarter and just google. I found later that [Wikipedia ](https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes) has it all:
> 
> > The optimization of starting at the square of each prime and only culling for primes less than the square root
> 
> Edit: you can aslo just make the change and run and validate it

That's true! Thanks!

I did want to understand the reason for it for my own curiosity, and it didn't take long to work out. If anyone's wondering about it too, consider when we're on factor = 5: 
- we're going to go ahead and mark all multiples of 5 -- `10, 15, 20, 25, 30, ...` 
- anything that's a multiple of 5 has another factor, the thing we multiplied with 5: `2, 3, 4, 5, ...`  
- we've _already_ processed all multiples of all numbers below 5: `10, 15, 20` because we've done `2,3,4`.
- so we only need to start the next one that we haven't processed yet: `5*5 = 25`

I'll fix the code and do a quick re-run ;)

EDIT: Nice! It's about a 7% uplift on Ryzen, and about 10% on the Pi.

mike-barber:
> That's one gorgeous example of a readme ;)

Thanks! Much appreciated :)

Guiorgy:
@mike-barber,
Glad to have helped. 1 question though, peviousely in the `stdout` section you had:
```
bit-storage;126534;5.0010280609;24
```
but below:
```
Bit storage     Passes: 141430, Threads: 24, Time: 5.0008363724, Average: 0.0000353591, Limit: 1000000, Counts: 78498, Valid: Pass
Bit storage     Passes: 141365, Threads: 24, Time: 5.0008583069, Average: 0.0000353755, Limit: 1000000, Counts: 78498, Valid: Pass
Bit storage     Passes: 141206, Threads: 24, Time: 5.0008640289, Average: 0.0000354154, Limit: 1000000, Counts: 78498, Valid: Pass
```
So the `stdout` would suggest that was more like 20% uplift. I am going to assume that the `stdout`, `bit-storage`, `multi-threaded` was previousely invalid (below what it should've been)

mike-barber:
> @mike-barber,
> Glad to have helped. 1 question though, peviousely in the `stdout` section you had:
> ...
> So the `stdout` would suggest that was more like 20% uplift. I am going to assume that the `stdout`, `bit-storage`, `multi-threaded` was previousely invalid (below what it should've been)

Well spotted -- I hadn't checked the results very carefully. It was probably just a bad run with me messing around with something else in Windows, which affects the multithread case more. Running single-thread is fine because I've got 11 other cores to handle Windows, but multi-thread is a bit more sensitive. I wasn't being very careful :)

PrimeRust/solution_1/README.md
mike-barber: Thanks Rolf -- I can update the solution to do what we want it to. Luckily, I already have support for command-line parameters, so we can  run it with `--bits --threads=1` for instance. The output here is just the default if you run the docker image without supplying any parameters.

Passing command-line parameters has some complexities though:
- solutions would need to support them
- the script calling the solution would need to know what parameters to pass
   - in particular, some solutions will support threads=X and others won't
   - would need to know which variants to ask for, e.g. `--bits` or `--bytes` is specific for this solution.

We'd probably need to figure out how to bridge the gap between the script running solutions and the solution itself. Maybe having a series of `run-x.sh` scripts in the solution folder do run each case? 

It might make sense, or it might be easier to deal with this on the reporting side as there's already a nice spec for the output. Not sure what's better/easier, though!

rhbvkleef: I think being able to pass these parameters is enough, for now.

mike-barber: Thanks @rhbvkleef -- happy to go with the parameters for now ;) 

I'm a bit unclear on what I should do with the PR for now, though. Couple of options:
1. Just make the default (no parameters) output a single line for single-threaded, e.g. `byte-storage;16849;5.0003237724;1` (I'm guessing most solutions will be like this, anyway).
2. Find out which parameters you'd like to send from the automated benchmark process, and hook them up 

I'm guessing (1) is the easier option to get us going; should integrate fine with the process. We can always tackle (2) later when we want to.

What do you think?

mike-barber:
Thanks @marghidanu 

We've still got a small outstanding question regarding the output format and/or input parameters for the benchmarking; happy to make changes to resolve this before we merge. Thread is in the conversation with Rolf above -- let me know your thoughts too :)

marghidanu:
@mike-barber is this about sending parameters to the Docker image?

mike-barber:
> @mike-barber is this about sending parameters to the Docker image?

@marghidanu yep I'm wondering what I should do with the output and input parameters. Currently the Docker image produces 4 results if no parameters are supplied, and Rolf mentioned we're not ready to parse results like that yet.

The code already supports command line parameters, but I'm not sure how to supply them from the benchmark job. 

So wondering what the best way to proceed is. Had a couple of options that I could think of, mentioned above:

> 1. Just make the default (no parameters) output a single line for single-threaded, e.g. byte-storage;16849;5.0003237724;1 (I'm guessing most solutions will be like this, anyway).
> 2. Find out which parameters you'd like to send from the automated benchmark process, and hook them up
> 
>I'm guessing (1) is the easier option to get us going; should integrate fine with the process. We can always tackle (2) later when we want to.

marghidanu:
I haven't written the output parser yet so the output format could still be extended, I mean we should just it for our needs anyway. We need something that can be easily extendable in the future.

mike-barber:
Thanks @marghidanu that makes sense! And it sounds like the best bet for now is to leave the output as it is currently, and we'll work out how to process and parse it later. 

It currently looks like this: 
```
byte-storage;16849;5.0003237724;1
bit-storage;11171;5.0002226830;1
byte-storage;157714;5.0008201599;24
bit-storage;126534;5.0010280609;24
```
There are two different variants (`byte-storage` and `bit-storage`) and single thread (1) and multithread (24 -- on my PC at least) runs. 

Happy to make update the output to fit a revised format later -- just let me know when you're ready.

PrimeRust/solution_1/README.md
mike-barber: Further discussion on this below - marking this resolved again :)

marghidanu:
That's fine, fine-tuning can happen at a later stage :)

mike-barber:
Super, thanks @marghidanu -- much appreciated :)

@rhbvkleef I think we've got everything sorted out, and it's in a merge-able state :) Just let me know if anything needs fixing.

PR #138, opened 2021-04-16
Move all implementations to the ppropriate folders, according to CONTRIBUTING

rhbvkleef:
## Description

I hereby move all implementations to the correct folders, according to the changed contributing guidelines. The actual implementations are not provided with a fully appropriate readme yet, and are also not provided with a dockerfile yet. That should happen in separate PR's, and should be done on an individual basis, at least in my opinion.


rhbvkleef:
I've also just noticed I apparently added a LICENSE file. Should we do that?

marghidanu:
I see a lot of changes, some of these will cause some conflicts I believe. 

* The change from solution_1 to solution_0: Is it necessary? What's the benefit of starting from 0? :)
* One License file in the root of the repository should be ok. Everybody who contributes to this repo should be made aware of the License.
* I've already made changes to the README file. I see you also made changes on this branch, this will cause a conflict. No need for that, we should take care of these in the pull request.

I feel we somehow substituted the main branch with this drag-race branch which is kinda unusual. For example, the scheduled job for running the benchmark only runs on the default branch, the way this is set up I cannot use it on this separate branch.

rhbvkleef:
> The change from solution_1 to solution_0: Is it necessary? What's the benefit of starting from 0? :)

That was originally my plan. I worked on this branch before the contributing PR was merged, so that's mostly a holdover. I did solve those conflicts.

> One License file in the root of the repository should be ok. Everybody who contributes to this repo should be made aware of the License.

👍🏽 

> I feel we somehow substituted the main branch with this drag-race branch which is kinda unusual. For example, the scheduled job for running the benchmark only runs on the default branch, the way this is set up I cannot use it on this separate branch.

Perhaps we should talk to Dave on how we should deal with this. I'm not sure how to solve that.

> I've already made changes to the README file. I see you also made changes on this branch, this will cause a conflict. No need for that, we should take care of these in the pull request.

Ah, check. You are recommending to remove all changes README's, or just those where there was already a README? The reason I wanted to add the README's is because that's the only place (aside from git history) where you could see who originally implemented it.

mike-barber:
Would it make sense to put Dave's original solutions in an "original" folder maybe?

Could be all in a single original folder, or each language could have an original folder with Dave's code in it, with community ones then labelled solution_x as usual.

One thing that could make this merge easier is doing it in two phases - 
1. Merge main into drag-race
2. Then fix the paths in drag-race

Then maybe one day get rid of drag race entirely and just have main :)

rhbvkleef:
> Would it make sense to put Dave's original solutions in an "original" folder maybe?

I also think that merging `master` and `drag-race` is a good idea.  I don't think that that's the scope of this PR, so I'll go ahead and create a new issue, and try to get Dave to comment on it too.

EDIT: Created issue #148

rbergen:
I'd swear I saw a notification that this PR was already closed by @rhbvkleef 
Anyway, now that PR #168 has been merged, I really see no reason to keep it open.

PR #141, opened 2021-04-16
"Closest approximation" implementation in MIXAL

rbergen:
## Description

This is an implementation in the assembly language for the MIX computer, described in D.E. Knuth's book series The Art of Computer Programming.

As a computer, MIX has a number of specific characteristics:
* It never existed in physical form.
* It was based on or inspired by more or less typical computer designs in the 1960s.
* (Even) in the context of the previous point, it includes a number of quirks. I believe these were introduced on purpose by its designer, to make the solution to some of the exercise problems in the book series less straight-forward. (In the context of this repository, the absence of "regular" bitwise operators was somewhat painful, for instance,)

Due to this, some of the basic rules could not be complied with completely, as described in more detail in the README.md. I think one could argue that as such, it is not a "proper" drag race contender, and its inclusion in an "inspired" or "alternative" branch would be more intuitive.

## Contributing requirements


micwoj92:
Third time's a charm

rbergen:
Yeah, doing this around midnight on a Friday, I should have known better. I apologize for the noise.

marghidanu:
I tried running this using `mdk` under Ubuntu with no success.

rbergen:
I should have mentioned in the README.md that there are small deviations in the MIXAL format that various emulators can digest; this is due to the fact that no actual computer is MIX. That, in turn, means that "cosmetic changes" to the source may be necessary to make it assemble across emulators. It is for this reason that the run instructions specifically mention using MixEmul. 
If we want to use MDK as a verification tool, then I can add a version that runs on that.

marghidanu:
That would be great as `mdk` is able to run in Docker. Therefor this can be Dockerized and included in in the actual drag race :)

rbergen:
I understand. 

I've started working on it. I'll update the PR when I have a version that I have seen assemble and run on Ubuntu/mdk.

rbergen:
Updated. I've also:
* switched output from printer to console
* made the number of runs to execute configurable
* updated the README.md, accordingly

rbergen:
Do let me know if there's anything you'd like me to change about the output, whether concerning device and/or format. There's only so much I can do, because:
* MIX's character set (available [here](https://esolangs.org/wiki/MIX_(Knuth)#Character_codes) for reference) is severely limited by its byte size
* On the console I have a limit of 70 characters per line. On the printer that number is 120. 

Within these boundaries, I'm happy to make any changes that will make it easier to adopt this implementation.

rhbvkleef:
> Do let me know if there's anything you'd like me to change about the output, whether concerning device and/or format. There's only so much I can do, because:

Looking at the character set, you should be able to comply with the format set out in the contributing file, right?

rbergen:
> > Do let me know if there's anything you'd like me to change about the output, whether concerning device and/or format. There's only so much I can do, because:
> 
> Looking at the character set, you should be able to comply with the format set out in the contributing file, right?

No, in the sense that the implementation itself can't keep track of time (MIX is unaware of the concept), so I can't report on time spent from within the implementation. In fact, even the "ticks" that are spent on execution are inaccessible from within the execution context itself. On the other hand, I can imagine that the sieve size **is** of interest, as it is not 1,000,000, by definition.
Furthermore, at least in case MDK is used for execution, there is leading and trailing text that I can't suppress. 

If I stick to the format indicated, then I can report name, the run count that is configured within the implementation itself, and the thread count. Taking the submitted run count of 27, again in case of MDK, that would basically translate to the "static" text:
```
Program loaded. Start address: <number>
Running ...
rbergen;27;;1
... done
```
Is this then what I should make the output be?

rhbvkleef:
Yes, I think that's better. At least we can then time this externally (which is imperfect, but at least something) and read the rest of the values from the output.

rbergen:
Updated as requested. The user name is in upper case because MIX doesn't support lower case, as discussed earlier.

marghidanu:
Was trying to run this again in a Docker container but I get an error. I was using the following container definition:

```
FROM ubuntu:18.04

RUN apt-get update -qq \
    && apt-get install -y mdk

WORKDIR /opt/app
COPY *.mixal .

RUN mixasm primes.mixal

ENTRYPOINT [ "mixvm", "-r", "/opt/app/prime" ]
```

To build it use `docker build -t mixal .` and to run in use `docker run -ti --rm mixal`. Unfortunately, this does not pass the build phase, it fails with the following error:

```
#9 0.183 primes.mixal:286: error: invalid expression
#9 0.183 (0 warning(s), 1 error(s))
```

marghidanu:
If everything is ok you can also commit the Dockerfile. It will automatically be built in the near future. :)

rbergen:
That's interesting. I obviously tested the implementation using the run commands in the README before pushing, and I did not get an error then (in fact, I got the output I copy/pasted into the README). 
When looking at differences between your situation and mine:
- I'm not using Docker
- I'm running on (WSL2) Ubuntu 20.04.2 LTS, as lsb_release -a reports

`mixasm -v` yields `mixasm, MIX assembler (GNU MDK 1.2.10)` on my system, it could be there is another difference, there.

Looking at the error, line 286 contains the last directive of the program (`END START`), the "expression" in which is a very straightforward symbol pointing to the first "real" instruction of the program. I know for sure that there is nothing invalid about the expression, as such.
What I do know is that 98% of the problem with getting the thing to run in MDK in the first place, was related to the classical line break character issue. I originally submitted a version with CR LF line breaks, and MDK can only handle LF. The error being reported on the last line of the program makes me wonder if there is a weird/very localized version of that problem happening.

What I've done is push another version with a slightly different framing of the `END START` directive. Maybe you can test again with that one? If that doesn't work then I may need some help replicating your Docker setup at my end. As you can tell, I'm basically debugging an assembly error that I know is "unjustified" from a source code perspective, that I also can't reproduce, here.

marghidanu:
Yep, `dos2unix` fixed it. Here's the updated Dockerfile.

```
FROM ubuntu:18.04

RUN apt-get update -qq \
    && apt-get install -y mdk dos2unix

WORKDIR /opt/app
COPY *.mixal .

RUN find . -name *.mixal -exec dos2unix {} \; 
RUN mixasm primes

ENTRYPOINT [ "mixvm", "-r", "/opt/app/primes" ]
```

Here's the output:

```
Program loaded. Start address: 3429
Running ...
RBERGEN;27;;1                                                         
... done
```

marghidanu:
Please add the Dockerfile to the commit, the CI will pick it up and build the image for us.

rbergen:
Added as requested. Anyway, great news that you got it to work! I appreciate the help with putting the Dockerfile together.

Amazing how line breaks occasionally still trip me up after 25 years...

marghidanu:
Honestly, it didn’t seem even occurred to me that line endings might be problematic. Thanks for suggesting that!

rbergen:
You're very welcome. It was actually my second guess, after ruling out some sort of character encoding mix-up. 
In my experience, character encoding/line ends and time zones/synchronisation (the latter two are clearly not an issue here) are some of the most common instigators of "problems that don't make any sense" with software.

marghidanu:
Let's change the path for your solution according to the CONTRIBUTING.md and we can merge this.

rbergen:
My bad, I completely overlooked that part of the document. Fixed in my latest push.

PR #143, opened 2021-04-17
Performance tweak and simplification

CorrodedCoder:
Modified PrimePY implementation as follows:
1) Change divide by 2 to bitshift right for performance boost.
2) Simplify while/for/if using a filter to select appropriate factor candidates.
3) Simplify sum of condition to just sum container.
4) Removed some spurious semi-colons.

### DIFF
--- a/PrimeSievePY/PrimePY.py
+++ b/PrimeSievePY/PrimePY.py
@@ -29,7 +29,7 @@ class prime_sieve(object):
 
     def __init__(this, limit):
         this.sieveSize = limit
-        this.rawbits = [True] * (int((this.sieveSize+1)/2))
+        this.rawbits = [True] * ((this.sieveSize+1)>>1)
 
     # Look up our count of primes in the historical data (if we have it) to see if it matches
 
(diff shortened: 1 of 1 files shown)


rhbvkleef:
I'm going to close this, because it is not necessarily a simpler, or better way to do this. Additionally, no idiomatic improvements are made. I understand there are a few code-smells in Dave's code, but using bitshifts for them is not the right way to go about it, as it significantly obscures functionality for beginner programmers, which is what this is (at least partly) aimed at.

CorrodedCoder:
Makes sense, thanks for taking a look.

CorrodedCoder:
There is a bit of a challenge here in comparing apples with apples - for example: the Java implementation both bit shifts and does not assert that the index is equal in ClearBit.
With the same changes to the Python version I see nearly double the performance.
I think the rejection of my change is valid for reasons given (in fact I didn't remove the assert for the same reasoning above), but does that mean the the Java version should be changed to match Dave's original more accurately?

PR #151, opened 2021-04-19
Drag race kinematics 2

Kinematics:
## Description

Bug fixes, performance optimization, and extra commandline option.

## Contributing requirements

* [X] I placed my solution in the correct category folder.
* [X] I added a README.md or Dockerfile.


rbergen:
Thank you for these updates, particularly the "run all sieves" option is very helpful.

I do wonder if it would be possible to add one more option that makes the run output conform to the drag race format? That being one line of the following format, per sieve type/run:
```
kinematics2_<sieve_tag>;<pass_count>;<runtime>;<pthread_count>
```

In this:
* The sieve tag could in principle be what you find best suited for a sieve type. However, considering that you already defined command-line options for the various types, it occurred to me that it may be easiest to just copy those. 
* The runtime should be measured in seconds (i.e. 5 + fraction) in en_US format
* The pthread count should reflect the number of threads actually used for running the sieve (meaning it will always be 1 for non-parallel runs, regardless of the number of cores/threads available)

What's behind this request is that adhering to this format will make it possible to automatically process the results of solutions and start reporting on them, at some point in the future.

In the same context, it would actually be ideal if you could also provide a Dockerfile with your solution. Alternatively, I can add one for you, which would then be based on the mcr.microsoft.com/dotnet/sdk:5.0-focal (build) and mcr.microsoft.com/dotnet/runtime:5.0-focal (run) images provided by Microsoft.

Kinematics:
>     * The pthread count should reflect the number of threads actually used for running the sieve (meaning it will always be 1 for non-parallel runs, regardless of the number of cores/threads available)

Can you clarify whether you mean multi-threaded running threads, or parallelized running threads?  Dave's versions have used multithreading (use multiple threads, and run one sieve per thread), whereas the parallelized algorithms I have are running one instance of a sieve with multiple threads employed to break up the work.

I've added an extra line to the output for the requested format. Can you tell me if this is acceptable?

```
@Kinematics: Starting (original)...
Passes: 2929, Time: 5.00012 s, Per Loop: 1.707067 ms, Sieve Size: 1000000, Thread Count: 1, Primes Found: 78498, Valid: True
kinematics_original;2929;5.00012;1

@Kinematics: Starting (standard)...
Passes: 3663, Time: 5.00125 s, Per Loop: 1.365274 ms, Sieve Size: 1000000, Thread Count: 1, Primes Found: 78498, Valid: True
kinematics_standard;3663;5.00125;1

@Kinematics: Starting (bool array)...
Passes: 5122, Time: 5.00014 s, Per Loop: 0.976181 ms, Sieve Size: 1000000, Thread Count: 1, Primes Found: 78498, Valid: True
kinematics_bool;5122;5.00014;1

@Kinematics: Starting (inverted bool array)...
Passes: 6344, Time: 5.00052 s, Per Loop: 0.788146 ms, Sieve Size: 1000000, Thread Count: 1, Primes Found: 78498, Valid: True
kinematics_ibool;6344;5.00052;1

@Kinematics: Starting (direct access inverted bool array)...
Passes: 9034, Time: 5.00016 s, Per Loop: 0.553465 ms, Sieve Size: 1000000, Thread Count: 1, Primes Found: 78498, Valid: True
kinematics_dbool;9034;5.00016;1

@Kinematics: Starting (raw bits)...
Passes: 6183, Time: 5.00075 s, Per Loop: 0.808669 ms, Sieve Size: 1000000, Thread Count: 1, Primes Found: 78498, Valid: True
kinematics_raw;6183;5.00075;1

@Kinematics: Starting (raw bits uint)...
Passes: 6371, Time: 5.00019 s, Per Loop: 0.784806 ms, Sieve Size: 1000000, Thread Count: 1, Primes Found: 78498, Valid: True
kinematics_raw32;6371;5.00019;1

@Kinematics: Starting (raw bits direct)...
Passes: 6439, Time: 5.00057 s, Per Loop: 0.776518 ms, Sieve Size: 1000000, Thread Count: 1, Primes Found: 78498, Valid: True
kinematics_rawd;6439;5.00057;1

@Kinematics: Starting (raw bits [2 of 6])...
Passes: 6112, Time: 5.00011 s, Per Loop: 0.818063 ms, Sieve Size: 1000000, Thread Count: 1, Primes Found: 78498, Valid: True
kinematics_raw6;6112;5.00011;1

@Kinematics: Starting (raw bits parallel version)...
Passes: 4277, Time: 5.00089 s, Per Loop: 1.169044 ms, Sieve Size: 1000000, Thread Count: 1, Parallel Thread Count: 16, Primes Found: 78498, Valid: True
kinematics_rawp;4277;5.00089;1

@Kinematics: Starting (array pool)...
Passes: 6100, Time: 5.00048 s, Per Loop: 0.819672 ms, Sieve Size: 1000000, Thread Count: 1, Primes Found: 78498, Valid: True
kinematics_pool;6100;5.00048;1

@Kinematics: Starting (array pool [2 of 6])...
Passes: 9644, Time: 5.0002 s, Per Loop: 0.518457 ms, Sieve Size: 1000000, Thread Count: 1, Primes Found: 78498, Valid: True
kinematics_pool2of6;9644;5.0002;1

@Kinematics: Starting (parallel array pool [2 of 6])...
Passes: 5516, Time: 5.0002 s, Per Loop: 0.906454 ms, Sieve Size: 1000000, Thread Count: 1, Parallel Thread Count: 16, Primes Found: 78498, Valid: True
kinematics_pool6p;5516;5.0002;1

@Kinematics: Starting (array pool [8 of 30])...
Passes: 12493, Time: 5.00034 s, Per Loop: 0.400224 ms, Sieve Size: 1000000, Thread Count: 1, Primes Found: 78498, Valid: True
kinematics_pool30;12493;5.00034;1

@Kinematics: Starting (array pool [8 of 30] with bitmasking)...
Passes: 17123, Time: 5.0001 s, Per Loop: 0.292005 ms, Sieve Size: 1000000, Thread Count: 1, Primes Found: 70750, Valid: False
kinematics_pool30m;17123;5.0001;1
```

I've also set up a dockerfile that will be included in the next push, however running that is significantly slower than normal.  For example, the Standard sieve gave 1750 passes instead of 3660.  Array pool (2 of 6) gave 5200 passes instead of 9600. Etc.  This is running from either the Powershell terminal or the Ubuntu terminal.

It may be useful to simply be _able_ to run the programs without needing the setup environment, and comparing between different sieves of the same language origin, but it's not great for getting a performance potential, and questionable for comparing languages.

rbergen:
> Can you clarify whether you mean multi-threaded running threads, or parallelized running threads? Dave's versions have used multithreading (use multiple threads, and run one sieve per thread), whereas the parallelized algorithms I have are running one instance of a sieve with multiple threads employed to break up the work.

The conceptual answer to your question is that the number should reflect the number of threads that are actively involved with getting to the sieve run count you're also outputting. I think that translates to:
- Multi-threaded running threads, if you use multiple threads to run sieves that are themselves sequential (i.e., don't use a parallelized algorithm)
- Parallelized running threads if your sieves are run using a parallelized algorithm, but the runs are executed sequentially
- "All threads involved" if you were to use a parallelized algorithm that you run multiple times at the same time.
 
> I've added an extra line to the output for the requested format. Can you tell me if this is acceptable?

Yes, that's wonderful, thank you.

> I've also set up a dockerfile that will be included in the next push, however running that is significantly slower than normal. For example, the Standard sieve gave 1750 passes instead of 3660. Array pool (2 of 6) gave 5200 passes instead of 9600. Etc. This is running from either the Powershell terminal or the Ubuntu terminal.

I'm deducing from your mention of Powershell that you're running on Windows. It is well known that performance of Docker (Desktop) on Windows can indeed be about half of what you get without it. And with "well known", I mean that this is recognized well outside the boundaries of this repo. The exact reasons I don't know, but with Docker doing some sort of balancing act between Windows and the VM that WSL2 of course still is, they're not that difficult to imagine.
I think consensus is that Docker on Windows is fine for development, but not for production runs.

> It may be useful to simply be _able_ to run the programs without needing the setup environment, and comparing between different sieves of the same language origin, but it's not great for getting a performance potential, and questionable for comparing languages.

Duly noted. And exactly as you say, if nothing else, the Dockerfiles do provide runnable, testable documentation of what it takes to get solutions to run. 

I think that bottom-line, you raise a good point about what should, or should not, be the context of the actual drag race solution runs. That said, by definition, a drag race is a statement of comparison, in whatever context the drag race takes place in.

PrimeCSharp/solution_1/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM mcr.microsoft.com/dotnet/runtime:5.0
+
+COPY . App/
+WORKDIR /App
+ENTRYPOINT ["dotnet", "PrimeCSharp.dll"]
```
rbergen: I just want to check if my understanding is correct that a push of the "final" Dockerfile is still underway? This Dockerfile doesn't build the solution, and hence also can't run it.

Kinematics: > I just want to check if my understanding is correct that a push of the "final" Dockerfile is still underway? This Dockerfile doesn't build the solution, and hence also can't run it.

Yeah, I realized that this morning. I'll work on fixing it tonight.

Kinematics:
Also, hold off on merging. There's another bug I need to fix.

marghidanu:
On hold :)

Kinematics:
All right, should be good now.

The rented array in the bitmasking algorithm had a risk of non-zero data if it was run after another algorithm.  The problem didn't show up when it was tested individually.  Zeroing out the array fixed things, though it dropped performance from 17250 loops to 15600 loops.

rbergen:
> The rented array in the bitmasking algorithm had a risk of non-zero data if it was run after another algorithm. 

That's a good find, I can imagine this leading to issues that make one pull their hair out when analyzing...

PrimeCSharp/solution_1/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+WORKDIR "/src/."
+RUN dotnet build "PrimeCSharp.csproj" -c Release -o /app/build
+
+FROM build AS publish
+RUN dotnet publish "PrimeCSharp.csproj" -c Release -o /app/publish
+
+FROM base AS final
+WORKDIR /app
+COPY --from=publish /app/publish .
+ENTRYPOINT ["dotnet", "PrimeCSharp.dll"]
```
rbergen: The Dockerfile builds and runs perfectly now, but only the standard implementation. That means that only that implementation would be included in the automated runs and therefore only the result of that implementation would be included in the automated result reports that will be generated on those runs in the (near) future.

Maybe you want to consider passing the `--all` argument in/to your `ENTRYPOINT`?

Kinematics: I've been trying to get it to work, but for some reason it's not passing in the command parameters when I build it into the Dockerfile.  If I run the Dockerfile with the additional parameters on the commandline, it works as expected. Any advice on that?

rbergen: I'd swear I've seen it "work on my machine". I'll take another look and let you know how I fare.

rbergen:
```
rutger@host:~/work/Primes/PrimeCSharp/solution_1$ cat Dockerfile
#See https://aka.ms/containerfastmode to understand how Visual Studio uses this Dockerfile to build your images for faster debugging.

FROM mcr.microsoft.com/dotnet/runtime:5.0 AS base
WORKDIR /app

FROM mcr.microsoft.com/dotnet/sdk:5.0 AS build
WORKDIR /src
COPY ["PrimeCSharp.csproj", "."]
RUN dotnet restore "./PrimeCSharp.csproj"
COPY . .
WORKDIR "/src/."
RUN dotnet build "PrimeCSharp.csproj" -c Release -o /app/build

FROM build AS publish
RUN dotnet publish "PrimeCSharp.csproj" -c Release -o /app/publish

FROM base AS final
WORKDIR /app
COPY --from=publish /app/publish .
ENTRYPOINT ["dotnet", "PrimeCSharp.dll", "--all"]
rutger@host:~/work/Primes/PrimeCSharp/solution_1$ docker run -ti --rm csharp
@Kinematics: Starting (original)...
Passes: 1339, Time: 5.00218 s, Per Loop: 3.735624 ms, Sieve Size: 1000000, Thread Count: 1, Primes Found: 78498, Valid: True
kinematics_original;1339;5.00218;1

@Kinematics: Starting (standard)...
Passes: 1715, Time: 5.00074 s, Per Loop: 2.915452 ms, Sieve Size: 1000000, Thread Count: 1, Primes Found: 78498, Valid: True
kinematics_standard;1715;5.00074;1

@Kinematics: Starting (bool array)...
Passes: 2688, Time: 5.00043 s, Per Loop: 1.860119 ms, Sieve Size: 1000000, Thread Count: 1, Primes Found: 78498, Valid: True
kinematics_bool;2688;5.00043;1
[...]
```

Obviously, I did run a `docker build -t csharp .` just before the `cat Dockerfile`. 

This means that with the `"--all"` added to the `ENTRYPOINT` in this way, it does work for me.

Kinematics: I'm not sure why it wasn't working, but now it is. Possibly a glitch in the VS building process. Pushed the revised version of the Dockerfile.

rbergen (approved):

rbergen:
I think this is ready to be merged. 

@marghidanu or @rhbvkleef Please have a look and merge if you agree.

marghidanu:
LGTM

PR #152, opened 2021-04-20
Single-threaded C entry.

mckoss:
## Description

A pure C version with a couple optimizations over the naïve sieve.

## Contributing requirements


rbergen:
@mckoss Would it maybe be possible to add a Dockerfile to this solution? Considering the language it is not difficult to add one myself, but this solution seems heavily optimized for performance. I can therefore imagine you prefer to create one, yourself. 

Also, the output is in the right format, but it will be just a little easier to process if it ends with a newline.

mckoss:
sure

mckoss:
> @mckoss Would it maybe be possible to add a Dockerfile to this solution? Considering the language it is not difficult to add one myself, but this solution seems heavily optimized for performance. I can therefore imagine you prefer to create one, yourself.
> 
> Also, the output is in the right format, but it will be just a little easier to process if it ends with a newline.

Dockerfile added.  I don't know why, but this runs substantially slower in a container on my machine.

rbergen:
> Dockerfile added. I don't know why, but this runs substantially slower in a container on my machine.

Thank you. Your README mentions you're running on Windows. It's a known fact that using Docker (Desktop using WSL2) on Windows comes with a notable performance impact.

rbergen:
I've tested your solution with Docker, and I think the Dockerfile needs a minor adjustment. The run fails because the run script is not executable. I've added a `RUN chmod a+x run` immediately after the COPY, and that made it run successfully on my system.

mckoss:
> I've tested your solution with Docker, and I think the Dockerfile needs a minor adjustment. The run fails because the run script is not executable. I've added a `RUN chmod a+x run` immediately after the COPY, and that made it run successfully on my system.

It's odd.  It was already marked executable in my image - but I added the suggested chmod so your configuration works as well (hopefully).

PrimeC/solution_1/README.md
```diff
@@ -0,0 +1,57 @@
+# C solution by mckoss
+
+Single threaded implementation in C of prime number sieve.
```
rbergen: I'm aware the following request is not yet mentioned in the contribution guidelines, but could you please add the following text between the header and the first line of text?
```
![Category](https://img.shields.io/badge/Category-unfaithful-yellowgreen)
```

PrimeC/solution_1/sieve.c
```diff
@@ -0,0 +1,303 @@
+/*
+Various prime sieve algorithms speed tested.
```
rbergen: I believe this solution actually includes one algorithm, being a (highly performance-optimized) 8 of 30 implementation.

rbergen (commented):
Thanks for adding this. I can confirm that the solution now builds and runs on my machine as well.

With that now being the case, I've performed a more detailed review of your solution. I found two small things; could you please look at these? Once those have been addressed, I think I'm ready to approve this for merge.

PrimeC/solution_1/README.md
mckoss: What is it that categorizes this algorithm as  "unfaithful"?

rbergen: As indicated in the contribution guidelines, a faithful implementation needs to keep track of full sieve state in a class/object. In C I suppose this would translate to a struct. Currently, your implementation is purely procedural, without this isolation of sieve state.

mckoss: I wrote this version to return the full sieve state in an allocated block of memory, as the C equivalent of an object.  I could put it in a struct.  Would that make it conformant?

rbergen: What we've seen the various solutions across languages converge to is a data structure that includes the prime boolean/bit array and the sieve size. The logic of the implementation should basically need nothing but the structure('s contents) to prepare the sieve runs, to execute the runs themselves and to output the results. In that, the structure needs to be created "from scratch" for every sieve run.
I don't mean to force you towards a particular solution, but you could take a look at the approach taken by the other C solution in this repo.

mckoss: Update to allocate and free a containing sieve object as part of each pass.

PrimeC/solution_1/sieve.c
mckoss: Ooops - old comment fixed.

rbergen:
Thank you. It now builds, runs and carries the correct category badge.

@marghidanu , @rhbvkleef : can either of you take a look to see if you agree that we can merge this?

rbergen (approved):

marghidanu:
LGTM

PR #154, opened 2021-04-21
C implementation serial and parallel with optimizations

danielspaangberg:
A C implementation with both an implementation similar to the original code and serial optimizations and parallelization.

## Description

Several different versions of the sieve implemented in C. Some optimizations for the serial version, and two different parallelization strategies - one that makes the sieve run faster, the other one to run multiple sieves in parallel.

## Contributing requirements


marghidanu:
Quite some work here! If you include some Dockerfile definitions the solutions will be included in the scheduled build.

danielspaangberg:
> Quite some work here! If you include some Dockerfile definitions the solutions will be included in the scheduled build.

Thanks. I'll see if I can do that. Never used docker before, but I'm eager to learn :-)

marghidanu:
Happy to help if you bump into any problems

danielspaangberg:
@marghidanu I think I got the Dockerfile correct. It was not very hard once I found a reasonable tutorial. Now it also passes the build. The parallel versions are quite a bit slower under docker though. Not sure if I do anything wrong (I tested under windows with wsl2, and on a chromebook (arm64, 8 cores, not the most typical config I guess).

danielspaangberg:
Got docker installed on my raspberry pi. There the parallel versions runs just fine and just as good as natively.

PrimeC/solution_2/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM alpine:3.13
+
+RUN apk add --update build-base
+
+WORKDIR /opt/app
+
+COPY compile.sh .
+COPY run.sh .
+COPY sieve_1of2.c .
```
marghidanu: `COPY *.c ./`

PrimeC/solution_2/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+COPY sieve_48of210_only_write_read_bits.c .
+COPY sieve_48of210_par.c .
+COPY sieve_480of2310_epar.c .
+COPY sieve_480of2310_only_write_read_bits.c .
+COPY sieve_480of2310_par.c .
+COPY sieve_5760of30030_epar.c .
+COPY sieve_5760of30030_only_write_read_bits.c .
+COPY sieve_5760of30030_par.c .
+
+RUN /bin/sh ./compile.sh
```
marghidanu: If you set the permissions ahead then you shouldn't need to run with the sh interpreter:
`chmod a+x compile.sh`

and in the Dockerfile this should be only: `RUN ./compile.sh`

marghidanu:
Super cool, you got it! Super glad that worked for you, and you nailed it, I've only suggested some small changes hope they make sense.

PrimeC/solution_2/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+ENTRYPOINT [ "/bin/sh","./run.sh" ]
```
marghidanu: Same as above, make sure you set the execution flag.

marghidanu:
The slash at the end of `COPY *.c ./` is important it denotes that you are copying a group of files into a directory.

PrimeC/solution_2/Dockerfile
danielspaangberg: > Super cool, you got it! Super glad that worked for you, and you nailed it, I've only suggested some small changes hope they make sense.

Yes it definitiely helps. Unfortunately I have the files on a windows system, and I'll read up on how to tell git they are executable. Still very new to both git (and docker).

PrimeC/solution_2/Dockerfile
danielspaangberg: It

> The slash at the end of `COPY *.c ./` is important it denotes that you are copying a group of files into a directory.


I'll try to fix it, did not show up when I executed docker locally. Thanks for checking and telling me!

PrimeC/solution_2/Dockerfile
marghidanu: You can do it directly in the Dockerfile if you want: `RUN chmod a+x *.sh`

marghidanu: Also if I recall correctly git has a Bash console under Windows which allows you to use all the Linux Shell commands like: `chmod`, `chown`, etc...

danielspaangberg: > You can do it directly in the Dockerfile if you want: `RUN chmod a+x *.sh`

True. That's a bit neater. Will try!

marghidanu:
Super cool! That's a wrap!

marghidanu:
@danielspaangberg  If it's ok with you I will merge this.

danielspaangberg:
> @danielspaangberg If it's ok with you I will merge this.

Yes, of course! Thanks for all the help! :)

PR #155, opened 2021-04-21
MIXAL implementation update with drag-race compliant output format

rbergen:
## Description

This is an updated version of the MIXAL implementation submitted before. It adds a wrapper shell script that:
* Times the execution of the core implementation run
* Embeds the measured time in the core implementation output, using a newly introduced placeholder in the latter
* Filters out all other lines of the output
* Turns the output into lower case.

With this, the wrapper shell script output conforms to the drag-race output format requirements.

Also:
* An updated Dockerfile that uses the wrapper script has been provided.
* An updated README.md has been provided.
* A bug in the end result output code was fixed. 
* All source files (MIXAL program, wrapper script and Dockerfile) have been marked as binary, to prevent further line end character (conversion) issues.

## Contributing requirements


rhbvkleef:
Awesome :)

PR #150, opened 2021-04-19
PrimeRust: PrimeSieve Rust implementation.

Azgrom:
PrimeSieve Rust implementation.

Based this code on PrimeSievePy, PrimeCPP and on #107 @joshallen64's Rust pull request.

Just got the following results:
Passes: 11630, Time: 10.000201941s, Avg: 859.862µs, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: true

My specs are: i7 9750H, 32GB RAM, Manjaro KDE

Project created and managed with [Cargo](https://www.rust-lang.org/tools/install)

Test was run with: `$ cargo run --release`
Code formatted with: `$ cargo fmt`

PS: _There is a little bug with the counts at 1000 limit. When run at 1_000_000 limit the counting goes perfect._

PR #161, opened 2021-04-23
Rust Simple implementaition

Blui42:
## Description

This simple rust implementaion is very strongly oriented towards the original code.

Expected output is:
- std-err: Author; Passes; Time; Threads
- std-out: Blui42; (passes); (time); 1
- std-err: (passes/time) Passes per second
- std-err: For Validation: (count) Primes were found
If further validation is needed, line 61 can be uncommented to print all the primes to std-err.

## Contributing requirements


Azgrom:
You see, there was another solution_2 going on at #150...

Even so, your code is quite straightforward. I would only make the printed results more clear, and separate declaration and definition concerns in a different file.

```rust
$ cargo run --release
   Compiling primes v0.1.0 (/home/rafael/Documents/Repos/Blui/PrimeRust/solution_2)
    Finished release [optimized] target(s) in 1.08s
     Running `target/release/primes`
Blui42; 2532; 5.000130272; 1
78498
```
For example. I am assuming `2532` is the number of passes. But I also coded this exercise. Other person might not be able to infer it.

marghidanu:
I think just renaming this to solution_3 should be enough to resolve this conflict. The output format is documented in CONTRIBUTING.md, this is to facilitate the automated reporting script to parse the output.

Blui42:
- Changed name from solution_2 to solution_3
- Clarified output in std-err

New output:
```
Author; Passes; Time; Threads
Blui42; 4032; 5.000549; 1
806.31 Passes per second
For Validation: 78498 Primes were found
```

I cannot change `Blui42; 4032; 5.000549; 1` as it is required by this rule:

> The test code outputs the following text to standard output:
> 
> `<name>;<iterations>;<total_time>;<num_threads>`

PrimeRust/solution_3/src/main.rs
```diff
@@ -0,0 +1,68 @@
...
+            }
+            for num in (factor*factor..=self.size).step_by(factor*2){
+                self.bits[num] = false;  // mark all multiples of this number as not prime
+            }
+            factor += 2;
+        }
+    }
+    pub(crate) fn print_results(&self, duration: f64, passes: u32){
+        eprintln!("Author; Passes; Time; Threads");  
+        println!("Blui42; {}; {}; 1", passes, duration);  // print required output to standard out
```
rhbvkleef: Ideally, it should be this:

```suggestion
        println!("Blui42;{};{};1", passes, duration);  // print required output to standard out
```

rhbvkleef:
I want to either see that a Dockerfile is added here, or wait for #168 and add a Dockerfile from it.

rbergen:
I think either Dockerfile in the existing Rust solutions are a good starting point for this solution; it should not be difficult to create a working Dockerfile for this solution starting with either of those. 

@Blui42: will you add one?

Blui42:
After getting frustrated trying to read the docs of docker for way longer than I should, I realized I only need to add a dockerfile, not a docker build.

Oh well...

rbergen:
It seems like the Docker run image can't find/copy the executable it's looking for. Could it be that it's called "primes" instead of "prime_rust", in line with the package name you specified in Cargo.toml?

Blui42:
Oh, right. Docker for some reason is not recognized as installed even though it is, so I had no way of testing. Will be fixed in less than five minutes.

PrimeRust/solution_3/README.md
```diff
@@ -0,0 +1,38 @@
+# Rust solution by Blui42
+
+Faithful
```
rbergen: Please replace "Faithful" with:
```
![Category](https://img.shields.io/badge/Category-faithful-green) 
```

rbergen:
I can confirm that the Dockerfile builds and runs. I found one last tiny thing in the README.md that I just commented on. Once that's in place I think we can merge this.

rbergen (approved):
@rhbvkleef or @marghidanu To me this looks good to merge. Please merge if you agree.

PR #125, opened 2021-04-12
Add a Prime Sieve written in Haxe.

eth0fox:
Tried to match the Python implementation as closely as possible.

# Readme:

## What's Haxe?

See: https://haxe.org/

Haxe is a programming language that while it does have it's own interpreter, can also be cross-compiled to other languages such as C++, Python, C# & more.

## Running the Prime Sieve

1. Make sure you have a working [Haxe](https://haxe.org/) installation.

### Haxe Interpreter

2. `haxe interp.hxml`

### Python

2. `haxe python.hxml`
3. `python3 bin/py.py`

### C++

2. `haxe cpp.hxml`
   1. If you get a `Error: Library hxcpp is not installed`, run `haxelib install hxcpp`
3. `./bin/cpp/Main`

## My personal results.

Take these with a grain of salt. My machine is a MacBook Pro 2020 (M1, 8GB RAM) running macOS Big Sur.

**Haxe Interpreter**: `Passes: 48, Time 10.0865981578826904, Avg: 0.21013746162255606, Limit: 1000000, Count: 78499, Valid: false`

**C++**: `Passes: 2071, Time 10.0035741329193, Avg: 0.00483031102506968, Limit: 1000000, Count: 78499, Valid: false`

**Python**: `Passes: 27, Time 10.352768182754517, Avg: 0.38343585862053764, Limit: 1000000, Count: 78499, Valid: false`

PR #162, opened 2021-04-24
Add Haxe Prime Sieve

eth0fox:
## Description

I was unsure of the requirements when I first created the pull request (#125). This is an updated PR that merges to drag-race instead of main.

## Contributing requirements


marghidanu:
Would it be possible to include a Dockerfile for the implementation?

eth0fox:
@marghidanu I’ve never used Docker so I’m not exactly sure how to

rbergen:
Coincidentally, I was just looking into this. It turns out that Haxe is not that easy to get to run in Docker, but I did get it to work in the end. 

A file called "Dockerfile" will need to be added to the solution directory, containing the following:
```
FROM haxe:4.2.1-alpine3.13

RUN apk update && apk add --no-cache g++ python3 musl-locales bash
RUN ln -s /usr/include/locale.h /usr/include/xlocale.h

WORKDIR /opt/app
COPY . .

RUN haxelib install hxcpp
RUN haxe cpp.hxml
RUN haxe python.hxml

COPY run.sh .
RUN chmod a+x run.sh

ENTRYPOINT [ "/opt/app/run.sh" ]
```

A small shell script needs to be added, as well. It needs to be called "run.sh" and contain the following:
```
#!/bin/bash

cd $(dirname "${BASH_SOURCE[0]}")

./bin/cpp/Main
haxe interp.hxml
python3 bin/py.py
```

eth0fox:
I have added a Dockerfile, which *should* work. The interpreter and Python versions run, but the CPP failed to build, as I was running it on an ARM machine and Haxe generated build files for x86_64.

marghidanu:
> I have added a Dockerfile, which _should_ work. The interpreter and Python versions run, but the CPP failed to build, as I was running it on an ARM machine and Haxe generated build files for x86_64.

Are you using an M1, or ARM CPU?

marghidanu:
> Coincidentally, I was just looking into this. It turns out that Haxe is not that easy to get to run in Docker, but I did get it to work in the end.
> 
> A file called "Dockerfile" will need to be added to the solution directory, containing the following:
> 
> ```
> FROM haxe:4.2.1-alpine3.13
> 
> RUN apk update && apk add --no-cache g++ python3 musl-locales bash
> RUN ln -s /usr/include/locale.h /usr/include/xlocale.h
> 
> WORKDIR /opt/app
> COPY . .
> 
> RUN haxelib install hxcpp
> RUN haxe cpp.hxml
> RUN haxe python.hxml
> 
> COPY run.sh .
> RUN chmod a+x run.sh
> 
> ENTRYPOINT [ "/opt/app/run.sh" ]
> ```
> 
> A small shell script needs to be added, as well. It needs to be called "run.sh" and contain the following:
> 
> ```
> #!/bin/bash
> 
> cd $(dirname "${BASH_SOURCE[0]}")
> 
> ./bin/cpp/Main
> haxe interp.hxml
> python3 bin/py.py
> ```

Why do you need the Bash file? You can always inline commands with `bash -c "..."`

eth0fox:
> > I have added a Dockerfile, which _should_ work. The interpreter and Python versions run, but the CPP failed to build, as I was running it on an ARM machine and Haxe generated build files for x86_64.
> 
> Are you using an M1, or ARM CPU?

I use an M1 machine.

rbergen:
> Why do you need the Bash file? You can always inline commands with `bash -c "..."`

"Need" is indeed the wrong word. I prefer to use a script if the run encompasses multiple commands, because I personally tend to lose oversight more quickly when using the inline command approach. One might say it's a matter of taste.

rbergen:
LGTM. @marghidanu, I suggest we merge despite @TayIorRobinson's comment on issues on M1. That's because
1. the solution works on amd64
2. compatibility with specific architectures is not a contributing requirement
3. I already know that other merged solutions _will_ have similar issues on non-amd64 architectures.

PR #163, opened 2021-04-24
4 FreeBASIC implementations

rbergen:
## Description

This PR adds four implementations in FreeBASIC, exploring four different approaches:
- One based on the "8 of 30" implementation that [mckoss](https://github.com/mckoss) has written in C
- One based on manipulating the bits in 64-bit unsigned integers
- One based on manipulating the bits in 32-bit unsigned integers
- One using a straightforward array of booleans

Main conclusion for me based on the results, as documented in the READMEs is that I shouldn't try to be too clever when tracking "trues" and "falses", at least in this particular language.

On a personal note, I wrote my first computer programs on graphing paper 35 years ago when I was 9, in MSX BASIC. I wrote them on graphing paper because I found a book in the library about MSX BASIC but had no computer available to try them on. It was oddly nostalgic and actually quite enjoyable to once again type in things like For and Next, and While and Wend.

## Contributing requirements


rbergen:
I just noticed the PrimeC/solution_2 structure, where multiple approaches are submitted in one solution. If you prefer that setup then I can do the same here, also wrapping up the build and execution of the various solutions in one Docker image. For one, it will create less bloat on hard disks related to FreeBASIC .tar.xz's.
Of course, the solution output will then consist of multiple lines, each still conforming to contribution guidelines. 

Just let me know what your take on this is.

rbergen:
I gave it some more thought, and decided to just roll with it. If you end up disagreeing, I can easily revert.

marghidanu:
I wonder how we can solve the `.tar.xz` file issue. Maybe an external image that would be re-used here. I’m not sure about binaries in the repo.

rbergen:
Yeah, I don't like it either, hence the remark. I guess you just confirmed it really is a bad idea. Let's park this PR until I've looked into this.

marghidanu:
Should we build a separate free basic image and use that as a base for this implementation?

rbergen:
I think I'll do exactly that, in the shape of a Docker Hub base image that is built from a public GitHub repo, under my account. That way it is easy to see there is no "black voodoo" going on in the background, while a) keeping this repo clean of collateral and b) leaving responsibility for making this commit work where it belongs (i.e., with me). I just have to see how all of that works with the CI workflows that are in place, here. As I've enabled those workflows in my fork, I can do that without having to bother the maintainers of this repo. 
I'll just leave another comment here when I think I've sorted it out. :)

rbergen:
Looking at the workflow run in my fork, it seems it's working as intended.

marghidanu:
LGTM!

PR #165, opened 2021-04-26
Implementation in Visual Basic

rbergen:
## Description

This is an implementation in Visual Basic.

This is the final programming language I will submit a solution for. As indicated, I am very happy to help with the project in other ways, if desired.

## Contributing requirements

### DIFF
--- a/PrimeBASIC/solution_2/.gitattributes
+++ b/PrimeBASIC/solution_2/.gitattributes
@@ -1,0 +1,3 @@
+PrimeVB.* binary
+*.vb binary
+Dockerfile binary
--- a/PrimeBASIC/solution_1/prime_bit32.bas
+++ b/PrimeBASIC/solution_1/prime_bit32.bas
@@ -114,7 +114,7 @@ Function PrimeSieve.RunSieve() As BitArray Ptr
 
         If (Number > SieveSqrt) Then Exit For
 
-        For Number = Factor * 3 To SieveSize Step Factor * 2
+        For Number = Factor * Factor To SieveSize Step Factor * 2
             This.NotPrimeArray->SetBit(Number \ 2)
         Next
     Next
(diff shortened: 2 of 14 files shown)


eth0fox:
This would be interesting to test the C# compiler V the VBNet compiler, however I suspect the results will be the same, if not the same as they both compile down to IL.

rbergen:
If I remember correctly, performance of VB vs C# is similar, but not quite identical. I guess we'll see more reliable numbers when both get tried in the drag race proper. What I did remember noticing is that the F# implementation (obviously also a .NET language) was noticeably faster than VB, on the same machine.

PrimeBASIC/solution_2/.gitattributes
```diff
@@ -0,0 +1,3 @@
+PrimeVB.* binary
+*.vb binary
+Dockerfile binary
```
rhbvkleef, line 3: These attributes are wrong? Why are you marking them as `binary`? I reckon they are all text?

rbergen: I've been having weird-ish errors due to CR/LF line ending conversions in a number of solutions. I'm using the "binary" attribute to just force things to stick to LF whenever I suspect that to cause an issue I come across. That's also because it doesn't seem to hurt anything else down the line. If it bothers you then I'm happy to pull the file out of the commit and see what happens.
Just for sake of completeness, I do note that my F# solution was merged earlier with the .gitattributes file in place.

rhbvkleef: Ah... The joys of line-endings :) I see why you did it. I believe you can force CRLF endings using gitattributes too, but I can live with this.

rhbvkleef: It's
```
GLOB	text eol=crlf
```

rbergen: Thanks for going through the trouble of finding this for me, I do appreciate it. To be frank, I still think I will stick to marking things as binary when things go haywire. My experiences with things getting clever about what's text and what isn't (even things as clever as git) are just not that positive, overall.

PR #168, opened 2021-04-28
Make "older" solutions conform to current contribution guidelines

rbergen:
## Introduction
I apologize up-front for the wall of text that this sentence is the start of. Considering the goal and extent of the PR, I do however think it is important to motivate its contents.

## Purpose
This PR aims to make solutions that were submitted to the repo before the current contributing guidelines were in place/applied, conform to those guidelines.

## Approach
When authoring this PR, I used the following approach:
1. **README.md**: 
    * I added a skeleton README.md to all solutions that didn't have one. It consists of a header mentioning the language and the author of the solution, as well as a category badge (see next point). The main reason for adding these README.md's is that category and author otherwise get lost when the "old" directory naming scheme (&lt;category&gt;/&lt;user&gt;) is replaced with the new one (solution_&lt;n&gt;).
    * I added a category badge to all README.md's that did not have one, based on the badge that was present in the README.md for the Crystal solution.

2. **Directory names**: I renamed all solution directories to solution_&lt;n&gt;, with n starting at 1. I am aware of [PR 138](https://github.com/davepl/Primes/pull/138) where a 0-based numbering scheme was chosen, but I believe the repo has settled on 1-based numbering, by now. This is probably partly due to the fact that the contributing guidelines are most easily interpreted as stating this to be the rule.

3. **Result output**: for those solutions that did not output a result line in the format stipulated by the guidelines, I added three lines to the solution's output/print routine (as in, the source code):
    1. The first line is a comment that states that I added the next two lines, and why. 
    2. The second line adds an empty line to the output. If the solution already included code to print an empty line/line break, I copied that. Otherwise, I used the approach that was documented in a representative piece of documentation/example code for the language in question.
    3. The third line outputs the execution result in guideline format. For this line, I followed the approach the solution used for its existing output. I did add format string and parameters (pass count and duration) on one line, to fix the number of added source code lines to 3, with the comment line included.
  
4. **Dockerfile**: I added a Dockerfile for the solutions that didn't contain one. When composing these, I used the following considerations:
    * When choosing a base image, my "selection algorithm" was as follows:
        1. If an official Docker Hub image exists for the language, I picked that. If multiple images are available with different underlying Linux distros, I picked the distro based on the next steps in this list.
        2. Otherwise, if I could get the solution to build and run on Alpine 3.13 using Alpine 3.13 packages, I did that.
        3. Otherwise, if I could get the solution to build and run on Ubuntu 18.04, I did that.
        4. Otherwise, I chose a base image that I was able to get the solution to build and run with. I've mentioned solutions where I ended up applying this choice, under "Notes/exceptions", below.

    * I tried to use common sense when choosing between explicit file inclusion in the image, vs `COPY . .` with an appropriate .dockerignore.

## Notes/exceptions

- **PrimeAda/solution_1**: I had to make a minor change to the original output code to make the solution build/run using GPRbuild/GNAT. The error I got indicated that the Image attribute had to be applied to a type and not to functions, where it was in two cases. As the Ada documentation I could find corrobarated this and the change I had to make was very limited, I concluded it was justified to apply it.
Also, I renamed the file README to README.md and added the language and author header.
- **PrimeAssemblyScript/solution_1**: the solution did include a README.md, but it lacked a language and author header, which I added.
- **PrimeCPP/solution_2**: this solution originally resided in the /PrimeCPP_PAR directory. As it is a C++ implementation, I felt it was best to move it to a second solution folder under the existing C++ language directory.
- **PrimeCSharp/solution_1**: I did not compose a Dockerfile, as there was an [open PR (151)](https://github.com/davepl/Primes/pull/151) for this solution when I was preparing this PR.
- **PrimeDelphi/solution_1**: I did not compose a Dockerfile for this solution, because it depends on Delphi, which is a commercial product. The terms of use of the Community Edition that is available are such that I consider the risk too great that we may end up unintentionally violating the license. 
- **PrimeGo/solution_1**: the solution did include a README.md, but it lacked a language and author header, which I added.
- **PrimeJava/solution_1**: the solution was setup to be built using [Gradle](https://gradle.org/). I made an attempt at putting a Dockerfile together that would indeed use Gradle to perform the build, but was unable to do so. As the actual source code is contained in one .java file, I decided to perform the compile using a "plain" javac invocation, instead. The run command has been adapted accordingly. 
- **PrimeLisp/solution_1**: the source code timed the run in such a way that the measured/reported time had a granularity of 1 second. I changed the code to use a more fine-grained unit for measuring and reporting time.
- **PrimePHP/solution_1**: the source code reported a duration that was effectively rounded to full seconds, where it intended to report duration rounded to milliseconds. I have modified the code so that the existing output reports rounded milliseconds, and the guideline-compliant output reports unrounded time.
- **PrimeSwift**: I deleted the .swiftpm directory, which seems to contain IDE settings that are unnecessary for the solution to build and run. My guess is that this directory should have been included in the .gitignore file of the original submitter.
- **PrimeZig/solution_1**: I used Archlinux as the base image for Docker. "Regular" Alpine 3.13 packages for Zig don't exist, and on Ubuntu, Zig is available as a snap package. Snap is practically impossible to get to work within Docker.  

## Solution status

With the changes in this PR applied, all "older" solutions now build and run in the CI and Benchmark workflows, except PrimeDelphi/solution_1 for reasons indicated.


PrimeJava/solution_1/.gitattributes
```diff
@@ -0,0 +1,3 @@
+* binary
```
rhbvkleef: I really don't like this kind of thing, At least with Java there is (or at least should not be) a problem with line-endings chainging, right?

rbergen: I understand. I never add this sort of thing unless on my machine, I hit a snag that relates to line endings. Again, I can drop the .gitignore and then we can see how far we get without it.

PrimeGo/solution_1/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM golang:1.16-alpine3.13
+
+WORKDIR /opt/app
+
+COPY go.mod .
+COPY main.go .
+
+ENTRYPOINT [ "go", "run", "main.go" ]
```
marghidanu: This has a performance penalty. We should build it separately and then run it.

marghidanu (commented):
I see a lot of permissions changes from 755 to 644

rbergen:
> I see a lot of permissions changes from 755 to 644

I'm not sure I know exactly what you're referring to.

PrimeGo/solution_1/Dockerfile
rbergen: Good spot, thank you. I've just pushed a new Dockerfile that does precisely that.

marghidanu:
In the Changes tab I see some files have the permissions changed:

```
PrimeCrystal/solution_1/run.sh  100755 → 100644
```

also 

```
PrimeCPP/stl-faithful/davepl/run.sh → PrimeCPP/solution_1/run.sh  100755 → 100644
```

rbergen:
That was no deliberate action on my part. My guess is that this is caused by the fact that I originally cloned to an NTFS partition, which is "blissfully unaware" of execution permission bits and may have cleared them when pushing. 

I'm now hosting the repo on an ext4 partition which means I could restore permissions to include a+x. However, to be honest, I'm actually not sure if relying on execution permissions to be set is the "fault-tolerant" thing to do, partly because of what I just said.

marghidanu:
It might be that some of the functionality in Dockefiles relies on that, also it's a feature in Git I think these should stay as they are.

rbergen:
The Dockerfiles have been tested to work with the permissions as they are in my commit (both CI *and* benchmark), so we already know nothing got broken there.

What remains then is I think a more conceptual consideration. Your perspective on that is clear, mine is different. Besides what I already said, from a security perspective I actually don't like anything to be executable that I pull from Github (or any other CSM, for that matter).  
Maybe @rhbvkleef can provide his perspective on this, too? I will up-front accept a majority vote on this, whatever the outcome.

marghidanu:
The fact that NTFS is altering the permissions it's a bit concerning especially if it's not intended.

rhbvkleef:
I'm not sure why this is happening. I also develop a lot with Windows devs, and I've never seen NTFS do this. Oh well...

rbergen:
Just to be clear, the NTFS thing is what I can think of, considering that I did not consciously do this. I may be barking up the wrong tree, entirely.

EDIT: I noticed that in all cases, the permission change goes together with a move (rename) related to the directory structure (old vs new). I think that does make the file system the most likely culprit.

Anyway, now that that is out of the equation, what shall we do with the executable bit?

marghidanu:
I would suggest keeping those as they were. Looks like the move operation is no preserving the permissions, good to know that in the future. We learn something everyday :)

rbergen:
Done.

PR #169, opened 2021-04-30
Updated version of the contribution guidelines

rbergen:
This is an update to the contribution guidelines that are currently part of the main branch. The following notes apply:
* I have based this version on the current guidelines, "established standards" that have developed within the drag-race branch over the past few weeks, and changes that were applied to the source tree with PR #168. 
* After review, approval and merge of this PR on the drag-race branch, I will open a PR for the updated guidelines on the main branch, to ensure we publish the same guidelines on both branches.
* Consistency across branches is also why the addition to the README.md included in this PR, refers to the drag-race branch when the README.md in this PR will itself be on the drag-race branch.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -134,7 +134,7 @@ Note that we cannot guarantee that we can help make your solution mergeable if i
    where:
    * `name` is **at least** your username, and if you have multiple implementations, a short keyword to discriminate each implementation
    * `iterations` is the amount of times your code ran in the allotted time
-   * `total_time` is the total time it took to run, which would be sligtly more than 5 seconds, in an `en_US` formatted decimal value (so, use `.` (period) and not `,` (comma) as the decimal separator)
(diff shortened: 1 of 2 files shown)


marghidanu (commented):
If we're planning on having a Raspberry Pi runner we should consider that the Docker images should support arm64 architecture.

rbergen:
> If we're planning on having a Raspberry Pi runner we should consider that the Docker images should support arm64 architecture.

I'm a big proponent of (also) running solutions that are arm64-compatible on arm64, but I think that making arm64 compatibility an actual requirement could well exclude too many (existing and future) solutions; arm64 is simply not as widely supported as amd64 is. With amd64-based runner hardware available, I also don't think it is needed to exclude solutions that don't support arm64; they can still run on the amd64 hardware we have.

All that said, I **am** very interested to see how the two architectures compare, performance-wise :)

marghidanu:
Correct, I know that Alpine images support the arm64 architecture, I've just double-checked now and it seems that Ubuntu also supports multiple architectures. In the case of Crystal lang, I intentionally didn't use the Crystal image base because there was no support for arm64.

PR #170, opened 2021-04-30
Report generator

marghidanu:
## Description

Initial work on the report generator.


_tools/bin/report.py
```diff
@@ -0,0 +1,127 @@
...
+formatter = logging.Formatter("%(levelname)s: %(message)s")
+handler.setFormatter(formatter)
+log.addHandler(handler)
+
+# Line parsing rules
+# pylint: disable=line-too-long
+RULES = [
+    r"^(?P<label>.+)\s*;\s*(?P<passes>\d+)\s*;\s*(?P<duration>\d+([.]\d+)?)\s*;\s*(?P<threads>\d)$",
+    r"^Passes:\s+(?P<passes>\d+),?\s+Time:\s+(?P<duration>\d+[.]\d+),?\s+Avg:\s+(\d+[.]\d+),?\s+Limit:\s+(\d+),?\s+Count1:\s+(\d+),?\s+Count2:\s+(\d+),?\s+Valid:\s+(true|false|1|0|Yes|No)$",
+]
```
rbergen: I see you've included a rule to parse a common variation of the "original" output format. I'm wondering if:

1. This is (still) necessary? I ask this because all solutions now have a line of output in place that conforms to the contributing guideline format, which is covered by the first rule. Furthermore, generating an output line in guideline format is a requirement for any new solutions that are added.
2. If this rule being activated would actually lead to a "working" report. I ask this because the "label" cannot be parsed from this format, and because I've seen some variations in the (format of) the values in this "old" format, in the solutions that create output according to it. A prime example of that is a solution I came across that outputs the Time in (rounded) milliseconds instead of seconds.

marghidanu: > 1. This is (still) necessary? I ask this because all solutions now have a line of output in place that conforms to the contributing guideline format, which is covered by the first rule. Furthermore, generating an output line in guideline format is a requirement for any new solutions that are added.

During development, I had access to both output types, for convenience I've added the original output too. If the output is consistent now I'll gladly remove the rule and attached logic.

> 2. If this rule being activated would actually lead to a "working" report. I ask this because the "label" cannot be parsed from this format, and because I've seen some variations in the (format of) the values in this "old" format, in the solutions that create output according to it. A prime example of that is a solution I came across that outputs the Time in (rounded) milliseconds instead of seconds.

Yes, labels are a bit inconsistent, some people (including ourselves) added the GitHub handle as a label, others included the solution type. As for the duration format, we can detect if the number is milliseconds or seconds just by looking at the number of digits. I have no preference for seconds or milliseconds, I do wish there would be only one type of value.

rbergen: > During development, I had access to both output types, for convenience I've added the original output too. If the output is consistent now I'll gladly remove the rule and attached logic.

In my opinion, it is better to remove the second regex. If there are still any solutions that do not write their output in the correct guideline format, then we should fix that within those solutions (i.e., at the root of the problem.) If such solutions are still in the drag-race branch, that would basically mean that I overlooked them when working on PR #168. Therefore, I would be more than happy to correct that myself.
I would suggest dropping the second regex, and seeing if there are any solutions that then "drop out" of the results. 

> Yes, labels are a bit inconsistent, some people (including ourselves) added the GitHub handle as a label, others included the solution type. 

That is true, but it is also in line with our own guidelines. The guidelines say that **at least** one's username should be added, and it is optional to add a short keyword to distinguish between multiple solutions of one submitter. Personally, I have added a keyword (only) if I submitted multiple solutions/implementations *for one programming language*. 
In any case, my original comment was not about inconsistencies in the labels, but about the fact that the second parsing rule (the one I recommend we drop) can't parse a label at all. That is actually one of the reasons why I think the second parse regex should be dropped. 

> As for the duration format, we can detect if the number is milliseconds or seconds just by looking at the number of digits. I have no preference for seconds or milliseconds, I do wish there would be only one type of value.

All output lines that conform to the guideline format do output in seconds. If we drop the regex that parses the old format, we immediately don't have to worry any more about the solutions that use milliseconds *in their old output*.

marghidanu: Ok, I will remove the additional rule for parsing old output.

rbergen:
LGTM

rbergen (approved):

PR #172, opened 2021-05-01
TypeScript implementation

marghidanu:
## Description

TypeScript implementation

## Contributing requirements

### DIFF
--- a/PrimeTypeScript/solution_1/src/index.ts
+++ b/PrimeTypeScript/solution_1/src/index.ts
@@ -89,7 +89,7 @@ while (true) {
   sieve.runSieve();
 
   passes++;
-  const duration = Math.floor((Date.now() - startTime) / 1000);
+  const duration = (Date.now() - startTime) / 1000;
   if (duration >= 5) {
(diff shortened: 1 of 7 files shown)


PrimeTypeScript/solution_1/src/index.ts
```diff
@@ -0,0 +1,97 @@
...
+
+let passes = 0;
+const startTime = Date.now();
+
+while (true) {
+  const sieve = new PrimeSieve(1e6);
+  sieve.runSieve();
+
+  passes++;
+  const duration = Math.floor((Date.now() - startTime) / 1000);
```
rbergen: I don't understand the use of Math.floor. This makes you round the duration measurement to whole seconds, which will always be a flat 5. The actual durations I've seen on my system after taking the Math.floor out, are all some milliseconds more than that.

marghidanu: You're right, I need to take it out because we might end up losing almost an entire second (or close to that).

rbergen (approved):
LGTM

PR #173, opened 2021-05-03
Explicitly convert duration to double.

OMGtechy:
## Description

Since we're only expecting values in the order of 5, this should be
fine. This fixes the warning C4244 on MSVC Version 19.28.29914.

Machine code produced is the same before and after.

## Contributing requirements

* [Yes] I read the contribution guidelines in CONTRIBUTING.md.
* [N/A] I placed my solution in the correct category folder.
* [N/A] I added a README.md or Dockerfile.
* [N/A] I selected `drag-race` as the target branch.
* [Yes] All code herein is licensed compatible with BSD-3.

### DIFF
--- a/PrimeCPP/PrimeCPP.cpp
+++ b/PrimeCPP/PrimeCPP.cpp
@@ -127,7 +127,7 @@ int main()
         passes++;
         if (duration_cast<seconds>(steady_clock::now() - tStart).count() >= 5)
         {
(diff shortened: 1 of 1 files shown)


rbergen:
As indicated in the repository README, we don't merge PRs that are opened on the "original" branch. 

With relation to the change you're proposing, the issue in the code is that it performs an integer divide where it shouldn't. For the drag-race branch, we're going to fix this with #220.

rbergen:
Closing this after merging #220.

PR #176, opened 2021-05-03
Faster OCaml Docker builds

marghidanu:
Initially, the OCaml Docker builds were between 8-9 minutes. I've created a separate image to deal with the dependencies and used it as a base image. The repository for the base images is available here: https://github.com/marghidanu/docker-ocaml

Pre-built images are available for **amd64** and **arm64**.


rbergen:
Size-wise that image is still a monster, but it sure builds a *lot* quicker!

LGTM.

marghidanu:
I traded space over time, don’t know what to do about that. It looks like the entire stdlib is one package :/

rbergen:
Space over time usually is the trade-off, of course. :) 
This works fine.

PR #180, opened 2021-05-09
Added Dart Implementation (v2)

eagerestwolf:
## Description

This is attempt number 2 at creating a Dart implementation. I made some critical git errors on the last PR and count not amend it to rectify them. So, here is a new implementation that fixes everything and has all of the drag-race rules adhered to. This algorithm is STL-faithful

## Contributing requirements

* [X] I placed my solution in the correct category folder.
* [X] I added a README.md or Dockerfile.


PrimeDart/solution_1/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+WORKDIR /app
+
+ADD pubspec.* /app/
+RUN pub get
+ADD . /app
+RUN pub get --offline
+
+CMD []
+ENTRYPOINT [ "/usr/bin/dart", "bin/PrimeDart.dart" ]
```
rbergen: I see you chose not to use the compiled option for your Docker image. I just want to make sure that is indeed what you want to do, as it is very likely that whatever approach you use in your Dockerfile is also the one that will be used for the benchmark runs we are in the process of setting up. If you want the best performance that Dart can offer to show up in the benchmark reports, the compiled option may therefore be a better option, here.

eagerestwolf: Valid point, I actually fixed that and I'm glad I did. It now runs 126% faster. I have updated the README as well to reflect the new results. It never occurred to me to do that.

PrimeDart/solution_1/bin/PrimeDart.dart
```diff
@@ -0,0 +1,218 @@
...
+    // This is just for code readability. Since stdout.write doesn't print a new
+    // line at the end of the string, we can split a large string across a few
+    // print requests to keep the code to under 80 columns (personal preference).
+    stdout.write('Passes: $passes, Time: $duration, ');
+    stdout.write('Avg: ${duration / passes}, Limit: $_sieveSize, ');
+    stdout.write('Count1: $count, Count2: ${countPrimes()}, ');
+    stdout.write('Valid: ${_validateResults()}\n');
+
+    // These 2 lines are for the drag race format
+    stdout.write('\n');
```
rbergen: My apologies for missing this earlier, but would it be possible to send this output to stderr? It makes it easier for us to distinguish between drag-race and non drag-race output when reporting.

eagerestwolf: Done deal, that's probably also partly my fault I must have missed that in the drag race rules. On the bright side, I did find and squash a bug in the process. I forgot to mark the `bits` list as being `late` since it's not initialized until the constructor is called.

eagerestwolf:
Sorry, don't mean to keep pushing things. I should probably rebase and squash all my commits, but I figured out why `Count1` was always displaying `1`, and this is the reason:

```cpp
for (int num = 3; num <= sieveSize; num+=2)
          {
              if (Bits[num])
              {
                  if (showResults)
                      printf("%d, ", num); // <-- here is the problem
                  count++;
              }
          }
```

This is why I ALWAYS use braces around block statements. For whatever reason, my brain interpreted that the `count++` statement belonged with the `printf` statement, but it doesn't. Those are 2 independent things. So, that's fixed. Also, the Dark SDK wanted to complain (it throws a warning) that `bits` wasn't marked as late, so I just marked it as late and didn't bother testing it. Turns out with `bits` marked as late it generates an error (lists aren't null safe), so I just removed it. It'll throw a warning when compiled that's all.

PrimeDart/solution_1/bin/PrimeDart.dart
rbergen: I'm sorry for the confusion/back and forth, but the change I meant to ask for is the other way around. :) As in, _(only) the drag-race format output_ goes to `stdout`, _everything else_ goes to `stderr`. That's why I marked the "classic output" lines in my last review, but I could have been more clear on exactly what I meant.

rbergen:
I just wanted to note that there's no need to rebase and squash, as I'll squash on merge.

PrimeDart/solution_1/bin/PrimeDart.dart
eagerestwolf: Oh, my apologies. No worries man, give me just a few moments and I'll have it fixed.

eagerestwolf: Fixed

rbergen (approved):
LGTM

julemand101:
Hi @eagerestwolf 

I have some improvements to your solution which does make it quite a bit faster. Do you prefer a pull request on top of your own solution or should I summit my changes as a separate forked solution in its own directory?

eagerestwolf:
By all means, feel free to merge over my implementation, I am by no means a Dart expert.

PR #187, opened 2021-05-11
CI QEMU builds

marghidanu:
This should allow building Docker images for various platforms, for now just amd64. Later we can probably add arm64. Also, the jobs should now be listed alphabetically in the CI.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Build amd64
         uses: docker/build-push-action@v2
         with:
-          image: ${{ steps.solution-name.outputs.normalized }}
+          image: "${{ steps.solution-name.outputs.normalized }}:latest"
           push: false
           load: true
           context: ${{ matrix.solution }}
@@ -60,7 +60,6 @@ jobs:
       - name: Scan image
         uses: anchore/scan-action@v2
         with:
-          path: ${{ matrix.solution }}
-          # image: ${{ steps.solution-name.outputs.normalized }}
+          image: "${{ steps.solution-name.outputs.normalized }}:latest"
           severity-cutoff: critical
           # fail-build: true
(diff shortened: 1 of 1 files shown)


marghidanu:
Hold the merge, I will also add container scanning.

rbergen:
Fantastic! I'll hold on #186 until we've merged this. I'll then re-enable the Dockerfile in that PR, so we can test this with an actual solution. That also means that we'll then have an arm64 assembly solution in the branch without any missing pieces.

marghidanu:
Ok, let's skip the image scan; for now, I will add that separately on a branch that addresses a few security concerns. Please update your arm64 branch after this merge. We will need to add a special case for it and make sure we build the image using the linux/arm64 platform

rbergen:
> Ok, let's skip the image scan; for now, I will add that separately on a branch that addresses a few security concerns.

Ok. I will review this PR, later today.

> Please update your arm64 branch after this merge. We will need to add a special case for it and make sure we build the image using the linux/arm64 platform

I was wondering about that. Do I need to do anything specific in the arm64 assembly solution Dockerfile or the solution directory to mark it as arm64?

marghidanu:
> I was wondering about that. Do I need to do anything specific in the arm64 assembly solution Dockerfile or the solution directory to mark it as arm64?

We can mark it directly in the CI, if the solution name matches our label then we run the arm64 job otherwise de use the default.

rbergen:
That's certainly a pragmatic approach. I can't suppress the feeling that conceptually I prefer a more "dynamic" approach, but I guess it's okay to worry about that if and when we have more architecture-specific solutions.

rbergen (approved):
LGTM

marghidanu:
Actually I found my bug for the image scanning configuration and corrected it. Everything works fine, let's merge this and have it available for future builds.

rbergen:
LGTM, again

PR #189, opened 2021-05-13
Solution in arm64 assembly, including CI

rbergen:
## Description

This is an implementation in arm64 assembly. Notes:

- The PR also includes a few optimizations and fixes in the amd64 assembly solution I submitted earlier. This makes that both solutions are algorithmically the same.
- I included a new version of CI.yml, to make it select and use arm64 as the build platform for specific solutions (this being the furst and currently only). 

## Contributing requirements

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -48,7 +48,28 @@ jobs:
       - name: Setup Docker Buildx
         uses: docker/setup-buildx-action@v1
 
+      - name: Select arm64 build for specific solutions
+        id: arch-arm64
+        if: matrix.solution == 'PrimeAssembly/solution_2'
+        run: echo "::set-output name=build::true"  
+
+      - name: Select amd64 build for other solutions
+        id: arch-amd64
+        if: "!steps.arch-arm64.outputs.build"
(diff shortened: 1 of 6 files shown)


rbergen:
@marghidanu: I can also commit a CI.yml that checks for an "arch-arm64" flag file in the solution directory and selects the arm64 build if the file exists. If you'd like to take that approach, then I'll update the PR accordingly.

rbergen:
The commit builds arm64 and/or amd64 based on respective flag files, defaulting to amd64 if no flag files are present.

rbergen:
By the way, some unrequested information on why I trigger the arm64 and amd64 builds individually: I actually created a CI version that runs one, multi-arch build when both archs are requested (it's [here](https://github.com/rbergen/Primes/blob/arm64flagfile/.github/workflows/CI.yml) for your reference). I then ran into [this issue in buildx](https://github.com/docker/buildx/issues/59), however.

TL;DR: if you build a multi-arch image, you **must** --push it. Until that gets fixed (_if_ it gets fixed), the recommended approach is to build every arch individually for all purposes except production runs.

marghidanu:
Looks good! Let's merge this

PR #191, opened 2021-05-15
Small conformity fixes

rbergen:
- Based the PrimeRust/solution_4 Dockerfile on the same base image as the other Rust solutions
- Made PrimeZig/solution_1 put the guideline-compliant output line to stdout


marghidanu:
LGTM

PR #192, opened 2021-05-16
Running benchmarks locally

marghidanu:
Make sure you have Docker and make installed on your machine. Type `make` and watch it run!


marghidanu:
@rbergen I also need to retrieve the machine information and generate a JSON file with all the information required.

tools/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM alpine:3.13
+
+RUN apk add --no-cache build-base bash \
+    python3 python3-dev \
+    py3-pip py3-click py3-numpy py3-pandas
+
+WORKDIR /opt/app
+
+# COPY requirements.txt .
+# RUN pip3 install -r requirements.txt
```
rbergen, line 10: These are present but commented out in preparation for a future activation of the requirements.txt?

tools/tools/commands/report.py
```diff
@@ -1,49 +1,23 @@
-#!/usr/bin/env python3
```
rbergen, line 1: Is the execute bit set on this file? I ask because I remember that I experienced an error on the initial run of the benchmark solution, due to report.py not being executable.

rbergen:
> @rbergen I also need to retrieve the machine information and generate a JSON file with all the information required.

Is this a change you want to add after this PR is merged, or one we should wait for before merging?

Makefile
```diff
@@ -0,0 +1,25 @@
...
+ARCH_FILE  := ${shell case $$(uname -m) in x86_64) echo arch-amd64 ;; aarch64) echo arch-arm64 ;; esac}
+
+all: report
+	@echo "Output files available in $(OUTPUT_DIR)"
+
+benchmark: $(SOLUTIONS)
+	@for s in $(SOLUTIONS); do \
+		NAME=$$(echo "$${s}" | sed -r 's/\//-/g' | tr '[:upper:]' '[:lower:]'); \
+		ls $${s}/arch-* > /dev/null 2>&1; \
+		if [[ -z "$${s}/$(ARCH_FILE)" || -f "$${s}/$(ARCH_FILE)" || "$$?" -ne 0 ]]; then \
```
rbergen: I know I'm commenting on my own contribution, but this should be
`if [[ -z "$(ARCH_FILE)" || -f "$${s}/$(ARCH_FILE)" || "$$?" -ne 0 ]]; then \`

As in: the solution directory should not be considered when checking if ARCH_FILE is empty because with it, the check will always yield "false".

tools/Dockerfile
marghidanu: Yes, pandas took forever to install so for now I’ve used the alpine packages. But in the future we will need to use pip to install packages.

tools/tools/commands/report.py
marghidanu: report.py is not used anymore. It should be deleted

Makefile
marghidanu: :) will make the change for you directly on my branch

tools/tools/commands/report.py
rbergen: I now see that the combination of click and an ENTRYPOINT in the Dockerfile indeed remove the need for an executable report.py (and in fact, anything with +x at the Python level).

Makefile
rbergen: Thanks, I wasn't close to anything that could pull, edit, commit, push and open a PR when I noticed and wrote that. :)

rbergen (approved):
LGTM

PR #193, opened 2021-05-21
Follow-up changes to repo transfer

rbergen:
This PR includes:

- Changes applied to the old main branch, after the drag-race branch was created, In effect, this is a manual merge of those changes into drag-race, bringing it to par with "original"
- FreeBASIC, FreePascal and Rust solutions are now based on newly composed multi-arch Docker base images
- The amd64 assembly solution is flagged as amd64 only


marghidanu:
LGTM!

PR #194, opened 2021-05-22
Make C++ reported times accurate.

chriselrod:
## Description
On the drag-race branch, the C++ benchmarks round the reported times.
This PR reports them accurately.
On this PR, I get:
```sh
> g++ -march=native -mprefer-vector-width=512 -Ofast PrimeCPP.cpp
> ./a.out
Passes: 8584, Time: 5.000523, Avg: 0.000583, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

davepl;8584;5.000523;1
> cd ../solution_2
> g++ -march=native -mprefer-vector-width=512 -Ofast PrimeCPP.cpp -pthread
> ./a.out
Passes: 97272, Threads: 36, Time: 5.000744, Avg: 0.000051, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

davepl_par;97272;5.000744;36
```

### DIFF
--- a/PrimeCPP/solution_1/PrimeCPP.cpp
+++ b/PrimeCPP/solution_1/PrimeCPP.cpp
@@ -129,9 +129,11 @@ int main()
         prime_sieve sieve(1000000L);
         sieve.runSieve();
         passes++;
-        if (duration_cast<seconds>(steady_clock::now() - tStart).count() >= 5)
+	auto tEnd = steady_clock::now() - tStart;
+        if (duration_cast<seconds>(tEnd).count() >= 5)
         {
-            sieve.printResults(false, duration_cast<microseconds>(steady_clock::now() - tStart).count() / 1000000, passes);
+  	    std::chrono::duration<double, std::micro> tEnd_micro = tEnd;
+            sieve.printResults(false, tEnd_micro.count() / 1000000, passes);
             break;
         }
     } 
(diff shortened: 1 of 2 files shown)


rbergen:
@chriselrod I have two comments regarding this PR:
1. Extending the precision of the time measurement has not more than a minute impact on the accuracy of the key result of the benchmark runs (being the number of sieve runs per second, per thread). With that I'm also saying that I think the accuracy of the current results is sufficient. That in turn makes that I don't consider this PR a fix for a bug, but a potential accuracy optimization. I'll therefore leave it up to @davepl, the original author of the solutions in question, if he wants to apply these changes or not. 
2. This PR crossed another one that modified the contents of the PrimeCPP/solution_2 directory to the point that this PR is no longer mergeable. If, considering the first point, you do still want to keep this PR open, then please modify it to align with the current second solution.

chriselrod:
1. If you prefer the timings be inaccurate, then at least don't print `5.0` instead of `5.000000` when the time is actually `5.000744`, to avoid misleading about the accuracy of the times.

rbergen:
I see your point when it comes to the integer divide of the measured duration. What would be sufficient to address that in solution_1 is to change the line
```cpp
sieve.printResults(false, duration_cast<microseconds>(steady_clock::now() - tStart).count() / 1000000, passes);
```
to
```cpp
sieve.printResults(false, duration_cast<microseconds>(steady_clock::now() - tStart).count() / 1000000.0, passes);
```

That's a change I would be happy to merge. That still leaves the second point, which is that the current PR causes a merge conflict.

rbergen:
I've fixed the rounding error in #220.

chriselrod:
Great, thanks! Sorry for not getting back to this sooner.

Issue #177, opened 2021-05-05
Discuss how to report on alternate algorithms


mike-barber:
I think we need to have a discussion to figure out how to deal with reporting for different algorithms, as this adds an extra dimension of fun. 

Currently, all of the implementations (except for the C ones) are very similar to the "reference" C++ version, where we consider all odd numbers as potential primes. 

The C implementations from @mckoss and @danielspaangberg are really interesting additions, and make reference to [this wikipedia section](https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes#Computational_analysis). The same applies to the C# implementation by @Kinematics. Thanks for these! The important thing to note here is that these are algorithmically quite different from all other implementations as they avoid needing to check higher known primes factors besides just 2: variants include 3,5,7, and upwards. 

However, this does raise some issues:
- we should _not_ just exclude these from reporting - I think it's great work, and it would be a disservice to not include them. We definitely should continue showcasing these.
- we also can't compare these results directly to every other implementation, as it would be a disservice to all the work done in other languages
- one of the C solutions, `1of2`, is the equivalent of the "reference" C++ implementation, so this one is comparable
- considering that `1of2` is part of a more general set of algorithms, I don't think it would make sense to move it to a separate project.
- there's a potential arms race as other languages implement this, resulting in a lot of code :)

In essence, I think we just need to work out what do to on the reporting side. My initial thoughts are along the lines of this: 
- have a category for "normal" solutions, which is where most stuff lives, including C `1of2`
- have a special "different algorithm" category where we can put the more interesting stuff like `8of30` and up

I'm not sure if this is the right approach, or how to deal with tagging specific lines in the output as "different algorithm". It'll be great to get everyone's input! I know you're working on reporting currently @marghidanu and it's looking pretty cool so far.

marghidanu:
Hello! We do include a label in the standard output which should account for the solution "category". Right now this label is not taken into account but as we progress with multiple solution types we will include that as a dimension in the report.

Kinematics:
I was experimenting with breaking up the sieves into components, and run things as an arbitrary combination of those components.  The components in question were:
[...]

PR #195, opened 2021-05-22
Introduction of implementation characteristics/tags

rbergen:
This PR is drafted to initiate the introduction of defined implementation characteristics. The following is used as a premise:

- The substance of issue #177.
- Conversation I've had with @davepl over the past number of weeks.

I'm opening this PR as a draft, for now. The reason is that I think we should first be fully aligned on the contents of the contribution guidelines related to the specification of characteristics, before we start making changes to existing solutions to make them conform to the new setup.


CONTRIBUTING.md
```diff
@@ -46,7 +45,7 @@ In the  `drag-race` branch, see what the highest numbered solution is for the la
 If no solution yet exists for the language you're submitting, put yours in `Prime<Language>/solution_1/`
 
 ### Dockerfile
-Please add a `Dockerfile` that is configured to build and run your solution. It should output the run result to standard output (stdout), using the format specified in the basic [rules](#Rules). Any "auxiliary" output should be directed to standard error (stderr), if the language and toolkit you chose allows.
+Please add a `Dockerfile` that is configured to build and run your solution. It should output the run result to standard output (stdout), using the format described under [output](#output). As indicated there, any "auxiliary" output should be directed to standard error (stderr), if the language and toolkit you chose allows.
 
 When composing the Dockerfile, please use the following as a reference for selecting the base image:
 * If an official image exists on [Docker Hub](https://hub.docker.com/) for the language you chose, use that. If multiple images are available with different underlying Linux distributions, select the one to use in accordance with the next steps in this list.
```
marghidanu, line 51: We should specify that we also desire images that support multiple platforms.

CONTRIBUTING.md
```diff
@@ -67,74 +66,245 @@ Finally, submit a pull request **targeting the branch `drag-race`**, and place a
...
+  ```
+  ![Bit count](https://img.shields.io/badge/Bits-unknown-yellowgreen)
+  ```
+
+## Output
+
+Your solution should write the following text to standard output for each implementation that it runs:
 ```
-![Category](https://img.shields.io/badge/Category-faithful%2Cparallel-green)
+<name>;<iterations>;<total_time>;<num_threads>;<labels>
```
marghidanu: I feel that this format is becoming a bit complex, lets's standardized all through labels, even for the first bits of information we can use a label:

```
name=<name>;iterations=<iterations>;total_time=<total_time>;threads=<threads>;other=value;extra=value
```

This way we have a readable consistent format; the order doesn't matter and it's portable; we don't depend on any library for serialization, deserialization. No mix-up between punctuation between the "first part" and the "second part" of the output. Let's keep things as simple as possible.

CONTRIBUTING.md
rbergen: I'm okay with stating that we prefer that. I'm still against making it a requirement.

marghidanu: We prefer images that also support the ARM64 platform but it's not required.

CONTRIBUTING.md
rbergen: I see your point, but your proposal would require a complete overhaul of all output for all solutions. My proposal is to add the labels as an extension to what we now have. I'm obviously not unbiased, but I think choosing a level of backward compatibility is preferable, certainly considering the large number of implementations we already have in place. My proposal also makes it clear that some output is mandatory/non-negotiable (i.e., the "old" output) and other output is optional, in principle (i.e., the labels).

As a secondary argument: I think we also need to keep an eye on the length of the output. If we were to stick everything in labels then we may run into languages that have issues with line lengths beyond a certain number of characters.

marghidanu: > I see your point, but your proposal would require a complete overhaul of all output for all solutions. My proposal is to add the labels as an extension to what we now have. I'm obviously not unbiased, but I think choosing a level of backward compatibility is preferable, certainly considering the large number of implementations we already have in place. My proposal also makes it clear that some output is mandatory/non-negotiable (i.e., the "old" output) and other output is optional, in principle (i.e., the labels).

Yes, that would be the major disadvantage of what I'm proposing, but maybe it's worth doing it in the long run just to avoid technical debt. Anyway, I guess it's something to think about, not rea

> As a secondary argument: I think we also need to keep an eye on the length of the output. If we were to stick everything in labels then we may run into languages that have issues with line lengths beyond a certain number of characters.

That's a very good point!

CONTRIBUTING.md
rbergen: I've just pushed a version that expresses a preference for multi-arch images.

CONTRIBUTING.md
rbergen: > Yes, that would be the major disadvantage of what I'm proposing, but maybe it's worth doing it in the long run just to avoid technical debt. Anyway, I guess it's something to think about, not rea

Maybe we could make this a two-step process then:

- First we add optional labels for additional attributes, as the current proposal describes. 
- In a second step we review and change the whole output structure. 

If we do this, when we take the second step, we will already have experience with parsing and using the labels for (initially) non-critical output. That means we will then also know by experience how to use these additional dimensions in our benchmark reports.

marghidanu: OK, I will add labels parser as an optional step but would really like to get this right in the near future and have a generic enough format to support any changes in the future.

rbergen: I don't think that anybody can disagree with that goal, and I certainly don't. Once the parser has been updated I will first add the labels to my own implementations, so we (a) have a representative test, and (b) already have a start with modifying existing solutions.

rbergen:
I've continued adding labels to existing solutions, and just pushed the changes up to and including PrimeJulia. I implemented but stashed the changes for PrimeCSharp/solution_1, as a proper addition of the labels to that solution requires a different set of changes than just a literal text change to (the) drag-race output line(s). I'll compose a separate PR for that solution (after this PR is merged), that I would like to then receive a review on from @Kinematics, if possible.

I'll keep working down the list of solutions over the next few days, and push again once I've reached the end of the list.

rbergen:
All solutions except for one or two exceptions have now had the labels applied. I'm thus marking this PR ready for review. Obviously, we shouldn't merge until the benchmark implementation is updated to process the new output.

marghidanu (approved):

mike-barber:
Wow, nice work!

rbergen:
> Wow, nice work!

Thank you! I'm sharing the credits with @marghidanu and @rhbvkleef, because it really took a coordinated effort to make the whole thing work, in the end. :)

rbergen:
Going to merge this now that I've confirmed that the new benchmark tool report includes tags for all solutions except [PrimeCSharp/solution_1](https://github.com/PlummersSoftwareLLC/Primes/tree/drag-race/PrimeCSharp/solution_1), for which I will open a separate PR.

PR #196, opened 2021-05-22
add second zig solution

ManDeJan:
## Description
Hi!, I finally got around to making this pull request. This version of the zig code uses aligned memory and code execution for more consistent and slightly improved performance over the existing solution. I made it to show off the fine grained yet easy to use control you get over the code execution in Zig. Also added some additional sieves if people want to have a look at different implementations :-).

## Contributing requirements


rbergen:
Adding a link to [this comment](https://github.com/PlummersSoftwareLLC/Primes/pull/200#issuecomment-847082698) in PR #200, as it relates to this PR.

marghidanu:
Any reason why this is not merged yet?

ManDeJan:
Yes, see (https://github.com/PlummersSoftwareLLC/Primes/pull/200). I'll push the new version soon :)

ManDeJan:
I pushed the new version! It contains two sieves, one working on a bitset and one working with bytes. The latter is the fastest faithfull implementation on my system so far! It gets `18250`+ iterations on my system while the bitset implementation  comparable to `std::vector<bool>` in C++ gets around `11000` iterations. All the tests were run on a Ryzen 3700x @ 4.1GHz. Big thanks to @ityonemo for helping out and collaborating :-).  Should be good to merge now!

marghidanu:
#200 has been closed since yesterday, that's why I'm asking :)

ityonemo:
@marghidanu just fyi we're going to probably have a multithreaded one by the weekend (day jobs, and all that).

marghidanu:
Looks like the build failed. :(

ManDeJan:
I see the CI failed because it uses an relatively "ancient" version of zig that doesn't support the bitset sieve yet, is there an easy way to use a newer version of the language?

marghidanu:
CI is building your Dockerfile, I suggest switching to alpine:3.13 and install zig(0.7.1-r0) in there. Let me know if you need any help

ManDeJan:
We need a even more recent version, zig 0.7.1 is ~6 months old, I found [this](https://hub.docker.com/r/euantorano/zig) docker image that has zig with a build of the master branch. I have no experience with docker, but would it be as simple as changing the docker file to:

```dockerfile
FROM euantorano/zig:master

WORKDIR /opt/app
COPY . .
RUN zig build -Drelease-fast

ENTRYPOINT [ "./zig-cache/bin/PrimeZig" ]
```

marghidanu:
Do you need to use the nightly build for Zig? Otherwise, I strongly suggest you use a stable verified image, alpine is small stable, and verified. Also, it has support for ARM64.

ManDeJan:
0.7.1 is the latest stable release, which as said unfortunatly doesn't support the bit sieve in this PR. I think the options are using a nightly build of zig or removing the bit sieve from the build for now. The image I proposed is also based upon alpine linux so it should be small and stable as well :). Let me know what you'd prefer!

marghidanu:
Ok, let's just go with the image proposed by you then.

rbergen:
Builds are still failing. Is this expected at this time?

ManDeJan:
Closed in favor of #241

PR #197, opened 2021-05-23
Made Nim project faster and more idiomatic

beef331:
## Description
Doing more optimized bit operations.
Previous impl: `Passes: 3970, Time: 5.00019383430481, Avg: 0.001259494668590632, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: true`

New impl: `Passes: 4388, Time: 5.000984191894531, Avg: 0.001139695577004223, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: true`
## Contributing requirements

### DIFF
--- a/PrimeNim/solution_1/primes.nim
+++ b/PrimeNim/solution_1/primes.nim
@@ -1,7 +1,4 @@
-import times
-import math
-import strformat
-import tables
+import std/[times, math, strformat, tables]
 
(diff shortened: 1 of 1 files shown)


beef331:
Did not realize the restriction on class, so will close this.

marghidanu:
Thanks for this, but why did you close it?

PR #198, opened 2021-05-23
Added Nim solution #2

beef331:
## Description

Copy of solution 1 with the following changes:
- Added a `bitlist` for more optimized storage/access
- Flicked the `danger` switch, enabled link time optimizations, and native march.  Since speed is the game
## Contributing requirements


marghidanu:
Ok, got it why you close it!

marghidanu:
Please keep in mind that we also run solutions on ARM64, the current Docker image you are using doesn't have support for the arm64 platform. Please consider using Alpine as a base image and install nim at build time.

beef331:
The reason I closed the other one was that I did not realize the "class" restriction (makes no sense in this case as it's a single object without any overriding)  so guessed it wasnt right as it didnt use anything close to classes (array and static dispatch). No clue about docker, so I might eventually get that done, but no clue when

marghidanu:
Well, I guess we can ignore the ARM64 build for your solution, but otherwise, this looks good to me.

PR #201, opened 2021-05-24
Powershell

crowbar27:
## Description
This PR adds two implementations for PowerShell. The first one is as close as possible to davepl's C# implementation. The second one inlines two functions to reduce the massive overhead caused by function calls in PS.

## Contributing requirements


crowbar27:
Sorry, I just noticed that I have checked in testing parameters in the Docker file. Will open a new PR.

PR #202, opened 2021-05-24
Powershell

crowbar27:
## Description
This PR adds two PowerShell implementations. The first is as close as possible to davepl's original C# implementation. The second reduces the enormous cost of function calls in PS by inlining the access to the bit array.

## Contributing requirements
* [X] I added a README.md with the right category badge(s).


rbergen:
Thanks for this submission. There's a few things I notice:

- The sample output included in your README shows a comma as the decimal separator. Please make sure that the actual output uses a period as the decimal separator, regardless of the locale/regional settings that apply to the system where the solution is run.
- You're using an mcr.microsoft.com image as the base Docker image. I'm not exactly sure what the situation is with arm64 support in those images, but we do try to have as many solutions as possible running on that architecture, too. If possible, make sure that your Docker image does include arm64 support.

crowbar27:
@rbergen: I have forced the thread-culture to en-US before the output. Wrt. the Docker image, this is the one listed as the official PS image on Docker, so I used it. TBH, this is my very first container ever, so I have no clue whether it supports ARM. I have only tested it on an Intel KNL.

crowbar27:
I have just checked https://hub.docker.com/_/microsoft-powershell, and it seems that switching to the preview version of the container would support ARM. I am unsure wether it would not be more reasonable to wait until ARM support gets pushed to the release branch.

rbergen:
@crowbar27 Thanks for the update re the culture. I got an e-mail notification of a comment you made about arm being supported in a preview version of the container you used, but I don't see that comment here anymore. If an mcr.microsoft.com/powershell container with arm support is underway, then I'm happy to wait for that to be promoted to the release branch. Arm64 support is not a hard requirement, but we do want to support the architecture where reasonably possible. Certainly with M1 on the table, that dimension in comparisons has become very interesting indeed.

Anyway, with all that said, this LGTM.

crowbar27:
I was a bit too quick - I thought that I have seen ARM64 on the preview list, but that was actually AMD64 ...

rbergen:
That does explain it, yes. Anyway, I've looked into the whole arm64 thing irt PowerShell, and it looks like arm64 builds of PowerShell will not be available on platforms other than Windows, for some time to come. That obviously makes a conversation about arm64 Docker images rather moot. 

Considering the fact that it then looks like PowerShell solutions will be amd64-only for some time, it may be best to explicitly mark them as such. This is not documented in the guidelines (yet), but you can do so by adding an (empty) file named `arch-amd64` to the solution directories. That means our CI and benchmark implementations will not even try to build or run on arm64.

rbergen:
@crowbar27 Could you do me a favour and 1) delete the global.json file from the PrimeCSharp/solution_1 directory and then 2) commit and push? I know this file has nothing to do with your solution, but it will allow me to check if the CI breakage that we're seeing in this PR is caused by what I think.

rbergen:
@crowbar27 Please remove PrimeCSharp/solution_1/global.json from your powershell branch and then commit and push. It will clear up the CI failures that this PR now runs into.

crowbar27:
> @crowbar27 Please remove PrimeCSharp/solution_1/global.json from your powershell branch and then commit and push. It will clear up the CI failures that this PR now runs into.

Hmm, I thought I had made sure that I did not touch anything than my new subdir. Anyway, it's now gone.

rbergen:
No, this was in no way your fault. Microsoft bumped up the patch version of the .NET SDK 5.0 yesterday, and the file I asked you to remove made our CI only work with the previous SDK patch version. We've already removed the file in question from the branch, but open PRs that are based on an earlier commit still have this problem. As we like to see a successful CI run before we merge, I will have to ask a few people to remove the file in their commits.

rbergen:
LGTM

rbergen (approved):

marghidanu:
LGTM

marghidanu (approved):

PR #204, opened 2021-05-25
Switching OCaml images

marghidanu:
We are now using images from our Docker repository.


rbergen:
LGTM. I did open [an issue](https://github.com/PlummersSoftwareLLC/OCaml-docker/issues/1) on the repo that contains the OCaml Dockerfile, but we don't have to hold back on merging this PR because of it.

PR #205, opened 2021-05-26
Added NodeJS implementation

fvbakel:
## Description
Added a NodeJS implementation

## Contributing requirements


PrimeNodeJS/solution_1/PrimeNode.js
```diff
@@ -0,0 +1,385 @@
...
+                        parseFloat(duration/passes).toFixed(8), 
+                        this.size, 
+                        count, 
+                        new Boolean(this.validateResults()).toString()
+                    );
+        }
+
+        // Following 2 lines are to conform to drag race output format
+        console.log();
+        console.log("fvbakelnodejs; %i;%f;1",  passes, duration);
```
rbergen: Please remove the space before the pass count.

PrimeNodeJS/solution_1/PrimeNode.js
```diff
@@ -0,0 +1,385 @@
...
+*/
+'use strict';
+
+/*
+Global parameters, set by command line parameters
+*/
+var showResult      = false;
+var isHelp          = false;
+var isVerbose       = false;
+var timeLimit       = 10;
```
rbergen: Please use a time limit of 5 seconds for your drag-race run duration, unless there is a specific reason to deviate.

PrimeNodeJS/solution_1/Dockerfile
```diff
@@ -0,0 +1,11 @@
+#
+# To conform to the drag race standard using the official node
+#FROM node:current-alpine3.13
+
+FROM node:latest
```
rbergen: I'm curious, what is your consideration for choosing node:latest over node:current-alpine3.13?

marghidanu: Let’s try to avoid using the latest tag. Changes to this tag can occur frequently.

fvbakel: I think I misread the  [CONTRIBUTING.md](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md). I thought that the latest official image was preferred over current-alpine3.13. Reading it again more carefully I realize that was my mistake. 
I am more than happy to change it to node:current-alpine3.13, because that would also be my personal preference.

marghidanu: Maybe "latest" is a bad choice of words. Is always better to have explicit versions declared, we don't want to have the solution failing because of some sudden change of the Docker image. :)

fvbakel: Fully agree, changed it to node:current-alpine3.13.

PrimeNodeJS/solution_1/PrimeNode.js
fvbakel: Changed this to setting to 5 seconds

PrimeNodeJS/solution_1/PrimeNode.js
fvbakel: Removed space

rbergen (approved):
LGTM

marghidanu (approved):

Issue #203, opened 2021-05-24
There are no instructions for running this suite


rayray:
Please add instructions to the README for running the entire suite. Specifically Windows, although any amount of instructions for any platform would be an improvement!

marghidanu:
We haven’t really tested on Windows. Probably it would work under WSL, assuming you’ve got Docker and make installed you just need to type: `make` and wait :)

rbergen:
@rayray I can confirm it works on WSL2 (and yes, the 2 matters), with Docker Desktop, bash and make installed; I've tested this myself on Ubuntu 20.04 LTS. However, within that context, there is a significant performance impact (up to 50%) concerning whatever runs within Docker, which is basically everything. We therefore recommend running the benchmark on a "bare" Linux distro. 

The basic instructions really are as simple as @marghidanu indicates. On Windows: 

- install Docker Desktop, WSL2, and a WSL2 Linux distro (again, Ubuntu 20.04 LTS is known to work)
- within that distro: install bash and make, run `make`, wait

As this is a collaborative project, feel free to draft a document (my suggestion would be a separate BENCHMARK.md that we can then link to from the README) with more elaborate instructions.

fvbakel:
I was looking for the same information. I think it is a good idea to write such a document. I have made a first draft proposal here:
https://github.com/fvbakel/Primes/commit/b96772c24f15db7568f2133e63070e297325cb67. I can add more details on the Windows instructions and create a pull request to solve this issue.

rbergen:
@fvbakel Thank you very much for picking this up, it's highly appreciated!

I just look at your first draft proposal, and I think it's a good start. Have you thought on how (and when) would you like to receive feedback/input?

fvbakel:
@rbergen Thank you for your quick response. If you feel that this first draft is in the right direction, then I can further extend it next week with the windows and mac os instructions in a similar way.

rbergen:
Yes, I think it's definitely in the right direction. Some comments I can make off the cuff:

- Although I have seen the benchmark suite run on Windows 10 within WSL2 and with Docker Desktop, the impact on performance is really significant in that setup (up to 50%, as I wrote before). If you ask me, the section on Windows should start with a clear statement that using that setup is not recommended. 
[...]

PR #213, opened 2021-06-01
Added instructions on how to run all benchmarks for different operating systems

fvbakel:
## Description
Added instructions on how to run all benchmarks for different operating systems. This is a solution for issue #203.

## Contributing requirements
Not applicable.

### DIFF
--- a/BENCHMARK.md
+++ b/BENCHMARK.md
@@ -1,0 +1,122 @@
+# Benchmarks
+
+This document describes how to run all benchmarks (solutions) for different operating systems.
+
+## Table of contents
+
+1. [General working mechanism](#general-working-mechanism)
+2. [Linux](#linux)
+3. [Windows](#windows)
+4. [Mac OS](#mac-os)
+5. [Example output](#example-output)
+
+## General working mechanism
+
+The main directory contains a file named "```makefile```". This makefile is a linux shell script written in bash. This script has the following basic:
+
+- A temp directory is created to write the output of the benchmarks to
+- All solution directories that have a Dockerfile are processed one by one
+- If a solution requires a spefic hardware architecture that does not match your hardware then this solution is skipped. For example, ```PrimeAssembly\solution_1``` contains a file named ```arch-arm64```. If you are running on different hardware then this solution is skipped
+- Each solution is build and run with the Dockerfile that is provided in the solution directory.
+- After all solutions have been run a summary report is created in the temp directory that was created in the first step.
+
+## Linux
+
+### Linux run instructions
+
+Take the following steps to run all benchmarks:
+
+1. Make sure you have installed the [required software](#linux-installation-and-prerequisites)
+2. Open a terminal.
+3. ```git clone https://github.com/PlummersSoftwareLLC/Primes.git```
+4. ```cd Primes```
+5. ```make```
+6. The run of all benchmarks will now start. Note that this will take a while, so please wait untill this task is complete.
+7. Once complete a report is written to the terminal, see [example output](#example-output).
+
+### Linux installation and prerequisites
+
+The following software must be installed:
+
+- /bin/bash
+- make
+- git
+- docker
+
+The details for each required software are described below:
+
+#### **/bin/bash**
+
+Most Linux distributions include ```/bin/bash``` by default. If this is not the case for your distribution then please check your distribution specific documentation.
+
+#### **make**
+
+Most Linux distributions include ```make``` by default. If this is not the case for your distribution then please check your distribution specific documentation. Below are some examples:
+
+Debian based distributions:
+
+```bash
+sudo apt-get install make
+```
+
+Fedora based distributions:
+
+```bash
+sudo dnf make
+```
+
+#### **git**
+
+Most Linux distributions include ```git``` by default. If this is not the case for your distribution then please check your distribution specific documentation. Below are some examples:
+
+Debian based distributions:
+
+```bash
+sudo apt-get install git
+```
+
+Fedora based distributions:
+
+```bash
+sudo dnf git
+```
+
+#### **Docker**
+
+For the installation of Docker follow the instuctions as described in <https://docs.docker.com/engine/install/>.
+
+## Windows
+
+[] TODO: to be detailed. Options:
+
+- through WSL2
+- with virtualbox / VMware
+- with cywin?
+
+Performance considerations.
+
+## Mac OS
+
+[] TODO: I don't have Mac OS, don't know but gues it is simular to Linux
+
+## Example output
+
+Below is an example of the benchmark output.
+
+```bash
+Generating Software Drag Race (single-threaded)
+         implementation    solution                              label  passes    duration  threads  passes_per_second
+1                primec  solution_2  danielspaangberg_5760of30030_owrb   13471    5.014557        1        2686.378877
+2                primec  solution_2    danielspaangberg_480of2310_owrb   11037    5.016852        1        2199.985170
+3                primec  solution_1                        mckoss-c830   10716    5.000000        1        2143.200000
+... etc
+
+Generating Software Drag Race (multi-threaded)
+   implementation    solution                              label  passes  duration  threads  passes_per_second
+1          primec  solution_2  danielspaangberg_5760of30030_epar   30572  5.015792        4        1523.787270
+2          primec  solution_2    danielspaangberg_480of2310_epar   28214  5.030535        4        1402.137148
+3          primec  solution_2      danielspaangberg_48of210_epar   20572  5.001535        4        1028.284317
+... etc
+
+Output files available in /tmp/tmp.kXe95aYI6R
+```
--- a/BENCHMARK.md
+++ b/BENCHMARK.md
@@ -87,7 +87,20 @@ sudo dnf git
 
 #### **Docker for Linux**
 
-For the installation of Docker follow the instuctions as described in <https://docs.docker.com/engine/install/>.
+For the installation of Docker first follow the instructions as described in <https://docs.docker.com/engine/install/>.
+
+After the installation you need to enable docker as non-root user. Take the following steps:
+
+1. ```sudo groupadd docker```
+2. ```sudo usermod -aG docker $USER```
+3. Log out and log back in so that your group membership is re-evaluated.
+4. Verify that you can run docker:
+
+   ```bash
+   docker run hello-world
+   ```
+
+ The docker website describes more [post-installation steps for linux](https://docs.docker.com/engine/install/linux-postinstall/). These additional steps are optional for running the benchmarks.
 
 ## Windows
 
@@ -98,7 +111,7 @@ The preferred method to run the benchmark is with native Linux. Running the benc
 Take the following steps to run all benchmarks:
 
 1. Make sure you have installed the [required software](#windows-installation-and-prerequisites)
-2. Open a **bash** terminal.
+2. Open an **Ubuntu** terminal (  Start --> Ubuntu 18.04 LTS )
 3. ```git clone https://github.com/PlummersSoftwareLLC/Primes.git```
 4. ```cd Primes```
 5. ```make```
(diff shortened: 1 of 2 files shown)


BENCHMARK.md
```diff
@@ -0,0 +1,221 @@
...
+4. [macOS](#macos)
+5. [Example output](#example-output)
+
+## What operating system to use?
+
+A Unix-like operating system is the preferred operating system to run this benchmark. Linux and macOS are Unix-like operating systems. This will result in the best performance because the benchmark is based on Unix technologies. Running with Windows or other operating systems is possible but will always require some extra layer of virtualization that impact the performance. Running the benchmark with Windows can have a significant impact on the performance, up to 50%.
+
+## General working mechanism
+
+The main directory contains a file named "```makefile```". This makefile is a linux shell script written in bash. This script has the following basic flow:
```
rbergen: I'll make this remark here, but it applies throughout the document:
* `makefile` really needs to be called `Makefile` (as in, with a capital M), at least to work on Unix-like systems
* For in-line code quotes, a single back-tick ( ` ) is enough. The triple back ticks are meant to be used to create separate (often multi-line) code blocks.

BENCHMARK.md
```diff
@@ -0,0 +1,221 @@
...
+
+- A temp directory is created to write the output of the benchmarks to
+- All solution directories that have a Dockerfile are processed one by one
+- If a solution requires a spefic hardware architecture that does not match your hardware then this solution is skipped. For example, ```PrimeAssembly\solution_1``` contains a file named ```arch-arm64```. If you are running on different hardware then this solution is skipped
+- Each solution is build and run with the Dockerfile that is provided in the solution directory.
```
rbergen: I admit fully that we're well into language OCD here, but the past and perfect tense forms of "build" is "built" (ending with a t). Again, I'll note it once here, but it obviously applies to all locations where "build" is used in a past or perfect tense.

BENCHMARK.md
```diff
@@ -0,0 +1,221 @@
...
+### Linux run instructions
+
+Take the following steps to run all benchmarks:
+
+1. Make sure you have installed the [required software](#linux-installation-and-prerequisites)
+2. Open a terminal.
+3. ```git clone https://github.com/PlummersSoftwareLLC/Primes.git```
+4. ```cd Primes```
+5. ```make```
+6. The build and run of all benchmarks will now start. Note that this will take a while, so please wait untill this task is complete.
```
rbergen: It may be good to mention that the build of all solutions only happens the first time. On subsequent runs, only solutions that have changed since the last benchmark run will be rebuilt, which makes those runs a lot faster.

marghidanu:
Please check the new `tools` implementation, the system dependencies have changed. This document needs to be updated, the benchmark tooling is still a work in progress.

BENCHMARK.md
```diff
@@ -0,0 +1,221 @@
...
+4. ```cd Primes```
+5. ```make```
+6. The build and run of all benchmarks will now start. Note that this will take a while, so please wait until this task is complete.
+7. Once complete a report is written to the terminal, see [example output](#example-output).
+
+### Windows installation and prerequisites
+
+The following software must be installed:
+
+- Windows Subsystem for Linux 2 (WSL2) with the Ubuntu 18.04 LST distribution
```
rbergen: We recommend the use of Ubuntu 18.04 LTS as an alternative base image for Docker images that cannot be made to work with Alpine (as described in the contributing guidelines). When it comes to the Docker *hosts*, being the machines that the Docker images are built and run on, we actually use Ubuntu 20.04 LTS.

BENCHMARK.md
```diff
@@ -0,0 +1,221 @@
...
+- make
+- Docker Desktop for Windows
+
+WSL2 has special [hardware requirements](https://docs.microsoft.com/en-us/windows-server/virtualization/hyper-v/system-requirements-for-hyper-v-on-windows).
+
+#### **Install WSL2**
```
rbergen: Personally, I would not "force" the header to bold. The translation of "heading 4" to a specific visual style should in principle be left to the Markdown renderer.

rbergen:
The remark that @marghidanu made just before I posted my review translates to Node.js being an additional requirement for the benchmark to run, next to make, bash, Docker (and WSL2 on Windows). This requirement only applies to the new tools setup (see PR #214), but as the new tools are likely to be merged on the short term and Node.js being installed doesn't break the current tools, I agree that it's best to add this to BENCHMARK.md from the beginning.

BENCHMARK.md
fvbakel: OK

BENCHMARK.md
fvbakel: This one always hits me, but you are right. Applied a change to the text.

BENCHMARK.md
fvbakel: Agree, added that.

BENCHMARK.md
fvbakel: OK, changed to Ubuntu 20.04 LTS.

BENCHMARK.md
fvbakel: Agree.

fvbakel:
@rbergen Thanks for your review. I applied the comments. I don't know the details about PR #214, but I hope it doesn't break this description to much :-).

rbergen:
> I don't know the details about PR #214, but I hope it doesn't break this description to much :-).

It shouldn't. Like I said, it basically adds Node.js (and npm) as an additional prerequisite for the benchmark to run. On Windows, it is important that the nodejs and npm packages are installed *inside* the WSL2 Ubuntu distro (just like make), not on Windows itself.

**Edit:** I forgot to mention that the Node.js/npm versions that are included in many OS-provided package managers are too old. This means that it will generally be required to download and install an appropriate version of Node.js from https://nodejs.org/en/download/. Personally, I have successfully tested the new tools setup with the current LTS version (that being 14.17.0 at the time of writing).

README.md
```diff
@@ -10,4 +10,8 @@ benchmarking the same prime number sieve in Python, C#, and C++.
 
 ## Community contributions
 
-Community contributions, fixes and improvements are accepted on this branch. If you want to add your own solution, please read [CONTRIBUTING.md](CONTRIBUTING.md).
\ No newline at end of file
+Community contributions, fixes and improvements are accepted on this branch. If you want to add your own solution, please read [CONTRIBUTING.md](CONTRIBUTING.md).
+
+## Running all benchmarks
+
+All benchmarks (solutions) in this repository can be build and run with one makefile. This is described in [BENCHMARK.md](BENCHMARK.md).
```
rbergen: In this sentence, "build" needs to be spelled as "built", and "makefile" as "Makefile" (with capital M).

fvbakel: ok

rbergen:
@fvbakel We have just merged the new benchmark tool (PR #214), which means the Node.js dependency now applies. Would it be possible to add the installation instructions for Node.js to your PR?

fvbakel:
Yes, I can include the Node.js instruction somewhere next week.

rbergen:
That's great news and highly appreciated!

fvbakel:
I have run the new reporting on Linux and that works fine. I also tested on Windows with WSL2, Ubuntu 20.04 and the same Linux binary installed, The solution are all run fine, but the reporting step fails. This is the error:
```
npm ci && npm run start -- report -d /tmp/tmp.fShEnvV3CO
/home/fvbakel/node-v14.17.0-darwin-x64/bin/node: 1: : not found
/home/fvbakel/node-v14.17.0-darwin-x64/bin/node: 1: cannot open : No such file
/home/fvbakel/node-v14.17.0-darwin-x64/bin/node: 1: : not found
/home/fvbakel/node-v14.17.0-darwin-x64/bin/node: 1:  not found
/home/fvbakel/node-v14.17.0-darwin-x64/bin/node: 1: ��#: not found
/home/fvbakel/node-v14.17.0-darwin-x64/bin/node: 1: : not found
/home/fvbakel/node-v14.17.0-darwin-x64/bin/node: 2: __nl_symbol_ptr__DATAaav__got__DATaaw__la_symbol_ptr__DATAsbb__mod_init_func__DATAbb: not found
/home/fvbakel/node-v14.17.0-darwin-x64/bin/node: 6: Syntax error: "(" unexpected
```

Anny idea how this can be fixed?

rbergen:
I'm a bit confused. Darwin is usually aimed at MacOS. How exactly did you install Node within WSL2?

fvbakel:
:-) Indeed, by accidentally extracted the macOS binaries on the WSL2. Tested again with the Linux distribution, that works fine.

fvbakel:
@rbergen Added the Node.js instructions.

BENCHMARK.md
```diff
@@ -0,0 +1,284 @@
...
+4. [macOS](#macos)
+5. [Example output](#example-output)
+
+## What operating system to use?
+
+A Unix-like operating system is the preferred operating system to run this benchmark. Linux and macOS are Unix-like operating systems. This will result in the best performance because the benchmark is based on Unix technologies. Running with Windows or other operating systems is possible but will always require some extra layer of virtualization that impact the performance. Running the benchmark with Windows can have a significant impact on the performance, up to 50%.
+
+## General working mechanism
+
+The main directory contains a file named "`Makefile`". This Makefile is a linux shell script written in bash. This script has the following basic flow:
```
rbergen: > This Makefile is a linux shell script written in bash.

That is actually not entirely the case; Makefiles have a format of their own that is most commonly used to coordinate build and installation tasks. In this case it mostly does execute shell commands in Bash, but it is not a shell script as such. (Background of the decision to create a Makefile instead of a "regular" shell script is that make is extremely good at tracking and managing dependencies. We think that will be helpful, because we expect to extend the benchmarking and reporting capabilities in the future.)

> This script has the following basic flow:

Maybe a phrasing like "The Makefile implements the following basic flow:" is better?

rbergen (commented):
I think the Node.js installation instructions you've added are a good addition, indeed. The one comment I'm making is basically to avoid confusion about Makefiles vs. shell scripts, for those who are not fully aware of the differences between the two.

BENCHMARK.md
fvbakel: Agree

rbergen (approved):
LGTM

BENCHMARK.md
```diff
@@ -0,0 +1,284 @@
...
+
+## Linux
+
+### Linux run instructions
+
+Take the following steps to run all benchmarks:
+
+1. Make sure you have installed the [required software](#linux-installation-and-prerequisites)
+2. Open a terminal.
+3. Make sure Node.js is added to the path. `PATH=/path/to/node/bin:$PATH`.
```
rbergen: @fvbakel I know I already approved, but it just occurred to me that I'm not sure this will work without `export`. As in: `export PATH=/path/to/node/bin:$PATH`. Have you tested this?
This question obviously also applies to Windows/WSL2

BENCHMARK.md
```diff
@@ -0,0 +1,284 @@
...
+
+The benchmarks makes use of Node.js to create a report. This requires Node.js version 14.17.0 or higher. The Node.js/npm versions that are included in many OS-provided package managers are too old. This means that it will generally be required to download and install an appropriate version of Node.js from <https://nodejs.org/en/download/>.
+
+Take the following steps to install Node.js on Linux:
+
+1. Navigate to <https://nodejs.org/en/download/>
+2. Download the Linux Binaries, depending on your CPU architecture
+3. Unzip the downloaded file: `tar -xf /path/to/downloaded/file.xz>`
+   For example: `tar -xf ~/Downloads/node-v14.17.0-linux-x64.tar.xz`
+4. The Node.js binaries are now extracted to a directory. This directory is refereed to as `/path/to/node` in the run instructions.
```
rbergen: "refereed" should be "referred". Same under Windows/WSL2

BENCHMARK.md
fvbakel: I am not to sure if I used the export when I was testing it, and wrote something different in the text. To be on the save side I will add export to this command.

BENCHMARK.md
fvbakel: ok

PR #214, opened 2021-06-01
Tools refactoring

marghidanu:
Tools are now running locally instead of a Docker container, this is to allow the collection of data about CPU, OS, and System. Also, the new implementation has been rewritten in TypeScript and Node.js is now a build dependency. 

This implementation needs to be tested on Windows.


tools/src/models.ts
```diff
@@ -0,0 +1,34 @@
...
+  serverVersion: string;
+}
+
+export interface Report {
+  version: string;
+  metadata: { [key: string]: string | number };
+  machine: {
+    cpu: si.Systeminformation.CpuData;
+    os: si.Systeminformation.OsData;
+    system: si.Systeminformation.SystemData;
```
rbergen: The machine.os and machine.system objects do contain some fields I think we need to take out because they are too specific/identifiable. I would argue that these include, at least:

- machine.os.hostname
- machine.os.fqdn
- machine.os.serial
- machine.system.serial
- machine.system.uuid

Alternatively/actually, we may want to actively pick out only those properties that we feel are safe to include. That way, we won't start inadvertently leaking sensitive information if we start using future versions of systeminformation.

tools/src/services/result.ts
```diff
@@ -0,0 +1,50 @@
...
+        { name: 'solution', title: 'Solution', alignment: 'left' },
+        { name: 'label', title: 'Label', alignment: 'left' },
+        { name: 'passes', title: 'Passes', alignment: 'center' },
+        { name: 'duration', title: 'Duration', alignment: 'center' },
+        { name: 'threads', title: 'Threads', alignment: 'center' },
+        {
+          name: 'passesPerSecond',
+          title: 'Passes/Second',
+          alignment: 'center'
+        }
```
rbergen: I'm wondering if we could also add some "known" tags as columns to this table (which would now be 'algorithm', 'faithful', 'bits').

rbergen (commented):
This PR now includes an empty tools/README.md. Do you intend to include a README at a later time?

rbergen:
I've tested this on Windows 10/WSL2/Ubuntu 20.04, with Node.JS 14.17.0 installed within the WSL2 Ubuntu distro. On my machine, it works with and without tags present.

tools/src/services/result.ts
marghidanu: Yes we can!

tools/src/models.ts
marghidanu: Agreed, I’ll create a partial structure and remove this information.

marghidanu:
> This PR now includes an empty tools/README.md. Do you intend to include a README at a later time?

Most definitely, I want to include development environment setup, installation notes. Anything that would be related to this project separately.

tools/src/models.ts
rbergen: I looked at the commit for this, and it looks good!

tools/src/services/result.ts
marghidanu: The changes should be in place.

rbergen: I ran a test with this version and I can confirm they are. Really cool to see this!

rbergen (approved):
Personally, I would either put content in the README.md or remove it, for now. I'm approving as I don't consider this a blocking issue for merging.

marghidanu:
Let’s delete it for now, I can always add it later.

PR #215, opened 2021-06-02
Added a implementation for Tcl.

fvbakel:
## Description
Added a implementation for Tcl. with a 32 bit string array as buffer
for the bit array. Native Tcl does not have objects, so this implementation uses an variable for the bit array.

## Contributing requirements

### DIFF
--- a/PrimeTcl/solution_1/Dockerfile
+++ b/PrimeTcl/solution_1/Dockerfile
@@ -1,0 +1,9 @@
+FROM alpine:3.13
+
+RUN apk add --update tcl
+
+WORKDIR /opt/app
+
+COPY primes.tcl .
+
+ENTRYPOINT ["tclsh","primes.tcl" ]
--- a/PrimeTcl/solution_1/README.md
+++ b/PrimeTcl/solution_1/README.md
@@ -2,8 +2,8 @@
 
 ![Category](https://img.shields.io/badge/Category-faithful-green)
 
-First implementation in Tcl, with a 32 bit string array as buffer
-for the bit array. This implementation is based on the logic from:
+Implementation in Tcl, with a 32 bit string array as buffer
+for the bit array. Native Tcl does not have objects, so this implementation uses an variable for the bit array. This implementation is based on the logic from:
 
 - Python/solution_2, by ssovest
 - PrimeCPP,          by Dave Plummer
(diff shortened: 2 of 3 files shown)


rbergen:
First off. thanks for submitting this. As I recently referred to in a conversation with some others, I love that we can add Tcl to the collection of languages. 

I've been thinking about the "faithful implementation without object" aspect, and I think we can bring the implementation one small step closer to "classical faithful". That could be achieved by creating and initializing the prime array outside of the run_sieve procedure (but within the while loop), and pass it to that procedure as a parameter.

Would it be possible to apply this change to the implementation?

fvbakel:
Thanks once more for your quick review. I have been doing lots of Tcl programming at the beginning of the 2000's. At the time it was in the context of eMatrix, a PLM system. So it was more than 15 years ago since my last line of Tcl. :-). But today I came to the same conclusion as years ago, Tcl is just really slow.

Yes, I can make the change you suggested. I moved the `primes` bit array to the main procedure. I pass the variable by reference now and I made that also consistent in the `print_results` procedure. For performance it does it does not really matter, but it is a bit cleaner now.

rbergen:
> Thanks once more for your quick review. 

You're welcome. :)

> I have been doing lots of Tcl programming at the beginning of the 2000's. At the time it was in the context of eMatrix, a PLM system. So it was more than 15 years ago since my last line of Tcl. :-). But today I came to the same conclusion as years ago, Tcl is just really slow.

I've revisited quite a few former programming contexts, myself, in the context of this project. The last Pascal code I wrote before the implementation I submitted here was at university, 25 years ago. The last "classic-style compiled BASIC" I wrote before *that* implementation was in QuickBASIC, a couple of years before that. :)

> Yes, I can make the change you suggested. I moved the `primes` bit array to the main procedure. I pass the variable by reference now and I made that also consistent in the `print_results` procedure. For performance it does it does not really matter, but it is a bit cleaner now.

Thank you. I realize that my request can be considered nitpicking, but under the hood `this` is passed as the first parameter to member functions in C++ and other languages, so I gather this is the closest we can get to that aspect of "faithful", in Tcl.

rbergen (approved):
LGTM

marghidanu:
LGTM

PR #85, opened 2021-04-06
Update PrimeSieveJava.java

rich1051414:
Up to a 20% performance increase using (index & 1) instead of index % 2

PR #233, opened 2021-06-05
Potentially 20% increase to performance

rich1051414:
Up to a 20% increase to performance doing bitwise operation to check if even instead of modulus.

## Contributing requirements


rbergen:
Merging with approval of @MansenC (see #85).

PR #234, opened 2021-06-05
Go: match bit twiddle memory handling to reduce cache pressure

Kunde21:
## Description

Go's boolean slices are not compact bitmaps, but instead uint8 in
memory.  This means 7 of 8 bits are clogging cache without being
useful for the calculation.  Set and get using bit masking and
shifting.

Because Go memory is zeoroed on allocation, the bits are considered
"set" when they are zero.  This is the inverse of many other
implementations, due to how the Go runtime initializes memory.

Small bugfix included:  Array was only being initialized once, so the
subsequent passes were making non-op writes.  Each pass now allocates
new memory.

## Contributing requirements

### DIFF
--- a/PrimeGo/solution_1/main.go
+++ b/PrimeGo/solution_1/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"flag"
 	"fmt"
 	"math"
 	"time"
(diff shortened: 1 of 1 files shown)


rbergen:
As this changes the Go solution submitted by @bundgaard, I'm allowing time for them to review and respond if they want to.

rbergen:
While we wait for input from @bundgaard, I'd already like to float one idea you could consider, which is creating a second Go solution with your implementation, instead of modifying the one that's in place. The reason I'm suggesting this is that it changes one of the "defined" implementation characteristics, being the number of bits per prime. It may actually be interesting to see how both approaches compare between each other in the benchmark.

rbergen:
In lieu of input from @bundgaard, I'm going to convert my idea into a decision. That is, I'm not comfortable merging this PR as a change to PrimeGo/solution_1, but I would be happy to merge your solution as a second one (PrimeGo/solution_2) next to the one this PR now aims to modify.

marghidanu:
LGTM

rbergen:
I'm still of the opinion that this would need to be added as a separate solution (PrimeGo/solution_2), not as a change to the Go solution that is already in the codebase.

rbergen:
Closing this. With the merge of #319 we have a 1-bit Go solution in place.

PR #236, opened 2021-06-08
Add an implementation in SQLite

fvbakel:
## Description
This is an implementation in SQLite. See README.md for more details and considerations.

I realize that this implementation is on the edge of the rules, however it is a close as I could get in SQL and SQLite. I still think it would be a great addition to the existing languages. I would be nice to compare this implementation with implementations in other SQL engines (Oracle, MySQL) and SQL languages (PL/SQL, TSQL).

## Contributing requirements


PrimeSQLite/solution_1/README.md
```diff
@@ -0,0 +1,102 @@
+# SQLite implementation of the Sieve of Eratosthenes algorithm
+
+![Algorithm](https://img.shields.io/badge/Algorithm-other-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-no-green)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-8-green)
```
rbergen: Due to their values, the Algorithm, Faithfulness and Bits labels need to be be -yellowgreen instead of -green.
Furthermore, because this is an approximation of the Sieve of Eratosthenes, I think one more badge should be included:
```
![Deviation](https://img.shields.io/badge/Deviation-approximation-blue)
```

PrimeSQLite/solution_1/README.md
```diff
@@ -0,0 +1,102 @@
...
+In the second stage the elimination is done for the found prime numbers of the first stage. This is done with the following logic:
+
+1. First all natural numbers between the square root of the limit and the limit are listed
+2. For all primes that are found in stage 1, all factors are calculated and listed
+3. Then list2 is subtracted from list1 and the primes found in stage1 are added
+4. The results from 3 are stored in a table, and this is considered the end result.  
+
+### Alternative Algorithms
+
+Alternative algorithms and formulations for this implementation have been considered. The different variants are stored in the [attic](./attic/README.md) of this solution.
```
rbergen: I think the `[attic]` link is probably intened to point to ./attic/ instead of ./attic/README.md; the directory is the attic, in which the README.md is one file that GitHub will show below the directory file listing anyway.

PrimeSQLite/solution_1/README.md
```diff
@@ -0,0 +1,102 @@
+# SQLite implementation of the Sieve of Eratosthenes algorithm
```
rbergen: I believe this solution does not actually implement the Sieve of Eratosthenes, but an approximation. It would probably be best if the title of the README reflects this.

rbergen (commented):
This is definitely an interesting submission.

Strictly speaking, this doesn't comply with the contribution guidelines, but I'm inclined to accept it if it is clearly marked as a deviation (I've made some specific pointers, below.)

@marghidanu What's your opinion?

PrimeSQLite/solution_1/README.md
fvbakel: Agree

PrimeSQLite/solution_1/README.md
fvbakel: Agree

PrimeSQLite/solution_1/README.md
fvbakel: oke

fvbakel:
@rbergen Thanks for your quick review. It would be great if it will be accepted in the race :-).

rbergen (approved):
LGTM

marghidanu (approved):

PR #240, opened 2021-06-11
Added implementation in R

fvbakel:
## Description
This is an implementation in R. Details see README.md

## Contributing requirements

### DIFF
--- a/PrimeR/solution_1/README.md
+++ b/PrimeR/solution_1/README.md
@@ -5,13 +5,13 @@
 ![Parallelism](https://img.shields.io/badge/Parallel-no-green)
 ![Bit count](https://img.shields.io/badge/Bits-1-green)
 
-This is an implementation of the [Sieve of Eratosthenes](https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes) algorithm in R. This implementation makes use of the reference class object oriented method that is available in R. This results in a solution that is closer to the rules defined in [CONTRIBUTION.md](../../CONTRIBUTION.md).
+This is an implementation of the [Sieve of Eratosthenes](https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes) algorithm in R. This implementation makes use of the reference class object oriented method that is available in R. This results in a solution that is closer to the rules defined in [CONTRIBUTING.md](../../CONTRIBUTING.md).
 
 ## Run instructions
 
 ### Run native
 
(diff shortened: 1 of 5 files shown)


PrimeR/solution_1/primes.R
```diff
@@ -0,0 +1,121 @@
...
+    start <- as.numeric(format(Sys.time(), "%OS3"))
+    passes <- 0
+    while (TRUE) {
+        passes <- passes + 1
+        sieve <- prime_sieve$new()
+        sieve$init_fields(limit)
+        sieve$run_sieve()
+        now <- as.numeric(format(Sys.time(), "%OS3"))
+        duration <- now - start
+        if (duration > time_limit || passes==1) {
```
rbergen: The `if (... || passes==1)` makes that the loop is only executed once, regardless of the time passed.

rbergen (commented):
I prefer your previous naming scheme for solution PRs :)

PrimeR/solution_1/primes.R
fvbakel: Ah, well spotted, thanks. I used that during my testing, but forgot to remove on the final checkin.

fvbakel:
I resolved the review remark. I also realized that the number of bits reported was incorrect. I did some additional research and testing and found that in the drag-race conditions 32 bits are used.

PrimeR/solution_1/README.md
```diff
@@ -0,0 +1,68 @@
+# Implementation in R
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-32-green)
```
rbergen: If the bit count is not 1, the badge colour to use is -yellowgreen instead of -green.

rbergen (commented):
Thank you for the new commit, and correcting the bit count per prime. I did find one small thing related to the latter, see below.

PrimeR/solution_1/README.md
fvbakel: ok

fvbakel:
> I prefer your previous naming scheme for solution PRs :)

Solved that too, didn't notice that :-)

rbergen (approved):
LGTM

marghidanu (approved):

PR #241, opened 2021-06-12
Zig kitchen sink

ityonemo:
## Description
the kitchen sink of optimizations for zig.  explores multiple forms of paralellism, bytesieves and bitsieves (of various sizes) and wheels.

## Contributing requirements


ityonemo:
this supercedes https://github.com/PlummersSoftwareLLC/Primes/pull/196 since it is branched from its tip, and also fixes the docker issue.

rbergen:
> this supercedes #196 since it is branched from its tip, and also fixes the docker issue.

So I guess #196 will be closed then?

ityonemo:
I recommend it @ManDeJan

ManDeJan:
Closed in favor of this PR, thanks @ityonemo for putting in the work :-)

PrimeZig/solution_2/src/main.zig
```diff
@@ -0,0 +1,84 @@
...
+    const elapsed = timer.read();
+    try printResults("ManDeJan&ityonemo-zig-bit-sieve", passes, elapsed, size);
+}
+
+fn printResults(backing: []const u8, passes: usize, elapsed_ns: u64, limit: usize) !void {
+    const elapsed = @intToFloat(f32, elapsed_ns) / @intToFloat(f32, time.ns_per_s);
+    const stdout = std.io.getStdOut().writer();
+    try stdout.print("{s};{};{d:.5};1\n", .{
+        backing, passes, elapsed
+    });
```
rbergen: This output does not include the `algorithm`, `faithful` or `bits` tags. Including them is not required, but will lead to default values being assumed when collecting and reporting on results, as indicated in the contributing guidelines.

PrimeZig/solution_3/README.md
```diff
@@ -0,0 +1,43 @@
+# Zig solution by ManDeJan and ityonemo
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Algorithm](https://img.shields.io/badge/Algorithm-<wheel>-yellowgreen)
```
rbergen: Please remove the angle brackets around `wheel` in the badge. Also, based on the contents of the output, I think the following 9 badges should be added:

- Faithful: yes (green)
- Parallel (2): yes, no (both green)
- Bits (6): 1, 8, 16, 32, 64, unknown (first green, latter 5 yellowgreen)

rbergen (commented):
In addition to the more detailed points that are listed below, I have two main questions:

- The PR includes 2 solutions. I just want to double-check that it is indeed your intention to add both of them to the codebase?
- With the combination of different parameters that solution_3 explores, the total number of permutations is significant, and therefore the number of implementation runs is also. I was wondering if it would be possible to "focus" the automated benchmark run on a subset of the total field of permutations, being those that highlight the most interesting/relevant differences between parameter settings?

Furthermore, as I don't believe the bit size of a Zig `bool` is defined by the language, I think the `bits` tag should not be included in the output lines for runs that use `bool`s for storage (meaning it will be interpreted as being unknown).

PrimeZig/solution_3/README.md
ityonemo: will get to these final changes hopefully by next weekend.

ityonemo:
> bit size of a Zig bool

I'm blindly using the @bitSizeOf operation to ascertain how many bits are used by zig's bool type, so if it is not defined by the language (and is architecture-dependent), that will be correctly reflected in the output.

marghidanu:
Is this considered stable at this point?

rbergen:
With the changes that were committed on June 18th, the only point/question that remains for me is the one relating to the (large) number of implementation permutations included in solution_3 (second bullet in my review of June 14th).

ityonemo:
ah, ok, I will decrease the number of permutations, and possibly put the rest of them under some sort of "all" flag.

ityonemo:
reduced it to the minimum set of architectural considerations: bool vs u8, 32 vs 64, HT vs non HT, and, for wheel, u8, if the system provides a memcpy optimization.

rbergen:
I'm now getting errors when trying to build and run the solutions in this PR locally, in Docker:

- Solution_2 builds, but fails to run with the following error:
  ```
  ~/work/Primes/PrimeZig/solution_2$ docker run -ti zig
  docker: Error response from daemon: OCI runtime create failed: container_linux.go:367: starting container process caused: exec: "./zig-cache/bin/PrimeZig": stat ./zig-cache/bin/PrimeZig: no such file or directory: unknown.
  ERRO[0001] error waiting for container: context canceled
  ```
  I think this is caused by the default location of the out binary of a zig build moving from zig-cache/bin to zig-out/bin in the latest release(s) of Zig. I recently modified the Dockerfile for solution_1 in PR #247 to deal with this.

- Solution_3 fails to build, with the following error:
  ```
     => ERROR [5/5] RUN zig build -Drelease-fast                                                                                                                     7.6s
  ------
   > [5/5] RUN zig build -Drelease-fast:
  #9 7.516 ./src/prime.zig:96:63: error: runtime value cannot be passed to comptime arg
  #9 7.516                 thread_ptr.* = try std.Thread.spawn(ThreadInfo{.sieve = self, .index = index}, workerLoop);
  #9 7.516                                                               ^
  #9 7.517 ./src/prime.zig:96:52: note: referenced here
  #9 7.517                 thread_ptr.* = try std.Thread.spawn(ThreadInfo{.sieve = self, .index = index}, workerLoop);
  #9 7.517                                                    ^
  #9 7.517 ./src/main.zig:63:5: note: referenced here
  #9 7.517     try sieve.parallelInit();
  #9 7.517     ^
  #9 7.524 PrimeZig...The following command exited with error code 1:
  #9 7.524 /usr/bin/zig build-exe /opt/app/src/main.zig -OReleaseFast --cache-dir /opt/app/zig-cache --global-cache-dir /root/.cache/zig --name PrimeZig --enable-cache
  #9 7.524 error: the following build command failed with exit code 1:
  #9 7.524 /opt/app/zig-cache/o/85320af569895f7f64e738d9f3f06be0/build /usr/bin/zig /opt/app /opt/app/zig-cache /root/.cache/zig -Drelease-fast
  ------
  executor failed running [/bin/sh -c zig build -Drelease-fast]: exit code: 1
  ```

rbergen:
Solution_2 generates the following output for me:
```
ManDeJan&ityonemo-zig-bit-sieve;7030;5.00014;1;faithful=yes,algorithm=base,bits=1
ManDeJan&ityonemo-zig-byte-sieve-type-bool;5689;5.00096;1;faithful=yes,algorithm=base,bits=1000000
```

I think the `bits` value in the second line is not entirely accurate?

ityonemo:
Haha probably not.  I'll fix it this weekend

rbergen:
That would be great. Certainly with the changes you made in your last 2 commits (for which I'd like to thank you), I think we're very close to being able to merge this. :)

rbergen (approved):
LGTM

PR #243, opened 2021-06-13
Added an implementation in COBOL

fvbakel:
## Description
Added an implementation in COBOL. See README.md for the details.

## Contributing requirements


rbergen:
This is mainly an OCD point for me, but please do me a favour and remove the spaces in the PR template checkboxes in your next PR. :)
`[X ]` renders as [X ], while `[X]` renders as a checked checkbox.

(For completeness: it is not just OCD; in the PR list, only actual "checked checkboxes" are counted as completed PR tasks.)

fvbakel:
Ah, thanks for your reminder [X] I did't realize that. It was getting late yesterday evening :-). I just noticed that there are some mistakes in the readme  and code comments. So I am planning for an update on that later this week. Maybe it is better for you to wait with your review until after these updates.

PrimeCOBOL/solution_1/README.md
```diff
@@ -0,0 +1,61 @@
+# COBOL implementation
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
```
rbergen: I'm afraid that I will have to argue that this solution is not faithful. I recognize fully that this is not your fault but almost certainly due to limitations in the language, but the fact remains that the sieve is currently not contained in a "class-like" object that is reinitialized from scratch for every run,

PrimeCOBOL/solution_1/primes.cbl
```diff
@@ -0,0 +1,159 @@
...
+           COMPUTE MAX_TIME_HS = 100 * MAX_TIME_SEC.
+           MOVE ZERO TO PASSES.
+           PERFORM GET-NOW THROUGH END-GET-NOW.
+           MOVE NOW-HS TO START-HS.
+           PERFORM UNTIL DURATION-HS >= MAX_TIME_HS
+               PERFORM RUN_SIEVE THROUGH END-RUN_SIEVE
+               ADD 1 TO PASSES
+               PERFORM GET-NOW THROUGH END-GET-NOW
+               COMPUTE DURATION-HS = NOW-HS - START-HS
+           END-PERFORM.
```
rbergen, line 63: In the loop (and the procedures it uses) I can't seem to find a place where the BIT-ARRAY is (re)set to ones. I think that means that all runs after the first operate on a "solved" prime sieve, unless I'm missing something?

rbergen:
> Ah, thanks for your reminder [X] I did't realize that. It was getting late yesterday evening :-). I just noticed that there are some mistakes in the readme and code comments. So I am planning for an update on that later this week. Maybe it is better for you to wait with your review until after these updates.

Too late. :) Anyway, I'm very pleased we will be adding a COBOL solution to the collection, so I'm happy to review it more than once.

PrimeCOBOL/solution_1/README.md
fvbakel: Yes, I agree on that. The default program style in Cobol seems to be to use globals. I could not find ways to dynamically allocate it based on the MAX_LIMIT variable. So I will change the faithful to no-yellowgreen.

PrimeCOBOL/solution_1/primes.cbl
```diff
@@ -0,0 +1,159 @@
...
+           DIVIDE 2 INTO MAX_LIMIT
+               GIVING BIT_SIZE ROUNDED
+               REMAINDER IS-EVEN.
+           IF IS-EVEN = 0 THEN 
+                   ADD -1 TO BIT_SIZE.    
+           PERFORM VARYING Z 
+                   FROM 1 BY 1 
+                   UNTIL Z>BIT_SIZE
+                MOVE 1 TO FLAG (Z)
+           END-PERFORM.
```
fvbakel: This part is where all flags are set to 1.

rbergen: Thank you for pointing that out. I missed that.

PrimeCOBOL/solution_1/primes.cbl
fvbakel: The deceleration of `BIT-ARRAY` at the top is a kind of place holder for a table with one column `FLAG` and half a million rows plus an Index that can be accessed by `Z`. The actual bit-array is named `FLAG`. It is reset in the beginning for RUN_SIEVE.   I wanted to keep that in place with the actual calculation of the BIT-SIZE, which is also done at the RUN_SIEVE. I think it would not be good to have these upfront calculations outside RUN_SIEVE as it is part of the calculation in my opinion. 

Furthermore I found that all of this seems to matter for Cobol performance. These inline `PERFORM` statements are faster than calling subroutines.  I used subroutines for all inline `PERFORM` statements in earlier versions but that had about 10% impact on the performance.

```COBOL
           PERFORM VARYING Z                    
                              FROM 1 BY 1                    
                              UNTIL Z > BIT_SIZE                
                    MOVE 1 TO FLAG (Z)           
           END-PERFORM.
```

PrimeCOBOL/solution_1/primes.cbl
fvbakel: No problem, i think it is great that you can review these different languages so quickly.

rbergen: Thank you. I think you and I have both seen a fair share of different languages in our time (and are getting acquainted with a few more in this project), and most languages do have recognizable structures. That said, if a solution in Brainfuck is submitted, I do hope it comes with a good amount of documentation. :)

fvbakel:
@rbergen I improved the initialization of the Bit array. I now copy a static array filled with 1 values over the bit array in each run. This gives a big performance improvement ( ~40%) but I am not sure if this is allowed by the rules. So in the previous code, half of the time was spent in setting the bit values to 1!

rbergen:
>  I now copy a static array filled with 1 values over the bit array in each run. This gives a big performance improvement ( ~40%) but I am not sure if this is allowed by the rules.

In my interpretation of the rules, they don't stipulate exactly how bit array initialization can or cannot be done, only that it needs to happen for every run.

rbergen (approved):
LGTM

marghidanu (approved):

PR #244, opened 2021-06-14
add second Julia solution

epithet:
## Description
This is another Julia solution, which runs much faster on my machine.

## Contributing requirements
I hope I did everything right. I added a `Dockerfile`, ~but I still need to verify whether it works~.


rbergen (approved):
LGTM

marghidanu (approved):

PR #250, opened 2021-06-25
add a PostScript solution

epithet:
## Description
I'll try and add a 1 bit variant at some point.

Unfortunately, for me, this is about 5x faster on my Linux laptop without Docker compared to my Windows PC at work with Docker (and newer `gs`). I haven't tried to make the situations more comparable, I'm just mentioning this because my Julia solution gave comparable results on those machines.

## Contributing requirements


PrimePostScript/solution_1/README.md
```diff
@@ -0,0 +1,22 @@
+# PostScript solution by @epithet
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
```
rbergen: For an implementation to be labeled "faithful", it needs to use a class-like construct (object) to contain the prime sieve plus whatever else is needed to run it. That object then needs to be created and initialized from scratch for every sieve run. 

I'm no expert in PostScript, but I'm not sure I can identify these characteristics in this implementation (nor do I know if PostScript contains the facilities for them).

PrimePostScript/solution_1/README.md
```diff
@@ -0,0 +1,22 @@
...
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-8-green)
```
rbergen: If the Bits count is anything else than 1, the badge color to use is -yellowgreen, instead of green.

rbergen (commented):
First off, I think it would be really cool to add PostScript to the collection of languages. :)

I only have a couple of notes concerning the tagging of the solution, as indicated in below comments.

PrimePostScript/solution_1/README.md
epithet: Sure, no problem. I was unsure about this. My understanding was that the criterion for non-OOP languages was that no buffers were reused (cf. #78), and I do put a fresh buffer (a `string`) on the top of the stack in the main procedure `primesieve`, which is then left on the stack for subsequent operations (see e.g. `countprimes`) and is explicitly `pop`ped off in each iteration of the test loop in `run.ps`.

I also saw assembly solutions flagged as faithful. I haven't studied those closely, but I assumed that the situation must be very similar.

As far as I know, there is no such thing as a user-defined class in PostScript, and the equivalent of C structs would be to collect related data in an `array` or dictionary. In this case, the full state of the sieve is contained in the `string` (the length of the string can be retrieved with the `length` "method") and wrapping it in a 1-element `array` in the name of faithfulness seems excessive, so I'm happy to flag this solution as unfaithful instead. The built-in types `string` and `array` could be considered class-like constructs (with similar interface, e.g. they both have pretty much the same `get` and `length` "methods").

rbergen: My last comment on #78 actually explains why that implementation (that also hasn't been merged yet) is correctly labeled as unfaithful. With regards to the assembly implementations, some are marked as faithful; those that are do indeed create and initialize from scratch a new equivalent of a C struct for every run. There are also assembly implementations that don't, and those are therefore also categorized as unfaithful.

epithet: I changed the badge to 'unfaithful' for the time being. I can always change it back if any of this is convincing.

rbergen:
Thank you for the changes in your last commits. I have one remark left, which is that the output in run.ps also needs to reflect the fact that the implementation is not faithful. Concretely, the line that requires change is the following:
```
(1;algorithm=base,faithful=yes,bits=8\n) print
```

I would have made this a review remark, but due to the fact that the file is marked as binary, GitHub won't let me.

epithet:
Sorry, It's getting late in Europe :)

rbergen:
> Sorry, It's getting late in Europe :)

I know, I'm in the Netherlands myself :D

rbergen:
LGTM, I'm happy to merge this.

rbergen (approved):

rbergen:
@epithet May I ask you to let us know when this PR is stable? It will allow us to do one final review on your final version. :)

epithet:
Fair enough, just wanted to sneak one last change in.. :)
I declare it stable now and I'll make a new PR if I really think it's worth it.

rbergen:
Still LGTM

PR #251, opened 2021-06-26
"Single-file" solutions in Bash, Lua and Octave

rbergen:
This is a set of three solutions (languages) that effectively consist of one source code file, for which PRs were opened on the `original` branch (formerly `main`) before the current contributing guidelines were in place.

It concerns the following:
- A solution in Bash, originally submitted via PR #59. The Dockerfile is based on Ubuntu, because on my machine the Bash script seemed to run indefinitely on Alpine with Bash installed.
- A solution in Lua, originally submitted via PR #50
- A solution in Octave, originally submitted via PR #72. This solution technically consists of two source files because Octave rejects the sources otherwise (yes, I tried). Also, it's marked as arch-amd64 because there is currently no (semi-)official Docker image available for arm64, and creating one is no mean feat. I will try to add support for arm64 at a later time.

Note that: 
- I have pulled from the original PRs only what is needed to be able to add the languages to the drag-race collection (i.e., the bare source files). 
- I've made changes to the sources where necessary to adhere to drag-race defaults and output format
- I've added Dockerfiles and skeleton README.md files, similar to what we've done earlier with solutions that were merged before the current contributing guidelines were in place.

I will close the aforementioned PRs once this one has been merged.

### DIFF
--- a/PrimeBash/solution_1/Dockerfile
+++ b/PrimeBash/solution_1/Dockerfile
@@ -1,0 +1,8 @@
+FROM ubuntu:20.04
+
+RUN apt-get update
+
+WORKDIR /opt/app
+COPY PrimeBash.sh .
+
+ENTRYPOINT [ "./PrimeBash.sh" ]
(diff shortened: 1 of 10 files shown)


marghidanu (approved):

marghidanu:
LGTM

PR #252, opened 2021-06-27
Wren

marghidanu:
## Description

Wren language implementation

## Contributing requirements

### DIFF
--- a/PrimeWren/solution_1/Dockerfile
+++ b/PrimeWren/solution_1/Dockerfile
@@ -1,11 +1,14 @@
-FROM alpine:3.13
+FROM ubuntu:20.04
 
-RUN apk add --no-cache curl libc6-compat \
-    && curl -sL https://github.com/wren-lang/wren-cli/releases/download/0.3.0/wren_cli-linux-0.3.0.zip | unzip -d / - \
+RUN apt-get update \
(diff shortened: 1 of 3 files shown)


PrimeWren/solution_1/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+RUN apt-get update \
+    && apt-get install -y curl busybox 
+
+RUN curl -sL https://github.com/wren-lang/wren-cli/releases/download/0.3.0/wren_cli-linux-0.3.0.zip | busybox unzip - -d / \
+    && cp /wren_cli-linux-0.3.0/wren_cli /usr/local/bin \
+    && rm -rf /wren_cli-linux-0.3.0 \
+    && chmod a+x /usr/local/bin/wren_cli 
+
+WORKDIR /opt/app
+COPY *.wren .
```
rbergen: I'm thinking that it may be better to add a slash to the target, considering that it's a wildcard copy?

marghidanu: Should be ok now

rbergen: Yep, that's even cleaner. :)

rbergen (approved):
LGTM!

PR #255, opened 2021-06-30
Cpp

mikey-b:
## Description
* Moved std::map out of class to move need for allocation
* moved to bitset over vector<bool>, in combination with compile flags, seeing an improvement.
* bitset requires compile-time size, adjusted

## Contributing requirements

### DIFF
--- a/PrimeCPP/solution_1/PrimeCPP.cpp
+++ b/PrimeCPP/solution_1/PrimeCPP.cpp
@@ -29,13 +29,14 @@ const std::map<const long long, const int> resultsDictionary =
 
 };
 
+template<long sieveSize>
 class prime_sieve
 {
   private:
 
-      long sieveSize = 0;
-      vector<bool> Bits;
-
+      //long sieveSize = 0;
+      //vector<bool> Bits;
+      bitset<sieveSize> Bits;
 
       bool validateResults()
       {
@@ -47,9 +48,10 @@ class prime_sieve
 
    public:
 
-      prime_sieve(long n) 
-        : Bits(n, true), sieveSize(n)
+      prime_sieve() 
+        //: Bits(n, true), sieveSize(n)
       {
+            Bits.set();
       }
 
       ~prime_sieve()
@@ -128,7 +130,7 @@ int main()
 
     while (true)
     {
-        prime_sieve sieve(1000000L);
+        prime_sieve<1000000L> sieve{};
         sieve.runSieve();
         passes++;
         if (duration_cast<seconds>(steady_clock::now() - tStart).count() >= 5)
--- a/PrimeCPP/solution_1/PrimeCPP.cpp
+++ b/PrimeCPP/solution_1/PrimeCPP.cpp
@@ -14,26 +14,28 @@
 using namespace std;
 using namespace std::chrono;
 
+const std::map<const long long, const int> resultsDictionary = 
+{
+      {          10LL, 4         },               // Historical data for validating our results - the number of primes
+      {         100LL, 25        },               // to be found under some limit, such as 168 primes under 1000
+      {        1000LL, 168       },
+      {       10000LL, 1229      },
+      {      100000LL, 9592      },
+      {     1000000LL, 78498     },
+      {    10000000LL, 664579    },
+      {   100000000LL, 5761455   },
+      {  1000000000LL, 50847534  },
+      { 10000000000LL, 455052511 },
+
+};
+
 class prime_sieve
 {
   private:
 
       long sieveSize = 0;
       vector<bool> Bits;
-      const std::map<const long long, const int> resultsDictionary = 
-      {
-            {          10LL, 4         },               // Historical data for validating our results - the number of primes
-            {         100LL, 25        },               // to be found under some limit, such as 168 primes under 1000
-            {        1000LL, 168       },
-            {       10000LL, 1229      },
-            {      100000LL, 9592      },
-            {     1000000LL, 78498     },
-            {    10000000LL, 664579    },
-            {   100000000LL, 5761455   },
-            {  1000000000LL, 50847534  },
-            { 10000000000LL, 455052511 },
 
-      };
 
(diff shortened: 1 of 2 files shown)


mikey-b:
Hi

I am seeing a 5-10% improvement when using bitset over vector<bool>, some of that is in combination with compiler flag changes which do not help the vector<bool> case.

Requires moving sieve size to compile-time constant, I see bitset is still included as a header, is this a valid approach or not allowed in the contest?

Kind regards,
Mike Brown

mikey-b:
Great wins if we half the memory usage of the bitset 😅

rbergen:
I've aligned with @davepl, and he is not comfortable with switching from a dynamic sieve size to a compile-time constant; it's a potential optimization he had considered himself before, and decided to not apply. One way to put it is that the dynamic nature of the sieve size is an intrinsic part of what is now a "standard" implementation.

mikey-b:
Hi rbergen,

Ok, yes I accept that.

I have reviewed the asm, its pretty tight already. My only suggestion at this point would be an NT store in the hot loop (line 89) with a single fence to commit, this would stop us from trashing the cache.

To do this, will require intrinsics, is this acceptable in the contest?

Kind regards,
M

mikey-b:
In fact, no, we'd still need to load due to the bit-fiddling, but we could load to L1 only instead of trashing all levels

rbergen:
> To do this, will require intrinsics, is this acceptable in the contest?

That's a very fair question. To be honest, I'm not sure we can be sure until we have "potentially offending" code to look at. A key question will be if we can still consider the solution a C++ solution, or if it really turns into something else. In the latter case, we may still be able to accept the solution, but maybe under a different label.
Again, I for one won't know until I have an implementation to look at.

mikey-b:
> 
> 
> > To do this, will require intrinsics, is this acceptable in the contest?
> 
> That's a very fair question. To be honest, I'm not sure we can be sure until we have "potentially offending" code to look at. A key question will be if we can still consider the solution a C++ solution, or if it really turns into something else. In the latter case, we may still be able to accept the solution, but maybe under a different label.
> Again, I for one won't know until I have an implementation to look at.

Trying to control the cache has not resulted in any gains. Trading-off vector<bool> and bitset has however. this is a dynamic sieve size. Is this acceptable? Here is the gist:

```
static constexpr size_t _log2(size_t n) { return ( (n<2) ? 0 : 1+_log2(n/2)); }
static constexpr bool is_powerof2(size_t v) { return v && ((v & (v - 1)) == 0); }	

class dynamic_bitset {
	static constexpr size_t bucket_size = 32;
	static_assert(is_powerof2(bucket_size));
	std::vector<std::bitset<bucket_size>> data;
		
	static constexpr size_t mod = bucket_size - 1;
	static constexpr size_t div = _log2(bucket_size);	
 public:
	dynamic_bitset(unsigned long n): data((n >> div)+1, 0) {}
	~dynamic_bitset() = default; 
	
	void clear(size_t position) noexcept {
		//size_t position = pos >> 1;
        size_t block_id = position >> div;
        size_t offset = position & mod;
		
		data[block_id].set(offset, true);
	}
	
	bool operator[](size_t position) const noexcept {
		//size_t position = pos >> 1;
        size_t block_id = position >> div;
        size_t offset = position & mod;		
		
		return (data[block_id][offset] == 0);
	}
};
```

Kind regards,
M

davepl:
Two brief comments on this one:

1)  I’m not convinced that moving the scope of a STATIC variable is meaningful, is it?  I prefer it declared as close as possible to its use while remaining static, and I think the original was better in that sense.  I’d have to try it with just that change and before/after the binary, but I’m fairly confident that static is static.  If it's not a binary optimization, keep the old way.

2) The sieve cannot preserve allocation between passes, and must handle a different value being passed as the sieve size, which means it cannot be a static or constant or similar.  But as long at the sieve allocates all needed memory each pass and releases it too, that's what matters.  No "caching" of data between passes, in other words.

I had played with bitset<> early on but ruled it out because of the allocation issue.  But we should likely understand WHY it's a little faster, perhaps we can improve the specialization of vector<bool> in that same way?

mikey-b:
Hi @davepl , @rbergen 

I have submitted the adjusted code, which trades off vector and bitset, to help review. Some feedback

**1. I’m not convinced that moving the scope of a STATIC variable is meaningful**
Im not sure which variable you mean? I think you mean the std::map resultsDictionary? It wasn't marked as static in the branch I forked. It does look to have moved from the validateResults() in the multithreaded solution into the class scope in the single threaded solution. 

In both cases, when they were not static, would have allocated and constructed each time. I think this was just a booboo, and probably removed at class scope because you can't have a static initialised without using newer features. We have 3 options:

1. static at file scope.
2. Move it back to class scope, but mark it as static inline std::map<> - An inline variable.
3. Move it back to validateResults(), and as static.

Option 3 will still allocate on the first invoking and construct, but the function is called once and doesn't add to the overall time. I see no harm in this option if clearer to the reader

**2. The sieve cannot preserve allocation between passes, and must handle a different value being passed as the sieve size,**
This solution should now be acceptable to those requirements

**I had played with bitset<> early on but ruled it out because of the allocation issue. But we should likely understand WHY it's a little faster, perhaps we can improve the specialization of vector in that same way?**
I have attempted to do a specialised vector - I know a few tricks to optimise, and there is a few ways to set and clear a bit - but I wasn't able to beat vector<bitset<>> - I've even read the code in vector and bitset, nothing seems new or enlightening to me. Some serious wizardry in the stdlib.

There are a few other submissions to improve the CPP solution - I think some are doing their own container. They may have better luck than I did. 

I think you've some work on your plate reviewing all these pull requests with this repo! Happy to help with this vector<bitset<>> solution if you wish to pursue further. 

Kind regards,
M

rbergen:
Optimizations to CPP/solution_1 have been merged with #290. A PR on the current solution_1 can be opened if you feel further optimizations are possible.

PR #258, opened 2021-07-02
Added implementations in Kotlin with various target platforms

fvbakel:
## Description
Added implementations in Kotlin with various target platforms. More details are in the README.md files 

## Contributing requirements


rbergen:
Thanks for yet another contribution!

I do think we'll have to create a Kotlin-native base Docker image for solution-2. Due to the dependencies it pulls in to run the actual compile, the build stage took almost 3.5 minutes when I ran it on my machine. That's a little too long for comfort.

In case you're not familiar with setting this up, I am happy to help. We have created base images for a number of other languages/solutions, before.

fvbakel:
@rbergen thanks once more for the quick response. I have made an external Dockerfile for the native build. It is still a real big image, but it should be a bit faster now when you build for the first time.

PrimeKotlin/solution_1/README.md
```diff
@@ -0,0 +1,61 @@
+# Kotlin implementation
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-1-green)
```
rbergen: This question applies to all 3 solutions: I was wondering how you established that the number of bits per prime is 1? As I understand it, the type used for storage is Boolean, the memory size of which is undefined as far as I can tell.

rbergen:
> I have made an external Dockerfile for the native build. It is still a real big image, but it should be a bit faster now when you build for the first time.

Thanks, it did indeed save some time at my end. :)

PrimeKotlin/solution_1/README.md
fvbakel: Ah, good question. I think I misread the Kotlin documentation on this.

For solution_1 the Kotlin documentation says it is mapped to a boolean[] in JVM. See [https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-boolean-array/](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-boolean-array/)

I think a boolean[] in JVM uses one byte per value, based on [https://www.baeldung.com/jvm-boolean-memory-layout#object-sizes](https://www.baeldung.com/jvm-boolean-memory-layout#object-sizes). So I think for solution_1 we should set it to 8 bits.

For the native and JS target it is indeed no known. So I propose to set these to unknown. I am not sure if these are then still faithful.

rbergen: I think your proposal makes sense to me. 

Faithfulness and bit count are both independent characteristics of an implementation.

fvbakel: Updated the README files and source files as proposed.

PrimeKotlin/solution_2/README.md
```diff
@@ -0,0 +1,62 @@
...
+Or do step 2 and 3 with `go.sh`.
+
+## Output
+
+Below is an example of the output on my machine, running with Docker.
+
+```bash
+Passes: 1245, Time: 5.004, Avg: 0.004019277 (sec/pass), Limit: 1000000, Count: 78498, Valid: true
+
+fvbakel_Kotlin_native;1245;5.004;1;algorithm=base,faithful=yes,bits=1
```
rbergen: This really is nitpicking, but with the change in output (dropping the bits tag, see my other comment) this will also change.

PrimeKotlin/solution_2/primes-native.kt
```diff
@@ -0,0 +1,105 @@
...
+    val avg: Float = duration / passes
+
+    if (show_results) {
+      print_primes()
+    }
+    
+    println("Passes: $passes, Time: $duration, Avg: $avg (sec/pass), Limit: $limit, Count: $count, Valid: $valid")
+    // Following 2 lines are to conform to drag race output format
+    println("")
+    println("fvbakel_Kotlin_native;$passes;$duration;1;algorithm=base,faithful=yes,bits=unkown")
```
rbergen: If the bit count is unknown, the bits tag can/should be left out altogether. That is: unknown is the non-numeric default.
This obviously also applies to solution_3.

rbergen (commented):
I can smell the finish line :)

PrimeKotlin/solution_2/primes-native.kt
fvbakel: Did not know that.

PrimeKotlin/solution_2/README.md
fvbakel: I think it is great that you notice these small things, that's what helps in the quality.

rbergen: Thank you, I appreciate you saying that. I don't see everything though. For example, I recently missed something much more significant that required a PR (#277) to fix...

PrimeKotlin/solution_2/primes-native.kt
rbergen: It's in the guidelines, but easy to miss. And quickly corrected, luckily :)

rbergen (approved):
LGTM

rbergen:
@wulkanat and @fvbakel: I am hereby inviting you to discuss how to combine #258 and #309 into one Kotlin PR, preferably containing less than 5 solutions. The two PRs can't both be merged anyway due to the merge conflict that would lead to.

davepl:
Hi all... pardon my naivete on this one, but I'm curious to hear input on whether to count Kotlin implementations that use coroutines as single or multithreaded.  The latter sounds more accurate!

I was almost tempted to say "If the language supports it, you're free to use it" but then my own CPP multithreaded impl uses only language features and no operating system calls, so it would technically meet that requirement as well.

Thoughts?

fvbakel:
@davepl I am not aware that this implementation uses Kotlin `coroutines` under the surface. However if an implementation would use it, it becomes multi threaded. However I am not that familiar with Kotlin `coroutines` to be able to tell what happens under the surface.

I think a solution counts as multi threaded if multiple threads are used in the prime calculation it self. By what ever means those extra threads where created. In Oracle SQL for example this could be done with deferred  transactions.

Note that languages that have garbage collection, like Java and .NET, can have multiple threads running because of that. I would not consider that multi threaded for the drag-race definition.

jakobkmar:
I have added a solution using coroutines in #326 and I would definitely call it a multithreaded solution.

In my case, I am using the `Dispatchers.Default` coroutine dispatcher, which is backed by as many threads as you have cores (but at least 2 in any case).

Singlethreaded coroutines are also possible, but don't make any sense in this case as that thread is busy all the time and coroutines are all about making use of the time where a thread just waits for something to happen.

krzysiek-zgondek:
I think it would be great to use kotlin multiplatform and run this code natively targeting C then compare it with JVM target

Theaninova:
@krzysiek-zgondek #309 already does this (Kotlin/Native is vastly slower than JVM)

rbergen:
I'm closing this after the merge of #309. As you know, if you feel anything in this PR should be applied to that solution, you can open a PR that aims to improve it. :)

PR #262, opened 2021-07-03
Rust solution_1 tweaks: docker performance regression + new bit reset algorithm

mike-barber:
## Description
It looks like we're getting close to the finale, and there are a couple of things I've been meaning to do for a while. 
1. `Alpine` image causes a nasty performance regression due to some memory allocation inefficiency; Alpine has a different libc implementation and it doesn't appear to be playing well. This PR reverts to using the "normal" `Debian`-based Rust images for build and deploy. @rbergen I know you're overseeing the Docker images; this might be worth doing for the other Rust solutions too. Hope it doesn't create too much of a headache for you. 
2. Replaced the storage for the bit vector from u8 (bytes) to u32 (words) for a very slight performance bump, possibly due to alignment -- still bits, but in larger words.
3. Added a new bit storage variation that resets bits slightly differently by rotating a mask instead of having to do the shift every time. It outputs its own line. This one is quite fun. 

I don't think any of this will catch up with those Zig solutions though :) 

## Contributing requirements

### DIFF
--- a/PrimeRust/solution_1/Dockerfile
+++ b/PrimeRust/solution_1/Dockerfile
@@ -2,7 +2,11 @@
 #   - run unit tests 
 #   - build executable
 #
-FROM primeimages/rust:1.52.1 AS build
+
+# Note: Debian based, not Alpine; for some reason, the 
+# Alpine malloc call is slower than it should be.
+#
+FROM rust:1.52.1 AS build
 
 WORKDIR /app
 COPY . .
@@ -14,7 +18,8 @@ RUN cargo build --release
 #   - copy built executable 
 # 
 
-FROM alpine:3.13
+# Debian base image, not Alpine, for reason noted above
+FROM debian:buster-slim AS runtime
 
 WORKDIR /app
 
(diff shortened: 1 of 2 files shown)


rbergen:
Saying that I'm overseeing the Docker images is giving me too much credit (or blame, depending on one's perspective :) ). It just so happens that I've Dockerfiles to pre-guideline solutions that we grandfathered, and created Docker base images for some languages, for various reasons.

To be frank, I don't remember why we decided to create our own Docker base image for Rust, but looking at the current state of standard Rust images on Docker Hub, I can't come up with any objections to using those, now.

I will not apply the Docker base image switch to other Rust solutions "just because I can". Of course, I will consider (a) PR(s) that seek to make that change, if they are opened. :)

rbergen (approved):

marghidanu (approved):

marghidanu:
LGTM

mike-barber:
Thanks guys!

> Saying that I'm overseeing the Docker images is giving me too much credit (or blame, depending on one's perspective :) ). It just so happens that I've Dockerfiles to pre-guideline solutions that we grandfathered, and created Docker base images for some languages, for various reasons.

@rbergen to be clear -- it's definitely credit :) You guys have all been doing an awesome job with this repo.

rbergen:
(Note to self: click "Quote reply", not "Edit"...)

> @rbergen to be clear -- it's definitely credit :) You guys have all been doing an awesome job with this repo.

@mike-barber I really appreciate that you say that, thank you. I'm just having so much fun doing this that it must be bad for my health. :)

PR #264, opened 2021-07-03
Update PrimePas.dpr

KimMadsen:
Performance improved version that nearly doubles performance.

## Contributing requirements

### DIFF
--- a/PrimeDelphi/solution_1/PrimePas.dpr
+++ b/PrimeDelphi/solution_1/PrimePas.dpr
@@ -1,24 +1,27 @@
 program PrimePas;
 
+// Optimized by Kim Madsen/C4D
+// www.components4developers.com
+
 {$APPTYPE CONSOLE}
 
 uses
-	System.SysUtils,
-	Classes,
-	System.Generics.Collections,
-	System.Timespan,
-	Math,
-	Windows;
+  System.SysUtils,
+  Classes,
+  System.Generics.Collections,
+  System.Timespan,
+  Math,
+  Windows;
 
 type
 	TPrimeSieve = class
 	private
-		FSieveSize: Integer;
+		FSieveSize: NativeInt;
+		FSieveSize2: NativeInt;
+		FSieveSizeSqrt: NativeInt;
 		FBitArray: array of ByteBool; //ByteBool: 4644. WordBool: 4232. LongBool: 3673
-		FMyDict: TDictionary<Integer, Integer>;
+		FMyDict: TDictionary<NativeInt, NativeInt>;
 
-		function GetBit(Index: Integer): Boolean;
-		procedure ClearBit(Index: Integer);
 		procedure InitializeBits;
 	public
 		constructor Create(Size: Integer);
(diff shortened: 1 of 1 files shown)


KimMadsen:
Performance improved Delphi code.

glenkleidon:
See my pull request #365 - Adding the Inline compiler hint  for getbit and clear bit also doubles performance.  I see you have elminated those methods in any case so that seems consistent.

I note also there is some discussion the comments that `mod 2` is faster than `and 1` - I have not found this to be the case.

glenkleidon:
@KimMadsen I made a pull request against your patch-1 branch for converting the dictionary to static.

mathias-burbach:
I am new here and I have not made any contributions yet. But I found a way to improve the current Delphi implementation.

Instead of

procedure TPrimeSieve.ClearBit(Index: Integer);
begin
  if (Index mod 2) = 0 then
    Exit;
  FBitArray[Index div 2] := False;
end;

we could use

procedure TPrimeSieve.ClearBit(const Index: Integer);
begin
  if (Index mod 2) = 1 then
    FBitArray[Index div 2] := False;
end;

I have observed a 2.1853% increase of passes without the Exit; call.

Sorry if the indention isn't working all that well in this editor.

Salut,
  _Mathias_

glenkleidon:
Hey Mathias,

 

There are branches far ahead of the master.

 

See Issue 264 https://github.com/PlummersSoftwareLLC/Primes/pull/264

 

and I also have a pull request https://github.com/KimMadsen/Primes/pull/1 against KimAMadsens branch from my (misspelled ) branch https://github.com/glenkleidon/Primes/tree/path-1_gk

 

Rutger has made an issue for us to resolve all of our updates into one release.

 

#316 <https://github.com/PlummersSoftwareLLC/Primes/discussions/316> 

 

So far Kim has not responded.

 

Glen.

mathias-burbach:
Hello Glen,

Thanks your reply. You are right I was not aware of the more recent branch [#264](https://github.com/PlummersSoftwareLLC/Primes/pull/264).

I have downloaded it and find it extremely educational to compare the old version against the newer version. :-)

If I run the latest branch on my Delphi _VMware_ virtual machine I get to 9046 passes for 32bit.

I wondered if there is a chance to further improve this by using unsigned integers like _NativeUInt_ and convert the passes into _NativeUInt_ too? But that slowed it down slightly. I thought it could help to increment the counter on less bytes, but it seems _Inc_ works best on _dword_.

So with my limited assembler knowledge it seems that there is not much more that I can contribute.

But thanks for all the improvements anyhow and giving me the chance to study them.

Salut,
  _Mathias_

glenkleidon:
Probably,

 

But the biggest bottleneck is setting the bit array.

 

I am just testing re-implementation of Dave’s original GetBit and ClearBit to be identical to the C++ code (and  then inlining them) – I cant see why it wouldn’t run at the same speed as the C++ code then.

 

Glen.

glenkleidon:
I built the current C++ version on the same machine as my patch for Kim’s solution.

 

My latest Delphi version runs as fast as the c++ solution now.  But its no longer strictly apples for apples as Kim’s solution reduces the array size by 50% by not allocating memory for the unus(able) even number array records.  The C++ one doesn’t do that.  If it did, then likely it would be faster again.

 

https://github.com/PlummersSoftwareLLC/Primes/discussions/316

 

 

Glen.

 

 

 

From: ***@***.*** ***@***.***> 
Sent: Sunday, 11 July 2021 6:03 PM
To: 'PlummersSoftwareLLC/Primes' ***@***.***>
Subject: RE: [PlummersSoftwareLLC/Primes] Update PrimePas.dpr (#264)

 

Probably,

 

But the biggest bottleneck is setting the bit array.

 

I am just testing re-implementation of Dave’s original GetBit and ClearBit to be identical to the C++ code (and  then inlining them) – I cant see why it wouldn’t run at the same speed as the C++ code then.

 

Glen.

rbergen:
@KimMadsen Do you still want to pursue getting this PR merged? The same question applies to #281.

rbergen:
Closing this based on comments made in #281.

PR #265, opened 2021-07-03
Startup checks for docker and node

cinderblock:
## Description
If you're running these without docker installed (or running), you get a whole lot of noisy errors that seem daunting. Let's make it more clear.

To be nice, we can also check if Node.js is installed at the start. The `Makefile` uses `npm` and `node` at the end to generate the report. It's a bummer to run the whole script only to get a node error at the end, especially since the work is done in a `mktmp` folder that is possibly difficult to find again.

## Contributing requirements

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ report: sanity-node benchmark
 	@cd tools/; \
 	npm ci && npm start -- report -d "$(OUTPUT_DIR)"
 
-one:
+one: sanity-checks
 	@if [[ ! -z "$${SOLUTION}" ]]; then \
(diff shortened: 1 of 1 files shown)


marghidanu:
Hey, @cinderblock thanks for this PR. This was a task on my list but I didn't get around to do it, everything looks good here.

rbergen:
I agree that this makes the benchmark tool more helpful. I was thinking that adding a link in the "check-*-works" messages to [the benchmark run instructions](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/BENCHMARK.md) might make it a little more helpful, still?

rbergen:
I'll merge this, and follow it up with a PR that replaces the links to the download pages with a link to the BENCHMARK.md.

PR #266, opened 2021-07-03
Cpp constexpr

flo80:
## Description

A solution heavily using compile time optimisation.

Note: I did not see anything in the rules against compile time optimisation but it might be considered a deviation from the base algorithm or not faithful 

## Contributing requirements

* [ X ] I read the contribution guidelines in CONTRIBUTING.md.
* [ X ] I placed my solution in the correct solution folder.
* [ X ] I added a README.md with the right badge(s).
* [ X ] I added a Dockerfile that builds and runs my solution.
* [ X ] I selected `drag-race` as the target branch.
* [ X ] All code herein is licensed compatible with BSD-3.


rbergen:
Thanks for opening this, it really made me scratch my head a few times while figuring out what I thought of it.

Just speaking for myself, I'm happy to add this as a third solution, because I think it covers a very interesting C++ edge case that has come up a few times before., However, I don't think we can add it under the faithful moniker. The reason for that is that the sieve buffer is effectively set (maximized) at compile time, even though the extent of use of that sieve buffer strictly isn't. I admit that the guidelines don't explicitly rule out this approach, but I've aligned with Dave on this point in relation to other recent PRs (#255 and #256), and we agree that fixing the sieve size (limit) at compile time isn't what the definition of faithful intended to express.

So, I will:
1. happily check and review this PR for inclusion, provided that it is submitted as not faithful
2. clarify the contributing guidelines on this specific point.

flo80:
I like this clarification, this should shake up the leaderboard (not an expert but e.g. zig uses the same mechanism, comptime, to perform tasks during compile time). I assume a few solutions now will need to be declared not faithful.

I have updated my solution tags and also added a single threaded performance

rbergen:
I now see there is another, and more fundamental, reason why at least one solution needs to be marked as unfaithful, being that the sieve objects are not created from scratch for each sieve run. I simply missed that when reviewing it.

PrimeCPP/solution_3/.gitignore
```diff
@@ -0,0 +1,2 @@
+.vscore/**
```
rbergen: I'm guessing you meant to .gitignore .vsco**d**e/**. That would also have removed the .vscode/tasks.json from the commit :)

PrimeCPP/solution_3/README.md
```diff
@@ -0,0 +1,68 @@
...
+## Output
+
+All on Apple M1 (Macbook Air)
+
+### Native performance
+
+     primes to 1000000 on 8 threads for 5 seconds.
+     Passes: 637960, Time: 5.002253, Avg: 0.000008, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1
+
+     flo80;637960;5.002253;8;algorithm=base,faithful=yes,bits=1
```
rbergen: I'm not going to make this a blocking review comment, but with the switch to `faithful=no` in the output, this will change accordingly when you run it.

ityonemo:
please clarify what you mean by "from scratch", e.g. does the memory need to be fully released to the operating system?

PrimeCPP/solution_3/.gitignore
flo80: That was actually a copy & paste from solution 2, will fix it

PrimeCPP/solution_3/README.md
flo80: updating

rbergen (approved):
LGTM

marghidanu (approved):

PR #268, opened 2021-07-03
Create Project1.dpr

KimMadsen:
Delphi solution using only 1 bit per prime candidate as required.
Using optimized bits class by Kim Madsen/Components4Developers (www.components4developers.com)

## Contributing requirements


KimMadsen:
Delphi single bit per prime candidate solution.

PR #269, opened 2021-07-03
LaTeX port

tjol:
## Description

This is a bit silly, but I've ported the program to LaTeX. I translated the original PrimePython as faithfully as I could, but LaTeX doesn't exactly offer a lot by way of easy-to-use encapsulation, so there's no class or class equivalent.

## Contributing requirements

### DIFF
--- a/PrimeLaTeX/solution_1/.gitignore
+++ b/PrimeLaTeX/solution_1/.gitignore
@@ -1,0 +1,6 @@
+*~
+*.aux
+*.dvi
+*.pdf
+*.log
+result.txt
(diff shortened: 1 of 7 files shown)


rbergen:
I don't think this is silly. I really value the submissions of "proper contestants" for the top spots on the leaderboard (and the cleverness of the improvements that people come up with towards those languages), but I also enjoy including less obvious candidates. Languages in the "clearly not made for this" category are very much included in that latter category.

rbergen (approved):
LGTM

marghidanu (approved):

marghidanu:
Love it! LGTM

PR #270, opened 2021-07-03
reset bitArray values between runs

MikeAWilliams:
## Description
Go makes a shallow copy of slices. So, the existing go solution only performs the prime sieve once and after that gets the slice with the solution already in place. I just made a quick fix to initialize the slice before every run.

## Contributing requirements

### DIFF
--- a/PrimeGo/solution_1/main.go
+++ b/PrimeGo/solution_1/main.go
@@ -98,10 +98,10 @@ func main() {
 	startClock := time.Now()
 
 	initBitArray := make([]bool, 1e6)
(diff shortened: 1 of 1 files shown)


marghidanu:
Well spotted!

marghidanu (approved):

marghidanu:
LGTM

rbergen (approved):
LGTM too, and if I'm honest I feel remiss for not seeing this earlier.

marghidanu:
I know, me too :) but we have a lot of languages on our hands.

PR #271, opened 2021-07-03
Refactored PrimeC/solution2

TheMatt2:
## Description
Some of the files had a mix of tabs and spaces. All spaces changed so they are only spaces.

Other whitespace was also changed to make the source code more readable.

One variable was changed for readability "t" was changed to "t1". One expression was expended for readability, this technically changes the binary when no optimizations are applied.

It was checked that these changes do not change the generated executable when compiled for x86 with gcc and optimizations.

### DIFF
--- a/PrimeC/solution_2/sieve_480of2310_epar.c
+++ b/PrimeC/solution_2/sieve_480of2310_epar.c
@@ -23,8 +23,8 @@
 
 // Steps array for finding the next number not divisible by 2,3,5,7,11
 // Pre halved
-static unsigned int steps[480]={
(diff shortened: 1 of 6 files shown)


rbergen:
As far as I can tell, these changes are purely cosmetic. As such, I don't think we should merge this.
@marghidanu, do you have a different opinion?

TheMatt2:
They are intended to only be cosmetic. Though this does fix the mixing of spaces and tabs.

rbergen:
For a number of reasons, we do consider it a key requirement for accepting changes to existing solutions that they improve the performance of those solutions. As this PR doesn't concern a change of that type, I'm closing this.

PR #273, opened 2021-07-04
Single run fix

marghidanu:
The `make one` target contained a bug that was creating an additional `tee` file in the project's directory.


rbergen (approved):

rbergen:
I do wonder, how did you find this one? 😄

marghidanu:
I used the target to test Pony, and when I wanted to commit the solution VS Code showed me an extra file in there named… tee.

Needless to say that I made the fix immediately.

PR #274, opened 2021-07-04
Clarifying dynamic sieve size/memory requirement

rbergen:
This aims to clarify the requirements regarding faithful implementations, in the sense that sieve and memory buffer size must be set at runtime. The need for this clarification is illustrated by PRs #255, #256 and #266.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -162,6 +162,7 @@ At a technical level, an implementation is considered faithful if it complies wi
 
 * It uses no external dependencies to calculate the actual sieve.
 * It uses a class to encapsulate the sieve, or an equivalent feature in your language. This class must contain the full state of the sieve. Each iteration should re-create a new instance of this class.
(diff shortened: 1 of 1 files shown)


ityonemo:
> Each iteration should re-create a new instance of this class.

Can you clarify what you mean by this?  Does all the memory need to be released *and* freed to the operating system?

marghidanu (approved):

rbergen:
> > Each iteration should re-create a new instance of this class.
> 
> Can you clarify what you mean by this? Does all the memory need to be released _and_ freed to the operating system?

Actively releasing the memory to the OS is not something we can demand, because there are quite a few languages/runtimes that handle memory management and garbage collection around the actual program source code. C#, one of the original 3 languages in the drag race, is one of them. Those languages often don't allow the programmer to directly control this.

That said, in languages where memory is explicitly allocated and released by the programmer, I personally think it is very bad practice not to do so. As it is required  to create from scratch the sieve for every run and dynamically allocate memory for it, not releasing that memory means you leak the memory size of a sieve with every run. I've accidentally done this in my Pascal implementation once, and that wasn't pretty. It didn't perform either, because growing the heap to gigabytes in the end took quite a lot of time.

PR #277, opened 2021-07-04
Marking PrimeZig/solution_3 unfaithful

rbergen:
The implementations in PrimeZig/solution_3 need to be marked unfaithful, for two reasons:

- The prime sieve objects are not recreated from scratch for every sieve run, but reset
- The memory for the sieve objects is effectively statically allocated at compile time. This also violates the "from scratch" principle. #274 has been opened to clarify the contributing guidelines on this point

The reason this PR is necessary is simply that I missed this in my review of #241.

### DIFF
--- a/PrimeZig/solution_3/README.md
+++ b/PrimeZig/solution_3/README.md
@@ -2,7 +2,7 @@
 
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Algorithm](https://img.shields.io/badge/Algorithm-wheel-yellowgreen)
-![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-no-green)
 ![Parallelism](https://img.shields.io/badge/Parallel-no-green)
 ![Parallelism](https://img.shields.io/badge/Parallel-yes-green)
 ![Bit count](https://img.shields.io/badge/Bits-1-green)
--- a/PrimeZig/solution_3/src/main.zig
+++ b/PrimeZig/solution_3/src/main.zig
@@ -145,7 +145,7 @@ fn printResults(backing: []const u8, passes: usize, elapsed_ns: u64, threads: us
     const stdout = std.io.getStdOut().writer();
     const algo = if (wheel) "wheel" else "base";
 
-    try stdout.print("{s};{};{d:.5};{};faithful=yes,algorithm={s},bits={}\n", .{
+    try stdout.print("{s};{};{d:.5};{};faithful=no,algorithm={s},bits={}\n", .{
         backing, passes, elapsed, threads, algo, bits
     });
 }


marghidanu:
I agree with you, upon closer inspection, it looks like the solution is indeed unfaithful.

marghidanu (approved):

ityonemo:
These criteria were instituted after submission.  Being pinged on these rules changes BEFORE marking as unfaithful would have been nice, so that I could change the code.

> The prime sieve objects are not recreated from scratch for every sieve run, but reset

Before each run, the entire working set is zeroed out; what do we mean by "create"?  Isn't it a smart thing to reuse memory and avoid a trip to the kernel?  Hell, a smart malloc implementation (maybe jemalloc?) does that anyways, so this is effectively disqualifying zig because of its explicitness gets you to actually read the code instead of "trusting libc".  I'm pretty sure the erlang implementation, for example, will not reallocate the space with a trip to the kernel, because the erlang VM hoards the memory for itself by punting to its internal allocator:  http://erlang.org/doc/man/erl_nif.html#enif_alloc

ityonemo:
https://github.com/jemalloc/jemalloc/issues/1128

> Typically, unused memory will get on a sliding scale, with the default being madvising it back to the OS after about 10 seconds.

So, over the lifetime of these runs, any system using jemalloc will not be freeing any memory back to the OS between runs.

ityonemo:
more important commentary:

https://unix.stackexchange.com/questions/53447/does-free-unmap-the-memory-of-a-process

> The C-library function free() can, but does not have to, return memory to the kernel.

https://www.geekyhacker.com/2019/01/04/jvm-does-not-release-memory/

rbergen:
> These criteria were instituted after submission. Being pinged on these rules changes BEFORE marking as unfaithful would have been nice, so that I could change the code.

That is not the case. The requirement that the sieve objects are created from scratch for every run is not new, it has been in the guidelines from the very beginning. There are also many examples in earlier PRs of solutions being marked unfaithful based on this rule being applied to this code pattern. A mistake was made here, but it is that I completely overlooked this when reviewing PrimeZig/solution_3. For that I apologize.

What we are adding to the guidelines now is a clarification on memory allocation, based on the definition of "from scratch" leaving room for interpretation. But again, PrimeZig/solution_3 should have been marked as unfaithful based on the old ruleset, too.

rbergen:
> Before each run, the entire working set is zeroed out; what do we mean by "create"? Isn't it a smart thing to reuse memory and avoid a trip to the kernel? 

Yes, it is very smart, just not within the boundaries of a faithful implementation. It's also important to be aware that this is not a value statement, but the assessment of one of the characteristics we distinguish in implementations. And the fact is that, with the faithfulness tag being what it is, the solution does still make a strong demonstration of the benefits of your approach.

Taking a few steps back, the key point of this repo is to facilitate a drag race between programming languages, where speed of memory management is an aspect that Dave explicitly includes in what he wants to compare. To be able to do that, the key characteristics of implementations need to be pretty tightly defined. That is also why algorithms based on wheel factorizations are accepted but so marked, and the number of bits per prime is also considered and made explicit, to name some examples.

Moving forward, I think PrimeZig/solution_3 is a great solution to have in the repo as it is. Making it faithful would require the init (and I personally think deinit, as I have indicated in my comment on #274) to be pulled into the loop, and the memory allocator not relying on a buffer that is effectively allocated at compile time.

PR #278, opened 2021-07-04
Python optimizations

chunheguo:
## Contributing requirements

```
$ python3 PrimePY.py 
Passes: 4327, Time: 10.001078526999999, Avg: 0.0023113192805639007, Limit: 1000000, Count: 78498, Valid: True

ssovest; 4327;10.001078526999999;1;algorithm=base,faithful=yes,bits=8
```

purely python improvements without using numpy or numba/jit

### DIFF
--- a/PrimePython/solution_2/PrimePY.py
+++ b/PrimePython/solution_2/PrimePY.py
@@ -5,9 +5,8 @@ MyFirstPython Program (tm) Dave Plummer 8/9/2018
 Updated 3/22/2021 for Dave's Garage episode comparing C++, C#, and Python
 """
 
-from math import sqrt
-
-
+from math import sqrt, ceil
+    
 class PrimeSieve:
 
     """This is the main PrimeSieve class. Call it with the number you wish as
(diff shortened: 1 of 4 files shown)


rbergen:
I'm wondering what is the nature of the improvements you intend to make? Phrased differently, what goal(s) are you aiming for?

chunheguo:
just using plain std python to improve the performance, without using numpy or jit.
While loops are slower than for loops, build in map functions are usually faster then for loops

PrimePython/solution_2/PrimePY.py
```diff
@@ -33,38 +32,40 @@ def validate_results(self):                      # Check to see if this is an up
...
 
-        factor = 1
         # sqrt doesn't seem to make any difference in CPython,
         # but works much faster than "x**.5" in Pypy
         q = sqrt(self._size) / 2
         bitslen = len(self._bits)
+        q_int = ceil(q)
 
-        while factor <= q:
+        for factor in range(1, q_int):
```
marghidanu: I believe the while loop might be faster than the for loop being used here.

marghidanu:
Can you provide any numbers that show some improvements between the original code and your changes?

chunheguo:
> Can you provide any numbers that show some improvements between the original code and your changes?

https://github.com/PlummersSoftwareLLC/Primes/pull/278/files#diff-ce75757ace6ff01ca0eeb44df7ceac668a8a04d8cce5391919d55940dcbb0f01R46

I was going off the values in the readme: 
https://github.com/PlummersSoftwareLLC/Primes/pull/278/files#diff-ce75757ace6ff01ca0eeb44df7ceac668a8a04d8cce5391919d55940dcbb0f01L57

rbergen:
I don't/no longer see a performance-improving change in the drag-race relevant part of the solution. As such, I think we should close this.

PR #279, opened 2021-07-04
Improved Tcl solution_2  by adding a second implementation

fvbakel:
## Description
Improved Tcl solution_2  by adding Tcl solution_3. The improvements are significant different that I decided to implement it as a separate solution. In the process I found some minor mistakes in the README of solution_2 that I solved as well. 
On my installation the solution_3 has a factor 2 performance improvement.

## Contributing requirements


rbergen:
I understand your proposal to add a third solution, but in my opinion, a significant improvement in performance is in itself not enough reason to add a new solution. That statement applies in general, but certainly if you're the author of the solution you seek to improve :) 

As far as I can see, the fundamental characteristics of both solutions are the same, so I would in fact propose to apply the improvements to solution_2. If you really want to demonstrate differences in implementation approaches, then you could consider adding a second implementation to that solution, generating a line of drag-race format output for each.

fvbakel:
I want to demonstrate differences in implementation and performance, so I refactored it to an extension of solution_2

rbergen (commented):
I am unable to run the Docker image with the current combination of Dockerfile and run.sh, because the image doesn't know how to execute the run.sh shell script. 
I think there are two approaches to solving this:
- changing the ENTRYPOINT in the Dockerfile to "sh","./run.sh", or
- adding a shebang to run.sh that points to /bin/sh

fvbakel:
@rbergen Sorry about the mistake, however it allowed me to sneak in one more performance improvement that I missed.

rbergen:
> however it allowed me to sneak in one more performance improvement that I missed.

Well, we do love those! :D

rbergen (approved):
LGTM

PR #280, opened 2021-07-04
Remove unnecessary unique_ptr from PrimeCPP_PAR.cpp

jruoff:
## Description
The prime_sieve can be stack allocated without a problem:
(1) The bulk of the sieve will still be heap allocated because vector<bool> does a heap allocation.
(2) The prime_sieve will be destructed at the same moment as the unique_ptr was destructed, after all, the unique_ptr was stack allocated just the same.
 
## Contributing requirements
* [Yes ] I read the contribution guidelines in CONTRIBUTING.md.
* [ N/A] I placed my solution in the correct solution folder.
* [ N/A] I added a README.md with the right badge(s).
* [ N/A] I added a Dockerfile that builds and runs my solution.
* [ Yes] I selected `drag-race` as the target branch.
* [ Yes] All code herein is licensed compatible with BSD-3.

### DIFF
--- a/PrimeCPP/solution_2/PrimeCPP_PAR.cpp
+++ b/PrimeCPP/solution_2/PrimeCPP_PAR.cpp
@@ -11,7 +11,6 @@
 #include <cmath>
 #include <vector>
 #include <thread>
-#include <memory>
 
 using namespace std;
 using namespace std::chrono;
@@ -242,7 +241,7 @@ int main(int argc, char **argv)
 
     if (bOneshot)
     {
-        std::unique_ptr<prime_sieve>(new prime_sieve(llUpperLimit))->runSieve();
+        prime_sieve(llUpperLimit).runSieve();
     }
     else
     {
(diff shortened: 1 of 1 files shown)


rbergen:
Unless this change achieves an actual improvement in performance, I think this is an implementation decision in the area of personal preference. I've been running both the stack and heap allocated versions of the solution for a few minutes now, and I don't see better performance in the stack-allocated implementation, here.

Do you see other numbers?

jruoff:
I don't expect any significant performance improvement for this change. (Though it does eliminate a couple allocations.) The main purpose of this change is to simplify the code, making it easier to read and avoid teaching anti-patterns (in this case unnecessarily using unique_ptrs) to people new to C++.

rbergen:
I understand. However, teaching people C++ is not the primary goal of this project/repo. In fact, I would say that the use of anti-patterns is actually recommended if that squeezes the most out of the language in question. Furthermore, the solution your PR targets is one of the original set. That means we can really only consider changes that do positively impact performance, and then only with the original author's approval.

Based on these considerations, I'm closing this PR.

PR #282, opened 2021-07-04
Added VB.NET implementation

jimbojim1997:
## Description
This is a VB.NET translation of davepl's C# implementation with minor differences.
1. The ```PrimeSieve``` implementation is in a separate class to the benchmarking code.
2. The benchmark uses a ```System.Diagnostics.Stopwatch``` to measure the execution time rather than a ```DateTime```.
3. The ```_knownPrimes``` dictionary is now ```Shared``` so it isn't re-instantiated for each test.
4. Method names changed to match VB convention.
5. Results reported from within ```Main``` rather than a separate method.

## Contributing requirements


rbergen:
We have an existing VB.NET solution in the repository, as [PrimeBASIC/solution_2](https://github.com/PlummersSoftwareLLC/Primes/tree/drag-race/PrimeBASIC/solution_2). Please review that solution, and if you feel the solution this PR aims to add should still be added, then either:

- Change this PR to one that improves the existing solution. Please explain what (performance) improvement you aim to achieve regarding the existing solution.
- Change this PR to use PrimeBASIC/solution_3 as a target directory. Please explain why you feel it is warranted to add this solution next to the one that is already in the codebase.

PR #283, opened 2021-07-04
F# performance improvements ~75%

dmannock:
## Description
Performance improvements to my previous F# solutions 2 & 3. Gives ~75% boost from changes involving sieve init, inlining & spans.

Test results updated in the readme files.

## Contributing requirements

### DIFF
--- a/PrimeFSharp/solution_2/README.md
+++ b/PrimeFSharp/solution_2/README.md
@@ -3,7 +3,7 @@
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
 ![Parallelism](https://img.shields.io/badge/Parallel-no-green)
-![Bit count](https://img.shields.io/badge/unknown-yellowgreen)
+![Bit count](https://img.shields.io/badge/Bits-unknown-yellowgreen)
 
 Port of the original C++ prime number sieve to F#. Matching as closely as possible by using while loops. 
 
--- a/PrimeFSharp/solution_2/README.md
+++ b/PrimeFSharp/solution_2/README.md
@@ -3,7 +3,7 @@
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
 ![Parallelism](https://img.shields.io/badge/Parallel-no-green)
-![Bit count](https://img.shields.io/badge/Bits-1-green)
+![Bit count](https://img.shields.io/badge/unknown-yellowgreen)
 
 Port of the original C++ prime number sieve to F#. Matching as closely as possible by using while loops. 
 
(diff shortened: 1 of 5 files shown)


rbergen:
Now that I'm reviewing this, I'm wondering if it is correct to mark the bit count per prime of these solution as 1. The type that is used to store the prime number flags is bool, which I think means the number of bits per prime is unknown.

dmannock:
Good spot. It appears bits isn't in the output when using BitArray / bool array elsewhere. Omitting it for consistency sounds like a plan 👍

rbergen:
> Good spot. It appears bits isn't in the output when using BitArray / bool array elsewhere. Omitting it for consistency sounds like a plan 👍

The use of bits=1 in solution_1 is accurate, because that implementation uses a BitArray class instead of a bool array. A BitArray guarantees that each item in the class does use one bit. That is to say, there is a relevant difference between the implementations of solution_1 on the one hand, and the other 2 solutions on the other. 

With that in mind, I think that: 
- the bits tag should indeed be removed from the output of solution_2 and solution_3, and the Bits badges in the README.md files of those solutions should be changed to `unknown-yellowgreen`, accordingly. 
- solution_1 should not be modified on this point.

dmannock:
Sorry, my bad. I Intended to update to be consistent with C# solution where thought it defaults to null when using BitArray and so doesn't output bits=1. However, on second glance the implementation that's used in the run all does indeed have BitsPerPrime = 1.  The consistency changes there would better suit another PR if there isn't one already open. I'll sort this PR out with what applies in this context.

PrimeFSharp/solution_3/README.md
```diff
@@ -3,7 +3,7 @@
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
 ![Parallelism](https://img.shields.io/badge/Parallel-no-green)
-![Bit count](https://img.shields.io/badge/Bits-1-green)
+![Bit count](https://img.shields.io/badge/unknown-yellowgreen)
```
rbergen: The full markdown for the Bits badge should be this:

```
```

Yielding this badge:
Bit count

PrimeFSharp/solution_2/README.md
```diff
@@ -3,7 +3,7 @@
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
 ![Parallelism](https://img.shields.io/badge/Parallel-no-green)
-![Bit count](https://img.shields.io/badge/Bits-1-green)
+![Bit count](https://img.shields.io/badge/unknown-yellowgreen)
```
rbergen: See my previous comment about the badge markdown.

PrimeFSharp/solution_3/README.md
dmannock: Cheers, glad someone has eyes today! 👀

rbergen:
I'm going to merge this as it concerns a scoped change by the original solutions' author.

PR #284, opened 2021-07-04
Forth solution

tjol:
## Description

I thought I'd do another relatively obscure language that I know little about.

I've marked it as “faithful” as the memory is reallocated and reinitialized every time, even though there are no classes or structs.

There are some extra spaces in the output – I hope these are ok. If not, I should be able to get rid of them with a bit of effort.

(It's slower than I thought it would be!)

## Contributing requirements

### DIFF
--- a/PrimeForth/solution_1/Dockerfile
+++ b/PrimeForth/solution_1/Dockerfile
@@ -1,0 +1,7 @@
+FROM mtrute/gforth-container:latest
+
+WORKDIR /opt/app
+
+COPY prime-bitarray.fs .
+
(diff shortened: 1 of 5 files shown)


rbergen:
I'll gladly add this new member to the collection of Primes languages, but I will have to insist that it is marked as unfaithful for the reason you have yourself indicated. And I do that acknowledging fully that this is due to limitations that are intrinsic to the language.

tjol:
Absolutely. That's why I pointed it out!

rbergen (approved):
LGTM. I've verified that the auxiliary spaces in the output don't prevent its parsing by the benchmark tool.

rbergen:
Still works on my machine

PR #285, opened 2021-07-04
optimize PostScript through early binding

epithet:
## Description
The main procedure is inlined and names are bound before executing the loop, so that they don't have to be looked up in each iteration.

With this simple change, I got between 6% and 245% more iterations, depending on the execution environment. It also seems to decrease the variance of the results.

## Contributing requirements


rbergen (approved):
LGTM. I'll merge this immediately as it's a small change to the contributor's own solution, with a confirmed performance impact.

PR #286, opened 2021-07-04
Common lisp based on mikehw's solution

mayerrobert:
## Description

This seems to give a 120x speedup.

Cheers, Robert.

## Contributing requirements

### DIFF
--- a/PrimeLisp/solution_2/PrimeSieve.lisp
+++ b/PrimeLisp/solution_2/PrimeSieve.lisp
@@ -54,13 +54,14 @@
        (start (get-internal-real-time))
        (end (+ start (* internal-time-units-per-second 5))))
 
-  (do () ((>= (get-internal-real-time) end))
-    (run-sieve 1000000)
-    (incf passes))
-
-  (let* ((duration  (/ (- (get-internal-real-time) start) internal-time-units-per-second))
-         (avg (/ duration passes)))
-    (print (list "Passes:" passes "Time:" duration "Avg" avg "Count" (count-primes (run-sieve 1000000))) *error-output*)
-    (terpri *error-output*)
-
-    (format t "mayerrobert-cl;~d;~f;1;algorithm=base,faithful=no,bits=1~%" passes duration)))
+  (prog (result)
+    (do () ((>= (get-internal-real-time) end))
+      (setq result (run-sieve 1000000))
+      (incf passes))
+  
+    (let* ((duration  (/ (- (get-internal-real-time) start) internal-time-units-per-second))
+           (avg (/ duration passes)))
+      (print (list "Passes:" passes "Time:" duration "Avg" avg "Count" (count-primes result)) *error-output*)
+      (terpri *error-output*)
+  
+      (format t "mayerrobert-cl;~d;~f;1;algorithm=base,faithful=yes,bits=1~%" passes duration))))
--- a/PrimeLisp/solution_2/PrimeSieve.lisp
+++ b/PrimeLisp/solution_2/PrimeSieve.lisp
@@ -61,5 +61,6 @@
   (let* ((duration  (/ (- (get-internal-real-time) start) internal-time-units-per-second))
          (avg (/ duration passes)))
     (print (list "Passes:" passes "Time:" duration "Avg" avg "Count" (count-primes (run-sieve 1000000))) *error-output*)
+    (terpri *error-output*)
 
     (format t "mayerrobert-cl;~d;~f;1;algorithm=base,faithful=yes,bits=1~%" passes duration)))
--- a/PrimeLisp/solution_2/PrimeSieve.lisp
+++ b/PrimeLisp/solution_2/PrimeSieve.lisp
@@ -63,4 +63,4 @@
     (print (list "Passes:" passes "Time:" duration "Avg" avg "Count" (count-primes (run-sieve 1000000))) *error-output*)
     (terpri *error-output*)
 
-    (format t "mayerrobert-cl;~d;~f;1;algorithm=base,faithful=yes,bits=1~%" passes duration)))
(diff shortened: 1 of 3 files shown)


PrimeLisp/solution_2/README.md
```diff
@@ -0,0 +1,8 @@
+# Lisp solution by mayerrobert
+
+Based on the solution by mikehw.
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
```
rbergen: I don't think this solution can be marked as faithful. It does not employ a class-like structure to encapsulate the sieve state.

PrimeLisp/solution_2/PrimeSieve.lisp
```diff
@@ -0,0 +1,65 @@
...
+
+  (do () ((>= (get-internal-real-time) end))
+    (run-sieve 1000000)
+    (incf passes))
+
+  (let* ((duration  (/ (- (get-internal-real-time) start) internal-time-units-per-second))
+         (avg (/ duration passes)))
+    (print (list "Passes:" passes "Time:" duration "Avg" avg "Count" (count-primes (run-sieve 1000000))) *error-output*)
+
+    (format t "mayerrobert-cl;~d;~f;1;algorithm=base,faithful=yes,bits=1~%" passes duration)))
```
rbergen: See my other comment concerning faithfulness.

rbergen (commented):
In addition to below comments, I am curious to what you would say is/are the (main) reason(s) for the performance improvement this solution shows?

PrimeLisp/solution_2/README.md
mayerrobert: The contributing guidlines say:

> It uses a class to encapsulate the sieve, or an equivalent feature in your language.

So since all state is contained in the function "run-sieve" I thought that counts as "equivalent feature. But I don't want to fight over this, should I change my branch?

mayerrobert: Re: speedup; almost all time went into the function get-bits. Apparently "aref" is super slow when the array could be any style of array. When I added

      (declare (simple-bit-vector rawbits))

things got a lot faster. I think this changed the access from a function call to inline instructions.

Also the various `fixnum` declarations helped so that add, multiply and compare became just a few instructions instead of a function call. At last it seems like local variables are considerably faster than globals, which in Common Lisp have added behaviour (google "special variables").

And of course `(declaim (inline...)` and `(declaim (optimize...)` helped, too. I'm not really experienced in Common Lisp so take everything I wrote with a grain of salt. It was fun, tough.

rbergen: The point here is that the state is not encapsulated in a transferable way (you can't "pass" the function to another function to work on its contents). For one, this is illustrated by the fact that you execute a separate sieve run to be able to verify the accuracy of the implementation.

Another point concerning faithfulness is that the sieve run function must return the prime (candidate) array, which your solution also doesn't do.

I want to stress that there is no accusation or value statement in the faithful/not faithful determination in the context of this project. It's just one of the 4 characteristics an implementation can have. And there are now quite a few languages in this repo that simply don't provide the facilities that a faithful implementation requires.
And we're not fighting, just exchanging perspectives and arguments. You're still welcome to provide arguments to counter mine. :)

As things stand, I do think the output and README.md should indeed be changed to reflect that this implementation is not faithful.

mayerrobert: Ok, I have changed both the output as well as the README.md. I guess now I'll have to learn CLOS (Common Lisp Object System) to add an object oriented Common Lisp version ;-)

Just FYI: `run-sieve` does return the result array, see line 48 (the last executed statement or more accurately the value of the last "form" of a function is the return value of the function).

rbergen: Thank you, and you are right in correcting me on the return value point.

rbergen (approved):
LGTM

rbergen:
@mayerrobert I'm still wondering what you believe is the main driver for the difference in performance between the existing solution and yours?

mayerrobert:
> 
> 
> @mayerrobert I'm still wondering what you believe is the main driver for the difference in performance between the existing solution and yours?

I don't think there is one single change that makes a huge difference, but there are several small changes that each make a substantial difference especially when combined. Among these are:

    (floor (/ x y))

and

    (floor x y)

compute the same primary result. But `(floor (/ 33 10))` allocates memory for the intermetiate result of `/` which is the rational 33/10, `(floor x y)` doesn't have this intermediate result. `(floor (/ 33 10))` has the primary result 3 and the secondary result 3/10, `(floor 33 10)` has the same primary result 3 but the secondary result 3. Making this change in get-bit and clear-bit made a 3x speedup.

Adding type information which is optional in Common Lisp helps, too. In Common Lisp variables don't have types by default, only values have a type. When the variables inside a function don't have a type then the code must be generic. But when `declare` is used to tell Lisp that only values of a certain type will be assigned to a variable then the generated code can be specific and more efficient. I observed these things by adding `(disassemble 'run-sieve)` to the end of the file. Each `(declare (fixnum... ))` made the assembly shorter and more efficient, and calls to `generic-=` or `generic-*` disappeared, much like C-programmers use Godbolt.

I guess you could put it this way: without added type information Lisp is similar to duck-typing and generated code contains lots of checks, and e.g. to add two variables a generic function is called.. With added type information (that you would add to performance critical functions) Lisp is more like an optimizing C-compiler.

rbergen:
Thanks for your extensive and clear answer! Those are interesting insights in a language I knew nothing about until fairly recently. :)

marghidanu (approved):

marghidanu:
LGTM

PR #287, opened 2021-07-04
PowerShell performance improvements by ~250%

BrooklynGabe:
## Description
Change function calls to advanced functions, then
use pipeline processing.

## Contributing requirements
_Unchanged from previous PowerShell `solution_2` implementation_

### DIFF
--- a/PrimePowerShell/solution_2/PrimePowerShell.ps1
+++ b/PrimePowerShell/solution_2/PrimePowerShell.ps1
@@ -79,8 +79,14 @@ found.
 
 [CmdletBinding()] 
 param(
-    [Parameter(Position = 0)] [string] $SieveSize = 1000000,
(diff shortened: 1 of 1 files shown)


rbergen:
This PR changes a key characteristic of the existing implementation, being its faithfulness. This is because a faithful implementation must encapsulate the full sieve state in a class-like structure. The original solution_2 implementation does this, the changes you propose remove this encapsulation. 

Because of this, I don't think we can merge this as a change to the existing solution. You could change your PR to one that adds a new solution, next to the ones that are now in place. Of course, in that scenario the new solution does need to comply with the [contributing guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/e8243bba5c0dcb1e8b2e56b1ae4c42005c3fac46/CONTRIBUTING.md).

BrooklynGabe:
Thanks for the quick feedback @rbergen

PR #289, opened 2021-07-04
CPP single-threaded implementation

Destroyerrrocket:
[from the first commit]
I've explicitly not used constexpr shennanigans that would go against
the spirit of the competition, or used intrinsics that would pretty much
ruin the fun of the game. I've seen that there's already someone that
has provided an implementation at compile time, so I feel that the power
of C++ already showcased ;D

I also wanted to see how the alignment of the underlying data would
affect the performance. It has been a pretty cool afternoon experiment,
showing that 32-bits was the sweet spot in my machine.

## Description
Hey! When I saw the implementation of the C++ version I saw there was 
a bit of room for improvement, so in order to defend my favorite language,
I decided to mess around a bit with this project.

I originally planned on making a fully compile time implementation, but I 
soon saw that it went against the faithfulness of the race. Since there was 
already one pull request in the works, I decided to just throw my hat at a 
simple runtime one.
 
## Contributing requirements

### DIFF
--- a/PrimeCPP/solution_3/Dockerfile
+++ b/PrimeCPP/solution_3/Dockerfile
@@ -1,0 +1,13 @@
+FROM ubuntu:21.04 AS build
+
+RUN apt-get update -qq \
+    && apt-get install -y clang g++
+
+COPY *.cpp /opt/app/
+WORKDIR /opt/app
+RUN clang++ -Ofast -std=c++20 PrimeCPP.cpp -o PrimeCPP
+
+FROM ubuntu:21.04
+COPY --from=build /opt/app/PrimeCPP /usr/local/bin
+
+ENTRYPOINT [ "PrimeCPP" ]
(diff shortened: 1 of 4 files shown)


Destroyerrrocket:
~~I clicked publish too early, one second please~~
Done!

rbergen:
Would I be right in concluding from the structure of the PrimeCPP.cpp file that you're really seeking to improve the performance of the existing PrimeCPP/solution_1 implementation? In that case I think it would be better to change your PR to one that indeed modifies that implementation. Dave, for one, prefers that if the fundamental characteristics of the implementation (algorithm, faithfulness, bits per prime, ...) stay the same.

Destroyerrrocket:
If you'd prefer for me to modify solution_1, I'd gladly do so. I can move it immediately.

rbergen:
I think it may indeed be best to open a PR that seeks to modify solution_1. We will of course still review your proposal and ask the author of the original implementation for a review, as well. :)

PR #290, opened 2021-07-04
Cpp single threaded improvement to solution 1

Destroyerrrocket:
(Original pr description, see [Original](https://github.com/PlummersSoftwareLLC/Primes/pull/289) for context)
I've explicitly not used constexpr shennanigans that would go against
the spirit of the competition, or used intrinsics that would pretty much
ruin the fun of the game. I've seen that there's already someone that
has provided an implementation at compile time, so I feel that the power
of C++ will be already showcased ;D

I also wanted to see how the alignment of the underlying data would
affect the performance. It has been a pretty cool afternoon experiment,
showing that 32-bits was the sweet spot in my machine.

## Description
Hey! When I saw the implementation of the C++ version I saw there was 
a bit of room for improvement, so in order to defend my favorite language,
I decided to mess around a bit with this project.

I originally planned on making a fully compile time implementation, but I 
soon saw that it went against the faithfulness of the race. Since there was 
already one pull request in the works, I decided to just throw my hat at a 
simple runtime one.
 
## Contributing requirements


Destroyerrrocket:
Once we are done with this one, I have already spotted one potential optimization for the multithreaded solution, so I'll change the second one accordingly to how this one goes :D

rbergen:
I think we'll need to focus this PR (and #295) on changes that objectively improve the performance of the solutions they're targeting, and then only to the extent that they concern the benchmarked (timed) parts of the code. 

The current PRs include changes in compilers, white spacing and the order of code, that feel more like cosmetic/coding style changes than anything else.

Destroyerrrocket:
If you wish so, I can check for whitespace and meaningless order changes, I'd like to indicate that  the compiler change is not at all a preference change, as it's necessary to generate the best (to my knowledge) assembly.

rbergen:
As the risk of sounding pedantic, I will have to insist on limiting changes to those that objectively improve "raw benchmarked performance". That means that qualifying a particularly version of assembly output as "best" really isn't specific enough unless it translates to faster execution. 

In line with this, it would be best if you can quantify the performance improvement that is achieved with the changes you propose in the next iteration of both of your PRs.

Destroyerrrocket:
Sure! :D I'll state what's my experienced improvement with each change! Most of it will translate to the parallel one, where there is only one additional change. 

I'll insist, that the latest gcc version generates an objectively faster assembly, and as such, as it is an improvement to the raw benchmarked performance, I must insist on keeping that change in. it was, if I remember, a 10% difference or so, and I'd prefer not leaving such a difference on the table when the change is that small.

rbergen:
> I'll insist, that the latest gcc version generates an objectively faster assembly, and as such, as it is an improvement to the raw benchmarked performance, I must insist on keeping that change in. it was, if I remember, a 10% difference or so, and I'd prefer not leaving such a difference on the table when the change is that small.

10% would indeed be worth the effort. Would be good if you can back up that percentage with numbers :)

Destroyerrrocket:
You can bet I'll do so! :P You'll hear from me in the next 4 to 10h depending on life circumstances!!! /s

Destroyerrrocket:
[DocumentationOfImprovements.txt](https://github.com/PlummersSoftwareLLC/Primes/files/6832852/DocumentationOfImprovements.txt)
I have compared multiple stages of improvements. Thanks to this, I've observed that halfing the size of the array accually seems to perform slightly worse, so I'll revert that.

Destroyerrrocket:
Diff slimmed down! I'm sure you'll find it a lot easier to review!

Destroyerrrocket:
Expect the parallel one soon, gotta do something IRL

rbergen:
I'll do some performance testing of my own over the weekend. Some of the improvements your report includes make no sense to me. For one, the resultsDictionary is used nowhere in the timed part of the code, so I would expect that any changes to it wouldn't impact benchmarked performance.

Destroyerrrocket:
It does make sense; it's a matter of where it is stored. Previously it was stored in the class, which required destroying and reconstructing the map each time. Meanwhile, with static, it will be initialized once "globally". Given that it is not part of the algorithm, it makes no sense to not make it static :D

rbergen:
That's a fair point.

So, I've ran some test on my machine, and my findings are that:

- pulling the reference map out of the class does improve performance
- clang++ generates _faster_ executables than g++ does, _until the switch to your BitArray class is introduced_.
- with your BitArray class in, performance in clang++ drops with about 40%, while g++ then generates the fastest executable of the set (original, global reference map, BitArray instead of vector<bool>).

In numbers:
- original implementation: 5140 (clang++) vs 5050 (g++)
- global reference map: 5220 (clang++) vs 5120 (g++)
- BitArray instead of vector<bool>: 3210 (clang++) vs 5360 (g++)

As this is one of Dave's original solutions, I will ask him to review your proposed changes, next.

Destroyerrrocket:
All you have reported is indeed true. I'll also add to the fact that in the report I've included a diff where the rotating loop is replaced with the original loop but I still use my custom array implementation, in which clang still performs the best, in case you want to check that out! It performs better than the global reference map but worse than the final implementation.
The only thing that kinda seems odd is the small amount of improvement you've reported, 5% is quite smaller than my 33%

rbergen:
@Destroyerrrocket, I've asked Dave for feedback on this PR and #295, and a question he asked is why you chose not to overload operator[] and use (a) member function(s) instead?

Destroyerrrocket:
Given that the behavior of operator[] is (usually) to provide access to an element of a container, I wanted to prevent the misconception that it would do that. My implementation intentionally returns the value, as that's more performant than what vector<bool> does to keep the behavior of []. As such expressions like Bits[n] = true; would be broken. Of course, after the great idea from rust of using a rotating mask, this distinction is a lot more irrellevant, as there is no longer a set() function, and all the accesses to the containers would not care if it's a reference or a value. Yet, given that the behavior of operator[] would not be quite correct (in the sence of what I consider a good implementation of operator[]) I did not consider changing it. If you want, I'm completely willing to change the get function for an operator[], the performance should be the same, and that should remove a few lines from the diff.

rbergen:
I don't have a strong opinion about it either way, but I'll quote what Dave wrote in an email on the topic:

> Stylistically I think I would have preferred he overload operator[] rather than GetBit() as it would have kept the rest the same, but maybe he had a reason?

As far as I'm concerned, that means it's up to you to decide.

Destroyerrrocket:
Well, if that's the case, I'm going to go with leaving it as it is. It'd be better to prevent confusions

rbergen:
Could you please sync your branch with this repo? It should allow the CI run to complete due to some merged PRs that fix some issues we've run into recently.

Destroyerrrocket:
I'm on my phone away from computers, let me see if I'm able to do it from github

PR #291, opened 2021-07-04
Octave: vectorize inner loop (500x faster)

tjol:
## Description
In Octave/Matlab/Numpy it is usual to write simple loops in vectorized form because this allows the interpreter to optimize many calculations more effectively.

Making this simple change to the loop marking off the non-primes makes the code about 500x faster on my machine.

## Contributing requirements


rbergen:
Good stuff, thanks!

PR #292, opened 2021-07-04
D Solution 2

BradleyChatha:
## Description
A slightly more idiomatic solution for D. Faithful, Bits 1. Base algorithm.

Doesn't use really any fancy tricks other than standard bit-wise stuff to set, unset, and get certain bits.

It uses the popcnt instruction if supported, otherwise it uses a software fallback.

No explicit vectorisation or crazy mathematics.

Crossing my fingers the DockerFile is correct, since my environment is slightly borked right now.

Contains a parallel version as well.

## Contributing requirements

### DIFF
--- a/PrimeD/solution_2/DockerFile
+++ b/PrimeD/solution_2/DockerFile
@@ -1,7 +1,0 @@
-FROM Alpine:3.31
-RUN apk update
-RUN apk add curl
-RUN curl -fsS https://dlang.org/install.sh | bash -s ldc
-WORKDIR /app
-ADD . /app
-ENTRYPOINT [ "dub build", "-b release", "--compiler ldc2" ]
--- a/PrimeD/solution_2/DockerFile
+++ b/PrimeD/solution_2/DockerFile
@@ -1,4 +1,6 @@
 FROM Alpine:3.31
+RUN apk update
+RUN apk add curl
 RUN curl -fsS https://dlang.org/install.sh | bash -s ldc
 WORKDIR /app
 ADD . /app
--- a/PrimeD/solution_2/DockerFile
+++ b/PrimeD/solution_2/DockerFile
@@ -1,0 +1,5 @@
+FROM Alpine:3.31
+RUN curl -fsS https://dlang.org/install.sh | bash -s ldc
+WORKDIR /app
+ADD . /app
+ENTRYPOINT [ "dub build", "-b release", "--compiler ldc2" ]
--- a/PrimeD/solution_2/source/app.d
+++ b/PrimeD/solution_2/source/app.d
@@ -27,6 +27,8 @@ final class Sieve(uint SieveSize)
     private
     {
         // Because SieveSize is a templated value, we know what it is at compile-time, so can stack allocate here.
+        // Most D functions can be ran at compile-time. This is called Compile Time Function Execution (CTFE).
+        // `alignTo` is being used in CTFE here.
         ubyte[alignTo!8(SieveSize) / 8] _bits = ubyte.max;
     }
 
--- a/PrimeD/solution_2/source/app.d
+++ b/PrimeD/solution_2/source/app.d
@@ -104,7 +104,7 @@ final class Sieve(uint _sieveSize)
         }
     }
 
-    void printResults(string tag, string attribs, bool showResults, Duration duration, uint passes) @trusted // `stderr` is unsafe apparently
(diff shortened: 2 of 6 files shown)


PrimeD/solution_2/source/app.d
```diff
@@ -0,0 +1,270 @@
...
+        //
+        // `$` within a slice means "length", so 0..length (exclusive).
+        //
+        // So essentially all we're doing is making a lazily evaluated, allocationless pipeline:
+        //  Take all the bytes we have from _bits.
+        //  Evaluate popcnt(n) where `n` is the next byte.
+        //  Find the sum of all the `popcnt(n)` evaluations.
+        return this._bits[0..$]
+                    .map!(num => popcnt(num & 0b1010_1010)) // We only care about odd numbers, so we mask out any bits that are even.
+                    .fold!((a, b) => a + b)(0);
```
pbackus: Could just use `std.algorithm.sum` here:
```suggestion
                    .sum;
```

BradleyChatha:
Q_Q I kept doing release builds, completely forgetting I needed to check that the assert in printResults needed to work.

BradleyChatha:
I don't know why, but it also always makes me happy when I see LDC inline stuff.

From:

```d
    private uint countPrimes() nothrow pure
    {
        import std.algorithm : map, sum;
        import std.range     : iota;
        return iota(0, SieveSize)
                .map!(num => this.getBit(num) ? 1 : 0) // Ternary operator is just to make it more clear, not actually needed.
                .sum;
    }
```

Into:

```asm
pure nothrow @safe uint example.Sieve!(1000000u).Sieve.countPrimes():
        xor     eax, eax
        xor     ecx, ecx
.LBB23_1:
        mov     edx, ecx
        shr     rdx, 3
        movzx   esi, byte ptr [rdi + rdx + 16]
        mov     edx, ecx
        and     dl, 6
        or      dl, 1
        movzx   edx, dl
        bt      esi, edx
        adc     eax, 0
        add     ecx, 2
        cmp     ecx, 1000000
        jne     .LBB23_1
        ret
```

PrimeD/solution_2/DockerFile
```diff
@@ -0,0 +1,7 @@
+FROM Alpine:3.31
```
rbergen: I'm guessing you meant to use alpine:3.13 as a base image.

PrimeD/solution_2/DockerFile
```diff
@@ -0,0 +1,7 @@
...
+RUN apk update
+RUN apk add curl
+RUN curl -fsS https://dlang.org/install.sh | bash -s ldc
+WORKDIR /app
+ADD . /app
+ENTRYPOINT [ "dub build", "-b release", "--compiler ldc2" ]
```
rbergen: This doesn't look right.

rbergen (commented):
I aborted my review because the Dockerfile doesn't work. I made two comments about it below, but if I'm honest then I think it would be best if you commit a new Dockerfile once you yourself have confirmed it to work.

BradleyChatha:
As a professional fool: all I had to do was reinstall Docker to get it to work. 😞

PrimeD/solution_2/source/app.d
```diff
@@ -0,0 +1,280 @@
...
+        scope sieve = new Sieve!PRIME_COUNT;
+        sieve.runSieve();
+        passes++;
+    }
+
+    const elapsedTime = timer.peek();
+    
+    auto s = new Sieve!PRIME_COUNT;
+    s.runSieve();
+    s.printResults("BradleyChatha", "algorithm=base,faithful=yes,bits=1,parallel=no", 1, false, elapsedTime, passes);
```
rbergen: Please remove the "parallel=no" tag, Parallelism is deduced from the thread count (1 == no).

PrimeD/solution_2/source/app.d
```diff
@@ -0,0 +1,280 @@
...
+        }
+    }
+
+    const elapsedTime = timer.peek();
+    
+    auto s = new Sieve!PRIME_COUNT;
+    s.runSieve();
+    s.printResults(
+        "BradleyChatha-Multi", 
+        "algorithm=base,faithful=yes,bits=1,parallel=yes",
```
rbergen: Please remove the "parallel=yes" tag, Parallelism is deduced from the thread count (>1 == yes).

rbergen (approved):
LGTM

BradleyChatha:
Sorry for the extra commit, but that'll be the last one now.

marghidanu (approved):

PR #293, opened 2021-07-04
PowerShell pipeline processing, 50% faster

BrooklynGabe:
## Description
Uses pipeline processing to be approx 50% faster. Sieve state is also moved from PSObject to Sieve class.

This _should_ account for the faithfulness issue in the closed PR https://github.com/PlummersSoftwareLLC/Primes/pull/287

## Contributing requirements
_Same as the previous PowerShell solution_2

* [ X] I read the contribution guidelines in CONTRIBUTING.md.
* [ X] I placed my solution in the correct solution folder.
* [ X] I added a README.md with the right badge(s).
* [ X] I added a Dockerfile that builds and runs my solution.
* [ X] I selected `drag-race` as the target branch.
* [ X] All code herein is licensed compatible with BSD-3.

### DIFF
--- a/PrimePowerShell/solution_2/PrimePowerShell.ps1
+++ b/PrimePowerShell/solution_2/PrimePowerShell.ps1
@@ -79,42 +79,55 @@ found.
 
 [CmdletBinding()] 
 param(
-    [Parameter(Position = 0)] [string] $SieveSize = 1000000,
-    [Parameter(Position = 1)] [int] $MinimumRuntime = 5,
+    [Parameter(Mandatory = $false,
+        Position = 0)]
+    [ValidateSet(10, 100, 1000, 10000, 100000, 1000000, 10000000, 100000000)]
+    [int]$SieveSize = 1000000,
+    
+    [Parameter(Mandatory = $false,
+        Position = 1)]
+    [int]$MinimumRuntime = 5,
+
     [switch] $ResetState,
+
     [switch] $ShowResults
 )
 
 
 # Known results for validating the output.
 $KnowResults = @{
-    10 = 1;
-    100 = 25;
-    1000 = 168;
-    10000 = 1229;
-    100000 = 9592;
-    1000000 = 78498;
-    10000000 = 664579;
+    10        = 1;
+    100       = 25;
+    1000      = 168;
+    10000     = 1229;
+    100000    = 9592;
+    1000000   = 78498;
+    10000000  = 664579;
     100000000 = 5761455
 }
 
+class Sieve {
+    Sieve([int]$Size) {
+        $this.SieveSize = $Size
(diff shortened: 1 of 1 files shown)


rbergen:
As this is a proposed change to a solution written by another author, I'm tagging @crowbar27 and allow some time for them to review and respond if they want to do so.

rbergen (approved):
LGTM

crowbar27:
This looks OK for me, too.

As classes require PS5+, I just wanted to take the opportunity to bring up a "philosophical" question of what a Powershell implementation actually is in the context of this project. I have been thinking quite a while about pasting Dave's C# code into a string and just invoking it, which would be a viable solution (an probably a faster one than acutally using PS commands) - but would it still be Powershell?

@BrooklynGabe I am just curious - does the performance improvement come from the pipeline or from the class?

BrooklynGabe:
@crowbar27 thanks for the question.

The performance improvement came from the pipeline processing when looping over the repeated calls to the `Invoke-Sieve` function. I had a previous PR (https://github.com/PlummersSoftwareLLC/Primes/pull/287) that did not use the `class` construct in PowerShell, which was actually faster than using the `class` via pipeline. Since passing the values without the `class` was not a faithful representation of the sieve in a class-like structure, I closed that PR rather than updating it again. 

I am not sure if there needs to be a separation of PowerShell solutions based on targeted version. Version 7 is the cross-platform that Microsoft is leaning towards currently. Version 5.1 is the oldest version still listed in their support documentation, but is also what is shipped with their most recent Windows Server (2019).

rbergen:
> I have been thinking quite a while about pasting Dave's C# code into a string and just invoking it, which would be a viable solution (an probably a faster one than acutally using PS commands) - but would it still be Powershell?

It would not be a _faithful_ PowerShell implementation. One of the requirements for a faithful implementation is that it doesn't use external dependencies for the actual sieve calculation, and if the sieve calculation requires a completely different language then that rule is clearly violated.

marghidanu (approved):

CyberSecDef:
Running both solution 1 and solution 2 with the same paramters, solution 2's final results include 'primes' that aren't prime.  This might explain part of the performance boost.

*** note the 'primes' hash includes 4 and several other non-primes.

Solution 1:
crowbar27_ps1;3321;5.00074;1;algorithm=base,faithful=yes,bits=1
SieveSize       : 100
Primes          : {2, 3, 5, 7...}
Passes          : 3321
AverageDuration : 0.00150579268292683
Valid           : True
Duration        : 00:00:05.0007375

Solution 2:
crowbar27_ps2;23721;5.00015;1;algorithm=base,faithful=yes,bits=1
SieveSize       : 100
Primes          : {2, 3, 4, 5...}
Passes          : 23721
AverageDuration : 0.000210789987774546
Valid           : True
Duration        : 00:00:05.0001493

PR #295, opened 2021-07-05
CPP solution_2 improvements multithreaded

Destroyerrrocket:
## Description
As I stated in my previous PR, I saw a few points of improvement.
Given that I felt inspired, I decided to finish it of today. I had a great time!

I'd like to note that there are a lot of similarities with my single 
threaded implementation.

## Contributing requirements


Destroyerrrocket:
The optimizations proposed on the single threaded improvements also apply here, and additionally keeping the threads working without destruction yields an improvement off:

Base with all optimizations except for better thread management:
Computing primes to 1000000 on 24 threads for 5 seconds.
Passes: 117552, Threads: 24, Time: 5.00001, Average: 4.25344e-05, Limit: 1000000, Counts: 78498/78498, Valid : Pass

davepl_par;117552;5.00001;24;algorithm=base,faithful=yes,bits=1

With thread management: [Speedup: 1.58]
Computing primes to 1000000 on 24 threads for 5 seconds.
Passes: 185267, Threads: 24, Time: 5.00074, Average: 2.69921e-05, Limit: 1000000, Counts: 78498/78498, Valid : Pass

davepl_par;185267;5.00074;24;algorithm=base,faithful=yes,bits=1

Destroyerrrocket:
Changes applied, now it should be a lot easier to review!

PR #299, opened 2021-07-05
Improve Ada speed

xpetit:
## Description

I mostly modified the Dockerfile to improve speed (compiler version & options)

## Contributing requirements


rbergen:
Tagging @BoopBeepBoopBeep and allowing time so they can review and respond to this PR on their original solution, if they want to do so.

BoopBeepBoopBeep:
Looks good. 👍
While you are at it, @MichaelAllenHardeman did double the speed.
That solution could be boiled down by adding a single line of code.
Could you add the line "Num := Factor" between the lines 42 and 43 in main.adb?

xpetit:
Thank you @BoopBeepBoopBeep , I added the commit 4549f2b3cfb5b1b08e12dda326864f989034ee26
Before:
```
Passes: 1693, Time: 5.002306400, Avg:  0.002954699, Limit : 1000000, Count1 : 78498, Count2: 78498, Valid :TRUE
```
After:
```
Passes: 3521, Time: 5.000292400, Avg:  0.001420134, Limit : 1000000, Count1 : 78498, Count2: 78498, Valid :TRUE
```
(On a busy MacBook (Retina, 12-inch, 2017))

michael-hardeman:
Looks good to me 👌

My branch is probably very out of date at this point but there is some good stuff available to copy pasta. 
https://github.com/PlummersSoftwareLLC/Primes/pull/127/commits

There are two important points I noticed about the implementation while i was working on it.

`type Bit_List`  is actually incredibly space inefficient. It allocates 1 byte for each boolean in the boolean array. I found I couldn't do more than calculations up to 10^7 because it allocates more space for the array than the default GNAT stack size. Moving the allocation onto the heap (new keyword) and freeing memory at the end (Ada.Finalization.Unchecked_Deallocation) allowed me to do larger numbers and the performance hit was negligible. I tried implementing it as a packed bit array, but sadly every try I made was far slower and I gave up on that idea.

Good luck to you 👍

onox:
I managed to increase the speed a bit from 5200 to about 5400/5500 by:

1. Removing function `Create_Bit_List`
2. Add:
```ada
True_Bits         : constant Bit_List (0 .. Sieve_Size) := (others => True);
Bits              : Bit_List := True_Bits;
```
3. In the while loop at the bottom replace `Bits := Create_Bit_List (Sieve_Size);` with `Bits := True_Bits;`

michael-hardeman:
Something else i did on my branch was move the calculation of Q out of the loop to a constant. This didn't seem to impact the preformance though. I suspect the compiler already preformed that optimization.

Glacia:
@xpetit Not sure why you deleted gpr file, if you want to add compiler switches you can do it like this:

```
`   package Compiler is
      for Switches ("ada") use ("-flto", "-O3", "-funroll-loops", "-gnatn", "-gnatX");
   end Compiler;`
```

thindil:
@Glacia I guess because docker image changed from Ubuntu to Alpine as base, and `gprbuild` isn't installed. Also, a few rants 😁 :

1. If you use _linking time optimization_, you don't need to add `-gnatn`. It is included in `-flto`.
2. If you want to add _linking time optimization_, you have to add flags also to linker, not just compiler.
3. Funny fact: sometimes you can get better result with using `-O2` instead of `-O3` when doing _linking time optimization_.

Glacia:
@thindil it's just an example, i just copied what GNAT Studio generated.

>     1. If you use _linking time optimization_, you don't need to add `-gnatn`. It is included in `-flto`.

Not sure about that one, inlinening can happen in front end and in back end, But whatever, LTO sucks anyway 🤷‍♂️

>     2\. f you want to add _linking time optimization_, you have to add flags also to linker, not just compiler.

>     3\. Funny fact: sometimes you can get better result with using `-O2` instead of `-O3` when doing _linking time optimization_.

Yeah, i know.

xpetit:
@onox unfortunately I can't replicate the performance boost on my MacBook 2017, so I tried on a Linux Server with an AMD Ryzen 7 3800X CPU and `scaling_governor` set to `performance` and I consistently get ~12250 passes with or without your change (using a constant instead of the return value of a function to set the variable).

@Glacia Yes sorry I removed the .gpr file, it was cleaner, but as @thindil pointed out, `gprbuild` wasn't available on the much smaller Alpine Docker image. I tried `-flto`, but did not observe a performance boost.

thindil:
I've done something similar to @onox with Bit_Lists:

1. Declaration of `Bits` replaced by `   Bits              : Bit_List(0 .. Sieve_Size);`
2. Removed function `Create_Bit_List`
3. In loop, I use `Bits := (others => True);`

It gives me boost around 10%, but YMMV. Anyway, good to check. 😄

xpetit:
> I've done something similar to @onox with Bit_Lists:
> 
> 1. Declaration of `Bits` replaced by `   Bits              : Bit_List(0 .. Sieve_Size);`
> 2. Removed function `Create_Bit_List`
> 3. In loop, I use `Bits := (others => True);`
> 
> It gives me boost around 10%, but YMMV. Anyway, good to check. 😄

```
Passes: 15571, Time: 5.000268869, Avg:  0.000321127, Limit : 1000000, Count1 : 78498, Count2: 78498, Valid :TRUE
```
(instead of 12250)

```
   🌔
🚀
```

Now faster than the first C++ solution (the second one is multi-threaded)

michael-hardeman (approved):

rbergen:
@xpetit Can I consider this PR stable and ready for review?

xpetit:
@rbergen Yes

michael-hardeman:
@BoopBeepBoopBeep @xpetit 

This might be a dead topic, but I just found this page:
https://www.adaic.org/resources/add_content/docs/95style/html/sec_10/10-5-7.html

And from my testing on the AdaCore Linux version of GNAT declaring the boolean array like this:

```
type Flag_State is array (1 .. 64) of Boolean;
pragma Pack (Flag_State);
...
Put (Natural'Image (Flag_State'Size)); -- prints 64
```

Does in fact create a true bit array of 64 bits long.

xpetit:
Indeed, Ada is so nice. You can also write:
```Ada
type Flag_State is array (1 .. 64) of Boolean with Pack;
```

PR #300, opened 2021-07-05
NodeJS solution with 32bit array and optimizations

rogiervandam:
## Description
NodeJS solution with 32bit array and optimizations

## Contributing requirements


rbergen:
The key characteristics of this implementation (algorithm, faithfulness, bits per prime) are the same as those of solution_1, Could you please explain:

- what are the optimizations in this implementation vs solution_1, from a performance perspective?
- what is the improvement in performance numbers is that these optimizations achieve?
- why you chose to open a PR to create a second solution, instead of applying these optimizations to the solution that is already in place?

rogiervandam:
3x to 5x performance gain due to using bit shifting and masking, like the PrimeC solutions.
Als some code restructuring helps in performance, while remaining faithful.
Feels like the restructure is educational in what performance is possible with nodejs (close to c)

rbergen:
@fvbakel It does seem that this implementation is significantly faster than the current PromeNodeJS/solution_1. With the credit that @rogiervandam has given you as the author of one of the solutions his is based on, would you be ok with replacing the current solution_1 with the one in this PR?

fvbakel:
@rogiervandam @rbergen I am perfectly fine with overwriting my solution_1. Great to see this performance improvement!

rbergen:
@fvbakel Thanks for responding so quickly (and supportive, I might add :) )

@rogiervandam Could you modify your PR to replace the existing PrimeNodeJS/solution_1 directory's contents with what it now aims to place in PrimeNodeJS/solution_2?

rogiervandam:
I changed the pull request to make it solution1

rbergen (approved):
LGTM

marghidanu (approved):

PR #302, opened 2021-07-05
Primes lazarus/solution 1

RexAlfes:
## Contributing requirements


marghidanu:
@RexAlfes  There seems to be a `.res` file in your commit, what does that file contain?

rbergen:
Contrary to what is indicated in the PR description, this solution does not conform to the contributing guidelines. At least It doesn't include a Dockerfile to build and run the solution, and it includes binary files.

Furthermore:, I believe the main point of this solution is not to add a language (which is Pascal), but add a different execution environment for an existing implementation. In the context of this project/repo, that is in itself not a reason to add a new solution.

If you feel a different execution environment would benefit the performance of the existing Pascal solution, then please compose a PR to modify the files in the existing Pascal/solution_1 directory. That PR would need to aim to modify that solution in such a way that:

- its performance is objectively improved compared to what it is now
- it conforms to the contributing guidelines. Amongst others, that means that a Dockerfile needs to be provided that builds and runs the solution in the execution environment of your choosing.

As it stands, we can't consider this solution for addition.

rbergen:
Closing this. A new PR can be opened that conforms to the guidelines as indicated in the comments.

PR #303, opened 2021-07-05
Remove dependency on Node.js for benchmark

HalloTschuess:
## Description
I forgot to install Node.js in WSL and got greeted with a nice error after almost an hour of building and running all programs.
Therefore I thought: "We have Docker at our disposal, might as well use the node docker image."

Tested on WSL and macOS Catalina

### DIFF
--- a/BENCHMARK.md
+++ b/BENCHMARK.md
@@ -33,13 +33,11 @@ Take the following steps to run all benchmarks:
 
 1. Make sure you have installed the [required software](#linux-installation-and-prerequisites)
 2. Open a terminal.
-3. Make sure Node.js is added to the path. `export PATH=/path/to/node/bin:$PATH`.
-   For example: `export PATH=~/node-v14.17.0-linux-x64/bin:$PATH`
-4. `git clone https://github.com/PlummersSoftwareLLC/Primes.git`
-5. `cd Primes`
-6. `make`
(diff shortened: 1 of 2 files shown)


rbergen:
That's not going to work, for a reason that I understand is not immediately obvious.

We're working towards a situation where benchmark results can be uploaded to an aggregation web service and then viewed with a web interface (there are some other repositories within the PlummersSoftwareLLC organisation related to this work in progress). For that, the benchmark tool also collects a number of anonymous properties of the execution environment, to help compare the benchmark results on different systems (or maybe more accurately, help explain the differences between them). These properties are not shown in the textual (console) output, but are included in the JSON file the report tool is also able to generate. 
Collecting these system properties doesn't work from within a Docker container because, by design, it shields key properties of the environment that Docker runs in.

HalloTschuess:
Oh well, that makes sense. Maybe should have looked at what the code actually does, my bad. Thanks for clarifying though.

marghidanu:
@rbergen, we probably need to merge the branch that performs the dependencies checks before the benchmark suite runs. It seems that this is a problem at the moment.

rbergen:
@marghidanu I'm fine with that, if we also add a link to the benchmark documentation in the messages. We didn't write it for nothing. :)

PR #305, opened 2021-07-05
Fortran: faster data types

tjol:
## Description

I've fiddled with the data types in the Fortran solution to achieve a ~+35% speed up on my machine:

 * Addressing the bit array as an int8 array is faster than addressing it as an int64 array (this is the main thing)
 * The implementation was already limited to 32-bit integers (like many others), making this a bit more consistent speeds it up another few percent.

## Contributing requirements

### DIFF
--- a/PrimeFortran/solution_1/PrimesFortran.f03
+++ b/PrimeFortran/solution_1/PrimesFortran.f03
@@ -5,12 +5,15 @@ program primes_fortran
   ! All bits are set to 1 for an integer value of -1 as in Fortran there are only signed numbers
   ! and the two's complement is used
   
-  integer(int64) , parameter :: all_true = int((-1), kind=int64)
+
   integer(int64), parameter:: num_possible_upper_limits = 8_int64
+  integer(int8) , parameter :: all_true = -1
+  integer, parameter:: bitfield_unit = 8
+  integer(kind=int8), dimension (:), allocatable :: array_of_bits ! Mapping for Bit-array
   integer(int64):: time_start, time_now, count_rate, count_max
-  integer(int64):: sieve_size
+  integer(int32):: sieve_size
   integer(int32):: passes
-  integer(kind=int64), dimension (:), allocatable :: array_of_bits ! Mapping for Bit-array
+  
 
   character(len=255), dimension(1:3) :: command_arguments
   
@@ -38,7 +41,7 @@ program primes_fortran
                               664579_int32,&        ! 10000000
                               5761455_int32/)       ! 100000000
 
-  sieve_size = int(1000000, kind = int64)
+  sieve_size = 1000000
       
   call allocate_bitfield(sieve_size)
 
@@ -51,7 +54,7 @@ program primes_fortran
      array_of_bits = all_true   ! sets each element of the array to the same value
      
      call run_sieve
-     passes = passes +1_int64
+     passes = passes + 1
      call system_clock(time_now, count_rate, count_max)
      if ((1.0_real64*(time_now-time_start)) / count_rate .ge. 5.0_real64  ) then        
         exit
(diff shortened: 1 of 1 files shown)


rbergen:
Tagging @johandweber and allowing time so they can review and respond to this PR on their original solution, if they want to do so.

johandweber:
I have tested the program and found it your modifications lead to a considerable speedup especially for 
newer versions of gfortran (I think, Ubuntu 18.04 still uses Version 7).

For my tests (Ryzen X 3700 with up to 3600 MHz, Ubuntu 20.04) I found that making the program consistent
with 64 bit led to another performance advantage of 20%. The disadvantage is that it will likely no longer 
work on 32-bit systems.

Additionally I have added a version of the code, where I allocate and deallocate the bitarray for each loop, for
a more similar approach to the a faithful one (it is still not object oriented so it remains unfaithful).
As expected the code gets slower, although the speeddown is less than 3%

I also have looked at your OO-version and found a speed increase consistent with your measurements.
It may be  too late for the drag race videos, but I think the OO-version should become the reference
implementation for Fortran.



Please find attached a table with my benchmarks:#
                             gfortran-7   gfortran-8   gfortran-9   gfortran-10
new_32                        5699           9085          9145             9430
new_64                        8711         11090        11186           12240
new_64_allocate          8570        11047        11035            11935
current                         4978          4877           4890              4877
solution_2                  17644        17731        17926             17617

rbergen:
> I have tested the program and found it your modifications lead to a considerable speedup especially for newer versions of gfortran (I think, Ubuntu 18.04 still uses Version 7).

The preference for Ubuntu 18.04 as written in the guidelines is not written in stone. In fact, I think it's time to rephrase it to "an LTS version of Ubuntu". If there is a meaningful speed increase to be gained from a bump to Ubuntu 20.04, then I suggest the Dockerfile is updated accordingly.

> I found that making the program consistent with 64 bit led to another performance advantage of 20%. The disadvantage is that it will likely no longer work on 32-bit systems.

The benchmark won't, anyway. Only x86-64/amd64 and aarch64/arm64 are supported in the context of the drag race. So again, if a switch to 64 bit increases performance on 64 bit systems, then I think making that change is a good idea.

johandweber:
Thanks for offering the possibility to update to a newer Version of Ubuntu.

I want to note that Thomas' object-oriented version  solution_2 appears to be much less compiler-dependent than my original solution_1 and the modifications to  solution_2 by both Thomas and me.  Every build of solution_2 is faster than any build of solution_1.

If you want, i can create a pull request based on Thomas' solution but (hopefully) consistent in 64, bit within 24-36 till tomorrow.

Would you prefer the fastest version, where the bit-array is allocated only once, or the slightly slower (by < 3%), but
more faithful version, where the memory allocation and deallocation is performed for each pass? The latter version would,
however, still not fully fulfill the "faithful"-criteria, as in the chosen procedural approach no sieve-object is created.

tjol:
> For my tests (Ryzen X 3700 with up to 3600 MHz, Ubuntu 20.04) I found that making the program consistent
with 64 bit led to another performance advantage of 20%. The disadvantage is that it will likely no longer
work on 32-bit systems.

Interesting. I tried going to 64 bit the other day and saw a (slight!) performance decrease. I imagine I made a mistake, but I suppose it's possible this is just a difference between Zen+ and Zen 2.

Do you want to open a new pull request with your changes? Then I'd close this one.

rbergen:
> Would you prefer the fastest version, where the bit-array is allocated only once, or the slightly slower (by < 3%), but
> more faithful version, where the memory allocation and deallocation is performed for each pass?

Thank you for asking this question. For me, definitely the more faithful version. :)

johandweber:
I will check my code again and hope to create a pull request till tomorrow.

Which build environment have you used.

As you told, I used a Zen2 CPU (I do not really think that this makes such a difference) and
gfortran-n -Ofast -o output_file input_file.f08.

I did not use the docker environment for my tests but ran the programs directly.

tjol:
I used gfortran-10 on OpenSUSE Tumbleweed. I might have another look at my patch this evening.

Overall, I think it'd be good to leave both solutions in:

* solution_1 witih small optimisations (faster choice of types, etc) and reallocation
* solution_2 (possibly switched to 64 bit if that turns out to be faster?)

tjol:
Correction: I was using mostly using gfortran-11, not -10. But with either compiler, it does look like a 64 bit version is significantly faster on my machine than the 32 bit. Don't know where I got the other idea.

johandweber:
@tjol : Is it O.K., if I create a pull request for solution_1 based on your pull request for soultion_1, but switched to 64 bit and 
using a more faithful memory management?

tjol:
Absolutely, please do!

johandweber:
@tjol O.K., my README.md will contain a statement that we are joint owners of the copyright and willing to license the code under the BSD-3 license. Do you agree?

tjol:
@johandweber Sounds good to me.

tjol:
Closing in favour of #327

PR #306, opened 2021-07-05
Delphi, Multi-Threaded, Based on JakeTapper's Delphi code submission

nmadani:
# Delphi solution by Navid Madani (modifies solution by Jake Tapper)

Parallelism

This application takes advantage of Delphi's support for parallel processing and runs the algorithm that was 
implemented in Delphi by Jake Tapper about 3-4 times faster than the original.

## Run Instructions 

```
Usage:
       PrmDelphi [runtime in seconds] [thread lag in milliseconds]
           Runtime defaults to 10 seconds and
           thread lag defaults to 50 milliseconds.
           The thread lag is subtracted from the runtime to allow
           lagging threads to finish within the allocated time.
```

## Output

```
Starting 10 second run with 50 millisecond thread lag ...

Passes: 14432, Time: 10.052 sec, Avg: 0.6965 ms, Limit: 1000000, Count: 78498, Valid: Yes

Press enter to close...
```

### DIFF
--- a/PrimeDelphi/solution_1/PrmDelphi.dpr
+++ b/PrimeDelphi/solution_1/PrmDelphi.dpr
@@ -1,0 +1,323 @@
+///  Modified Jake Tapper's submission:
+///  Removed the validation code into a separate static class.
+///  Inherited the class from TThread for multi-threaded execution called
+///  from a parallel loop in Main procedure.
+///  Call with "PrmHelp h" for usage instructions
+program PrmDelphi;
+
+{$APPTYPE CONSOLE}
+
+uses
+  System.SysUtils,
+  Classes,
+  System.Generics.Collections,
+  System.Timespan,
+  Math,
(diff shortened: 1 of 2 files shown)


joshdaniel87 (approved):

sglienke:
That code just creates a ginormous memory leak and eventually crashes.
The TParallel.For only creates new threads (TPrimeSieve inherits from TThread!)

nmadani:
> 
> 
> That code just creates a ginormous memory leak and eventually crashes.
> The TParallel.For only creates new threads (TPrimeSieve inherits from TThread!)

Oops - just needs a nil assignment to the dynamic byte array in the destructor to fix.

```
destructor TPrimeSieve.Destroy;
begin
  AtomicIncrement(passes);
  FBitArray := nil;
  inherited;
end;
```

PR #307, opened 2021-07-05
Improved speed ~4-5x for PHP

xeridea:
No longer uses class methods for basic operations.
Leaned the loop logic.
Uses factor * factor.
Uses SplFixedArray rather than normal array, they use less memory and are faster.
Enable PHP JIT.

## Description
@DennisdeBest Would you like to comment?

## Contributing requirements

### DIFF
--- a/PrimePHP/solution_1/Dockerfile
+++ b/PrimePHP/solution_1/Dockerfile
@@ -5,4 +5,4 @@ WORKDIR /opt/app
 COPY *.php .
 COPY opcache.ini /usr/local/etc/php/conf.d/
 
-ENTRYPOINT [ "php", "-d opcache.enable=1", "-d opcache.enable_cli=1", "-d opcache.jit=1255","-d opcache.jit_buffer_size=100M", "PrimePHP.php" ]
+ENTRYPOINT [ "php", "PrimePHP.php" ]
--- a/PrimePHP/solution_1/Dockerfile
+++ b/PrimePHP/solution_1/Dockerfile
@@ -4,4 +4,4 @@ WORKDIR /opt/app
 
 COPY *.php .
 
-ENTRYPOINT [ "php", "PrimePHP.php" ]
+ENTRYPOINT [ "php", "-dopcache.enable=1", "-dopcache.enable_cli=1", "-dopcache.jit=1255","-dopcache.jit_buffer_size=100M", "PrimePHP.php" ]
--- a/PrimePHP/solution_1/Dockerfile
+++ b/PrimePHP/solution_1/Dockerfile
@@ -3,5 +3,6 @@ FROM php:8.0-cli-alpine3.13
 WORKDIR /opt/app
 
 COPY *.php .
+COPY opcache.ini /usr/local/etc/php/conf.d/
 
-ENTRYPOINT [ "php", "-dopcache.enable=1", "-dopcache.enable_cli=1", "-dopcache.jit=1255","-dopcache.jit_buffer_size=100M", "PrimePHP.php" ]
+ENTRYPOINT [ "php", "-d opcache.enable=1", "-d opcache.enable_cli=1", "-d opcache.jit=1255","-d opcache.jit_buffer_size=100M", "PrimePHP.php" ]
--- a/PrimePHP/solution_1/PrimePHP.php
+++ b/PrimePHP/solution_1/PrimePHP.php
@@ -48,9 +48,10 @@ class PrimeSieve
                 }
             }
 
-            $ft2 = $factor * 2;
-            for ($i = $factor * $factor; $i <= $sieveSize; $i += $ft2) {
-                $rb[$i / 2] = 0;
+            $ft2 = $factor * 1;
+            $start = ($factor * $factor) / 2;
+            for ($i = $start; $i <= $this->rawBitsSize; $i += $ft2) {
+                $rb[$i] = 0;
             }
 
             $factor += 2;
--- a/PrimePHP/solution_1/PrimePHP.php
+++ b/PrimePHP/solution_1/PrimePHP.php
@@ -12,6 +12,7 @@ class PrimeSieve
     private array $rawbits;
 
     private int $sieveSize;
+    private int $rawBitsSize;
 
     public static array $primeCounts = [
         10 => 4,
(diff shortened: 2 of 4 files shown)


rbergen:
Tagging @DennisdeBest and allowing time for them to review and respond to this PR on their original implementation, if they want to do so.

DennisdeBest:
Hi ! Thanks for the optimization. Running this on my computer inside the docker container it goes from 177 passes to 392 !

Running it with or without JIT does not really make in difference for me.

To try it out you can change the CMD in the Dockerfile to `CMD [ "php", "-dopcache.enable=1",  "-dopcache.jit_buffer_size=100M", "PrimePHP.php" ]`

@rbergen I think this is a great improvement and it can definitly be merged ;)

rbergen:
@DennisdeBest Thanks for your positive feedback on this PR, and responding so quickly.

@xeridea I think there's nothing wrong with updating the README.md to mention the fact that you've optimized this solution, if you want to. Of course, the fact you have will also forever be document in GitHub's commit history. :)

I will hold back on my full review of this PR until you have updated it, if you want to do so.

xeridea:
@DennisdeBest Opcache for me improves about 30%. Will test in docker.

@rbergen I have another optimization I have theorized, I will have findings soon. Can existing PR be updated, or is it easier to open a new one?

rbergen:
@xeridea In that case I prefer you to update this PR and let me know when you think it's ready to be merged. It will save us reviewing your optimizations twice. We have quite a lot of PRs to go through already. :)

Petah:
What about inlining it, or does it have to be faithful? And what really is "the full state of the sieve"?

https://gist.github.com/Petah/71e7fa50c984888c3780eb9949e7dd61

rbergen:
> And what really is "the full state of the sieve"?

Simply put. whatever information your algorithm needs to run it. After reviewing 100+ implementations, I'd say that in most cases that seems to be the bit/flag array, and the sieve size. The point is that the algorithm must not depend on any information external to the sieve state object (like global variables, compile-time constants, etc.) for executing or controlling the sieve run.

DennisdeBest:
@Petah running it inside my docker container as it is I get the following results :
```
Passes: 389, Time: 10013ms, Avg: 25ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 372, Time: 10018ms, Avg: 26ms, Limit: 1000000, Count: 78498, Valid: True
```
If I apply the changes from your GIST I get the following : 
```
Passes: 309, Time: 10028ms, Avg: 32ms, Limit: 1000000, Count: 78498, Valid: True
Passes: 309, Time: 10025ms, Avg: 32ms, Limit: 1000000, Count: 78498, Valid: True
```
So it seems to slow things down a little bit.

Petah:
@DennisdeBest weird

https://3v4l.org/XV45v
https://3v4l.org/s7Ke0

xeridea:
@rbergen updated readme and docker, ready to merge.

@DennisdeBest Enabled JIT, gives small boost to Docker, not as big as it should :( My other optimization makes PHP slower for some reason. Testing in C++ gives good boost though. Will be committing that after this is merged if you want to see.

xeridea:
@rbergen @DennisdeBest Sped up another 15%!

rbergen:
@xeridea Please base your PR on the current drag-race branch commit. The diff now includes all merges to drag-race since you opened it.

xeridea:
@rbergen Been fighting git to rebase it for couple hours but finally got it. Git isn't pushing from Windows, so doing from my VM, but rebasing with tortoisegit, but files kept getting messed up. Ready now.

DennisdeBest:
@xeridea awesome!

rbergen:
I'm wondering, what does the getRawbitCount() function do if it is called when the Sieve class has been created, but runSieve has not yet been called?

DennisdeBest:
@xeridea I just retested this with the docker container.

Before the 15% improvements  without JIT: 
`Passes: 373, Time: 10017ms, Avg: 26ms, Limit: 1000000, Count: 78498, Valid: True`

Before 15 % improvement with JIT :
`Passes: 383, Time: 10022ms, Avg: 26ms, Limit: 1000000, Count: 78498, Valid: True`

With the improvements and JIT :
`Passes: 409, Time: 10008ms, Avg: 24ms, Limit: 1000000, Count: 78498, Valid: True`

nice impovements !

rbergen:
> I'm wondering, what does the getRawbitCount() function do if it is called when the Sieve class has been created, but runSieve has not yet been called?

I tested it, and it throws an error: 
```
Fatal error: Uncaught Error: Typed property PrimeSieve::$rawbits must not be accessed before initialization in /opt/app/PrimePHP.php:73
Stack trace:
#0 /opt/app/PrimePHP.php(87): PrimeSieve->getRawbitCount()
#1 {main}
  thrown in /opt/app/PrimePHP.php on line 73
```

May I suggest to use `isset()` to check `$this->rawbits` in `printResults()` and `getRawbitCount()` and have the functions bail out cleanly if it is uninitialized? That means the Sieve class acts consistently (has a comprehensive state) after construction without impacting benchmarked performance in any way.

xeridea:
@rbergen Strange, getRawBitCount() shouldn't have been called before run. Initialized as empty array in class definition, which fixed it, and added isset() for good measure.

rbergen:
> @rbergen Strange, getRawBitCount() shouldn't have been called before run. Initialized as empty array in class definition, which fixed it, and added isset() for good measure.

It isn't called before run. My point was that it _should be possible_ to call it before run without it throwing a fatal error, because otherwise the class' state is inconsistent. This is an aspect specific to this implementation, because it is the only implementation I have seen that doesn't initialize the actual bit array in the constructor but in the runSieve() function itself, and then assigns it to the class member.

Initializing as an empty array will fix it too, but you may want to check if that has a negative impact on performance. If it does, you can drop the empty array initialization again if both getRawBitCount() and printResults() have the isset() check in place.

xeridea:
@rbergen yeah it is better coding practice to have it initialized anyway. For some reason, using local variable is faster, and then assigning to class member is pretty free since PHP uses copy on write. Initializing empty doesn't seem to impact speed.

rbergen:
@xeridea Fine with me. :) Can I consider the PR stable again?

xeridea:
@rbergen Yes

rbergen (approved):
LGTM

duskwuff:
You haven't fully enabled JIT! You need to set `-d opcache.enable_cli=1` to make opcache (and JIT) work in a CLI context. By default, it's only enabled for web server contexts.

Expect another ~2x speedup from JIT.

xeridea:
@duskwuff Hmm, thought I had confirmed it, and had flags set in dockerfile. Changed dockerfile to just copy an opcache.ini file to where it needs to go. Now confirmed, +%65 speed! Thank you for noticing.

@DennisdeBest JIT for real now.

@rbergen still stable.

xeridea:
@DennisdeBest The method of enabling JIT in the docker entrypoint apparently wasn't working. There is now opcache.ini that gets copied to php conf.d directory, and JIT works now. The last commit just removes some unnecessary stuff from dockerfile. Can you confirm?

B-Sterner:
Good Day, I made a proposed update. Not  sure on the git procedures with this, but here it goes.

1) remove the even checks from the clearBit and the getBit
2) update the runSieve to no longer send even numbers to the getBit use $i += 2 instead of $i++
3) clearBit doesn't get an even number anyway and can just remove the even check

HugoSantiagoBecerraAdan:
I found that all other implementations use 5 sec. for execution time limit, but here it's set to 10 sec.

As I already proposed in https://github.com/PlummersSoftwareLLC/Primes/pull/409 replacing raw arrays with `SplFixedArray` will **increase performance about 20%**.

From:
```
Passes: 157, Time: 5005ms, Avg: 31ms, Limit: 1000000, Count: 78498, Valid: True

DennisdeBest;157;5.005188;1;algorithm=base,faithful=yes
```

To:
```
Passes: 193, Time: 5.018801, Avg: 0.026004, Limit: 1000000, Count: 78498, Valid: 1

HugoSantiagoBecerraAdan;193;5.018801;1;algorithm=base,faithful=yes
```

rbergen:
> I found that all other implementations use 5 sec. for execution time limit, but here it's set to 10 sec.

There are others, including at least Java, AssemblyScript, Haxe and Python. These are solutions that were in the repo before the 5 second rule was in place. We (the maintainers) are discussing setting all runtimes to 5 seconds.

xeridea:
@HugoSantiagoBecerraAdan Updated with SplFixedArray.
@DennisdeBest Speed!!!

@rbergen I know you have a lot to manage, wondering if this can be merged soon? Working on some C++ ideas and don't want to complicate merge requests, so holding off on sharing them for now.

rbergen:
> @rbergen I know you have a lot to manage, wondering if this can be merged soon? Working on some C++ ideas and don't want to complicate merge requests, so holding off on sharing them for now.

I can't merge a PR until it's stable, but I feel comfortable deducing from your question that this one is, now. :) 
As you mention C++ ideas, I do want to manage expectations concerning the merge time on those. We currently already have 5 C++ PRs open that cause merge conflicts between them. I would therefore highly recommend going through the ones that are open before you open your PR, to see if you can add your improvements to one of those.

PR #309, opened 2021-07-05
Add Kotlin Solution

Theaninova:
## Description

Add a fast and an idiomatic Kotlin Solution. Both solutions are roughly based around the Java solution. The fast solution focuses on producing clean bytecode while the idiomatic solution focuses on writing more real-world readable code.

## Contributing requirements


rbergen:
A few things initially come to mind when looking at this PR:

- We have an earlier PR #258 open that is close to being merged, which adds 3 Kotlin solutions. Adding 2 more would bring the total to 5, which I think is suboptimal. Could you please take a look at #258 and indicate how you think this PR relates to it?
- In relation to the previous point, this PR would create a merge conflict with #258.
- Using Gradle requires the inclusion of a number of binary files, which we really don't like doing. (I know the Java solution suffers from the same problem; that was a very early solution that we grandfathered. Removing the Gradle files in that solution is still on my to-do list.) Would it be possible to choose an alternative build approach that does not rely on binary files?

Theaninova:
> A few things initially come to mind when looking at this PR:
> 
> * We have an earlier PR [Added implementations in Kotlin with various target platforms #258](https://github.com/PlummersSoftwareLLC/Primes/pull/258) open that is close to being merged, which adds 3 Kotlin solutions. Adding 2 more would bring the total to 5, which I think is suboptimal. Could you please take a look at [Added implementations in Kotlin with various target platforms #258](https://github.com/PlummersSoftwareLLC/Primes/pull/258) and indicate how you think this PR relates to it?
> * In relation to the previous point, this PR would create a merge conflict with [Added implementations in Kotlin with various target platforms #258](https://github.com/PlummersSoftwareLLC/Primes/pull/258).

Yes, I saw that one, but in my humble opinion the Kotlin Code in that repository doesn't do a good job showing various language features or writing idiomatic Kotlin code, it's more of a Java file converted to Kotlin Syntax rather than a Kotlin version with Kotlin features in mind.

In my opinion, it doesn't really make sense to use Kotlin over Java when you just end up writing Java code, because you don't gain much readability and you still have to make sure the compiler output is clean bytecode. If the compiler didn't mess anything up, it's basically just Java code (Kotlin is about way more than just null safety). Maybe nice as a comparison for "should you really use Kotlin files over Java for high-performance code if you omit Kotlin language features anyways".

But then my code is also a mixed functional approach, I could see an argument from both sides. If I had to make a guess, their solution would probably be faster than mine, but my fast solution was already pretty much on par with the Java version.

----

Apart from the code quality, they did a really interesting thing and created implementations for Kotlin/JVM, Kotlin/Native, and Kotlin/JS, respectively compiling to JVM Bytecode, Native Code, and JavaScript. That's something I honestly totally forgot about, and something that should 100% be included in some form, be it their version or mine adapted.

I would honestly plead for either the full idiomatic version or both the full idiomatic and reduced version, and their version, as "Idiomatic", "Idiomatic Optimized", "Barebones Optimized".

As for Kotlin/JVM, Kotlin/Native, and Kotlin/JS, I think it makes sense to have them as separate languages just as we have Python and Cython separately since there isn't a 100% overlap between the three, and they might require different optimization approaches, and are for different use cases (JVM for server, JS for front end, Native for embedded).

> * Using Gradle requires the inclusion of a number of binary files, which we really don't like doing. (I know the Java solution suffers from the same problem; that was a very early solution that we grandfathered. Removing the Gradle files in that solution is still on my to-do list.) Would it be possible to choose an alternative build approach that does not rely on binary files?

Yes, with the Docker setup the Gradle binaries wouldn't be necessary, you should in principle be able to just remove the `gradle` folder and the `gradlew[.bat]` as the Docker container already comes with Gradle 6.8 bundled (you should also be able to copy that approach to your Java version, as you might have also noticed I purposely used the same JDK docker container for execution as the Java version), but I kept the gradle binaries to aid anyone opening it in an IDE.
As for alternative approaches, there is the Kotlin Command Line compiler of course, but I've never used that one before.

rbergen:
@wulkanat and @fvbakel: I am hereby inviting you to discuss how to combine #258 and #309 into one Kotlin PR, preferably containing less than 5 solutions. The two PRs can't both be merged anyway due to the merge conflict that would lead to.

rbergen:
> As for Kotlin/JVM, Kotlin/Native, and Kotlin/JS, I think it makes sense to have them as separate languages just as we have Python and Cython

This I don't agree with. Cython is generally considered as not just a different runtime environment for Python, but a superset of that language. In the case of Kotlin, the language website itself states that it is one language, that can be used in different runtime environments.

Theaninova:
> > As for Kotlin/JVM, Kotlin/Native, and Kotlin/JS, I think it makes sense to have them as separate languages just as we have Python and Cython
> 
> This I don't agree with. Cython is generally considered as not just a different runtime environment for Python, but a superset of that language (similar to Delphi and Object Pascal, if you will). In the case of Kotlin, the language website itself states that it is one language, that can be used in different runtime environments.

Hmm, okay, fair point, Kotlin feature differences might be too minuscule to consider them a separate language. I'm gonna work on a version that supports Kotlin Multiplatform instead... Thought that leaves the question of how to handle the Docker containers, I suppose three Dockerfiles?

rbergen:
> Thought that leaves the question of how to handle the Docker containers, I suppose three Dockerfiles?

Three execution environments tends to mean three Dockerfiles, which also means three solutions. This is basically why I am hoping @fvbakel and yourself can collaborate towards one PR: while it is not possible to combine multiple execution environments in one solution, it is possible to do that with multiple _implementations_ that run in the _same_ execution environment. In that context, adding more implementations to one solution is a lot "cheaper" in resources (primarily build-time) than separate solutions.

Theaninova:
> > Thought that leaves the question of how to handle the Docker containers, I suppose three Dockerfiles?
> 
> Three execution environments tends to mean three Dockerfiles, which also means three solutions. This is basically why I am hoping @fvbakel and yourself can collaborate towards one PR: while it is not possible to combine multiple execution environments in one solution, it is possible to do that with multiple _implementations_ that run in the _same_ execution environment. In that context, adding more implementations to one solution is a lot "cheaper" in resources (primarily build-time) than separate solutions.

Hm, okay. Sounds like there is no one good solution... I'm not that well versed in Docker, but I think it should be possible to compile all three JVM, JS, and Native solutions from one multiplatform project in a single container, and then subsequently redirect those builds to other containers... Totally forgot, but Kotlin Multiplatform Projects leave a way to have common implementations as well as platform-specific ones (why wouldn't they), so with that, it'd definitely be nicer to have one Kotlin implementation as one solution... Maybe with a command-line argument or something for the Docker container? Not sure if that is a good solution, unfortunately, 99% of the time spent is on the Gradle daemon starting.

rbergen:
> Maybe with a command-line argument or something for the Docker container? Not sure if that is a good solution, unfortunately,

I find it very difficult to have an opinion about source code/solutions that I don't have in front of me. (That's a general limitation at my end, not your fault. :) ) That said, in the automated drag-race benchmark tooling, we don't support the passing of parameters to Dockerfiles while building/running the solution images.

fvbakel:
@wulkanat @rbergen mm, I did not notice there was already a Kotlin Pull request pending.  I would like to bring forward in the Kotlin solutions the following two points:

- show that with simple code, without bit shifting and masking, you get a reasonable performance
- show the difference in performance for the target platforms Native and JS and show that Kotlin seems to be optimized for JVM.

Is it possible to merge solution 1 and 2 in one solution, with two different main classes and a run.sh that runs both?

So my proposal would be to use this pull request as the basis. Make the code from solution 1 or 2 multiplatform. In this solution target JVM. In solution_3 target native and solution_4 target JS. If building these with gradle is not possible, we can use the Dockerfile I have prepared.  Then during the build of the docker file we can just copy the code from solution_1 to the Dockerimage of solution_3 and 4. This would avoid any special needs for volume mounting in Docker, or a difference in interface compared to the other builds in this project.

I am not familiar with gradle, so I don't know if the Dockerfile from solution 1 would be able to build Native and JS?

@wulkanat I noticed that your solutions runs for 10 sec instead of the standard 5. Further more it outputs all the Primes, which is usually switched off. One last remark, both solutions output `wulkanat` as the identifier, maybe better to use a distinct name?

Theaninova:
> @wulkanat @rbergen mm, I did not notice there was already a Kotlin Pull request pending.  I would like to bring forward in the Kotlin solutions the following two points:

Was actually my fault, I overlooked yours 😑
> 
> - show that with simple code, without bit shifting and masking, you get a reasonable performance
> - show the difference in performance for the target platforms Native and JS and show that Kotlin seems to be optimized for JVM.
> 

Absolutely agree with those two. Additionally I'd like to add 
- Performance of Kotlin vs Java, Kotlin vs plain JS, and Kotlin/Native vs C, Go, Rust,... are also in my opinion very viable questions to ask
- Show that using language features like inline lambdas, extension functions, tail recursion, etc don't have to come at a hard performance hit, essentially that you can write both beautiful and performant code.
- Show that with some attention you can achieve Java-Level performance with minimal tradeoffs in readability

> Is it possible to merge solution 1 and 2 in one solution, with two different main classes and a run.sh that runs both?
> 

Seeing how similar my two solutions are, definitely, my grime with my two solutions is just that tad bit less readable but with that tiny bit already achieves Java level performance. I'm not sure what constitutes to a single solution, I'd definitely like to keep both of mine in some form, but if that is not possible I can also completely merge those two.

> So my proposal would be to use this pull request as the basis. Make the code from solution 1 or 2 multiplatform. In this solution target JVM. In solution_3 target native and solution_4 target JS. If building these with gradle is not possible, we can use the Dockerfile I have prepared.  Then during the build of the docker file we can just copy the code from solution_1 to the Dockerimage of solution_3 and 4. This would avoid any special needs for volume mounting in Docker, or a difference in interface compared to the other builds in this project.

Okay, I guess the multiplatform example will then just ouput all results at once?

With the separate JS and Native solutions I'd be okay with, just bugs me a little bit that it would probably be relatively easy once set up to have the option for all projects to compile them for either platform. I'm not quite sure what the requirements for leaving it comparable are, but of possible having a flag to compile them for each platform would be my preferred option.
> 
> I am not familiar with gradle, so I don't know if the Dockerfile from solution 1 would be able to build Native and JS?

I believe you can just use the Gradle Docker Image to build all.
> 
> @wulkanat I noticed that your solutions runs for 10 sec instead of the standard 5. Further more it outputs all the Primes, which is usually switched off. One last remark, both solutions output `wulkanat` as the identifier, maybe better to use a distinct name?

Oh, yes, oversights from my part. What name would you suggest?

fvbakel:
> but if that is not possible I can also completely merge those two

For me Kotlin is new, this was my first Kotlin program ever. But I know that in Java, one jar file can have multiple classes with a main and then you can run a specific class with `java -cp myjar.jar com.mypackage.myClass`. I am not sure if Kotlin has a similar construct. If this is possible the "merge" would be just two classes with two different names.

>Okay, I guess the multiplatform example will then just ouput all results at once? 

It would be great if it could output all the different build for the different target platforms

>I believe you can just use the Gradle Docker Image to build all.

Great, if could implement that.

>Oh, yes, oversights from my part. What name would you suggest?

I use names like wulkanat_Kotlin1, wulkanat_Kotlin-js, etc. Maybe you could make this a command line parameter/make it depend on the platform so we can distinct the different target platforms.

Theaninova:
> For me Kotlin is new, this was my first Kotlin program ever. But I know that in Java, one jar file can have multiple classes with a main and then you can run a specific class with `java -cp myjar.jar com.mypackage.myClass`. I am not sure if Kotlin has a similar construct. If this is possible the "merge" would be just two classes with two different names.

For Kotlin/JVM it's the exact same concept, for other platforms on the other hand we'd probably need to adjust the build script (and at that point might as well do it for Kotlin/JVM

> It would be great if it could output all the different build for the different target platforms

Okay, so something like a single Dockerfile that builds all platforms at once, and then three other Dockerfiles for the individual tests?
> Great, if could implement that.

Alright, already started working on it. Kotlin Multiplatform unfortunately is something I'm not greatly familiar with, but I'll make it work some how.
> I use names like wulkanat_Kotlin1, wulkanat_Kotlin-js, etc. Maybe you could make this a command line parameter/make it depend on the platform so we can distinct the different target platforms.

Okay, in that case maybe omit the credit and use `Kotlin/JS-Idiomatic` etc instead? I wasn't sure what the names were for in the beginning.

rbergen:
This is just a quick check if this PR is still moving forward?

Theaninova:
Yup, really sorry about the lack of updates in the last few days, I think it's probably the best solution to try not to do everything at once and focus on an idiomatic and a fast variant, and add multiplatform support (which is giving me a headache) sometime later... That way I could get the PR ready much much sooner (maybe within a day)

fvbakel:
I assume no further action is required from me for this Kotlin PR. If there is something I can help with, just let me know.

rbergen:
@wulkanat No need to apologize, I have a day job too :) It's just that with 4 PRs open aiming to add Kotlin to the set of languages and it not being in yet, I'm just eager to assess what is still being worked on.

rbergen:
This is just a heads-up: after the weekend (as in Monday) I will review the status of the Kotlin PRs that are now open. I'd like it best if we can merge a stable version of this PR at that time, due to the thorough discussion that has taken place here. If that's not possible then I will probably merge #258, instead.

If this PR stabilizes before Monday, I will proceed towards merging it when that happens.

Putting it simply, it's time to add Kotlin to our collection of languages. :)

Theaninova:
Alright, I've now added a commit with three main solutions:
* `solution_1`, a variant of `solution_2`, idiomatic, but very close to Java speed.
* `solution_2`, a fully idiomatic solution
* `solution_4`, a solution based on #326, the fastest of all solutions.

**All** solutions use the exacts same build script, Dockerfile, and timing solutions, that also means that **all** solutions are multithreaded based on #326 

What is missing now is a working multiplatform implementation. `solution_3` includes a draft of this, a single solution that should compile to Native, JS, and JVM, however not fully functional quite yet... Specifically creating a fat jar doesn't work, however compiling to Native & JS works fine. If someone were to get the build script from that solution to work, we could add multiplatform support to **all** solutions, `solution_3` would be omitted. I'm not sure how to handle this now, I have little hope that I myself will get it to work, should we include just a JS & Native solution then? Or the ones based on #258 ?

Other than that I'd consider this PR ready now

rbergen:
> **All** solutions use the exacts same build script, Dockerfile, and timing solutions, that also means that **all** solutions are multithreaded based on #326

If that's the case, would it be possible to convert the three solutions into three implementations that are rolled up into one solution? It would really make a difference in docker build time and resources during benchmark runs...

Theaninova:
> If that's the case, would it be possible to convert the three solutions into three implementations that are rolled up into one solution? It would really make a difference in docker build time and resources during benchmark runs...

Yes, should be no problem, what's the desired output...? Do all tests output one after another? Building all at once would **greatly** reduce build time due to Gradle ramp up, but with each test taking 5s, we are looking at the whole thing taking 30s to complete when it runs through all at once. I'm not that experienced with docker, would it be possible to separate out the build step and then have command-line arguments supplied, I'm thinking something like `-idiomatic:s -fast:s` aka every method you want to run... It's ofc not a question of is it possible in Kotlin, more how would it be handled by Docker.

rbergen:
Desired output is one "drag-race format" output line per implementation, period. :)

The benchmark tool is not setup to pass parameters to docker, but within the docker context you can basically do whatever is needed. An example of a solution combining multiple implementations is PrimeBASIC/solution_1. As you can see, that solution uses shell scripts to glue things together. 
In any case, if getting things to run in Docker is where things get stuck, I'm happy to help out where I can.

Theaninova:
> Desired output is one "drag-race format" output line per implementation, period. :)
> 
> The benchmark tool is not setup to pass parameters to docker, but within the docker context you can basically do whatever is needed. An example of a solution combining multiple implementations is PrimeBASIC/solution_1. As you can see, that solution uses shell scripts to glue things together.
> In any case, if getting things to run in Docker is where things get stuck, I'm happy to help out where I can.

Okay, so I suppose just output everything one after another, with no way to configure it in any way what is being run, so it just means that the test will always take 30s to execute. This might become an issue once (if) we get Kotlin Multiplatform to work... Not sure if it's desirable to wait a few minutes if you just want to re-run a single test (for example if you want to compare it to Java and just want to run the fast solution, not the multithreaded or JS or Native examples).

If that is the desired way to handle it though, it should pose no issue to combine them.

rbergen:
> Okay, so I suppose just output everything one after another, with no way to configure it in any way what is being run, so it just means that the test will always take 30s to execute.

You can make what's run configurable (as in, controlled by a command-line parameter passed to docker), as long as "everything in sequence" is then the default. The benchmark tool can then run all of them, while selections can be executed by manual use of docker build and docker run.

Theaninova:
Okay, I've now updated solutions 1 & 3:

`solution_1` is now a collection of three solutions from me and #326

BUT

I've now finally gotten Multiplatform to work, which means

`solution_3` includes the same solutions, but running on JVM, Native, and NodeJS.
This would in principle make `solution_1` and #358 obsolete, but building and execution in Docker are broken.
Specifically
- The native build doesn't work because of some missing dependencies
- The Gradle Daemon starts for each task over
- Running the jar and the js files in a Docker container just yields one output... I have no idea why, it works fine on my machine, just not in the container, and it's an issue with both the JS and the JVM variant...
- I can't seem to be able to run two containers one after another after the build stage is finished... I'm sure that's just my lackluster experience with Docker.

In any case, if you @rbergen would be able to get the Docker setup running, I've been unable so far to locate the issue, I'd propose to omit everything except `solution_3` (rename that to `solution_1`), and merge that as the all-encompassing Kotlin solution (unless someone got another radically different solution.

rbergen:
@wulkanat I'm happy to help where I can, but I feel I do need to manage expectations a little bit. I'm not a Docker expert by any means; I've learned what I know "on the job", myself.  

If getting all platforms to build and run in one docker image is not something we can make working, I am ok with merging three solutions, one for each platform. This would be akin to the structure of #258, which is the PR I would be inclined to merge otherwise, as I mentioned. However, I would really want to aim for a situation where we do not include binary files (like Gradle wrappers) in the commit, as we don't merge binary files if it can at all be avoided. As #258 does not rely on binary files either, maybe that PR could serve as a "template" from a solution and build perspective?

@fvbakel Could this be something you could provide assistance with?

Theaninova:
Alright... I want to mention that it is not necessary to include the Gradle binaries right now, but it makes it possible to open, edit and debug it in IntelliJ. The Docker version doesn't rely on them and instead uses just the Gradle provided by the image.

I suppose the solutions for each platform the would be a single solution with three docker files for each platform, maybe with the Kotlin CLI build tools (I've got exactly 0 experience with that though)?

rbergen:
To make the whole thing work with the benchmark tool that is part of this project, that being a requirement for them to be included in the leaderboard comparisons, 3 Dockerfiles does automatically mean 3 solutions. We only support one Dockerfile per solution directory, which must be called Dockerfile and stored at the "solution root".
I do tend to motivate people to combine implementations in one solution where a Dockerfile can be shared between them, but to me it really is more important that we have a solid representation of Kotlin in the repo than it is to stick to the "one Dockerfile principle" in a situation where it stops things from working at a practical level.

fvbakel:
@wulkanat @rbergen I can have a look into this this afternoon and see if I can make a proposal on how to solve the issues.

fvbakel:
@wulkanat @rbergen I had a look into this today. See outcome below.

**Get native build working**

The native build was missing one Ubuntu package. This can be solved by running `apt-get update && apt-get -y install libncurses5`. After this `gradle linkReleaseExecutableNative` produces an executable `/src/build/bin/native/releaseExecutable/rimeSieve.kexe`. Running `./PrimeSieve.kexe -t -if -i` produces:

```
native_kotlin_traditional_single;1280;5.002;1;algorithm=base,faithful=yes
native_kotlin_idiomatic_fast_single;1314;5.003;1;algorithm=base,faithful=yes
native_kotlin_idiomatic_single;1251;5.002;1;algorithm=base,faithful=yes
```

I am not sure if that is the expected output, I did not dig into the code.

**Build process proposal**

If I understand correctly solution_3 is the super set that contains all other solutions. under that assumption I propose the following: 

1. rename `solution_3` to `source_build` 
2. change the Dockerfile to the content below. This will produce a docker image that can build the sources, contains all dependencies prepacked, but does not contain the source or build products.

```
ARG GRADLE_VERSION=6.8
ARG JDK_VERSION=16

FROM gradle:${GRADLE_VERSION}-jdk-hotspot

COPY . /src
WORKDIR /src

RUN apt-get update && apt-get -y install libncurses5
RUN gradle shadowJar
RUN gradle jsNodeProductionRun
RUN gradle linkReleaseExecutableNative
RUN rm -rf /src/build/bin/native/releaseExecutable/*
RUN rm -rf /src/build/classes
RUN rm -rf /src/build/libs/*.jar
RUN rm -rf /src/build/js/packages/PrimeSieve/kotlin/*
RUN rm -rf /src/build/bin/native/releaseExecutable/*
RUN rm -rf /src/src/*

ENTRYPOINT ["bash"]
```
3. Publish the Dockerfile from step 2 on [https://hub.docker.com/](https://hub.docker.com/) and provide a tagged build there
4. Create solution_1 directory. With Dockerfile like this in pseudo code:

```
FROM <step 3 published Dockerfile> as BUILD
COPY ../source_build/src /src
RUN gradle shadowJar

FROM adoptopenjdk:${JDK_VERSION}-jdk-hotspot

COPY --from=BUILD /src/build/libs/PrimeSieve-1.0-all.jar /bin/runner/run.jar

WORKDIR /bin/runner
COPY ./run.sh ./

ENTRYPOINT ["./run.sh"]
```

where `run.sh` is:

```
java -jar run.jar -t -if -i -t:m -if:m -i:m
```

5. Create `solution_2` directory, similar as `solution_1`. But this time do the JS steps.
6. Create `solution_3` directory, similar as `solution_1`. But this time do the Native steps. 

This is the general concept, I think there might be some small mistakes or details. 

Please let me know what you think of this proposal. If oke, then I might be able to implement that on Thursday afternoon.

Edits:

- typos

Theaninova:
Great work! This sounds like a perfect solution to me!

rbergen:
@fvbakel Thanks for putting that together, it looks very well thought through. I only have one comment to make up-front, and I apologize for sort of repeating myself: the solution that is in the end lined up to be merged should itself not include binary files, like Gradle-related .jar files.

fvbakel:
@rbergen @wulkanat  Thanks for the feedback. I will try to do without the gradle binaries. 

@rbergen Do you have any tips how I can branch of the `wulkanat:drag-race` to my own branch where I can do the edits. Should I best fork his repository and create a PR to @wulkanat? or is there another, better approach?

Theaninova:
@fvbakel I invited you as a collaborator to this my PR repo

rbergen:
@fvbakel The approach @wulkanat has chosen is a great way to collaborate on any repo.
Otherwise, and for future reference, your idea also works. Technically, what works for me is locally adding a remote that points to the other person's fork, branching off of that, pushing that branch to your own fork on GitHub and then opening a PR off of your GitHub branch on the other person's fork, as you described.

fvbakel:
@rbergen @wulkanat Thanks for the tips and enabling the push access. I will workout my proposal on Thursday afternoon.

fvbakel:
Just to give you a status update, I tried my proposed solution today, only to find out that it does not work. Main reason is that the benchmark mechanism and Docker do not allow a copy from ../source_build directory. Next to that the java build seems broken.
I will have a fresh look an new ideas on Friday.

rbergen:
> Main reason is that the benchmark mechanism and Docker do not allow a copy from ../source_build directory

No, that is indeed not possible.

fvbakel:
@rbergen @wulkanat I did my best to get the multiplatform working. It does work, but it is still a bit slow and cumbersome. For the moment this is the best I can do. 

On my system I can now run `make DIRECTORY=PrimeKotlin/solution_1` and I get the output:

```
make[1]: Entering directory '/home/fvbakel/git/forks/1/Primes/tools/node_modules/node-uname/build'
  CC(target) Release/obj.target/uname/uname.o
  SOLINK_MODULE(target) Release/obj.target/uname.node
  COPY Release/uname.node
make[1]: Leaving directory '/home/fvbakel/git/forks/1/Primes/tools/node_modules/node-uname/build'
added 229 packages in 3.935s
info: [PrimeKotlin][solution_1] Building...
info: [PrimeKotlin][solution_1] Running...
┌───────┬────────────────┬──────────┬─────────────────────────────────────┬────────┬──────────┬─────────┬───────────┬──────────┬──────┬───────────────┐
│ Index │ Implementation │ Solution │ Label                               │ Passes │ Duration │ Threads │ Algorithm │ Faithful │ Bits │ Passes/Second │
├───────┼────────────────┼──────────┼─────────────────────────────────────┼────────┼──────────┼─────────┼───────────┼──────────┼──────┼───────────────┤
│   1   │ PrimeKotlin    │ 1        │ jvm_kotlin_idiomatic_fast_multi     │  5218  │ 5.00000  │    1    │   base    │   yes    │      │  1043.60000   │
│   2   │ PrimeKotlin    │ 1        │ jvm_kotlin_traditional_multi        │  4991  │ 5.00000  │    1    │   base    │   yes    │      │   998.20000   │
│   3   │ PrimeKotlin    │ 1        │ jvm_kotlin_idiomatic_multi          │  3988  │ 5.00000  │    1    │   base    │   yes    │      │   797.60000   │
│   4   │ PrimeKotlin    │ 1        │ jvm_kotlin_idiomatic_fast_single    │  3947  │ 5.01700  │    1    │   base    │   yes    │      │   786.72513   │
│   5   │ PrimeKotlin    │ 1        │ jvm_kotlin_traditional_single       │  3818  │ 5.06700  │    1    │   base    │   yes    │      │   753.50306   │
│   6   │ PrimeKotlin    │ 1        │ jvm_kotlin_idiomatic_single         │  2704  │ 5.00600  │    1    │   base    │   yes    │      │   540.15182   │
│   7   │ PrimeKotlin    │ 1        │ native_kotlin_idiomatic_fast_single │  730   │ 5.00400  │    1    │   base    │   yes    │      │   145.88329   │
│   8   │ PrimeKotlin    │ 1        │ native_kotlin_traditional_single    │  679   │ 5.00300  │    1    │   base    │   yes    │      │   135.71857   │
│   9   │ PrimeKotlin    │ 1        │ native_kotlin_idiomatic_single      │  640   │ 5.00100  │    1    │   base    │   yes    │      │   127.97441   │
│  10   │ PrimeKotlin    │ 1        │ js_kotlin_traditional_single        │  573   │ 5.00900  │    1    │   base    │   yes    │      │   114.39409   │
│  11   │ PrimeKotlin    │ 1        │ js_kotlin_idiomatic_fast_single     │  544   │ 5.02100  │    1    │   base    │   yes    │      │   108.34495   │
│  12   │ PrimeKotlin    │ 1        │ js_kotlin_idiomatic_single          │  535   │ 5.03500  │    1    │   base    │   yes    │      │   106.25621   │
└───────┴────────────────┴──────────┴─────────────────────────────────────┴────────┴──────────┴─────────┴───────────┴──────────┴──────┴───────────────┘
```

Theaninova:
Beautiful Output, and good job getting it all to work, though I believe we have to output it in the machine-readable format...

marghidanu:
> Beautiful Output, and good job getting it all to work, though I believe we have to output it in the machine-readable format...

 Use the formatter option: `make DIRECTORY=PrimeKotlin/solution_1  FORMATTER=csv` or `make DIRECTORY=PrimeKotlin/solution_1 FORMATTER=json`

fvbakel:
@wulkanat I did not create that beautiful output, it what is created by the make tool @marghidanu and @rbergen created based on the output of solution_1. The output of your Kotlin programs is unchanged. I forgot to mention some tasks for you. What used to be solution 3 is now solution 1. The other solutions have been prefixed with old. Can you check if everything is now included in solution_1 and if so, can you then delete the solutions that are prefixed with old.

I will fix the CI check that failed.

Once we have that in place, we can ask @rbergen to do the review.

Theaninova:
@fvbakel Okay, I've reviewed your changes and have now removed the obsolete solutions, I also updated the README!

rbergen (commented):
@fvbakel I appreciate the full transparency of including the CompileDockerfile and publish_to_docker.sh files in this PR, but I think you'll have to pull these into a separate repo (and out of this PR). Including a link to that repo in this PR's README and/or a comment in the Dockerfile would then still implement the transparency you do now.

The reason I say this is that people who don't have your Docker Hub credentials can't push the kotlin-primes base image anyway, and nobody but you should have those credentials.

Theaninova:
Hmm, I'd disagree, if he or his Docker account disappears all we'd have to do is change some links and someone else would publish the image and it would work again...

rbergen:
> Hmm, I'd disagree, if he or his Docker account disappears all we'd have to do is change some links and someone else would publish the image and it would work again...

To be honest, I think that's a rather unlikely scenario, but I'll obviously leave it up to @fvbakel to comment on that as well. Still, if that's a concern then we can add the Dockerfile repo to this organization and the image to the PrimeImages Docker Hub organization account that's linked to this project. We've done this for a number of other Docker base images as well.

fvbakel:
@rbergen Thanks for your remarks, I think it keeps the quality of this project at a high level.

@rbergen @wulkanat I did not know there is a PrimeImages organization  on Docker, or how to use that. The main reason I put the docker file here is because it depends on the source code in this repository. The main reason for that is that graddle starts to download the dependencies at first build.
If some dependency changes the published  image has to be updated. If there is no dependency change, but only a source code change, then a fresh publish is not needed.

If we move the compiledockerfile away from here, then where do we put it, and how can we still make it depend on the docker context here? 

But apart from this location discussion, is otherwise the Docker build acceptable for you?

rbergen:
Apologies for the #482 reference. I clicked on the wrong tab at one point in my review workflow. 

@fvbakel Thank you for your explanation. I guess that I didn't quite get the purpose of the CompileDockerfile, before. 
Now that I think I do, I don't think this approach is going to achieve what you intend to. By design, all solutions in this repo must be able to be compiled/built completely from source, without circular dependencies between the source and external components. 
The entire point of the benchmark tool is to allow anybody to clone the repo, make any changes locally that they want, and then build and run (their versions of) the solutions on their machine. This concept will break with the setup that is now chosen here.

I assume the CompileDockerfile setup was chosen to save docker build time for this set of implementations. As much as I appreciate the consideration you've given this, any pre-built base image that is stored in an external registry can't rely on the sources in this solution.

fvbakel:
@rbergen 
> I assume the CompileDockerfile setup was chosen to save docker build time for this set of implementations

Yes, that was the main reason, but it indeed comes at he cost of maintenance complexity. If you can live with a little bit longer build time, we can do without and make it less complex. We can just do the CompileDockerfile stuff in the regular Dockerfile, There is also an update of the graddle and Kotlin pending by @bluefireoly, see [https://github.com/wulkanat/Primes/pull/1](https://github.com/wulkanat/Primes/pull/1). That upgrade might reduce the build time also a bit, I have not tested that. 

So the question to you is, can you live with the longer build time?

rbergen:
> So the question to you is, can you live with the longer build time?

I'll know when I build the solution as part of my review. :) That maybe sounds like a bit of cheap shot, but I really can't decide on something I haven't yet seen. That said, you can trust that, when reviewing, I will keep in mind that the alternative to this multiplatform solution would be a number of separate solutions, each with _their own_ docker build time. 

I guess it's still possible to create and register a base image for the compile, that being an image that uses the Gradle image as its base images, and performs the apt-get update and install of libncurses5. I don't know how much time that would save, but I can imagine it not being worth the effort.

fvbakel:
The problem in the build time is with the first time graddle build that installs the dependencies depending on the source code here. Having separate solutions would not help, it would just increase the time.

So let's make the one Docker file build.

rbergen:
@wulkanat and @fvbakel I just want to quickly check that we're not in some sort of circular wait situation? As in, is my assumption correct that one or both of you will consolidate both Dockerfiles into one?

fvbakel:
@wulkanat @rbergen I am indeed waiting on the approval of @wulkanat  for the updates @bluefireoly and the go signal to update the Docker file after that decision.

fvbakel:
Oke, now the update of the Docker files is on me I guess. I hope I can do that on Tuesday because there is a major internet outage in my area.

fvbakel:
@rbergen I have now updated the Dockerfile. The build takes 10 minutes on my system.

rbergen:
So far I've found that for me, at home (100mbps consumer-grade downstream Internet connection, shared with a teenage daughter who likes watching Netflix when it's raining), the download of base images took about a minute.

After that, I got the following output:
```
------
> [build 2/5] RUN apt-get update && apt-get -y install libncurses5:
#11 0.597 Get:1 http://archive.ubuntu.com/ubuntu focal InRelease [265 kB]
#11 0.709 Get:2 http://archive.ubuntu.com/ubuntu focal-updates InRelease [114 kB]
#11 0.773 Get:3 http://security.ubuntu.com/ubuntu focal-security InRelease [114 kB]
#11 0.795 Get:4 http://archive.ubuntu.com/ubuntu focal-backports InRelease [101 kB]
#11 0.840 Get:5 http://archive.ubuntu.com/ubuntu focal/restricted amd64 Packages [33.4 kB]
#11 0.863 Get:6 http://archive.ubuntu.com/ubuntu focal/multiverse amd64 Packages [177 kB]
#11 0.900 Get:7 http://archive.ubuntu.com/ubuntu focal/main amd64 Packages [1,275 kB]
#11 1.248 Get:8 http://archive.ubuntu.com/ubuntu focal/universe amd64 Packages [11.3 MB]
#11 4.926 Reading package lists...
#11 5.552 E: Release file for http://archive.ubuntu.com/ubuntu/dists/focal-updates/InRelease is not valid yet (invalid for another 8h 14min 51s). Updates for this repository will not be applied.
#11 5.552 E: Release file for http://archive.ubuntu.com/ubuntu/dists/focal-backports/InRelease is not valid yet (invalid for another 8h 15min 31s). Updates for this repository will not be applied.
#11 5.552 E: Release file for http://security.ubuntu.com/ubuntu/dists/focal-security/InRelease is not valid yet (invalid for another 8h 14min 30s). Updates for this repository will not be applied.
------
executor failed running [/bin/sh -c apt-get update && apt-get -y install libncurses5]: exit code: 100
```

I guess that means I'll need to wait a little before I can come to further conclusions about build time.

rbergen:
Right. So, I didn't have the time to look into it when I posted that last comment, but I now know that was my first encounter with WSL2 time drift. 
With that sorted, the build's now running.

rbergen:
So, all in all download and build added up to about 6.5 minutes on my machine. That's just over 2 minutes per platform, which I think is perfectly acceptable.

rbergen:
@wulkanat and @fvbakel It seems that all implementations report a fixed thread count of 1, including the multithreaded ones. I don't think that can be accurate?

jakobkmar:
@rbergen all outputs should be correct now

rbergen:
@bluefireoly Thank you, and apologies for not tagging you in earlier comments.

rbergen (approved):
LGMT
@wulkanat, @bluefireoly and @fvbakel Thank you for your collaboration on this one.

marghidanu (approved):

PR #311, opened 2021-07-05
Fortran: Faithful + faster 2nd solution

tjol:
## Description

I know you're not really looking for new solutions where you already have one, but here's another Fortran solution which, unlike solution 1, uses a class rather than a global array that's allocated once.

It's also a fair bit faster compared to Fortran solution 1 (with or without #305), at the cost of the inner loop being less readable.

## Contributing requirements

### DIFF
--- a/PrimeFortran/solution_2/.gitignore
+++ b/PrimeFortran/solution_2/.gitignore
@@ -1,0 +1,4 @@
+*.mod
+/prime-8bit
+/prime-bitarray
+/prime-logical-array
--- a/PrimeFortran/solution_2/Dockerfile
+++ b/PrimeFortran/solution_2/Dockerfile
@@ -1,0 +1,10 @@
+FROM ubuntu:18.04
+
+RUn apt-get update && apt-get -y install make gfortran
+
+WORKDIR /opt/app
+COPY Makefile *.f08 ./
+
+RUN make all
+
+ENTRYPOINT [ "make", "run" ]
--- a/PrimeFortran/solution_2/Makefile
+++ b/PrimeFortran/solution_2/Makefile
@@ -1,0 +1,18 @@
+.PHONY: all run clean
+
+ALL=prime-bitarray prime-8bit prime-logical-array
+
+FORTRAN=gfortran
+#  -march=native just slows it down (?!)
+FORTRANFLAGS=-Ofast
+
+all: $(ALL)
+
+%: %.f08
+	$(FORTRAN) $(FORTRANFLAGS) -o $@ $<
+
+run: all
+	@for program in $(ALL); do ./$${program}; done
+
+clean:
+	rm -f $(ALL) *.mod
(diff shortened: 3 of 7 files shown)


rbergen:
> I know you're not really looking for new solutions where you already have one

Just to clarify on that: we do accept new solutions if there is a significant change between the solution that is already in place, and this one. A switch from not faithful to faithful is an example of that; it's educational from a performance perspective how one implementation type compares to another.

To complete this comment: in a scenario where having multiple implementations does make sense, there is a (slight) preference for adding multiple implementations to one solution instead of having a solution per implementation. The former approach will save creating and running a second (or third...) Docker image when running the benchmark.

Garklein:
Using gfortran kind (1, 2, 4, 8) isn't portable, perhaps consider using iso_fortran_env's int8 etc?
I know it's only going to be used here with gfortran but I still think it's a good idea.

tjol:
> Using gfortran kind (1, 2, 4, 8) isn't portable, perhaps consider using iso_fortran_env's int8 etc?

Is that a GNU extension? Well in that case this sounds like a good suggestion.

johandweber:
It is a pert of Fortran 2008.

Garklein:
In Fortran, using numbers for kind isn't standardized. In gfortran, 1, 2, 4, 8 could mean bytes, while in another compiler, 1, 2, 3, 4 could mean the same thing.

johandweber:
What I really like about this version is that it shows that modern Fortran can be used to write clean, well-structured
and object-oriented programs and Fortran programming no longer means writing fixed-form F77 style code optimized
for punch-card readers.

rbergen:
@tjol @johandweber @Garklein Can I consider this PR stable and ready for review?

tjol:
As far as I'm concerned, yes.

johandweber:
I also think that it is ready for review (Note that in contrast to Solution_1, I was not involved in the creation of Solution_2, so my OK is not really necessary).

rbergen:
@johandweber Thanks for replying. Just as a note, I wasn't asking for an OK, only for a statement if the PR is stable from your perspective. It could have been you were still brewing on some review feedback, which you're allowed to give whether you are involved in the creation or not. If that had been the case, I'd rather waited with moving towards merging.

Garklein:
Yep, I have no more review feedback.

Garklein:
tjol, in the primesieve_clear_bits subroutine, where the bitmask is made, it still uses kind=1 instead of kind=int8.

tjol:
> tjol, in the primesieve_clear_bits subroutine, where the bitmask is made, it still uses kind=1 instead of kind=int8.

Well spotted, thanks! I'll change that later today

rbergen:
@tjol Was that last commit the last one you planned to make in this PR?

tjol:
@rbergen Well really the one before that *was* the last one I was planning ;-)

I'm not planning any further changes unless something else comes up during review. Do your worst!

rbergen:
> Do your worst!

Never on a Tuesday! :)

rbergen (approved):
LGTM

marghidanu (approved):

PR #314, opened 2021-07-06
Support for multiple output formatters

marghidanu:
At the moment, all results are displayed in a text table. To make it more friendly towards other tools (like Excel & friends), the benchmark output can now be exported as CSV or JSON. The Makefile has also been modified to support the FORMAT variable; below is an example:

```
make one SOLUTION=PrimeCrystal/solution_1 FORMATTER=json
make one SOLUTION=PrimeCrystal/solution_1 FORMATTER=csv
make one SOLUTION=PrimeCrystal/solution_1 FORMATTER=table
```

If the variable is omitted, it will default to the `table` output formatter.


rbergen:
I had a feeling there was something brewing after some things I read yesterday :)

I think it's a very nice addition, and on my machine it works like a charm. For one it will make it much easier for me to generate test data for PrimeView. :D

I was thinking it may be good to add the FORMATTER capability to BENCHMARK.md as well?

marghidanu:
Thanks @rbergen! I will add the additional documentation, so we can let everybody know about this feature. :)

rbergen (approved):
Looks good to me!

PR #319, opened 2021-07-06
Made Pascal ver 64-bit capable

tedsmith:
## Contributing requirements

### DIFF
--- a/PrimePascal/solution_1/prime.pas
+++ b/PrimePascal/solution_1/prime.pas
@@ -15,13 +15,13 @@ type
         NotPrimeArray: PackedBoolArray;
 
     public
-        constructor Create(Size: Integer);
+        constructor Create(Size: Int64);
         function RunSieve(): PackedBoolArray;
         function CountPrimes(): Integer;
         function ValidateResults(var ReferenceResults: CheckMap): Boolean;
 end;
 
-constructor PrimeSieve.Create(Size: Integer);
+constructor PrimeSieve.Create(Size: Int64);
 begin
     SieveSize := Size;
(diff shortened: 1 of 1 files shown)


rbergen:
Thanks for your submission. 

I'm happy to change the sieve size data type to Int64, but that should then also be done for the PrimeSieve class' SieveSize variable, otherwise you're potentially assigning a 64-bit value to a 32-bit variable.

I'm not convinced the changes to variable initialization are a good idea. In most cases they effectively lead to a double initialization, which will not have a positive impact on performance. I would say possibly to the contrary, however negligible. In the case of the variable Count in the PrimeSieve.CountPrimes() function, I think the current explicit initialization to 0 is more expressive than an initialization to Default in the var block. 

In more general terms, beyond the sieve size change to In64, it is my opinion that the changes made are basically cosmetic and don't improve the performance of the benchmarked code. That means I don't think there is a reason to apply them.

tedsmith:
Hi there. I quite understand, and I suspect you are entirely correct. Thanks for considering it and looking over it though. And extra thanks for even including Freepascal...it often gets overlooked. :-)

rbergen:
> Thanks for considering it and looking over it though. And extra thanks for even including Freepascal...it often gets overlooked. :-)

You're welcome. I enjoyed programming in Pascal when it was taught in university (it was Turbo Pascal at the time, we're talking late '90s here), and simply had to add a solution in the language when I noticed there wasn't one, at the time. :)

tedsmith:
Me too. I was studying computer science in 1993 and used Turbo Pascal back then. I got into freepascal in 2010ish

PR #320, opened 2021-07-06
cleaned up fortran

Garklein:
## Description

Improved readability of the Fortran solution
Not sure if you're accepting improvements to existing solutions or not

### DIFF
--- a/PrimeFortran/solution_1/PrimesFortran.f03
+++ b/PrimeFortran/solution_1/PrimesFortran.f03
@@ -1,194 +1,160 @@
 program primes_fortran
-  use iso_fortran_env
-  implicit none
+    use iso_fortran_env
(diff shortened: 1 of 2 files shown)


rbergen:
We do. However:

- The changes need to objectively improve the performance of the solution. If your PR only improves readability, then I don't think we should merge it.
- We already have 2 PRs open with regards to Fortran. This PR would create a merge conflict with one of them (#305). If you think your changes should be applied to solution_1, then please discuss in #305 how to incorporate them in that PR.

Garklein:
Closing because this doesn't improve performance.
The main reason I wanted to change it was because the old comparison operators (.eq. etc.) kill my eyes, and it also looks like the second, faithful solution in the other PR doesn't use those.

PR #321, opened 2021-07-06
added Java implementation with BitSet

PratimGhosh86:
## Description

*This implementation follows the original implementation by Dave and uses BitSet rather than arrays and no longer limited to `Integer.MAX_VALUE - 2` (Requested array size exceeds VM limit).*

## Contributing requirements


rbergen:
This PR seems to contain a lot of files that are not part of the core solution (which seems to consist of one .java file), but development and/or build tooling that were used to create it. Some of these files are binary (jar) files, which we prefer not to introduce to the repo.

Could you please remove all files that are not needed to build and run the solution in the context of the Dockerfile provided with the solution, or at least the files in binary format?

PratimGhosh86:
@rbergen removed the maven wrapper and maven `pom.xml` files, let me know if any other change is required.

rbergen:
@PratimGhosh86 Thanks, that's a lot more manageable already. Looking at the set of files now, I think it's preferable if .editorconfig is removed as well, or alternatively added to .gitignore. The .editorconfig file is not required to build and run the solution, and furthermore editor preferences tend to be somewhat user-specific, by nature.

PratimGhosh86:
@rbergen done, it's just the source and dockerfile, and scripts to run it without docker.

rbergen (approved):
LGTM

marghidanu (approved):

MARKMENTAL:
Hi Pratim, your solution was really interesting in the way you used BitSet to keep track of primes and I'm trying to make something similar with Kotlin, I just have one question about a possible bug in your printResults() function. On my PC (Running Fedora 34 with IntelliJ and JDK 16) when the boolean parameter showResults is true it seems that the JVM throws an error about being unable to convert the long to a string. 

`Exception in thread "main" java.util.IllegalFormatConversionException: d != java.lang.String`

Is this related to using printf instead of println? If I change the printing object to println

`System.out.println(String.valueOf(num));`

 and pass num through it that way it seems to work then.

PR #322, opened 2021-07-06
adding Elixir implementation

cdesch:
## Description

Adding an implementation in Elixir. This implementation uses recursion in place of a while loop since Elixir does not have a true while loop. Elixir is also immutable requiring the bit array in Dave's implementation to be transformed and returned.

## Contributing requirements

* [x ] I placed my solution in the correct solution folder.
* [x ] I added a README.md with the right badge(s).
* [x ] I added a Dockerfile that builds and runs my solution.
* [x ] I selected `drag-race` as the target branch.
* [x ] All code herein is licensed compatibly with BSD-3.


PrimeElixir/solution_1/README.md
```diff
@@ -0,0 +1,34 @@
+# Ruby solution by rbergen
```
rbergen: I feel honoured you used my Ruby README.md as the template for yours. I do think the title of this one should probably be updated, though :)

PrimeElixir/solution_1/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-no-green)
```
rbergen: For a solution that is not faithful, the badge colour to use is -yellowgreen instead of -green

PrimeElixir/solution_1/prime.ex
```diff
@@ -0,0 +1,78 @@
...
+    100000000 => 5761455
+}
+
+size = 1000000
+primes = Enum.map(1..size, fn(_x) -> true end)
+# results = PrimeSieve.run_sieve(primes, 3, size)
+#   |> Enum.drop_every(2)
+#   |> PrimeSieve.count_primes()
+# IO.inspect results
+# IO.inspect PrimeSieve.validate_results(size, results, reference_results)
```
rbergen: If this commented-out code is not needed for the final solution, then please remove it.

PrimeElixir/solution_1/README.md
cdesch: Thanks! I updated it!

PrimeElixir/solution_1/README.md
cdesch: Thank you! Changed!

PrimeElixir/solution_1/prime.ex
cdesch: Thanks. Removed!

cdesch:
Thanks @rbergen ! I pushed the changes you noted!

rbergen:
I'm wondering if there is a way (like a command-line option to elixir) to suppress the warning messages about unused variables that are shown when running the implementation?

cdesch:
yes, @rbergen or I could just fix the warnings 👍

rbergen:
You know what, I think you're right :)

PrimeElixir/solution_1/prime.ex
```diff
@@ -0,0 +1,105 @@
...
+  Validate the prime count for the sieve_size with the reference results
+  """
+  def validate_results(sieve_size, count) do
+    Map.get(@reference_results, sieve_size) == count
+  end
+
+end
+
+# Setup
+size = 1000
```
rbergen: The sieve size does need to be set to 1,000,000, if possible.

cdesch:
@rbergen Thanks. I fixed the `sieve_count`.

I also wrapped everything into a `mix` project that elixir uses typically with tests and updated the docs. 

Running the docker container is exactly as it was before for integration with `make`

rbergen:
Do the latest commits mark the end of your work on this?

cdesch:
Yes. With the last push I did some final clean up the code (DRY and formatting) and readme.md

It is good to go. Thank you!

I'm sure someone will have some suggestions later on for optimization or fixing a mistake. Elixir doesn't do loops and iterations exactly the same as in Dave's base algorithm, but there might be a happy medium to be as faithful as possible with the language as well as other solutions that are closer to the `wheel` algorithm.  

thank you again for the code review @rbergen

cdesch:
One Moment. Ran the docker container again and got an error. I'll fix.

cdesch:
All set! Ready to go! Thanks

rbergen (approved):

marghidanu (approved):

PR #326, opened 2021-07-06
Add Kotlin solution

jakobkmar:
## Description
This PR adds a Kotlin implementation of the base algorithm.

There are two test outputs, one using coroutines and one running on a single thread.

This solution is using Kotlin 1.5.20 and Java 16.

## Contributing requirements


rbergen:
There already are two PRs aiming to add Kotlin solutions, being #258 and #309. The discussion between those PR's authors on how best to move forward with getting Kotlin added is taking place in the latter of those. I suggest you join that conversation to see if and how your ideas can be integrated in that ongoing effort.

jakobkmar:
Just want to note here that my PR achieves much higher results using coroutines and is even faster in the single threaded test (and I tested on a slower CPU according to the README of the other PRs).

If you want me to join the discussion of the other PRs anyways, I am willing to do so.

rbergen:
Yes, please do. We're trying to limit the number of solutions per language, which means that the conversation is basically inevitable anyway. Furthermore, as things stand, the various PRs would clash on merge.

Theaninova:
Coroutines, lovely! I'll have a look at it tomorrow (and please do join our discussion!)

jakobkmar:
> Coroutines, lovely!

the solution using coroutines is already quite fast, but it doesn't use any of the special coroutines features to make full use of it (I don't know if that would even make any sense here)
will have another look at it tomorrow whether it can be improved

fvbakel:
@bluefireoly great to see your coroutines solution. I think the single threaded solution has a performance in the same order of magnitude than my single threaded solution. I would have to do some more testing to see the difference. However, I did notice on my laptop that the measurements of any solution have a standard deviation of about 5% and a worst case difference up to 40%!. 

So you would have to run both solution 100 times on the same machine  and take the median of the measurement if you want to measure small differences. 

Next to this I was wondering the impact of the target run time environment that we use. I notice that you are using at least a different environment. I think that should also be the same in a compare test. Eg, we should both use the fastest. I did not really consider that, I just took the one that is the same as the Java implementation, to have a fair compare.

jakobkmar:
>  think the single threaded solution has a performance in the same order of magnitude than my single threaded solution.

Yes it is about the same, but I noticed an improvement in speed when
1. using `shl` and `shr`
2. using `BooleanArray` and declaring that true means sieved (instead of false)
3. I haved used things like `for (num in factor until size step 2)` at first, but replaced that with a plain while loop because it is faster

Of course this means that my base algorithm does not use any of the cool Kotlin features, instead it just goes for performance.

> So you would have to run both solution 100 times on the same machine and take the median of the measurement if you want to measure small differences.

Maybe it will be worth a try, yes.

> Next to this I was wondering the impact of the target run time environment that we use.

Yeah I just used the latest one, I think that this is appropriate in order to display the current status of JVM performance.

Another thing about my concurrent solution: it does not depend on my implementation of the algorithm, it can be very well used to execute other implementations concurrently as well.

Theaninova:
> > think the single threaded solution has a performance in the same order of magnitude than my single threaded solution.
> 
> Yes it is about the same, but I noticed an improvement in speed when
> 
> 1. using `shl` and `shr`

Weird, it'd have hoped the JVM would optimize stuff like that, shifting instead of multiplication/division is one of the most well known optimizations.
> 2. using `BooleanArray` and declaring that true means sieved (instead of false)
> 3. I haved used things like `for (num in factor until size step 2)` at first, but replaced that with a plain while loop because it is faster

Stupid (but praxis relevant) limitation of the Kotlin range loop. The offender here is the `step` function, which causes Kotlin not to optimize it to a plain for loop but to actually create a range object.

> Another thing about my concurrent solution: it does not depend on my implementation of the algorithm, it can be very well used to execute other implementations concurrently as well.

I think that would be nice, though I wasn't sure if this counts as a parallel implementation...

jakobkmar:
My suggestion would be to use this PR as the coroutines solution and then take the best from the two other PRs for:
- showing some nice Kotlin language features
- going for best performance (my implementation could be the fastest as well but that has to be tested)

This way, we could narrow it down to 3 Kotlin solutions.

Theaninova:
> My suggestion would be to use this PR as the coroutines solution and then take the best from the two other PRs for:
> 
> * showing some nice Kotlin language features
> * going for best performance (my implementation could be the fastest as well but that has to be tested)
> 
> This way, we could narrow it down to 3 Kotlin solutions.

Yup, that sounds good to me, I'm trying to get Kotlin Multiplatform working, right now what's missing is getting an executable jar, but maybe we use the Kotlin CLI compiler in the containers instead.

jakobkmar:
I have removed my single threaded solution, as there are enough other PRs adding Kotlin variants for that.

It seems as if my concurrent approach has been used in #309 in some way as well, is that correct? If that is the case, could we just add it with this PR.

I would suggest to merge this PR as the concurrent JVM coroutines solution and continue working on #309 to add different approaches. This PR really doesn't add that many solutions as the other PRs and therefore no debate is necessary which solution to add and which not.

@wulkanat and @fvbakel would you be okay with that?

I will rename the path for my solution if that leads to any conflicts.

Theaninova:
> I have removed my single threaded solution, as there are enough other PRs adding Kotlin variants for that.
> 
> It seems as if my concurrent approach has been used in #309 in some way as well, is that correct? If that is the case, could we just add it with this PR.

Correct, I used your Coroutines approach to provide Multithreading support to all JVM solutions (you're credited in the README of course), and I included your implementation as the `traditional` variant. It's altered a bit though, to accommodate Kotlin Multiplatform (we only have a single solution now, that includes three implementations, three platforms, and coroutines for JVM)
> I would suggest to merge this PR as the concurrent JVM coroutines solution and continue working on #309 to add different approaches. This PR really doesn't add that many solutions as the other PRs and therefore no debate is necessary which solution to add and which not.

If we merged this one, we'd have to delete it later again as it will become obsolete once #309 is merged. Also thanks to the work of @fvbakel we now have a fully working all-encompassing solution... The only thing that needs to be done now is I have to look over it again and then remove the obsolete solutions, and the code has to be reviewed before it's merged.

jakobkmar:
Ah I see! I will close this PR - and have a closer look at #309.

jakobkmar:
@wulkanat could you give me access to your PR repo as well? I will work on a separate branch if I want to change something.

Theaninova:
Yup, added you

PR #333, opened 2021-07-07
results from IBM compiler:

awierzch:
## Description
This code is based on https://github.com/PlummersSoftwareLLC/Primes/tree/drag-race/PrimeCOBOL/solution_1.

I took the above code and made some minor changes to properly compile and run it on a real mainframe.

## Contributing requirements

### DIFF
--- a/PrimeCOBOL/solution_1.1/JCL
+++ b/PrimeCOBOL/solution_1.1/JCL
@@ -1,0 +1,16 @@
+//PRIMES   JOB 1,NOTIFY=&SYSUID
+//***************************************************/
+//COBRUN  EXEC IGYWCL
+//COBOL.SYSIN  DD DSN=&SYSUID..DGARAGE(PRIMES),DISP=SHR
+//LKED.SYSLMOD DD DSN=&SYSUID..LOAD(PRIMES),DISP=SHR
+//***************************************************/
+// IF RC = 0 THEN
+//***************************************************/
+//RUN     EXEC PGM=PRIMES
+//STEPLIB     DD DSN=&SYSUID..LOAD,DISP=SHR
+//CEEDUMP     DD DUMMY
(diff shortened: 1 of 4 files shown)


rbergen:
As much as I am genuinely charmed by this, I don't think we can merge it into the solution tree. Not mentioning some smaller technicalities, the main problem is that we have no IBM mainframe to run the benchmarks on. We're now at a point where we really can't merge anything new that we can't include in the benchmark. 

May I suggest to open a "show and tell" post in Discussions, and post this there?

rbergen:
Tagging @fvbakel. I imagine you'll like this, too :)

fvbakel:
@rbergen Thanks for mentioning this!

@awierzch Great that you took the effort for the IBM compile, I don't have access to such hardware. I did notice that the time is 50 in your output. Is there a dot missing?

rbergen:
It seems GitHub lacks at least one feature, being the ability to convert a PR into a show and tell Discussions post...
Be that as it may, I will have to close this.

awierzch:
> As much as I am genuinely charmed by this, I don't think we can merge it into the solution tree. Not mentioning some smaller technicalities, the main problem is that we have no IBM mainframe to run the benchmarks on. We're now at a point where we really can't merge anything new that we can't include in the benchmark.
> 

No problem, just wanted to share my results and compare GNU compiler on PC with IBM compiler on mainframe.

awierzch:
> @rbergen Thanks for mentioning this!
> 
> @awierzch Great that you took the effort for the IBM compile, I don't have access to such hardware. I did notice that the time is 50 in your output. Is there a dot missing?

Yes. I was too lazy to fix this.

andrew890:
I have been putting together some specific IBM mainframe z/OS adaptations so I have included the changes in that:
https://github.com/andrew890/Primes-zOS

PR #334, opened 2021-07-07
R - Sieve of Eratosthenes solution 2

nobrien97:
## Description
An implementation of the sieve of Eratosthenes in R, using a functional programming paradigm.

## Contributing requirements


rbergen:
Thanks for this submission. Could you elaborate a little why this should be added as a second solution, instead of a modification of the R solution that is currently in place?

nobrien97:
> 
> 
> Thanks for this submission. Could you elaborate a little why this should be added as a second solution, instead of a modification of the R solution that is currently in place?

Sure!

R is quite a powerful and flexible language, and as such supports many programming styles. Base-R includes 2 styles of OOP for those inclined to use it, but also works as a functional language.  The other solution is written in a (more faithful to the original sieve) OOP style, whereas mine is written in a functional style, which is a more common way of using R. 
So basically, compared to solution 1, this solution is less faithful to the original sieve C++ code, but more faithful to the average programmer's experience of using R. I think together, the solutions show the breadth of options that R gives to its users.

rbergen:
Thanks for your explanation. The issue, if that's not too strong a word, is that from a benchmark characteristic respective both implementations are identical (base algorithm, not faithful, single-threaded, 32 bits per prime). 
To keep things manageable, we're trying to limit the number of solutions per language with one set characteristics, to 1.

I therefore see two possibilities:

1. You modify your PR to apply only the _drag-race relevant performance improvements_ in your implementation to the existing solution.
2. @fvbakel agrees to have his existing solution replaced by yours.

fvbakel:
@rbergen could we move my solution to the "atic", like I did for the SQL solution? Then the code difference can still be shown, without impact on the make runtime.

I choose my method because it is more faithfuI. I think the too many solutions can be solved by the atic like construction. I have more ideas on this that I was planning to add to the ideas section.

rbergen:
@fvbakel If @nobrien97 and yourself can work together towards one PR to replace the current solution_1, then I am fine with considering one that includes an "attic" construct with code of the current implementation. I will still have to insist that the drag-race relevant part of the code of the new actual solution_1 must perform better than the current one.

I am not (yet) comfortable with committing to the attic concept in a general sense. For that, I think a discussion does need to be had first, and some rules need to be drafted on when and how it can and should not be used. I really want to prevent a situation where a significant part of the code in this repo is there for "illustrative" purposes only.

For completeness: if, in your collaboration, the conclusion is that one of the key characteristics of the new implementation is actually different than those of the current one, then adding a second solution would be possible.

fvbakel:
Today I had some time to have a closer look at the code of this PR. I noticed that solution_1 runs twice as fast on the docker image of solution_2. Not sure what the cause is of this difference.

@rbergen 
> from a benchmark characteristic respective both implementations are identical (base algorithm, not faithful, single-threaded, 32 bits per prime).

I think that this is not true, the existing solution 1 is marked as faithful and the new is not. So from benchmark perspective they are different. I think we should keep both. Further more I am not even sure if the solution_2  use the base algorithm because it is not searching for the next prime. See details below

@nobrien97 Great to see your solution and the big difference in performance. The big difference in performance comes from the difference in the loop` while (fac <= q) `. You are just not searching for the next prime, as stated in the official algorithm, but just cross out all multiples of all odd numbers <1000.

By replacing my line to search for the next prime `factor <- factor + min(which(bit_array[(factor + 1) : bit_size]))` with `factor <- factor + 1` and using your Docker image I get a 10% better performance than  solution_2.

I have to think of the best way to continue, but is is interesting for sure.

Edits:

- comment out, should be replace by, 
- 2 times better is actually 10% better :-)

rbergen:
> I think that this is not true, the existing solution 1 is marked as faithful and the new is not.

@fvbakel You're absolutely right. The badge is "faithful-no-green", which put me on the wrong foot.
But, in that case, we don't need to change anything to the PR as it stands. I'll review it tonight in detail, then.

fvbakel:
@rbergen @nobrien97 Is it oke, if I submit an improvement of my solution with the idea's of @nobrien97? Of course I will mention you for that.

rbergen:
@fvbakel I think it would be very courteous of you to credit @nobrien97 for his code inspiring you.

Beyond that, I don't believe any developer has gotten anywhere without borrowing ideas from others, to the point that half my university curriculum consisted of algorithms devised by others :)
Speaking slightly less generally, this whole project is based on the concept of developers copying the idea of someone else, that being using a prime sieve to compare languages. And that idea was actually also thought of by others, before.

rbergen:
> The big difference in performance comes from the difference in the loopwhile (fac <= q). You are just not searching for the next prime, as stated in the official algorithm, but just cross out all multiples of all odd numbers <1000.

I think @fvbakel makes an interesting point here. I've thought about it, and my stance is as follows. The "find next prime" loop is actually meant to prevent running the loop that clears the multiples of a factor. The reason for that is that the latter loop is a very expensive loop in most languages; not having the "find next prime" loop would significantly impact overall performance. 
In the case of R, things seem to be different. It's much quicker to operate on a sequence that's encoded the way a for loop would be in other languages, than it is to run a "traditional" loop.

I therefore think this approach probably is the "R" way to go about it, and as such illustrates a very interesting difference between R and many other languages.
I'm confident @marghidanu will develop his own opinion when he does his review, and I am very interested in learning what that is.

PrimeR/solution_2/README.md
```diff
@@ -0,0 +1,58 @@
+# Sieve of Eratosthenes implemented in (functional) R
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-no-green)
```
rbergen: The badge colour to use for solutions that are not faithful is -yellowgreen instead of -green.

fvbakel:
Just one small update, in the meantime I found the fastest way to find the next prime in R. It is like this, based on solution_1:

```R
            while(factor <= maxroot_index){
                if (bit_array[factor] == TRUE) {
                    # cross out all factors
                    prime <- factor * 2 +1
                    start <- ( ((prime * prime) -1) /2 )
                    step <- factor * 2 + 1
                    
                    # propagate
                    bit_array[seq.int(from=start, to=bit_size, by=step)] <- FALSE
                } 
                factor <- factor + 1
            }
```
Implemented all of the above in solution_1 in PR #357.

Edit:

Friday = now

PrimeR/solution_2/README.md
nobrien97: Apologies for the confusion - I've updated the badge colour.

nobrien97:
> 
> 
> ```r
>             while(factor <= maxroot_index){
>                 if (bit_array[factor] == TRUE) {
>                     # cross out all factors
>                     prime <- factor * 2 +1
>                     start <- ( ((prime * prime) -1) /2 )
>                     step <- factor * 2 + 1
>                     
>                     # propagate
>                     bit_array[seq.int(from=start, to=bit_size, by=step)] <- FALSE
>                 } 
>                 factor <- factor + 1
>             }
> ```

Fantastic! This is really cool - R truly is unique. Well done on improving solution 1 @fvbakel!

rbergen:
@nobrien97 @fvbakel Can I consider this PR stable and ready for review?

fvbakel:
I think that this code would run faster if it used the same tweaks as solution_1.

Furthermore I think that the memory for the Bits vector is only allocated in in the first run. The second reuses the memory of the first run. I think that violates the rules, but I don't know how to avoid it in this solution. In solution_1 the memory is allocated in each run because a new class instance is created. 

@rbergen 
I am not sure if the above is an issue for an unfaithful implementation.

rbergen:
> I am not sure if the above is an issue for an unfaithful implementation.

No, it's not an issue for an unfaithful implementation, it just means it's unfaithful in this way (too). Unfaithful is defined as "everything that's not faithful but still sticks to the basic rules". That encapsulates a lot, including this.

fvbakel:
@nobrien97 How do you want to continue? Are you willing to implement the tweaks of solution_1 on top of this implementation? I suspect it becomes faster than solution_1 if you do. Or do you want to keep it the way it is now because it is the typical R way of doing things?

nobrien97:
> 
> 
> @nobrien97 How do you want to continue? Are you willing to implement the tweaks of solution_1 on top of this implementation? I suspect it becomes faster than solution_1 if you do. Or do you want to keep it the way it is now because it is the typical R way of doing things?

Apologies for the late response: I've implemented the changes from solution 1 into solution 2, with a few small tweaks as well: I replaced the while loop in `runSieve()` with a for loop to reduce condition checking, and replaced `/2` with `*0.5`. These are super minimal optimisations that don't do much, but might eke out a tiny amount of extra performance. Your intuition was correct @fvbakel, on my machine the current version of solution 1 gets ~508 passes whereas solution 2 with your `runSieve()` implementation gets ~670. I've pushed a commit including this change, and I think it's ready for review @rbergen.

fvbakel:
@nobrien97 thanks for the feedback. To be honest, my suspicion was based on a quick check with a global vector like you have.
I will add you ideas to mine, then the two will be closer, but yours will stil be faster I guess.

fvbakel:
@nobrien97 Just for your information an interest:

I implemented your  tweaks in solution_1.  I could not really measure a big improvement in performance, with the ` * 0.5` and `for in` instead of `while`. It had a `~0.5%` improvement over 10 runs.

However, I did find that assigning in place the FALSE value to the member variable of a refClass is slow. I found that if I declare the bit_array as ANY then R performs less checks on assignments. This change had a ~20% performance improvement in solution_1. This declaration improvement does not apply to this solution, because there is no declaration needed . See PR #386 for the full details. I get `~590` with this solution_2 and `~530` with the improved solution_1.

rbergen (approved):
LGTM

marghidanu (approved):

PR #337, opened 2021-07-07
AWK Implementation

DaviNakamuraCardoso:
## Description
Prime Sieve implementation in AWK

## Contributing requirements


PrimeAWK/solution_1/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ubuntu
```
rbergen, line 1: For purposes of longer-term predictability of behavior, I think it's better to pin on a specific (supported LTS) version of ubuntu.

PrimeAWK/solution_1/README.md
```diff
@@ -0,0 +1,45 @@
+# AWK Solution
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
```
rbergen: What is the bit size of the prime candidates within this solution?

PrimeAWK/solution_1/README.md
```diff
@@ -0,0 +1,45 @@
+# AWK Solution
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
```
rbergen: This solution cannot be marked as faithful. For one, it doesn't use a class-like structure that contains the sieve state. That also means that the solution's drag-race output line needs to be changed accordingly.

PrimeAWK/solution_1/Dockerfile
DaviNakamuraCardoso: Fixed

PrimeAWK/solution_1/README.md
DaviNakamuraCardoso: Fixed

PrimeAWK/solution_1/README.md
DaviNakamuraCardoso: Fixed

PrimeAWK/solution_1/primes.awk
```diff
@@ -0,0 +1,109 @@
...
+
+    if (count != count_primes())
+    {
+        print "Wrong count of primes";
+        exit;
+    }
+
+    print "Passes: " passes ", Time: " duration ", Average: " duration / passes ", Count: " count ", Valid: " validate_results();
+
+    print "DaviNakamuraCardoso;" passes ";" duration ";algorithm=base,faithful=yes";
```
rbergen: It seems the output line still needs to be modified to reflect that the solution is not faithful.

DaviNakamuraCardoso: Done.

PrimeAWK/solution_1/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ubuntu:18.4
```
rbergen: The docker build fails because ubuntu:18.4 is an unrecognized container image tag. Please make sure that the docker build and build run work before you commit.

DaviNakamuraCardoso: It works now.

PrimeAWK/solution_1/README.md
```diff
@@ -0,0 +1,46 @@
+# AWK Solution
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-no-green)
```
rbergen: I missed this before, but the badge color for implementations that are not faithful is -yellowgreen instead of -green

rbergen (approved):
LGTM

marghidanu (approved):

PR #338, opened 2021-07-07
Remove parseFloat in verbose output

icecream17:
I made a whole description in the commit


rbergen:
> I made a whole description in the commit

I noticed. It made the PR list an interesting thing to look at for a few seconds.

The possible performance improvement you mentioned relates to printing the result, which is not part of the benchmarked part of the implementation and as such not relevant. Even if it was, it does not warrant the other changes, which are all cosmetic.  

I think the right decision here is to close this.

PR #339, opened 2021-07-07
Optimizations for PrimeCPP

supsm:
## Description

adds a few optimizations which improve performance (13% faster on my machine)  
not sure why the diff is messed up, I might've accidentally changed some indentation  
**IMPORTANT:** I have not applied these optimizations to multithreaded since it doesn't seem to perform as well. I have also decided not to mess with constexpr solution.

## Contributing requirements

* [ ] I added a README.md with the right badge(s). **(does not apply)**
* [ ] I added a Dockerfile that builds and runs my solution. **(does not apply)**

### DIFF
--- a/PrimeCPP/solution_1/PrimeCPP.cpp
+++ b/PrimeCPP/solution_1/PrimeCPP.cpp
@@ -16,118 +16,107 @@ using namespace std::chrono;
 
 class prime_sieve
 {
-private:
+  private:
 
-    long sieveSize = 0, count = -1;
-    vector<bool> Bits;
-    const std::map<const long long, const int> resultsDictionary =
-    {
-          {          10LL, 4         },               // Historical data for validating our results - the number of primes
-          {         100LL, 25        },               // to be found under some limit, such as 168 primes under 1000
-          {        1000LL, 168       },
-          {       10000LL, 1229      },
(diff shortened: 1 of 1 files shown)


rbergen:
For some reason (line ends? whitespacing?) the diff makes it look like the whole file has changed, which it clearly hasn't. Please modify your commit so that the diff only shows the actual changes in code.

supsm:
It should be fixed now, can you check again?

rbergen:
It is fixed. And now it is, I can see that your algorithm effectively skips:
- checking if 5 is a prime
- establishing it is a prime, and 
- clearing multiples of 5

All of this effectively means the algorithm treats 5 as a known prime. That's a deviation of the base algorithm, which makes it's not the base algorithm anymore. 

Based on this, we can't merge this as an optimization to the existing solution.

supsm:
Why would this be any different from skipping multiples of 2?

rbergen:
Because 2 is the _only_ allowed exception. Dave has been very explicit about this up to this point, [including yesterday](https://github.com/PlummersSoftwareLLC/Primes/discussions/318#discussioncomment-972240), and will be again, soon.

supsm:
Alright, thanks!

PR #340, opened 2021-07-07
Sped up the octave implementation

AdamCooman:
## Description

The old implementation first created a vector of indices and then performed a sum and divide by two on that vector to compute the indices in the `raw_bits` vector. This was taking a lot of time. By avoiding a call into the function `clear_bits` and computing the indices in the original vector directly, I obtained a significant speedup. 
I replaced the raw_bits vector by a logical vector, as it seems more appropriate to use that here.
I also removed the get_bit function because it was performing a check that was unneeded.

Before this change, I obtained around 400 Passes. The new implementation has around 1500 Passes.

## Contributing requirements

* [ x ] I read the contribution guidelines in CONTRIBUTING.md.
* [ x ] I placed my solution in the correct solution folder.
* [ existed already ] I added a README.md with the right badge(s).
* [ existed already ] I added a Dockerfile that builds and runs my solution.
* [ x ] I selected `drag-race` as the target branch.
* [ x ] All code herein is licensed compatible with BSD-3.

### DIFF
--- a/PrimeOctave/solution_1/run_sieve.m
+++ b/PrimeOctave/solution_1/run_sieve.m
@@ -10,33 +10,14 @@ function count=run_sieve(limit)
   % Only store information about odd numbers, as all even numbers are
   % automatically prime
   % rbergen: wrapped 2 in int32() to fix idivide integer input error in Octave 6.2.0
-  raw_bits = ones(1,idivide(limit,int32(2)));
+  raw_bits = true(1,idivide(limit,int32(2)));
   
-  % Check if 'index' has been eliminated as non-prime yet. Returns 1 if
-  % it has not yet been eliminated (still considered a prime candidate)
-  function bit = get_bit(index)
-    if rem(index,2) == 0
-      % If multiple of 2, automatically return false and warn the user 
-      % that he is checking even numbers unnecessarily
-      'checking even bits!'
-      bit = 0;
-      return
-    else
-      bit = raw_bits((index+1)/2);
-      return
-    endif
-  endfunction
-
-  function clear_bits(range)
-    raw_bits((range + 1) / 2) = 0;
-  endfunction
-
   factor = 3;
   q = sqrt(limit);
   while(factor<=q)
     % Find the next prime (look for the first bit in raw_bits that is still set
     for num=factor:2:limit
-      if get_bit(num)
+      if raw_bits((num+1)/2)
         factor = num;
         break
       endif
@@ -44,8 +25,7 @@ function count=run_sieve(limit)
 
     % Mark all odd multiples of 'factor' as non-prime
     % Even multiples are already known to be non-prime
-
-    clear_bits(factor*3:factor*2:limit);
+	raw_bits((factor*3+1)/2:factor:end) = false;
     
     % Increment 'factor' to the next prime candidate
     factor += 2;


rbergen:
Tagging @fspigel and allowing time for them to review and respond to this PR on their original solution, if they want to do so.

PrimeOctave/solution_1/run_sieve.m
```diff
@@ -10,42 +10,22 @@
...
         factor = num;
         break
       endif
     endfor
 
     % Mark all odd multiples of 'factor' as non-prime
     % Even multiples are already known to be non-prime
-
-    clear_bits(factor*3:factor*2:limit);
+	raw_bits((factor*3+1)/2:factor:end) = false;
```
JohnnySammour, line 28: can you explain this? i tried to execute it and got incorrect results (some prime numbers indices were set to false after the factor = 3 pass) which makes me believe you are trying some optimization
according to [wikipedia](https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes):
>`for i = 2, 3, 4, ..., not exceeding √n do
        if A[i] is true
            for j = i2, i2+i, i2+2i, i2+3i, ..., not exceeding n do
                A[j] := false`

i suggest trying `raw_bits(factor*factor:factor:end) = false;`

tjol: How did you determine that the results were incorrect? This code does the exact same thing as the previous version (the division by 2 was just in function clear_bit before) and generates the correct number of primes.

The array doesn't store even numbers. 3 is in index 2, 5 in index 3, and so on. (Index 1 is unintentionally left blank)

JohnnySammour: > The array doesn't store even numbers. 3 is in index 2, 5 in index 3, and so on

thanks for pointing it out.

fspigel: This solution is faster by a factor of 25, that's rather dramatic. I didn't know abstracting functionality into functions was slowing it down so much. Is this a consequence of Octave being an interpreted language?

In any case this is fine by me, send it.

fspigel (approved):

rbergen:
@fspigel Thank you for responding so quickly. Based on your feedback, I will merge this.

PrimeOctave/solution_1/run_sieve.m
tjol: Yep, with Octave, and MATLAB, and Numpy, many other interpreted languages/environments designed for numerical computing, the general rule of thumb is to do as little as possible in the high-level language and as much as possibly in built-in functions since they're written in C and Fortran. Especially no loops!

I think the difference is especially pronounced with Octave as the interpreter is particularly slow (compared to MATLAB, at least by reputation)

tjol: also, @fspigel, speaking of dramatic, I'm actually not sure if you ever saw the previous PR on the octave solution: #291

JohnGarbarino13 (changes requested):
no go

PrimeOctave/solution_1/run_sieve.m
fspigel: Goodness, no I hadn't. I never got the notification for some reason. That is certainly humbling :)

PR #341, opened 2021-07-07
pythonic updates to python solution 1

JacobCallahan:
This update focuses on making things look more pythonic as well as makes the script more efficient overall.

before:
```
Passes: 55, Time: 10.067714939825237, Avg: 0.18304936254227702, Limit: 1000000, Count: 78498, Valid: True

davepl;55;10.067714939825237;1;algorithm=base,faithful=yes
```

after:
```
Passes: 78, Time: 10.085202750982717, Avg: 0.12929747116644508, Limit: 1000000, Count: 78498, Valid: True

davepl;78;10.085202750982717;1;algorithm=base,faithful=yes
```

## Contributing requirements

### DIFF
--- a/PrimePython/solution_1/PrimePY.py
+++ b/PrimePython/solution_1/PrimePY.py
@@ -1,140 +1,125 @@
-
 # Python Prime Sieve
 #
 # MyFirstPython Program (tm) Dave Plummer 8/9/2018
 #
-# This is the main prime_sieve class. Call it with the number you wish as an upper limit, then
-# call the runSieve method to do the calculation. printResults will dump the count to check validity.
+# This is the main PrimeSieve class. Call it with the number you wish as an upper limit, then
+# call the run_sieve method to do the calculation. print_results will dump the count to check validity.
 #
 # Updated 3/22/2021 for Dave's Garage episode comparing C++, C#, and Python
+# Updated 7/7/2021 to be more pythonic/efficient
 
-from sys import stdout                      # So I can print without an automatic python newline
-from math import sqrt                       # Used by the sieve
-import timeit                               # For timing the durations
+from math import sqrt  # Used by the sieve
+import timeit  # For timing the durations
 
-class prime_sieve(object):
 
-    rawbits = None   # Storage for sieve - since we filter evens, just half as many bits
-    sieveSize = 0    # Upper limit, highest prime we'll consider
+PRIMECOUNTS = {
+    10: 4,  # Historical data for validating our results - the number of primes
+    100: 25,  # to be found under some limit, such as 168 primes under 1000
+    1000: 168,
+    10000: 1229,
+    100000: 9592,
+    1000000: 78498,
+    10000000: 664579,
+    100000000: 5761455,
+}
 
-    primeCounts = { 10 : 4,                 # Historical data for validating our results - the number of primes
-                    100 : 25,               # to be found under some limit, such as 168 primes under 1000
-                    1000 : 168,
-                    10000 : 1229,
-                    100000 : 9592,
-                    1000000 : 78498,
-                    10000000 : 664579,
-                    100000000 : 5761455
-                  }
 
-    def __init__(this, limit):
-        this.sieveSize = limit
-        this.rawbits = [True] * (int((this.sieveSize+1)/2))
+class PrimeSieve(object):
+    # more of a memory optimization, but doesn't hurt to demo!
+    __slots__ = ["sieve_size", "raw_bits"]
 
-    # Look up our count of primes in the historical data (if we have it) to see if it matches
+    def __init__(self, limit):
+        # Upper limit, highest prime we'll consider
+        self.sieve_size = limit
+        # Storage for sieve - since we filter evens, just half as many bits
+        self.raw_bits = [True] * ((self.sieve_size + 1) // 2)
 
-    def validateResults(this):                      # Check to see if this is an upper_limit we can
-        if this.sieveSize in this.primeCounts:      # the data, and (b) our count matches. Since it will return
-            return this.primeCounts[this.sieveSize] == this.countPrimes() # false for an unknown upper_limit, can't assume false == bad
+    def validate_results(self):
+        """Look up our count of primes in the historical data (if we have it) to see if it matches"""
+        # Check to see if this is an upper_limit we can
+        if self.sieve_size in PRIMECOUNTS:
+            # the data, and (b) our count matches.
+            return PRIMECOUNTS[self.sieve_size] == self.count_primes()
+        # Since it will return false for an unknown upper_limit, can't assume false == bad
         return False
 
-    # GetBit
-    # 
-    # Gets a bit from the array of bits, but automatically just filters out even numbers as false,
-    # and then only uses half as many bits for actual storage
-
-    def GetBit(this, index):
-
-        if (index % 2 == 0): # even numbers are automaticallty returned as non-prime
-            return False
+    def get_bit(self, index):
+        """Gets a bit from the array of bits, but automatically just filters out even numbers as false,
+        and then only uses half as many bits for actual storage
+        """
+        if index % 2 != 0:  # even numbers are automaticallty returned as non-prime
+            return self.raw_bits[index // 2]
         else:
-            return this.rawbits[int(index/2)]
+            return
 
-    # ClearBit
-    #
-    # Reciprocal of GetBit, ignores even numbers and just stores the odds. Since the prime sieve work should
-    # never waste time clearing even numbers, this code will assert if you try to
-
-    def ClearBit(this, index):
-
-        if (index % 2 == 0):
-            assert("If you're setting even bits, you're sub-optimal for some reason!")
-            return False
+    def clear_bit(self, index):
+        """Reciprocal of get_bit, ignores even numbers and just stores the odds. Since the prime sieve work should
+        never waste time clearing even numbers
+        """
+        if index % 2 != 0:
+            self.raw_bits[index // 2] = False
         else:
-            this.rawbits[int(index/2)] = False
+            return
 
-    # primeSieve
-    # 
-    # Calculate the primes up to the specified limit
-
-    def runSieve(this):
-
+    def run_sieve(self):
+        """Calculate the primes up to the specified limit"""
         factor = 3
-        q = sqrt(this.sieveSize)
+        q = sqrt(self.sieve_size)
 
-        while (factor < q):
-            for num in range (factor, this.sieveSize):
-                if (this.GetBit(num) == True):
+        while factor < q:
+            for num in range(factor, self.sieve_size):
+                if self.get_bit(num):
                     factor = num
                     break
 
             # If marking factor 3, you wouldn't mark 6 (it's a mult of 2) so start with the 3rd instance of this factor's multiple.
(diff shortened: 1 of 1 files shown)


rbergen:
The solution you're seeking to modify is one of Dave's original entries in the drag-race. Because of this, we're not going to merge cosmetic and coding style changes to it. Because of this, I'm going to ask if you can either:
- modify your PR to only include the code changes that improve performance, then we can consider those;
- explain how the changes you now committed realize the performance improvement you claim.

JacobCallahan:
@rbergen I can add some comments as to why specific changes were made especially those that were changed for performance reasons. However, I challenge the historical style rigidity for these reasons:

1. I have not altered the original algorithm.
2. The style changes reflect what is most common in python projects.
3. Dave mentioned that he dabbled in Python, which is why the style is more reminiscent of other languages.
4. More modern style changes (f-strings) are actually more performant.
5. Dave asked for people to update the examples he came up with to improve them.
6. This is a better/clearer representation of the language for those unfamiliar with it.

PrimePython/solution_1/PrimePY.py
```diff
@@ -1,140 +1,125 @@
...
 # MyFirstPython Program (tm) Dave Plummer 8/9/2018
 #
-# This is the main prime_sieve class. Call it with the number you wish as an upper limit, then
-# call the runSieve method to do the calculation. printResults will dump the count to check validity.
+# This is the main PrimeSieve class. Call it with the number you wish as an upper limit, then
+# call the run_sieve method to do the calculation. print_results will dump the count to check validity.
 #
 # Updated 3/22/2021 for Dave's Garage episode comparing C++, C#, and Python
-
-from sys import stdout                      # So I can print without an automatic python newline
```
JacobCallahan, line 11: unnecessary as `print` can take a `sep` argument to suppress newline characters. This also eliminates an import.

PrimePython/solution_1/PrimePY.py
```diff
@@ -1,140 +1,125 @@
...
-    def validateResults(this):                      # Check to see if this is an upper_limit we can
-        if this.sieveSize in this.primeCounts:      # the data, and (b) our count matches. Since it will return
-            return this.primeCounts[this.sieveSize] == this.countPrimes() # false for an unknown upper_limit, can't assume false == bad
+# Updated 7/7/2021 to be more pythonic/efficient
+
+from math import sqrt  # Used by the sieve
+import timeit  # For timing the durations
+
+
+PRIMECOUNTS = {
```
JacobCallahan, line 15: primecounts didn't need to be nested under the class and is treated as a constant. This change better reflects that.

PrimePython/solution_1/PrimePY.py
```diff
@@ -1,140 +1,125 @@
...
+    1000: 168,
+    10000: 1229,
+    100000: 9592,
+    1000000: 78498,
+    10000000: 664579,
+    100000000: 5761455,
+}
+
+
+class PrimeSieve(object):
```
JacobCallahan, line 27: This capitalization pattern is more common in python. Also, `object` doesn't need to be explicit anymore, but I left it. could be removed.
```suggestion
class PrimeSieve:
```

PrimePython/solution_1/PrimePY.py
```diff
@@ -1,140 +1,125 @@
...
+    # more of a memory optimization, but doesn't hurt to demo!
+    __slots__ = ["sieve_size", "raw_bits"]
+
+    def __init__(self, limit):
```
JacobCallahan, line 31: `self` is almost always used instead of `this` unlike other languages. Technically any valid name could be used here, but this is the most common standard.

PrimePython/solution_1/PrimePY.py
```diff
@@ -1,140 +1,125 @@
...
+        # Upper limit, highest prime we'll consider
+        self.sieve_size = limit
+        # Storage for sieve - since we filter evens, just half as many bits
+        self.raw_bits = [True] * ((self.sieve_size + 1) // 2)
```
JacobCallahan, line 35: using a floordiv instead of a truediv and explicit conversion back to an integer is 10x faster.

PrimePython/solution_1/PrimePY.py
```diff
@@ -1,140 +1,125 @@
...
+
+    def validate_results(self):
+        """Look up our count of primes in the historical data (if we have it) to see if it matches"""
```
JacobCallahan, line 38: docstrings are used in python instead of block comments above the function/method definition

PrimePython/solution_1/PrimePY.py
```diff
@@ -1,140 +1,125 @@
...
-
-    # ClearBit
-    #
-    # Reciprocal of GetBit, ignores even numbers and just stores the odds. Since the prime sieve work should
-    # never waste time clearing even numbers, this code will assert if you try to
-
-    def ClearBit(this, index):
-
-        if (index % 2 == 0):
-            assert("If you're setting even bits, you're sub-optimal for some reason!")
```
JacobCallahan, line 61: This assertion does nothing. Assertions in python are checks that the following statement returns a "truthy" result. If not it will return an AssertionError. This will always assert True

PrimePython/solution_1/PrimePY.py
```diff
@@ -1,140 +1,125 @@
...
-            return False
+            return
+
+    def clear_bit(self, index):
+        """Reciprocal of get_bit, ignores even numbers and just stores the odds. Since the prime sieve work should
+        never waste time clearing even numbers
+        """
+        if index % 2 != 0:
+            self.raw_bits[index // 2] = False
```
tjol, line 60: This is where the performance improvement comes from: replacing the call to `int` with a much cheaper integer division.

(As for the rest, it's just a question of style and therefore shouldn't be merged)

PrimePython/solution_1/PrimePY.py
```diff
@@ -1,140 +1,125 @@
...
-
-    def printResults(this, showResults, duration, passes):
-
-        if (showResults): # Since we auto-filter evens, we have to special case the number 2 which is prime
-            stdout.write("2, ");
+    def print_results(self, showResults, duration, passes):
+        """Displays the primes found (or just the total count, depending on what you ask for)"""
+        if showResults:
+            # Since we auto-filter evens, we have to special case the number 2 which is prime
+            print("2, ", sep="")
```
JacobCallahan, line 93: an empty string passed into `sep` overrides the default newline character

PrimePython/solution_1/PrimePY.py
```diff
@@ -1,140 +1,125 @@
...
-                    stdout.write(str(num) +", ")
-                count+=1
-
-        assert(count == this.countPrimes())
-        stdout.write("\n");
-        print("Passes: " + str(passes) + ", Time: " + str(duration) + ", Avg: " + str(duration/passes) + ", Limit: " + str(this.sieveSize) + ", Count: " + str(count) + ", Valid: " + str(this.validateResults()))
-
+        for num in range(3, self.sieve_size):
+            # Count (and optionally dump) the primes that were found below the limit
+            if self.get_bit(num):
```
JacobCallahan, line 98: python allows for "truthy" checks. this is almost 50% faster than checking if the value is equal to True explicitly.

PrimePython/solution_1/PrimePY.py
```diff
@@ -1,140 +1,125 @@
...
+                count += 1
+
+        assert count == self.count_primes()
         # Following 2 lines added by rbergen to conform to drag race output format
-        stdout.write("\n");
-        print("davepl;" + str(passes) + ";" + str(duration) + ";1;algorithm=base,faithful=yes");
+        print(
+            f"\nPasses: {passes}, Time: {duration}, Avg: {duration/passes}, Limit: {self.sieve_size}, Count: {count}, Valid: {self.validate_results()}"
+        )
+        print(f"\ndavepl;{passes};{duration};1;algorithm=base,faithful=yes")
```
JacobCallahan, line 108: using f-string here makes the line more readable and is also faster than all the explicit strong conversion and concatenation calls.

PrimePython/solution_1/PrimePY.py
JacobCallahan: That alone added an additional 20 passes. However, the remaining 3 passes came from the other changes.

rbergen:
Thank you for your explanation.

> I challenge the historical style rigidity for these reasons:

That's fair enough. However, this "historical style rigidity" has been applied consistently, because when Dave said he was looking for improvements to existing solutions, he meant he was looking for improvements _to the drag-race performance of_ those existing solutions. 

Based on your explanation, my conclusion is that the vast majority of the changes in your PR are style changes or not relevant for the benchmark's result. Based on that, I will have to ask to limit your changes to those that improve the performance of the actual sieve implementation, for me to be able to recommend merging this PR.

As it stands, this is one of Dave's original solutions, so he would have to agree to any changes anyway. Would you like me to ask for his feedback on the current PR? Of course, in that case I will be transparent about my recommendation.

JacobCallahan:
@rbergen 
> That's fair enough. However, this "historical style rigidity" has been applied consistently, because when Dave said he was looking for improvements to existing solutions, he meant he was looking for improvements _to the drag-race performance of_ those existing solutions.

Also fair. While I do not believe my style changes fall outside the scope of what is requested, you are a maintainer so ultimately if you're not comfortable with the changes then I can close this. I would rather not separate the style changes from the performance changes, as that would be an implicit acceptance of the current style.

> Based on your explanation, my conclusion is that the vast majority of the changes in your PR are style changes or not relevant for the benchmark's result. Based on that, I will have to ask to limit your changes to those that improve the performance of the actual sieve implementation, for me to be able to recommend merging this PR.

If numbers alone were presented in the video and not a review of the language itself, I wouldn't be as concerned. However, as a would-be python educator, I'd prefer viewers to see a more faithful representation of the language style/use.

> As it stands, this is one of Dave's original solutions, so he would have to agree to any changes anyway. Would you like me to ask for his feedback on the current PR? Of course, in that case, I will be transparent about my recommendation.

If you wouldn't mind, would you reach out? Regardless of the decision you two make on accepting/rejecting this pr, he may at least be interested in seeing how someone more familiar with the language would change up his version.
Thanks!

davepl:
Without regard to whether its an improvement or not, because I'm sure it is, one reality I had to accept in my career was that you always wrote code that looked like the codebase you were working with; your code should follow the style of the code you're working with; you should not modify their bracing or indentation of variable naming or commenting scheme, etc.  When you're done it should look like the original author did the work, and not that someone imposed their own design sensibilities on it... unless that person is taking complete ownership of the deliverable.

I learned this going back and forth between Cairo (which had an updated coding standard that I liked) and Windows GDI, which was older, and User32, which was older still.  All three are very different, and you work in the style you find them in.  I know, because I once tried to "improve" one of them and got my hand slapped!

This is all a long-winded way of saying _any changes to this project should be for performance only and should be the minimum amount of change possible to achieve that perf gain_.

rbergen:
@JacobCallahan I'll leave to you how you want to proceed with this PR, based on Dave's input.

JacobCallahan:
@davepl @rbergen I'll respect your decisions and thank you for your time. Closing this as I'd rather not divorce the performance changes from what I think is truly valuable.

PR #342, opened 2021-07-07
Performance improvement for Ada solution_1

albertklee:
## Description
Adjusted the algorithm to match the current Python implementation,
and enabled compiler optimization.

## Contributing requirements


rbergen:
We already have #299 open that aims to improve the performance of the Ada solution, that the original author is already engaged in. Could you please discuss in that PR how to add your changes to it?

albertklee:
You know, you mentioned @MichaelAllenHardeman who has put in more work than the small changes I made here, and that "Num := Factor;" line that we both added is really the most important.  I'm happy to drop this pull request.

PR #343, opened 2021-07-07
optimize V solution_1

Ekopalypse:
@marghidanu 
Do not add a second solution, but improve the existing solution_1

### DIFF
--- a/PrimeV/solution_1/primes.v
+++ b/PrimeV/solution_1/primes.v
@@ -1,6 +1,12 @@
 import time
 import math
 
+const (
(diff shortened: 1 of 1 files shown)


marghidanu:
Looks good to me!

marghidanu:
@Ekopalypse how much does this improve the performance?

Ekopalypse:
@marghidanu Roughly around 40%, calculated by taking the average of 100 local runs, no docker.

marghidanu:
@rbergen LGTM

PR #347, opened 2021-07-08
Ada version with build optimizations enabled instead of debug build

heharkon:
## Description
The original solution was with default settings, i.e. a debug build. This one is the same with optimization flags enabled. 

## Contributing requirements

* [ x ] I read the contribution guidelines in CONTRIBUTING.md.
* [ x ] I placed my solution in the correct solution folder.
* [ x ] I added a README.md with the right badge(s).
* [ x ] I added a Dockerfile that builds and runs my solution.
* [ x ] I selected `drag-race` as the target branch.
* [ x ] All code herein is licensed compatible with BSD-3.

### DIFF
--- a/PrimeAda/solution_2/Dockerfile
+++ b/PrimeAda/solution_2/Dockerfile
@@ -1,0 +1,17 @@
+FROM ubuntu:18.04 AS build
+
+RUN apt-get update -qq \
+    && apt-get install -y gprbuild gnat
+
+COPY . /opt/app/
+WORKDIR /opt/app
+RUN mkdir obj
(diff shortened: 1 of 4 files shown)


onox:
@heharkon The implementation of solution_2 is identical to solution_1, so why not just simply add the missing compiler flags to solution_1?

Besides `-O3` you also might want to add `-march=native`, which gives another speed boost.

rbergen:
@heharkon We already have #299 open to improve the speed of the Ada solution. Please discuss in that PR how your optimizations can be added to that PR.

heharkon:
Ok. Better to continue there, let's close this one.

PR #348, opened 2021-07-08
New benchmark command

marghidanu:
Moved all the logic from the Makefile into a command inside the **tools** project. This new command is building the Docker image, running the container, parsing results, and generating output all in one. Additionally, I've extended the syntax to run all benchmarks, all solutions within an implementation, or a single solution.

Just typing make will build and run all solutions from all implementations:

```
make 
```

If you just want to run all solutions for the Zig implementation you can do something like this:

```
make DIRECTORY=PrimeZig/
```

Or if you want to run just one solution for Crystal you can do this:

```
make DIRECTORY=PrimeCrystal/solution_1/
```

The temporary files are not created anymore since all the data is now aggregated in memory. By moving the logic from the Makefile into TypeScript we have a little more flexibility. Also, the Makefile has been reduced by a lot and there's no need for the `make one` target anymore since the default one can do everything now.


marghidanu:
@rbergen this is still WIP. Would love to hear your feedback!

fvbakel:
@rbergen @marghidanu Great improvements! Would it also be possible to run the solutions based on an inputfile? For example, I make a file like this:

 ```
PrimeJava/solution_1
PrimeCSharp/
```

And then run ```make FILE=myfile.txt ``` to run the specified solutions. This would allow to run any combination of solutions. For example all garbage collecting languages.

**Comment by rbergen** [comment]
@marghidanu I see you dropped support for the arch-* flag files. That's fine in a WIP version, and I'm happy to (re)add it myself later.

Other than that: I absolutely love where this is going. The execution state and progress overview is so much clearer than it was that it looks (acts) like a completely different tool, indeed.

**Comment by marghidanu** [comment]
@rbergen I will add those back in, this is still under development, just wanted to quickly validate the idea so far. We will support all existing features.

@fvbakel what a great idea, will search for a way to implement it nicely. Maybe also support for multiple directories.

I also have in store a few new formatters for output. And the possibility to save the output to file.

**Comment by marghidanu** [comment]
@rbergen I've removed the WIP flag, I think if everything runs well under WSL we should be good to go.

**Review by rbergen — APPROVED** [review]
Works like a charm on WSL2/Ubuntu 20.04. Did I already mention I love how clean this iteration's output is? :)

**Comment by marghidanu** [comment]
Thanks @rbergen! If it’s ok with you I will merge this.

**Comment by rbergen** [comment]
Hang on. I just noticed some messages with some solutions:
```
[PrimeCPP][solution_2] Building ...
[PrimeCPP][solution_2] Running ...
Error running solution!
[PrimeCPP][solution_3] Building ...
[PrimeCPP][solution_3] Running ...
Error running solution!
[PrimeCrystal][solution_1] Building ...
```

What's the best way to investigate what's behind those?

marghidanu:
On it, let me check if I get errors and add some debug logs

rbergen:
Do let me know when I can help with testing again.

marghidanu:
@rbergen For some reason the CPP solutions exit with code 162! And that's because the main function returns the number of primes found(?!?) The thing is I use the return code to determine if the execution of the Docker container was correct or not, maybe I should discard this condition.

P.S.: I forsee a new rule!

rbergen:
@marghidanu That is an interesting finding, indeed...

> P.S.: I forsee a new rule!

Hang on, I'm still waiting on [the most recent clarification](https://github.com/PlummersSoftwareLLC/Primes/pull/398) being approved for merge! ;)

marghidanu:
Ok, one thing I could do is to detect if the container has exited with errors, signal that but attempt to parse the output anyway. The error message will change to something like: "Exited with abnormal code, results might be partial. Error code: xxx"

rbergen:
I think that's a great idea. It will process output that can still be parsed, while providing some pointer to the nature of the problem with a solution where there really is one (to the solution's author, at least).

marghidanu:
Right, I already have this in place, I will run a few tests and ping you when I'm done.

marghidanu:
@rbergen, I might have something functional

rbergen:
It took a little time to test it (these benchmark runs do take some time with this many solutions :) ), but works like a charm for me, now. I'd say: merge at will.

PR #351, opened 2021-07-08
Common lisp wheel

mayerrobert:
## Description

This adds PrimeLisp/solution_3 which is an approx. 2.4x speedup over solution_2. It implements the wheel algorithm and is a port of PrimeC/solution_2/sieve_5760of30030_only_write_read_bits.c by Daniel Spangberg.

## Contributing requirements


rbergen:
As you are also the author of solution_2, I would like to ask you to combine both implementations in one solution, that being a new version of solution_2. Akin to the way PrimeC/solution_2 also includes a range of different implementations, so to say.

mayerrobert:
Makes sense, will do that.

mayerrobert:
I have moved and merged the files from solution_3 to solution_2 and removed the now empty directory solution_3.

Also I have rebased the branch after your merging of PR #350 and it seems that now there are no conflicts.

PrimeLisp/solution_2/PrimeSieveWheel.lisp
```diff
@@ -0,0 +1,356 @@
...
+  (do () ((>= (get-internal-real-time) end))
+    (setq result (create-sieve 1000000))
+    (run-sieve result +steps+)
+    (incf passes))
+
+  (let* ((duration  (/ (- (get-internal-real-time) start) internal-time-units-per-second))
+         (avg (/ duration passes)))
+    (format *error-output* "Passes: ~d  Time: ~f Avg: ~f ms Count: ~d~%" passes duration (* 1000 avg)  (count-primes result))
+
+    (format t "mayerrobert-cl-wheel;~d;~f;1;algorithm=wheel,faithful=no,bits=2~%" passes duration)))
```
rbergen: How does this implementation use 2 bits per prime?

mayerrobert: Kinda sorta. It uses an array with element datatype fixnum (which uses 64 bits of memory but only 62 bits can be used by the programmer) and from these 64 bits only 32 bits are used to stick in bits using AND and OR. I tried to describe that in the README.

So to store 32bits it uses 64bits of memory and so on and so forth.

mayerrobert: I have since changed the array from fixnums to (unsigned-byte 32)s which means bits=1. also this gives another 10% speedup. should I push this latest change to the current PR?

rbergen: Please push whatever you want to end up in the merged solution, and then add a comment to indicate the PR is stable. It'll save PRs, review cycles and thus quite a bit of time :)

mayerrobert: Ok, I'm done, I declare the PR as stable ;-)

rbergen (approved):
LGTM

marghidanu (approved):

PR #352, opened 2021-07-08
Speedup of PrimePerl/solution_1. Same algorithm.

kjetillll:
## Contributing requirements

### DIFF
--- a/PrimePerl/solution_2/primes.pl
+++ b/PrimePerl/solution_2/primes.pl
@@ -1,0 +1,97 @@
+#!/usr/bin/env perl
+
+use strict;
+use warnings;
+
+package PrimeSieve {
+    my %DICT = (
+        10          => 4,
+        100         => 25,
+        1000        => 168,
+        10000       => 1229,
+        100000      => 9592,
+        1000000     => 78498,
+        10000000    => 664579,
+        100000000   => 5761455,
+        1000000000  => 50847534,
+        10000000000 => 455052511,
+    );
+
+    sub new {
+        my ( $class, $sieve_size ) = @_;
+	bless {
+            sieve_size => $sieve_size,
+	    bits       => '0' x $sieve_size,
+        }, $class;
+    }
+
+    sub run_sieve {
+        my $self = shift;
+
+        my $factor = 3;
+	my $bits   = \$self->{bits};
+	my $size   = $self->{sieve_size};
+        my $q      = sqrt $size;
+
+        while ( $factor <= $q ) {
+            my $num = $factor;
+	    while ( $num < $size ) {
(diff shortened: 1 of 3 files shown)


PrimePerl/solution_2/primes.pl
```diff
@@ -0,0 +1,97 @@
...
+                substr $$bits, $num2, 1, '1';
+                $num2 += $factor2;
+            }
+            $factor += 2;
+        }
+    }
+
+    sub primes {
+        my $self = shift;
+	my $bits=\ $self->{bits};
```
marghidanu: Why the double reference? I see you're dereferencing it immediately in the next function.

PrimePerl/solution_2/primes.pl
```diff
@@ -0,0 +1,97 @@
...
+        100000000   => 5761455,
+        1000000000  => 50847534,
+        10000000000 => 455052511,
+    );
+
+    sub new {
+        my ( $class, $sieve_size ) = @_;
+	bless {
+            sieve_size => $sieve_size,
+	    bits       => '0' x $sieve_size,
```
marghidanu: Strings? :) That's interesting...

PrimePerl/solution_2/primes.pl
kjetillll: So not to copy it. It's a long string, a million chars in the default case. Saves doing lots of hash-lookups on 'bits'. Even if it's just a two keys hash, hash-lookups can be slower than dereferencing a string-ref I think.

PrimePerl/solution_2/primes.pl
kjetillll: Perl is good at strings, got a speedup from it.

bbrtj:
From ~50 to ~440 passes on my machine. Good job!

rbergen:
@marghidanu What's your opinion on this PR, as it stands?

marghidanu:
It's ok, it's basically an optimization for dereferencing variables.

rbergen:
Do you agree this should be added as a second solution, or is this an optimization that should be applied to the existing solution_1?

marghidanu:
It's ok in solution_1, nothing else has changed.

bbrtj:
@marghidanu are you sure you're looking at the right PR? This one is about using string as a storage for sieve data, not about dereferencing. I think you might've confused it with the other PR that is now merged.

marghidanu:
My bad, I was talking about the other one :). Sorry about that, this one is ok too. Actually, I'm quite surprised this is so fast just by using a string as storage.

rbergen:
So this should be applied to solution_1 then?

marghidanu:
No, let's keep it as solution_2

PR #355, opened 2021-07-08
Pascal bitpacked implementation

circular17:
## Description

Bitpacked implementation for Pascal base algorithm

## Contributing requirements

### DIFF
--- a/PrimePascal/solution_2/Dockerfile
+++ b/PrimePascal/solution_2/Dockerfile
@@ -1,0 +1,11 @@
+FROM primeimages/freepascal:3.2.0 AS build
+
+WORKDIR /opt/app
+COPY *.pas .
+
+RUN fpc prime
+
+FROM ubuntu:20.04
+COPY --from=build /opt/app/prime /opt/app/
+
+ENTRYPOINT [ "/opt/app/prime" ]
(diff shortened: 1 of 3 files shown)


rbergen:
Exactly how does this differ from PrimePascal/solution_1?

circular17:
The difference is the type of NotPrimeArray here being PPackedBoolArray:

- the array is bitpacked (using bitpacked keyword) so it takes less memory
- it is a pointer to such array (allocated with GetMem)

Dynamic arrays ("array of" without size) cannot be bit-packed, at least with the version of FreePascal I am using. So I defined the type PackedBoolArray that is a static array of maximum size and allocate required memory manually with GetMem.

benibela:
Two trivial improvements might make both solutions much faster:

* Compile it with `fpc -O4 prime` in the Dockerfile

* Only use local variables, i.e. do not use NotPrimeArray and SieveSize in the loop.

rbergen:
@circular17 Thanks for clarifying that. :)

rbergen (commented):
LGTM, but I forgot the select "Approve" when submitting my review.

rbergen (approved):
LGTM

marghidanu (approved):

PR #356, opened 2021-07-08
Optimised Cython versions

tjol:
## Description

Now that, thanks to @rpkak, there is a Cython version, I thought it might be nice to show off what Cython can do if you let it.

The current version is little more than a copy of the Python original. This is great! It shows off what Cython can do without much manual performance tuning. I'd like to propose additional versions *with* a modicum of manual performance tuning

I've added two new copies of the code, while trying to keep the modifications as minimal as possible:

 * One, `PrimeCY_pointers.pyx`, that uses Cython's support for C pointers to access the bit array directly, eliminating all calls to Python functions in the loop. On the 1000000 prime benchmark this is something like 2-3 times faster
 * Two, `PrimeCY_bitarray.pyx`, a further modification that stores the flags as bits rather than as bytes. The performance is in between the two other versions.

I'm eager to hear what @rpkak thinks of me rudely mucking about in their solution. (Might also ping @ssovest while we're at it as the original author of Python solution 2, which this is all based on)

## Contributing requirements


rbergen:
If I'm honest, I think the PR should be updated to convert the rpkak+pointers implementation a modification to the original rpkak implementation, and only the rpkak+bit-array implementation should be made an additional one. The reason is that the key characteristics of the original pkak implementation and the rpkak+pointers implementation are identical.

Thank you for already tagging @rpkak. Let's give them some more time to respond to this if they want to.

tjol:
I think it's potentially interesting to have the contrast between Python solution 2 and rpkak's original cython, but I appreciate that that's really not the point. I'd be happy to change it, but of course let's wait a bit and see whether @rpkak has an opinion.

PrimeCython/solution_1/PrimeCY_bitarray.pyx
```diff
@@ -0,0 +1,158 @@
...
+                  }
+
+    cdef int _size
+    cdef int _nbits
+    cdef bytearray _bits
+
+    def __cinit__(self, limit):
+        self._size = limit
+        self._nbits = ((self._size + 1) // 2)
+        self._bits = bytearray(b'\xff' * (self._nbits // 2 + 1))
```
ssovest:
```suggestion
        self._bits = bytearray(b'\xff' * ((self._nbits + 7) / 8))
```

PrimeCython/solution_1/PrimeCY_bitarray.pyx
```diff
@@ -0,0 +1,158 @@
...
+            size  = bitslen - start
+            stop = start + size
+            # Cython doesn't understand range(start, stop, step), but provides special syntax:
+            for index from start <= index < stop by step:
+                byteindex = index // 8
+                bitindex = index % 8
+                bitmask = ~<char>(1 << bitindex)
+                bits[byteindex] &= bitmask
+
+            factor += 1
```
ssovest: Can i suggest a shorter version?
```suggestion
       """Calculate the primes up to the specified limit"""
        cdef size_t index, step
        cdef char* bits = <char*>(self._bits)

        cdef size_t factor = 1
        cdef size_t q = <size_t>(sqrt(self._size) / 2)

        while factor <= q:

            if bits[factor // 8] & ~BITMASKS[factor % 8]:
                # If marking factor 3, you wouldn't mark 6 (it's a mult of 2) so start with the 3rd instance of this factor's multiple.
                # We can then step by factor * 2 because every second one is going to be even by definition
                index = 2 * factor * (factor + 1)
                step  = factor * 2 + 1

                # Cython doesn't understand range(start, stop, step), but provides special syntax:
                for index from index <= index < self._nbits by step:
                    bits[index // 8] &= BITMASKS[index % 8]

            factor += 1
```
With BITMASKS defined somewhere else as `cdef const char* BITMASKS = b"\xfe\xfd\xfb\xf7\xef\xdf\xbf\x7f"`

ssovest:
Thanks for letting me know about this PR! Cython looks really great, it's like my 2 favorite programming languages got married :)

PrimeCython/solution_1/PrimeCY_bitarray.pyx
tjol: Shorter **and faster**. Nice!

tjol:
> Thanks for letting me know about this PR! Cython looks really great, it's like my 2 favorite programming languages got married :)

It has its moments. Cython is pretty messy and a bit of an abomination really. Basically any time you're tempted to use the Python C API just using Cython is a better choice, but in many cases where you might consider it, just using cffi to interface with C/C++ or Rust (or Fortran...) code is a better choice still IMHO.

But yes, it is pretty cool

tjol:
> If I'm honest, I think the PR should be updated to convert the rpkak+pointers implementation a modification to the original rpkak implementation, and only the rpkak+bit-array implementation should be made an additional one. The reason is that the key characteristics of the original pkak implementation and the rpkak+pointers implementation are identical.
> 
> Thank you for already tagging @rpkak. Let's give them some more time to respond to this if they want to.

Having heard no dissent so far, I have implemented this change and replaced rpkak with rpkak+pointers (renamed rpkak+byte-array for consistency with rpkak+bit-array).

rbergen:
Moving forward with merging.

PR #358, opened 2021-07-08
Deno implementation

GerbenRampaart:
## Description
Deno implementation. 

## Contributing requirements


rbergen:
If I understand the Deno website correctly, Deno is not a language but a runtime for JavaScript and TypeScript. That means that adding a PrimeDeno top-level directory is not appropriate, because those are reserved for distinguishing languages. 
It also means that if Deno is to be used, it should be used as the runtime for the existing TypeScript solution. That would (only) make sense if Deno is faster than the runtime that solution currently uses. 
@GerbenRampaart is the case?

GerbenRampaart:
Hi Rutger, thx for your efforts on this interesting project.

Deno is a runtime build on Rust, which uses typescript as a syntax (or javascript if you want), but is an entirely different runtime than Node and therefore has different benchmark results (https://deno.land/benchmarks).

I figured we are both testing the syntax as well as the runtime. For example both Java and Scala are part of primes, both are JVM languages.

I'd be happy to add the Deno solution to TypeScript though those two executions, Node vs Deno, are radically different and end at their syntactic similarities.

rbergen:
Thank you, and you're welcome. I'm just doing the best I can. :)

I know I'm at risk of starting a semantical discussion which is not my intention; I'm trying to explain the terminology this repo and I use when discussing these things:

- C, C++, Pascal, JavaScript, TypeScript, Java, Scala, V, Zig etc, are all languages. Some of these are, or were originally, based on  another: C++ was based on C, TypeScript was based on JavaScript, Scala was based on Java.
- Scala and Java _are_ different languages, but both run within the same (conceptual) virtual machine called JVM
- Every language has its own syntax, which describes the structure of the language in question
- All languages run in one runtime environment ("runtime") or another, some providing richer services to the program(mer) than others. 

Within that frame of definitions, NodeJS and Deno are runtimes, not languages. This repo is _not_ about comparing runtimes, but about comparing languages. However, of course we have to acknowledge that runtimes do have an influence on the performance of the language it hosts. It is for that reason that we do allow a solution to be switched to another runtime than the one it is setup to use, if that switch improves the _language's_ performance.

So, if the Deno runtime makes the existing solution in the TypeScript language run faster, you can open a PR to make that solution use Deno instead of NodeJS. We will then move to ask the original author of TypeScript/solution_1 to lend us his view on all of this.

GerbenRampaart:
It’s not faster currently, they are working hard on improving it. If those are the criteria, we can best remove the PR and I’ll keep it around in my fork. 

To be perfectly frank; I do think this is slightly odd. But then again, it’s not my repo and I don’t make the rules.

rbergen:
> To be perfectly frank; I do think this is slightly odd.

That is fair enough, I totally accept that. However, the criteria are indeed as I stated.

PR #359, opened 2021-07-08
implements a faithful elixir version

ityonemo:
…on_1

## Contributing requirements


rbergen:
@ityonemo Please refrain from using value statements about the subjective quality of the work of others, like saying your implementation is "4000x better". If there is an objective difference between your implementation and the one you compare it to (like its performance), it is of course fine to mention that.

Please change the title of this PR accordingly.

rbergen:
I've removed the part of the title that posed the problem for you.

ityonemo:
Thanks, sorry about that.

rbergen:
I've taken a first look at your implementation, and I wonder why you don't use a struct to contain the sieve state?

ityonemo:
Performance will be much slower.  Counters is an immutability escape hatch in the standard library, a struct will make a new copy of the entire thing on each modification, or at best, a copy on write.

rbergen:
Based on some confusion that has arisen in the context of other PRs, we've recently added a clarification to the contributing guidelines that a faithful implementation needs to use a class construct if it is available, or the construct closest to it if it isn't. I would argue that in the case of Elixir, struct is the construct closest to the class concept.

ityonemo:
I think this is a misunderstanding of elixir and erlang, but I will implement it.

ityonemo:
ugh, this pulled in some changes from the zig branch.  I'm have to reset some git history here, sorry.

rbergen:
@ityonemo Please let me know when you consider this PR ready for review.

rbergen:
@ityonemo Due to the earlier merge of #500, this PR now causes a merge conflict. If you still want to move towards getting this PR merged, it's probably easiest to rename the solution_2 directory to solution_3.

ityonemo:
oh this has always been ready.  I was waiting for a judgement call on whether it is actually necessary to wrap the content in a "struct" to fulfill the arbitrary requirement that it be wrapped in a "class-like" thing, and whether or not "records" are good enough.  If records are good enough, then ed22d5d should be pulled in, if they aren't then pull in aa6459b, if the requirement that it be wrapped is lifted (because it really does cause unfair performance regressions, then go with 9d8b6c0.  Maintaining the elixir version for me has become less fun because I feel like I am playing rules lawyer more than I am pushing the performance of the implementation and learning about the VM.

rbergen:
After reviewing the characteristics of both Elixir structs and records, my conclusion is that the arbitrary requirement for a faithful implementation that we do apply to all solutions, can indeed be fulfilled using records. The option to not use either and make your solution an unfaithful submission is still yours to choose, if you want to.

In any case, if you can update your PR to put your solution in `PrimeElixir/solution_3`, then we can proceed towards merging it.

rbergen:
@ityonemo Just a heads-up: if there's no follow-up to my last comment, I'll close this PR in a week or so.

ityonemo:
I decided I'm not happy with compromising the elixir implementation into unnecessarily wrapping it into a struct or record (moreover making a single-item struct or record is unidiomatic and something that pretty much no erlang or elixir develper would ever do), so go ahead and close it, just note that you won't have an implementation that accurately characterizes the capabilities of the VM, on account of rules laws that don't make sense for the language platform.

rbergen:
Ok, if that's what you want then that's what I'll do. I do think it's unfortunate, because your solution is elegantly written and indeed a "project-faithful" implementation that would have been a valuable addition to the project in my opinion.

Just for the record: this project has never been about accurately characterizing the capabilities of any VM (or language), except when it comes to providing the fastest implementation of the sieve of Eratosthenes. To qualify for certain specific characterizations in the context of the project ("faithful" being one of them), specific rules do have to be respected, and that applies to all languages.

In any case, when it comes to this PR I guess it just is what it is.

ityonemo:
I'm just saying that the rule that it must be 'wrapped in a struct' is a very silly one, especially for the BEAM languages.  Original erlang didn't even have records, so it wouldn't have been able to participate at all!

There's a lot to be learned about performance differences between mutable and immutable structures (when you want a mutability escape hatch, e.g.), and how wrapping a datatype in a cosmetic structure can have performance implications.

PR #361, opened 2021-07-08
add python/numpy based solution

emillynge:
## Description
This is a *unfaithful* python implementation that uses numpy array for bit sieve. It is *only* unfaithful due to the numpy dependency.

I still think it is a valuable solution to include due to how widespread numpy is and how close the solution is to the original.

## Contributing requirements

### DIFF
--- a/PrimePython/solution_3/PrimePY.py
+++ b/PrimePython/solution_3/PrimePY.py
@@ -1,0 +1,142 @@
+"""
+Python Prime Sieve using Numpy
+
+Based on MyFirstPython Program (tm) Dave Plummer 8/9/2018
+Adapted by Emil Sauer Lynge 08/07/2021
+
+This particular version is based on the PrimePython/solution2, whic itself is adapted from Dave Plummer original.
+"""
+import numpy as np
+from math import sqrt
+
+
+class PrimeSieve:
+
+    """This is the main PrimeSieve class. Call it with the number you wish as
+    an upper limit, then call the run_sieve method to do the calculation.
+    print_results will dump the count to check validity."""
+
+    prime_counts = { 10 : 4,                 # Historical data for validating our results - the number of primes
+                     100 : 25,                # to be found under some limit, such as 168 primes under 1000
+                     1000 : 168,
+                     10000 : 1229,
+                     100_000 : 9592,
+                     1000000 : 78498,
+                     10000000 : 664579,
+                     100000000 : 5761455
+                     }
+
+    def __init__(self, limit):
+        self._size = limit
+        self._bits = np.ones(((self._size + 1) // 2), dtype=np.bool_)
+
+    @property
+    def debug_sieve(self):
+        return np.vstack(((np.arange(0, self._bits.size) * 2 + 1) * self._bits , self._bits)).T
+
+    def validate_results(self):                      # Check to see if this is an upper_limit we can
+
+        """Look up our count of primes in the historical data (if we have it)
+        to see if it matches"""
+
+        if self._size in self.prime_counts:                              # the data, and (b) our count matches. Since it will return
+            return self.prime_counts[self._size] == self.count_primes()  # false for an unknown upper_limit, can't assume false == bad
+        return False
+
+    def run_sieve(self):
+
+        """Calculate the primes up to the specified limit"""
+
+        factor = 1
+
+
+        # sqrt doesn't seem to make any difference in CPython,
+        # but works much faster than "x**.5" in Pypy
+        q = sqrt(self._size) / 2
+
+
+        bits_view = self._bits[factor:]
+        while factor <= q:
+            for v in bits_view.flat:
+                if v:
+                    break
+                factor += 1
+
+            bits_view = self._bits[factor + 1:]
+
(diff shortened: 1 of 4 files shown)


tjol:
How does the performance compare to solution_2? Do you think it might make sense to add this as a variant of solution_2 rather than as a separate solution? They do both use array-based processing, the difference being that solution_2 uses stdlib features with their limitations

PrimePython/solution_3/PrimePY.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+        # sqrt doesn't seem to make any difference in CPython,
+        # but works much faster than "x**.5" in Pypy
+        q = sqrt(self._size) / 2
+
+
+        bits_view = self._bits[factor:]
+        while factor <= q:
+            for v in bits_view.flat:
```
tjol, line 60: `np.argwhere`?

rbergen:
> Do you think it might make sense to add this as a variant of solution_2 rather than as a separate solution? They do both use array-based processing, the difference being that solution_2 uses stdlib features with their limitations

Due to the fact that this implementation constitutes a different approach, and solution_2 was created by a different author, I am okay with adding this is as a separate solution.

PrimePython/solution_3/PrimePY.py
emillynge: Unfortunately, `argwhere` will return **all** indices to `True` entries - which will be close to `N` in the first iterations trending towards zero.

I looked for a long time, but numpy has *no* equivalent to `list.index`, which will return just the first occurence. I find that very odd, but there really uis no good solution.

`argmax` is the closest fit, but performs much worse than the current loop.

Given that the maximum number of iterations in the while loop is around 14-17, the problem performance hit is not too bad.

emillynge:
If the original author is OK with it, then I would be perfectly happy to merge my implementation into solution_2. They do after all share around 90-95% of the code.

I could just add `--with-numpy` flag that uses numpy for the bits array, and let the docker container execute the program twice with different flags.

There is an issue though, with the github tags since my implementationm is not faithful, whereas solution_2 is.

PrimePython/solution_3/PrimePY.py
tjol: Right! Makes sense.

tjol:
@rbergen Just to be clear, I didn't mean to step on your toes, I was just thinking aloud.

@emillynge What I meant to suggest was, if anything, to consider adding your program as a separate file to the solution_2 directory. But really it's unnecessary, and may just confuse things.

emillynge:
Regarding performance, as you can see from the included benchmark in the readme; Using numpy increases performance by 360% over solution_2 which leaves it at roughly 60% of the performance of PrimeC/solution_1!!

tjol:
Good lord, was that table right in front of me this whole time? I should really read things more carefully before asking stupid questions next time :-)

Very cool!

rbergen:
> @rbergen Just to be clear, I didn't mean to step on your toes, I was just thinking aloud.

@tjol I have decently-sized feet, but my toes aren't **that** long. :)

rbergen (approved):
LGTM

emillynge:
BTW, full disclosure, just to be clear.

The numpy dependency relies very heavily on C extensions and the speedup is most certainly due to critical parts being executed in C extensions.

The difference between this PR and #370 is mostly that the programmer doesn't need to know any C at all to write the program.
In fact, most users of numpy likely does not know that it is very much a C library.

PrimePython/solution_3/PrimePY.py
mgdotdev: If you want to, you could do this in C by iterating across the numpy array until you find the first `1`:
```C
#include <Python.h>
#include <numpy/arrayobject.h>

static PyObject* _find_first(PyObject *self, PyObject *args) {
    PyArrayObject* array;
    int item;

    if (!PyArg_ParseTuple(args, "Oi", &array, &item)) {
        return NULL;
    }

    npy_intp _len = PyArray_DIM(array, 0);
    npy_intp _stride = PyArray_STRIDE(array, 0);

    for (int i=0; i<_len; i++) {
        if (*(int*)(array->data+i*_stride) == item) {
            return Py_BuildValue("i", i);
        }
    }

    return Py_BuildValue("i", -1);
}

static char indexMethodsDocs[] = 
    "return the index of first instance of a provided integer value within a numpy array";


static PyMethodDef indexMethods[] = {
    {"find_first", _find_first, METH_VARARGS, indexMethodsDocs},
    {NULL, NULL, 0, NULL}
};

static struct PyModuleDef indexModule = {
    PyModuleDef_HEAD_INIT,
    "indexModule",
    NULL,
    -1,
    indexMethods
};

PyMODINIT_FUNC PyInit_indexModule(void) {
    return PyModule_Create(&indexModule);
}
```

which you could then call from python (after you set it up as an extension) via

```python
index = indexModule.find_first(array, 1)
```

emillynge: Even though I am calling numpy which I know uses C extensions i'd rather not write any actual C code for this solution.

What I'm trying to show, is that by clever use of a widely available package, you can get C-like performance without writing a single line of C yourself.
Also, keeping the program compile-free is nice.

marghidanu (approved):

PR #363, opened 2021-07-09
Cleaned up NodeJS solution a bit

Newms34:
## Description

Cleaned up the NodeJS code a bit. In particular, I:

1. Removed the call to a semi-external module ('perf-hooks'). This _shouldn't_ improve speed, but... every bit counts!
2. Moved the class instantiation for the batch run _outside_ our `do...while` loop. This should improve our speed by not forcing JS to create a new class instance every loop. Instead, I added  a `.clear()` method to "reset" the `bitarray` property.
3. "Reversed" one of the loops. In general, reversed loops in JS are faster, since a check for falseyness (i.e., is `myArray.length` equal to zero) is _faster_ than checking if the array length is a particular number.
4. Switched all calls to `performance.now()` to `process.hrtime()`. This _would_ seem to slow things down (hrtime stores stuff as an array, so we need to convert it), but it's important to note that we only actually use/evaluate the time _once_: at the end of batch run. As such, storing the  numbers as an array is fine.

## Contributing requirements

### DIFF
--- a/PrimeNodeJS/solution_1/PrimeNode.js
+++ b/PrimeNodeJS/solution_1/PrimeNode.js
@@ -11,20 +11,20 @@ Date:      2021-07-08
 'use strict';
 
 // get performance API (for older node versions)
-const { performance }                                  = require('perf_hooks');
+// const { performance } = require('perf_hooks'); //Why would we include this if node already has high rez time?
 
 // Historical data for validating our results - the number of primes
 // to be found under some limit, such as 168 primes under 1000
 const knownPrimeCounts = {
-    10 : 4,
-    100 : 25,
-    1000 : 168,
-    10000 : 1229,
-    100000 : 9592,
-    1000000 : 78498,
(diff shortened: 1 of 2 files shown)


rbergen:
For existing solutions, we can only accept changes that don't modify the [key characteristics](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md#characteristics) of the solution, and do improve the benchmarked/timed part of the code (i.e., the prime sieve run loop). Those performance changes then must be made [with the minimum amount of changes needed to implement them](https://github.com/PlummersSoftwareLLC/Primes/pull/341#issuecomment-875952572).

Of your changes, the second one will improve performance, but it also changes the solution from a faithful to an unfaithful one. This is due to the fact that a faithful implementation must create the sieve object from scratch for every run.

The other changes are either cosmetic or don't improve the speed of the timed code.

Due to this, we will not be able to merge this.

PR #364, opened 2021-07-09
Added Prolog solution

jimbxb:
A solution written in Prolog


rbergen:
I'll love adding Prolog to the collection of languages, but please mark this solution as unfaithful. It does not use a class-like object to encapsulate the full sieve state. And I do want to note that that statement is not an accusation, only an assessment of one of the key characteristics we distinguish.

jimbxb:
The state of the sieve is stored in an integer of unbound size. 

My question is: why is this considered unfaithful?

While I agree this is not a class-like data structure, I fail to see why this differs in practice from a regular bit-vector used in other implementations.

I'm happy to change this to unfaithful if I'm mistaken in my understanding of faithfulness :)

rbergen:
> I fail to see why this differs in practice from a regular bit-vector used in other implementations.

Could you point to an example of a solution that is marked as faithful and uses a bit vector that is itself not wrapped in a class-like structure?

PrimeProlog/solution_1/README.md
```diff
@@ -0,0 +1,40 @@
+# Prolog solution by jimbxb
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
```
rbergen, line 4: If you're marking your solution as unfaithful, the badge markdown to use is 
```
```

PrimeProlog/solution_1/README.md
```diff
@@ -0,0 +1,40 @@
...
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-1-green)
```
rbergen, line 6: Please don't take this as a challenge, as it is a genuine question: how does your implementation arrange the bit-level access?

jimbxb:
I have modified the solution to use a functor (`state/2`) to store the state of the sieve. This is the closest prolog has to a class-like structure. I believe this makes the solution faithful now.

PrimeProlog/solution_1/README.md
rbergen: I see it now. You're using getbit/2 for getting and bit shifting in setbits for setting. It took me a while to figure it out because Prolog's syntax is quite different from the languages I have grown accustomed to.

jimbxb: Would the solution still be considered faithful without the added `state/2` functor?

rbergen:
> I have modified the solution to use a functor (`state/2`) to store the state of the sieve. This is the closest prolog has to a class-like structure. I believe this makes the solution faithful now.

This is definitely a step in the faithful direction. What I'm not sure of is if the implementation creates the term from scratch for each sieve run, which is another part of the faithful requirements. I can add to that that I also don't know if the language even allows you to force it to do so.

jimbxb:
> This is definitely a step in the faithful direction. What I'm not sure of is if the implementation creates the term from scratch for each sieve run, which is another part of the faithful requirements. I can add to that that I also don't know if the language even allows you to force it to do so.

The requirement of rebuilding the sieve data from scratch each iterations is met. The `timed_primes` predicate acts recursively, timing each call to `primes` (with this predicate recalculating the sieve each time) until 5sc has been exhausted.

Happy to clarify further if there's any more confusion. I know Prolog is quite niche, and this use case isn't typical.

rbergen:
No, I'm satisfied with this explanation, I'm happy to add such a niche language to the repo, and appreciate your efforts towards making this atypical use case work.

PrimeProlog/solution_1/README.md
rbergen: No. Faithfulness does require the state/2 functor, the 1-bit claim requires the memory footprint of a prime candidate to really be one bit only. That is to say: both are independent characteristics of an implementation.

rbergen (approved):
LGTM

marghidanu (approved):

PR #365, opened 2021-07-09
Update PrimePas.dpr

glenkleidon:
As the GetBit and And Clearbit methods are trivial implementations, you can include the "inline" compiler hint to tell the compiler that there is no need to create a new method (ie consequently stack operations) for these.  Removing the function overhead doubles performance at least.

## Contributing requirements

### DIFF
--- a/PrimeDelphi/solution_1/PrimePas.dpr
+++ b/PrimeDelphi/solution_1/PrimePas.dpr
@@ -17,8 +17,8 @@ type
 		FBitArray: array of ByteBool; //ByteBool: 4644. WordBool: 4232. LongBool: 3673
 		FMyDict: TDictionary<Integer, Integer>;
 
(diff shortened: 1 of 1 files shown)


rbergen:
We already have #264 open that aims to improve this solution. Please discuss in that PR how your changes can be added to it.

glenkleidon:
It seems that Kim's change removes those methods anyway - If we are going to accept Patch-1 from @kimMadsen then this change is not needed.

As there are no code changes, only compiler hint.  If that pull request is rejected however, this tiny change achieves vast improvement with no code changes (only compiler hints)

rbergen:
I created #316 to request for consolidation for the various PRs on the Delphi solution. So bottom line, either #264 or this one will indeed need to be closed. :)

PR #366, opened 2021-07-09
Optimizations for dart solution 1

mmcdon20:
## Description

Improved performance of dart solution_1:
- Using `Uint8List` instead of `List<bool>`
- Made ` _resultsDictionary` a `const` so that it is not recreated every time you instantiate the `PrimeSieve` class.
- Updated dart language from version `2.10.0` to `2.13.0`.
- Declared variables as final whenever possible.
- Updated the `Stopwatch` code in the main method, `.start()` was being called in every iteration of the loop needlessly, and `timer.elapsed.inMicroseconds` allocates an additional `Duration` object whereas `timer.elapsedMicroseconds` does not.
- Use `.toInt()` instead of `.round()` (which I think may be a bug in original implementation, it's also slightly faster).
- Simplified `_validateResults()` method.

## Contributing requirements

* [X] [ALREADY EXISTED] I placed my solution in the correct solution folder.
* [X] [ALREADY EXISTED] I added a README.md with the right badge(s).
* [X] [ALREADY EXISTED] I added a Dockerfile that builds and runs my solution.

### DIFF
--- a/PrimeDart/solution_1/README.md
+++ b/PrimeDart/solution_1/README.md
@@ -3,7 +3,7 @@
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
 ![Parallelism](https://img.shields.io/badge/Parallel-no-green)
-![Bit count](https://img.shields.io/badge/Bits-unknown-yellowgreen)
+![Bit count](https://img.shields.io/badge/Bits-8-yellowgreen)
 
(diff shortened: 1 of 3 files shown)


rbergen:
Tagging @eagerestwolf and allowing time for them to review and respond to this PR on their original solution, if they want to do so.

eagerestwolf:
@rbergen looks good to me. I'm impressed how these simple changes improve Dart so much. @mmcdon20 good job man, I'm by no means a Dart expert, so I just did what felt natural; but you seem to have a very good grasp of the language.

PR #369, opened 2021-07-09
VB.Net - Reform and optimize without change algorithm, result almost twice from before.

RevensofT:
## Description

- Reverse almost loop back to pure do loop for JIT ASM optimize.
- Change a condition add in prime count method to bitwise add.
- Reverse dictionary back to select case.

## Direct impact performance changed
- Reverse almost loop back to pure `Do Loop` without `Until`, `While` or `For` for JIT ASM optimize.
- Split code into small methods for inline method optimize.

### Previous - rbergen's cpu
```
rbergen_vb;2457;5.0012308;1;algorithm=base,faithful=yes,bits=1
```

### Previous - AMD Fx-6300
```
rbergen_vb;1139;5.003093;1;algorithm=base,faithful=yes,bits=1
```

### Last update - AMD Fx-6300
```
rbergen_vb;2053;5.0011727;1;algorithm=base,faithful=yes,bits=1
```

## Contributing requirements

### DIFF
--- a/PrimeBASIC/solution_2/Dockerfile
+++ b/PrimeBASIC/solution_2/Dockerfile
@@ -7,7 +7,7 @@ RUN dotnet restore
 
 # copy and publish app and libraries
 COPY *.vb /source
-RUN dotnet publish -c release -o /app --no-restore
+RUN dotnet publish -c "release free" -o /app --no-restore
 
 # final stage/image
 FROM mcr.microsoft.com/dotnet/runtime:5.0-focal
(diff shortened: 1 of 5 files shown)


rbergen:
Thank you for your contribution. Could you please update your PR so that it only makes the changes that are required to increase the performance of the benchmarked part of the code (i.e. the sieve run loop)?

RevensofT:
Hi rbergen,
I'm not sure how to change my pull request (I don't think you mean pull request note), I look at files change, it's all I change without any duplicate, I need to add compile config to remove integer overflow check so I need to change .vbproj, .sln and docker along with it, in .vb I rewrite almost every line of code because all of them has to be optimize.

Edit: I add section "Direct impact performance changed" into pull request note.

rbergen:
> to remove integer overflow check

That's a change I don't agree with. This removes a boundary check that is intrinsic to the language.

> in .vb I rewrite almost every line of code because all of them has to be optimize.

I can see you've rewritten almost every line of code, which makes it completely impossible to see what you've optimized there. What I'm asking you to do is pull the original .vb file back into your fork, and change _only those lines_ that need to be changed to improve the performance of the .vb code.

PR #370, opened 2021-07-09
CPython implementation

mgdotdev:
## Description
This implementation makes full use of the CPython interpreter, by interfacing with the parameters of the problem in the Python layer and crunching the numbers in the C layer. The result is a ~6600% boost over the original interpretation. No other optimizations have been applied.

## Contributing requirements
* [ x] I read the contribution guidelines in CONTRIBUTING.md.
* [ x] I placed my solution in the correct solution folder.
* [ x] I added a README.md with the right badge(s).
* [ x] I added a Dockerfile that builds and runs my solution.
* [ x] I selected `drag-race` as the target branch.
* [ x] All code herein is licensed compatible with BSD-3.


rbergen:
I can't consider this a faithful implementation, based on the following:

- CPython is not a language in itself, but rather an implementation of Python. This is well documented, and illustrated by the fact that your PR aims to add your solution as a Python solution. This means that the solution must be implemented in the Python language.
- From the perspective of the Python part of the implementation, the C module is an external dependency. The fact that this dependency is provided in the same PR as the Python code is irrelevant to the point.
- As the contributing guidelines state, the first rule for a faithful implementation is that "It uses no external dependencies to calculate the actual sieve."

mgdotdev:
If the standard you established above was indeed standard, then we would also need to reject davepl's solution_1 as well. The `import math` module is a wrapper of the same style as my `import sieveModule` as it, to quote the Python documentation, "consists mostly of thin wrappers around the platform C math library functions" [1, 2]. Python is generally considered a "batteries included" language, where many of the tools available to Python devs are included in the standard CPython distribution as "part of the language". I believe that knowing how to strap extra batteries to this metaphorical battery pack is as fundamental to the Python language as anything else. We may be taking this idea to its logical extreme with this solution, but nevertheless I think it is consistent with the rules as applied to other implementations already included in the repository.

[1] https://docs.python.org/3/library/math.html
[2] https://github.com/python/cpython/blob/main/Modules/mathmodule.c#L3613

rbergen:
@1mikegrn As I stated, the requirements for a faithful implementation include that no external dependencies may be used _to calculate the actual sieve_. Dave's solution_1 doesn't do this, it only uses the math import to compute the square root of the sieve size before the actual sieve is run. That's a pattern we've indeed consistently accepted, and I am very happy to do that now as well. 

The issue is thus not with the use of an external dependency, but using one to do the entire sieve calculation. That makes that one could argue that the implementation is not a Python implementation of a prime sieve, but a C implementation thereof, put in a Python wrapper. 

If a PR was opened that aimed to add a "C" implementation that performed the actual sieve calculation in an asm block (that being an integrated capability of the C language), I would make exactly the same argument as I do now.

mgdotdev:
I would further argue that knowing when to prematurely stop iterating towards `this.sieveSize` is a function necessary to solution_1's ability to "calculate the actual sieve" as fast as possible (it was a drag race after all).

You could implement a "pure-python" fix for this, by defining
```python
def _is_valid_factor(this, factor):
    if factor**2 < this.sieveSize:
        return True
    return False
```
and then call `while this._is_valid_factor(factor)` in `this.runSieve`. This would allow you to remove the math module so as to be consistent with your rules.

As a compromise, would simply flipping the `faithfulness` flag to `false` be satisfactory? Though it initially started as simply a drag race I think that this repository has morphed into more of a teaching tool than a race; of the same vein as how Brian Kernighan oftentimes implements the same algorithm in different languages so to demonstrate to his students how different programming languages function. That's my ultimate motivation for writing this implementation, and ultimately why I would want to have it included.

mgdotdev:
@rbergen Did we ever reach some sort of consensus regarding a potential compromise?

rbergen:
@1mikegrn No, because I still haven't made up my mind about what it is that I'm looking at from a languages perspective. I will try to get there within the next few days.

rbergen:
Ok, so I've had some communication with Dave on the topic as well, and I think the best solution is creating a PrimeMixed "language" directory, within which this would be solution_1. The README can explain that it is a mix of Python and C, as it basically already does. My only suggestion would be to make the label field in the output (currently `1mikegrn`) also express that it concerns a solution that combines Python and C.
Concerning faithfulness: the solution would have to be marked as not faithful. It doesn't really matter which of the two languages you pick as a frame of reference: there's always an external dependency on the other language.

rbergen:
Quick update: PrimeMixed/solution_1 has now been taken by an implementation in #474. This means that the solution this PR seeks to add would need to be put in PrimeMixed/solution_2.

rbergen:
@1mikegrn Do you still want to move forward towards getting this merged?

mgdotdev:
> @1mikegrn Do you still want to move forward towards getting this merged?

Sorry, been out of town till recently. I'll get the necessary changes committed shortly.

PrimeMixed/solution_2/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM python:3-alpine
+
+USER root
+
+WORKDIR /code/
+
+RUN apk add build-base
+
+ARG CACHEBUST_APP=1
```
rbergen: I'm unclear what is the purpose of this argument?

PrimeMixed/solution_2/src/pysieve/sieve.py
```diff
@@ -0,0 +1,34 @@
...
+            self._results = sieveModule._run_sieve(self.sieve_size)
+        return self._results
+
+    def validate_results(self):
+        if self.sieve_size in PRIME_COUNTS:
+            return PRIME_COUNTS[self.sieve_size] == self._results
+        return False
+
+    def print_results(self, duration, passes):
+        print(f"1mikegrn/CPython;{passes};{duration};1;valid={self.validate_results()},algorithm=base,faithful=no,bits=32")
```
rbergen: Please remove the `valid` tag from your output line. It's not a known tag at the moment, and it may clash with a tag we may want to add in the future.

PrimeMixed/solution_2/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM python:3-alpine
+
+USER root
```
rbergen: This directive is not necessary. The containers run as root by default.

rbergen:
> Sorry, been out of town till recently. I'll get the necessary changes committed shortly.

That's not a problem. Thank you for picking this up now that you're able to.

PrimeMixed/solution_2/Dockerfile
mgdotdev: Oh, it's a docker hack; you can pass `--build-arg CACHEBUST_APP=$(date +%s)` and that'll force subsequent lines to rebuild instead of docker using a cached image. We can remove that, however.

PrimeMixed/solution_2/Dockerfile
mgdotdev: True. That being said, this Dockerfile really needs some love; I might clean this up a bit for best practices.

PrimeMixed/solution_2/src/pysieve/sieve.py
mgdotdev: Done.

mgdotdev:
hang on, just noticed that README needs to be updated

rbergen:
Please take a look at the Dockerfile linting findings made by CI. The warning and the error need to be fixed to be able to proceed, it is appreciated if you address the notices as well.

rbergen:
@1mikegrn I'll have to ask you to merge this branch into yours, one more time. It will re-enable CI to work for your PR. Also, please rename your solution directory to PrimeMixed/solution_3 (#646 was merged today, which uses PrimeMixed/solution_2).

rbergen:
@1mikegrn Please fix the issues with your Dockerfile.

mgdotdev:
🎉

rbergen (approved):
LGTM

PrimeMixed/solution_3/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM python:3-alpine
+
+RUN apk add build-base
```
marghidanu: Let's try to fix the warnings given by Hadolint in here.

marghidanu (approved):

PR #376, opened 2021-07-10
Markdown file containing result charts

rpkak:
## Description

If you call
```
make results
```

the `RESULTS.md` file and the charts of it will be build.

The workflow `Build Results` automatically calls this command so you can see the latest results of the implementations in GitHub without any building.

## Contributing requirements


rbergen:
I think this approach will no longer work once we move to the benchmark automation implementation of #348. @marghidanu, can you confirm/correct?

> The workflow Build Results automatically calls this command so you can see the latest results of the implementations in GitHub without any building.

We won't adopt this. The performance of GitHub Actions workflows vary so wildly that benchmark results using it are meaningless.

marghidanu:
In the past, we ran the benchmark via CI but the results were varying a lot, we even considered running our own GitHub Actions runners which turned out to have some security problems. 

In the future, we will allow users to upload their results and @rbergen has already built a UI for displaying the data.

P.S.: @rpkak, I appreciate the engineering and the way you used formatters to implement this :)

rpkak:
Ok

rpkak:
Or do you think adding the command without the workflow make sence?

marghidanu:
@rpkak I've already added a CLI chart formatter last week, the PR should be merged soon.

PR #379, opened 2021-07-11
Optimized Bash implementation

Blzut3:
## Description
I'm not entirely sure if using properties of uninitialized values is considered legal, but if so these changes do a little more than double performance of solution_1 on my machine (~12 seconds to ~5 seconds).  Solution_2 included here uses the same trick as PowerShell's solution_2 but only takes the run time for me from ~5 seconds to ~3 seconds.

As an aside I did try bitpacking to avoid the initialization slowness, but it seems that Bash's performance in mathematical code is basically a function of the number of variable accesses that are being done.  Doing the math, and creating the local variables required, to bit pack basically costs as much as it saves vs not bit packing.

I did try implementing a faster sqrt algorithm but once again the complexity of doing a smarter algorithm seemed to be outweighed by overheads on executing commands and variable access.  Plus it would only matter for smaller sieveSizes which the simple algorithm doesn't take long with anyway.  So I got rid of the subshell and called it a day.

Oh I saw a note somewhere that the reason Ubuntu was used for the container instead of Alpine was because the script wasn't exiting on Alpine.  This was likely due to the lack of coreutils for the date command, but even fixing that I found Alpine's Bash to be about half the speed of Ubuntu's.  Not sure why, but have to assume libc differences?

In summary the changes are:
- Invert meaning of 0 and 1 to avoid need for slow initialization. (Uninitialized values act as 0 for the purposes of this script.)
- Reduce complexity of get and clear functions for significant performance improvement.
- Use associative arrays for bitArray as they're actually faster in Bash (possibly due to avoiding the expression evaluator on the index)
- Upgrade to Bash 5.1 (Ubuntu 21.04) for improved performance. Note, Bash 5.1 in Alpine is about half the speed.
- Use a different method of limiting the runtime of the script which doesn't require spawning a process each loop.
- Removed the need for a subshell for the sqrt call.
- Set LC_ALL to C to avoid localization costs.
- Fixed typo on printing of times (9 digits instead of 8).
- Wrapped main in a function so that some of the parsing is done ahead of time instead of line by line.
- Added solution 2 which does the same inlining as PowerShell's solution 2 (albeit Bash's function call overhead isn't anywhere near as bad)

## Contributing requirements

### DIFF
--- a/PrimeBash/solution_1/Dockerfile
+++ b/PrimeBash/solution_1/Dockerfile
@@ -1,8 +1,8 @@
-FROM ubuntu:20.04
-
-RUN apt-get update
+# bash (with coreutils) in alpine seems to be less optimized than ubuntu
+# bash 5.1 seems to have much better performance than 5.0 shipped in 20.04
+FROM ubuntu:21.04
 
 WORKDIR /opt/app
-COPY PrimeBash.sh .
-
-ENTRYPOINT [ "./PrimeBash.sh" ]
+COPY *.sh ./
+ 
+ENTRYPOINT [ "./run.sh" ]
--- a/PrimeBash/solution_1/PrimeBashInline.sh
+++ b/PrimeBash/solution_1/PrimeBashInline.sh
@@ -60,23 +60,18 @@ function getBit {
 	(( $1 & 1 && !bitArray[\$1] ))
 }
 
-function clearBit {
-	# Mark as composite
-	bitArray[$1]=1
-}
-
 function runSieve {
 	local factor num q
 	sqrt q "$sieveSize"
 	for ((factor=3; factor <= q; factor+=2)); do
 		for ((num=factor; num < sieveSize; num+=2)); do
-			if getBit "$num"; then
+			if (( !bitArray[\$num] )); then
 				factor=$num
 				break
 			fi
 		done
 		for ((num=factor**2; num < sieveSize; num+=factor*2)); do
-			clearBit "$num"
+			bitArray[$num]=1
 		done
 	done
 }
@@ -127,7 +122,7 @@ function printResults {
 		"$count" "$valid"
 
 	# rbergen: added drag-race format output
-	printf "\nbash;%s;%s;1;algorithm=base,faithful=no\n" "$passes" "$dur_str"
+	printf "\nbash_inline;%s;%s;1;algorithm=base,faithful=no\n" "$passes" "$dur_str"
 }
 
 function main {
--- a/PrimeBash/solution_1/PrimeBashPacked.sh
+++ b/PrimeBash/solution_1/PrimeBashPacked.sh
@@ -1,0 +1,162 @@
+#! /bin/bash
+#
+# A simple prime sieve based on https://github.com/davepl/Primes
+# Written in bash.
+#
+# Tyler Hart (nitepone) <admin@night.horse>
+
+# rbergen -- changed number for 10 from 1 to 4
+readonly -A primeCounts=(
+	[10]=4
+	[100]=25
+	[1000]=168
+	[10000]=1229
+	[100000]=9592
+	[1000000]=78498
+	[10000000]=664579
+	[100000000]=576145
+)
+
+declare sieveSize=0
+declare -A bitArray=()
+# rbergen: changed runtime to drag-race default
+readonly RUNTIME_SEC=5
+
+function sqrt {
+	local -n i=$1 # Output
+	i=1
+	while ((i**2<=$2)); do
+		((++i))
+	done
+	((--i))
+}
+
+function initGlobals {
+	sieveSize=$1
+}
+
+function emptyBitArray {
+	bitArray=()
+}
+
+function validateResults {
+	local result=$1
+	(( primeCounts[\$sieveSize] == result ))
+}
+
+function countPrimes {
+	local i count=$((sieveSize >= 2))
+	for ((i=3; i < sieveSize; i++)); do
+		if getBit "$i"; then
+			((++count))
+		fi
+	done
+	echo "$count"
+}
+
+function getBit {
+	local index=$1 bit
+	(( index & 1 && !(bit=1<<(index>>1&63), index>>=7, bitArray[\$index] & bit) ))
+}
+
+function clearBit {
+	local index=$1 bit
+	(( index & 1 && (bit=1<<(index>>1&63), index>>=7, bitArray[\$index] |= bit) ))
+}
+
+function runSieve {
+	local factor num q
+	sqrt q "$sieveSize"
+	for ((factor=3; factor <= q; factor+=2)); do
+		for ((num=factor; num < sieveSize; num+=2)); do
+			if getBit "$num"; then
+				factor=$num
+				break
+			fi
+		done
+		for ((num=factor**2; num < sieveSize; num+=factor*2)); do
+			clearBit "$num"
+		done
+	done
+}
+
+function printResults {
+	local showresults dur_nano avg_dur_nano dur_str avg_dur_str passes count valid
+	showresults="$1"
+	dur_nano="$2"
+	passes="$3"
+	# create duration strings from nanosecond duration time
+	avg_dur_nano=$((dur_nano/passes))
+	dur_str="$(printf "%d.%09d"\
+		"$((dur_nano/1000000000))"\
+		"$((dur_nano%1000000000))"\
+	)"
+	avg_dur_str="$(printf "%d.%09d"\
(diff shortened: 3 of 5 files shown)


rbergen:
Please convert solution_2 to a second implementation of solution_1. It will save docker build time and resources during the benchmarks. Also already tagging @Nitepone so they can provide feedback on this PR on their original implementation, if they want to do so.

Nitepone:
These look like excellent changes to me! I had not spent much time optimizing my original implementation. I also have never even tried to optimize bash code for time. (I also hadn't much time to even fix my original PR last month. Thanks @rbergen for getting my original implementation merged in!)

A minor suggestion would be to remove the getBit and clearBit functions from the solution_2 solution. clearBit is dead code and getBit could be easily replaced (alas, for no test time performance benefit).

Also, you could update the readme/file to accredit your changes.

Awesome PR @Blzut3, I didn't expect to see a 4-5x improvement when running solution_2 on my machine! Made my morning to see this code get much needed improvements! :tada:

rbergen:
@Nitepone Thanks for responding so quickly!

@Blzut3 That only leaves my request to make solution_2 in your PR an extension to solution_1, instead of an additional solution.

Blzut3:
@rbergen The contributing instructions appear to use the term solution and implementation interchangeably so I'm not sure what your request is supposed to look like as far as the execution framework goes?  Is there an example from another language that I can reference?  My best guess is you want me to include both run runSieve functions in one file and run both main loops back to back, but not sure.

rbergen:
@Blzut3 PrimeBASIC/solution_1 demonstrates what my request aims for. It's one solution directory with one Dockerfile, that (in the BASIC case compiles and) runs multiple implementations (approaches) in one Docker image. As that solution also shows, it is not necessary to include both implementations in one file, as long as they are executed within the one Docker image.

Blzut3:
@Nitepone Took out the clearBit function, but left getBit in since as you say it would only affect code outside of the drag race.  Makes things slightly easier to change/experiment with if the other code is all the same.

@rbergen Done.  Also added the bit packed implementation as a separate commit.  Feel free to toss it out if you wish, but figured perhaps someone would be curious as to why it wasn't done, and it's just a matter of dropping in a different getBit/clearBit function pair.

Also, kind of curious what it would take for an implementation to be considered faithful?  Is the issue just that it uses global variables for the state instead of passing it through the functions?  Kind of feel like the rules are technically met by emptyBitArray tossing the old state (possibly more technically correct if there was an unset first?), but I can definitely see how such would be twisting the rules.  Creating a unique variable name for bitArray each iteration and passing it to the inlined runSieve wouldn't be particularly difficult and wouldn't cost much in runtime performance.  But I don't know if one would need to go as far as generating unique variable names, or if it's just the act of passing the variable as an argument that's enough to do it?

rbergen:
> Also, kind of curious what it would take for an implementation to be considered faithful?

If I'm honest, I do think the second rule in the "faithfulness" section of the contributing guidelines does answer your question, but as I'm typing anyway: what's primarily missing here is a "class-like" wrapper object that encapsulates the full sieve state, that being the prime candidate flags and whatever else the implementation needs to run the sieve. That wrapper object needs to be created (and torn down in languages where that is a developer action) for every sieve run,

Nitepone:
> what's primarily missing here is a "class-like" wrapper object that encapsulates the full sieve state, that being the prime candidate flags and whatever else the implementation needs to run the sieve. That wrapper object needs to be created (and torn down in languages where that is a developer action) for every sieve run,

Bash does not contain a class construct. The contributing guidelines specify that such languages should use the "closest thing" to such functionality. This sadly leaves some ambiguity in this case. 😥

```
It uses a class to encapsulate the sieve, or a (closest) equivalent feature in your language if a class construct is not available. 
```

Due to the nature of Bash, globals variables are often required. (i.e. Returning an array from a bash function).

Creating class-like wrappers is not something I've ever seen in the wild. It is common to maintain globals through functions though. By accessing the globals entirely through functions, you can achieve a (weak) level of encapsulation.

rbergen:
If I'm honest, my thinking is that there is no way to create a faithful implementation in Bash. As it has not been possible to do so in quite a few other purely procedural languages. It's important to keep in mind that in the context of this repo "not faithful" does not constitute a value statement. It's just a statement on one of the four key characteristics of an implementation.

Nitepone:
That makes sense, and is what I expected. It's a shame, but bash just isn't the best language to use to find prime numbers.

rbergen:
...which makes the fact that it's doing just that in this repo all the more interesting to me. :)

rbergen:
@Blzut3 Can I consider this PR stable?

Blzut3:
> @Blzut3 Can I consider this PR stable?

Yes, this PR should be ready to merge regardless of the faithfulness discussion.

> Due to the nature of Bash, globals variables are often required. (i.e. Returning an array from a bash function).

You can use references to pass local arrays in and out of functions.  There some care that needs to be taken since for whatever reason Bash doesn't ignore attempts to create circular references, but that's usually avoidable with careful variable naming, if not some simple boilerplate to check if the name is already what you want it to be can work around that problem.

> Creating class-like wrappers is not something I've ever seen in the wild. It is common to maintain globals through functions though. By accessing the globals entirely through functions, you can achieve a (weak) level of encapsulation.

It's definitely not true that it's not possible to do classes in Bash.  Let me introduce you guys to Bash Infinity: https://github.com/niieani/bash-oo-framework  A nifty bash script that leverages crafty aliases and functions to implement object oriented programming, exceptions, among other things in pure Bash!  (Yes you could probably dig through the internals and figure out how its storing class member variables to break the encapsulation, but you can do that in C++ too with pointer arithmetic.)

So there definitely exists a point where Bash can cross the faithfulness line, which is why I was curious on getting a judge's ruling on where that line is.  First let me state that I realize that the whole discussion is kind of silly since ticking that box basically means nothing, but that's the fun of it all right?

If we walk through things step by step I believe I can make a leading argument that the current implementation barely misses the faithfulness requirement in trivial ways.  First we accept that a language supporting object oriented programming like C++ is obviously meeting the requirement.  But we can look at C where we don't have OO classes, in that case we accept that a structure, without compiler enforced encapsulation, with functions that take that structure by reference is functionally equivalent and therefor pass the rules.  There exist languages, such as PHP, where one can convert between associative array and object notation pretty much freely, so I think it's easy to argue that an associative array is a structure.  What does the structure contain in C?  The size of the sieve and the array of bits determining if a number is prime.  Right now the bash script doesn't carry the size explicitly, but this could easily be changed by just writing it to bitArray (after all it's an associative array now).  Doing this would have no to negligible change in performance since we're already writing the size to a global and would mean that the variable bitArray contains the entire state.

Lets look at the dynamic allocation requirement.  Bash's array storage is dynamically allocated by definition.  Right now emptyBitArray discards all state, so the only point of contention I see is that it's possible it's reusing the hashmap structure and just clearing it vs constructing a new one.  This concern could be eliminated by just calling unset to remove the variable itself as well.  So in that sense we can allocate the "structure", populate it, and then free it.

But languages like C pass the structure by reference.  We can do that in bash as well.  We can treat the name of the variable as its pointer and use `local -n` to bind that pointer to a name that can be indexed on.  Now if we're treating the variable name as a pointer, do we need to have a unique one every run?  I would argue that it shouldn't matter, in C we assume that malloc is returning a unique pointer every time we call it, but there's nothing illegal that I'm aware of about a libc implementing an optimization where malloc looks at the last freed pointer and if the size is the same simply return that pointer again.  Thus while it would be possible to generate a unique variable name (say `local -A "bitArray_$count"` every iteration in bash, I don't think this is necessary.  Although it would have basically no performance penalty, I just don't know why it would matter.

So to recap here are the points of contention I see and all of them are trivially solved.
1. The size is stored in a separate global from bitArray.  Trivially solved by just storing in the array.
2. Bash could possibly be reusing the underlying hashmap structure.  Trivially solved with unset.
3. The "structure" isn't passed to the functions but referenced globally.  Trivially solved with references.  This would be a performance problem for the non-inlined implementations since having the pass by reference to gitBit/clearBit would be a significant performance hit.  However the change could be made to the inlined version with no significant change in performance.

Now personally I'm not certain the rules cover if point 3 needs to be solved.  In any case perhaps the requirement is "you'd have to use Bash Infinity," which is fine.  I wouldn't personally go through the effort of importing that here just to prove that Bash can be faithful, but if my interpretation of the rule is correct then crossing the line should be trivially obtainable.

rbergen (approved):
LGTM

marghidanu (approved):

PR #383, opened 2021-07-11
PrimeGo/solution_2: Cosmetic updates

Rellum:
Added a test and some cosmetic updates to the Go solution. No performance improvements, although I ran benchmarks to ensure no performance regression.

### DIFF
--- a/PrimeGo/solution_2/go.mod
+++ b/PrimeGo/solution_2/go.mod
@@ -1,0 +1,5 @@
+module prime_solution_2
+
+go 1.16
+
+require github.com/google/go-cmp v0.5.6
--- a/PrimeGo/solution_2/main.go
+++ b/PrimeGo/solution_2/main.go
@@ -1,41 +1,52 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
+	"log"
 	"math"
 	"math/bits"
 	"time"
 )
 
 var primeCounts = map[uint64]uint64{
-	10:        4,
-	100:       25,
-	1000:      168,
-	10000:     1229,
-	100000:    9592,
-	1000000:   78498,
-	10000000:  664579,
-	100000000: 5761455,
+	10:             4,
+	100:            25,
+	1_000:          168,
+	10_000:         1_229,
+	100_000:        9_592,
+	1_000_000:      78_498,
+	10_000_000:     664_579,
+	100_000_000:    57_61_455,
+	10_000_000_000: 455_052_511,
 }
 
-type Sieve struct {
+type sieve struct {
 	bits []uint8
 	size uint64
 }
 
-func (s Sieve) RunSieve() {
+func (s *sieve) RunSieve() {
 	var factor, start, step uint64
+
+	// We ignore all even numbers.
 	end := (s.size + 1) / 2
+
+	// The factors of the biggest number in the search range can't be bigger than
+	//  its square root, so we only need to cross off multiples of factors smaller
+	//  than this. Dividing by 2 because we're ignoring all even numbers.
 	q := uint64(math.Sqrt(float64(s.size)) / 2)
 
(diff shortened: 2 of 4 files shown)


PrimeGo/solution_2/go.mod
```diff
@@ -0,0 +1,5 @@
+module prime_solution_2
```
Rellum, line 1: Declaring a module makes using the tooling a lot simpler.

PrimeGo/solution_2/main.go
```diff
@@ -1,49 +1,60 @@
...
 var primeCounts = map[uint64]uint64{
-	10:        4,
-	100:       25,
-	1000:      168,
-	10000:     1229,
-	100000:    9592,
-	1000000:   78498,
-	10000000:  664579,
-	100000000: 5761455,
+	10:             4,
```
Rellum, line 14: I find the thousands separators easier to follow. Happy to revert if I'm in my own camp.

PrimeGo/solution_2/main.go
```diff
@@ -1,49 +1,60 @@
...
+	10_000:         1_229,
+	100_000:        9_592,
+	1_000_000:      78_498,
+	10_000_000:     664_579,
+	100_000_000:    57_61_455,
+	10_000_000_000: 455_052_511,
 }
 
-type Sieve struct {
+type sieve struct {
```
Rellum, line 25: types shouldn't be exported by default (this is a main package, so we wouldn't be able to reference this type from a different package, anyway).

PrimeGo/solution_2/main.go
```diff
@@ -52,35 +63,48 @@ func (s Sieve) CountPrimes() uint64 {
...
 
 	flag.Uint64Var(&limit, "limit", 1000000, "limit")
 	flag.DurationVar(&duration, "time", 5*time.Second, "duration")
 	flag.Parse()
 
-	stop := make(chan struct{})
 	passes := 0
 	start := time.Now()
-	time.AfterFunc(duration, func() { stop <- struct{}{} })
+	ctx, cancel := context.WithTimeout(context.Background(), duration)
```
Rellum, line 94: The canonical way of terminating processes using an external signal like the system clock is to use a context variable.

PrimeGo/solution_2/main.go
```diff
@@ -52,35 +63,48 @@ func (s Sieve) CountPrimes() uint64 {
...
+	defer cancel()
 
-loop:
 	for {
-		select {
```
Rellum, line 76: Removed the channel and select - it isn't necessary when there's no concurrency (although the standard library's context implementation does use a done channel, so this is really just a refactor).

PrimeGo/solution_2/main_test.go
```diff
@@ -0,0 +1,45 @@
+package main
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestCountPrimes(t *testing.T) {
```
Rellum, line 10: A couple of simple unit tests to make future improvements easy to make and review. 

I added a benchmark, too, but haven't committed this as I'm not sure it adds value.

PrimeGo/solution_2/main.go
```diff
@@ -52,35 +63,48 @@ func (s Sieve) CountPrimes() uint64 {
 	return t
 }
 
-func (s Sieve) ValidateResults() bool {
+func (s *sieve) ListPrimes() []uint64 {
```
Rellum, line 66: Add this mostly for my own benefit, to understand how the implementation works. It outputs the actual prime numbers after the sieve has been executed.

rbergen:
As is by now well documented in earlier PRs, we don't apply cosmetic updates to existing solutions.

PR #386, opened 2021-07-11
Improve performance R solution_1 by declaring member variables ANY

fvbakel:
## Description
Align with tweaks from PR #334 and make member variables of type ANY. Making the member variables of type ANY is the biggest improvement, the other changes have less impact.

## Contributing requirements


rbergen:
@fvbakel Why is this a new PR and not an update to #334?

rbergen:
...because it's an update to solution_1. So, nevermind.

PR #390, opened 2021-07-12
Comprehensive C++ solution finding optimal implementations by exhaustive search

BlackMark29A:
## Description

A generic templated algorithm that can be used to generate implementations using a configurable set of optimizations.

By generating all possible combinations of optmizations I found the configurations which yield maximum performance and managed to beat (or tie) all existing solutions in each category, at least on my hardware.

Also includes a non-faithful algorithm that generates the sieve at compile-time, further impoving upon the `constexpr` solution by @flo80.

## Contributing requirements


fcolecumberri:
It's impressive how the single threaded version of this solution is 3 times faster than the solution_1 (benchmarked on my own pc).
and the pre-generation algorithm is just impressive, 200+ times faster proving that C++ power is on the hands of the programmer.

BlackMark29A:
Well, the pre-generated version is kind of cheating, because the sieve is just copied to dynamic memory from the compile-time generated sieve. So this is really only limited by how fast `memcpy` is. But at least it should be a good upper bound. You can't go any faster, unless you make the `runSieve` operation a no-op.

rbergen:
@BlackMark29A I'm asking an open question here: I'm wondering if some of your implementations could be converted to improvements to the existing C++ solutions? We already have 3 C++ solutions in place, and with the language and solution count we're currently looking at, we need to be mindful of this.

BlackMark29A:
I fully understand that it should be a priority to improve existing solutions rather than adding new ones. And some of the things that my solution shows can absolutely be applied to other solutions as well, and not just the C++ ones. However, the methodology of my solution is very different than that of all other solutions, at least as far as I'm aware, and therefore, I would argue that adding it as a new solution is merited.
The whole point of my solution is not that C++ can be used to implement an efficient solution, but that knowing which implementation would be efficient is not trivial, not intuative, and also there is no overall best solution. It is very much dependant on the compiler and hardware used, and so my solution aims to be able to find the most efficient solution for any given hardware by generating and testing all possible solutions and selecting the best one(s). Taking just the results and applying it to existing solutions would lose this methodology, which I think merits inclusion as its own approach in the repository.

rbergen:
@marghidanu I'm wondering what's your opinion on this?

marghidanu:
@rbergen I agree with @BlackMark29A changing the previous implementations is not really suitable, I believe that this is a separate solution and should be treated as such.

PrimeCPP/solution_4/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM ubuntu:21.04 AS build
+
+RUN apt update -qq && apt install -y g++-11 clang-12 make
```
rbergen: Please use apt-get instead of apt.

PrimeCPP/solution_4/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+WORKDIR /opt/app
+
+COPY Makefile ./
+COPY *.cpp ./
+COPY *.hpp ./
+
+ENV CXX_CLANG=clang++-12
+ENV CXX_GCC=g++-11
+
+ENTRYPOINT [ "/usr/bin/make", "-j", "run_bench" ]
```
rbergen, line 28: Would it be possible to separate the build (compile) from the actual run? With the Dockerfile being what it is now, the compile would be part of every benchmark run.

rbergen (commented):
Please sync your fork with the current state of this repo. It will re-enable our CI to run on your solution.

PrimeCPP/solution_4/Dockerfile
BlackMark29A: Done.

PrimeCPP/solution_4/Dockerfile
BlackMark29A: Done.

BlackMark29A:
I also implemented additional storages to further optimize the results. The code is now also commented a lot better, to try to explain the things that are not obvious. I also added box plots that compare my solution against the current best solutions in the repository and managed to beat them in all but 1 category. My solution is still not optimal in the "8-bit base algorithm" category, although the difference is very small.

rbergen (approved):
LGTM. @marghidanu The PR includes a number of binary files (images), which we don't tend to merge. As these binaries are verifiably images, included in the README.md, I'm inclined to allow this. Please feel free to disagree when you review. :)

marghidanu:
I tend to agree about the binary images, can these be hosted somewhere else?

BlackMark29A:
The images are now hosted in my fork and only linked to from the README.

rbergen:
Looks even better to me, now :) @marghidanu can you take one last look, too?

marghidanu (approved):

PR #391, opened 2021-07-12
Optimizations in VB.NET metrics

NFITC1:
## Description
Tiny optimizations have been made to improve performance.
-Bitshifting is always faster than involving the division method of a processor, possibly as much as 10x faster. I would recommend that each language that supports bitshifting implement that. Integer division by a power of two is always simpler by performing a lbitshift.
-Using the stopwatch object instead of the function call to the system's timer is a much cheaper operation. Querying any system property is possibly an expensive process. The constructor for the stopwatch already does much of the system metrics so it doesn't impact the performance of the actual calculation of primes. Removing the querying of the system time and subsequent subtraction improves performance by allowing more time for additional iterations to perform.
-Moving the loop condition to the end allows the procedure to begin as soon as possible after timing begins. We already know we want to enter the loop, we want to know if we should continue to loop (pre-conditional loop vs post-conditional loop). Looks functionally similar on paper, except for this post conditional loop I eliminated one unnecessary comparison that does affect time.

## Contributing requirements

### DIFF
--- a/PrimeBASIC/solution_2/PrimeVB.vb
+++ b/PrimeBASIC/solution_2/PrimeVB.vb
@@ -28,7 +28,7 @@ Module PrimeVB
 			For factor = 3 To sieveSqrt Step 2
 
 				For number = factor To sieveSqrt Step 2
-					If primesArray(number \ 2) Then
+					If primesArray(number >> 1) Then
 						factor = number
 						Exit For
 					End If
@@ -37,7 +37,7 @@ Module PrimeVB
 				If number > sieveSqrt Then Exit For
 
 				For number = factor * 3 To sieveSize Step factor * 2
-					primesArray(number \ 2) = False
+					primesArray(number >> 1) = False
 				Next
 
 			Next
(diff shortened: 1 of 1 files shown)


rbergen:
Could you provide some numbers on the actual performance improvement these changes make?

NFITC1:
I can't run Docker because it's a VM and my laptop doesn't support that, but both of these combined increased pass count from an average of 550 to nearly 680 while running in debug mode. That was running ten consecutive times each before and after alterations.

rbergen:
I've actually looked at \ 2 vs >>1 and DateTime vs Stopwatch before, and I've found no advantage in performance either way. When testing your changes vs my original solution, I don't see a performance impact either; the variations for both implementations average out to a difference that is statistically irrelevant by any standard. 

I haven't looked at the IL let alone what the JIT compiler makes of that, but my guess is that the \ 2 is optimized to >> 1 somewhere in the compilation chain. Conclusions on the comparison between Stopwatch vs DateTime seem to differ from one article to another, but as I said the performance gain of the former over the latter is impossible to identify as far as I can tell.

Therefore, I think I will stick to the premise of this solution staying as close to Dave's C# solution(_2) as possible, at least until changes are proposed that have a positive impact on performance that I can confirm.

NFITC1:
It's likely a compiler option in my VS that is causing the discrepancy on my end especially since I'm testing these in debug mode.

The DateTime vs Stopwatch probably differs greatly from one processor to the next. It only yielded about 3% more passes on its own. That's probably optimistic too. I suppose it depends on what you're doing with the classes that would make one more advantageous over the other.

I can't attest to the compiling of \ 2 into >> 1. Every independent test I've run shows that bitshifting is significantly faster so it is treated differently at runtime.

I believed the goal of this project was to determine which language can achieve the goal of calculating primes as often as possible in 5 seconds using whatever tools the language offered. These changes definitely increased the pass count. Based on my metrics (as measured by the VS) reduced tick count by several thousand ticks per pass.

I'm going to reexamine these results and see if there's something I'm doing differently. I'll have to compile both in release mode and see if there are any differences.

NFITC1:
Update: Yikes! I compiled both my changes and the base code in release and ran them independently 10 times. My changes are actually a performance HIT and made the process about 20% slower averaging out just under a thousand passes where the base was pushing 1.2k passes. That's very peculiar to me! My changes perform (comparatively) better in debug mode and I thought that would translate into better release performance. I'll close this pull request and see if I can find optimizations elsewhere.

PR #378, opened 2021-07-11
Added Solution_3 to PrimeGo

rmasci:
## Contributing requirements

* [ x] I read the contribution guidelines in CONTRIBUTING.md.
* [x ] I placed my solution in the correct solution folder.
* [ x] I added a README.md with the right badge(s).
* [ x] I added a Dockerfile that builds and runs my solution.
* [ x] I selected `drag-race` as the target branch.
* [ x] All code herein is licensed compatible with BSD-3.

PR #367, opened 2021-07-09
zanicar-go multi threading solution_3

zanicar:
## Description
Adapted 1-bit go solution_2 by ssovest to include multi threading.

## Contributing requirements

* [X ] I read the contribution guidelines in CONTRIBUTING.md.
* [ X] I placed my solution in the correct solution folder.
* [ X] I added a README.md with the right badge(s).
* [ X] I added a Dockerfile that builds and runs my solution.
* [ X] I selected `drag-race` as the target branch.
* [ X] All code herein is licensed compatible with BSD-3.

PR #392, opened 2021-07-12
Moved to solution 4, added loop count

rmasci:
## Contributing requirements

* [ x] I placed my solution in the correct solution folder.
* [ x] I added a README.md with the right badge(s).
* [ x] I added a Dockerfile that builds and runs my solution.
* [ x] I selected `drag-race` as the target branch.
* [ x] All code herein is licensed compatible with BSD-3.

### DIFF
--- a/PrimeGo/solution_4/Dockerfile
+++ b/PrimeGo/solution_4/Dockerfile
@@ -1,0 +1,9 @@
+FROM golang:1.16-alpine3.13
+
+WORKDIR /opt/app
+
+COPY main.go .
+
+RUN go build main.go
+
+ENTRYPOINT [ "./main" ]
--- a/PrimeGo/solution_4/README.md
+++ b/PrimeGo/solution_4/README.md
@@ -1,0 +1,30 @@
+# Go Prime Sieve solution by Richard Masci (rmasci)
+
+The reason I am highlighting this is show how easy it is to write concurrent in Go as concurrency, and Go was designed to be concurrent -- it wasn't added in later. Today computers don't scale like they did when most of these languages were written, where you could wait a few months and you had a faster processor and more RAM. Today, it's more about the number of cores.
+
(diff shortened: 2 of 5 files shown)


rbergen:
Could you indicate why this should be a new solution instead of an improvement on solution_3?

rmasci:
> Could you indicate why this should be a new solution instead of an improvement on solution_3?

I wrote solution_4 before solution_3.  I submitted it originally as solution_3, that pull request was rejected because another had submitted a pull request that I wasn't aware of -- that today is solution_3. So I pulled my version, made it solution_4. I didn't build off of solution_3 at all, and the way I've implemented goroutines is a lot different.

Since the submission I've further 'tweaked' the program so that I can limit the number of CPUs and by default it's 1. It also shows the number of go routines -- it's a bit more interesting as you can see when the 'Stop' is issued, and the go routines filtering down.  This solution runs faster than solution 1 and 2, while still only using 1 cpu. When you open it up to all cpus on my system it runs faster than solution_3 -- but again is different in that when one go routine ends, it starts another, and tries to keep a number of them running at all times -- which can be shown in a version I've not uploaded yet.

rbergen:
The PR that was merged was #367, which was open before you opened #378, meaning you could have been aware of it if you had looked at the list of open PRs. If I had seen your original PR when #367 was merged then I would have told you to discuss how to integrate your proposed changes into that PR. With that PR now merged, the goal should be improving its performance, rather than moving towards adding a new solution.

marghidanu:

rbergen:
@rmasci Just to clarify my previous comment: I will not move towards merging this as a new solution unless the scenario of improving solution_3 has been exhausted as a possibility.

rbergen:
Closing for reason indicated. If desired, it can be reopened with my previous comment in mind.

PR #395, opened 2021-07-12
primes.pl: gate printout behind $show_results

greg-kennedy:
The existing way of doing this was interfering with my profiling output, changed to just a block if() statement instead.

### DIFF
--- a/PrimePerl/solution_1/primes.pl
+++ b/PrimePerl/solution_1/primes.pl
@@ -53,17 +53,19 @@ package PrimeSieve {
     sub print_results {
         my ( $self, $show_results, $duration, $passes ) = @_;
 
-        print "2, " if ($show_results);
+        if ($show_results) {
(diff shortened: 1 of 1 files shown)


marghidanu:
I don't see any improvements here.

greg-kennedy:
Profiling with Devel::NYTProf shows 500,000 calls made during the `for` loop that never actually print anything.  This makes the prof. output harder to parse when attempting to do optimization work.

It also spares 1.5 seconds of time doing nothing at the end of the job.

marghidanu:
`print_results` is not part of the benchmark it just prints the results what's the advantage of optimizing this?

greg-kennedy:
because it makes output from devel::nytprof harder to read by adding a ton of noise

marghidanu:
You can also comment on the function since is not relevant to the output while running your profiling.

PR #397, opened 2021-07-12
Make prime_sieve type-stable

lungben:
## Description
`BitArray` is an abstract type because dimensionality is not defined. Change it to `BitVector` for type stability.
This gives a speedup by a factor of 5 on my machine (350 passes before compared to 1500 passes after this change).
See https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-abstract-container for background info.

## Contributing requirements


lungben:
Btw even after this change this is still much slower than solution_2 for Julia, but it was a quick-win therefore I posted it.

rbergen:
Tagging @dcbi and allowing time for them to review and respond to this PR on their original implementation, if they want to do so.

lungben:
To be included into #401

Issue #382, opened 2021-07-11
Comparability between implementations


dnlmlr:
## The problem
I just took a look at the implementations for C, C++ & Rust as those are commonly known for getting highly optimized versions in benchmarks. And indeed there are quite highly optimized versions to be found.
But after looking at the C implementations, I was wondering at what point the implementations are no longer comparable. 

In my opinion it is not really possible to compare the languages in __any meaningful__ way,  when some versions optimize the algorithm itsef in the way that most of the C implementations do. Of course the meaningful comparison is debatable in any case, but I would argue that it may still produce interesting results when all implementations use the same base algorithm.

The _original_ implementation was optimized to skip even numbers which is quite reasonable and results in no cpu time overhead while calculating or counting since it just increments by 2 instead of 1. 

Almost all C versions on the other hand further optimized the skipping part, with not only skipping even numbers but instead also skipping higher multiples of other primes (2, 3, 5) or even skipping multiples of all primes up to 13 in one example. This of course provides a significant boost to performance (at least the first few) but is now no longer comparable to the original version (and in turn any language that doesn't use the same optimizations). This optimization is of course quite clever and would be valid in any other case, but here it is basically just using existing knowledge to highly decrease the workingset to a much smaller size than the other implementations have to deal with. (Just skipping the multiples of 3 decreases the __unchecked values__ from about `500_000` to only about `166_667`)

Besides the counting part now no longer being trivial (and not being timed), this of course makes C/C++ and in turn all languages with implementations that use this optimization seem much faster in relation to other languages than they are. 

A language that is inherntly slower than another language could implement this change an in turn __look faster__ on paper, while the other just doesn't have this optimization.

## Considerations for a solution
Ok I don't really have a solution, but here are some of my considerations: 
- Constructs and optimizations that a language provides are fine and wanted (for example using iterators / filters in rust instead of the __normal__ for loops, or C++ vector find) since those represent the languages features
- The base algorithm should have to be the same if any cross language comparison is wanted (outer loop, find next factor, set values for this factor, increment factor)
- The factor checking __HAS__ to start with 3 and not 5, 7, or even higher to stay comparable
- The way how the smaller tasks in the algorithm, like setting / clearing or storing the values can be changed as needed

flo80:
> ## Considerations for a solution
> Ok I don't really have a solution, but here are some of my considerations:
> 
[...]

PR #398, opened 2021-07-12
Clarifies that class must be used if available

rbergen:
Clarifies the contributing guidlelines on the points that:
- a class is the (only) allowed construct if a language provides one
- if it isn't but alternatives are, the one closest to the class concept must be used
- a class instance needs to be created _from scratch_ for every sieve run

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -161,7 +161,7 @@ If you choose to include badges in your `README.md`, then:
 At a technical level, an implementation is considered faithful if it complies with the following:
 
 * It uses no external dependencies to calculate the actual sieve.
-* It uses a class to encapsulate the sieve, or an equivalent feature in your language. This class must contain the full state of the sieve. Each iteration should re-create a new instance of this class.
+* It uses a class to encapsulate the sieve, or a (closest) equivalent feature in your language if a class construct is not available. This class must contain the full state of the sieve. Each iteration should re-create a new instance of this class from scratch.
 * The sieve size and corresponding prime candidate memory buffer (or language equivalent) are set/allocated dynamically at runtime. The size of the memory buffer must correspond to the size of the sieve.
 * It conforms to the base [rules](#Rules).
 


bbartels:
I really don't want to come across as pedantic, but this class rule seems rather arbitrary. There is no language independent definition of a class. I feel like the primary objective here was to require: _some language construct that allows for the encapsulation of state_. Which is a significantly more precise definition, than class.

I don't particularly care about this rule being one way or another, just seems like a well defined ruleset would avoid confusion going forward :)

rbergen:
Of course I accept your opinion as your opinion, but the new phrasing of the rule actually more accurately expresses the intent of the original rule. Generally speaking, classes are the most "heavily burdened" by OO concepts (like support for inheritance and default maximum isolation) in those languages where there is a difference between it and similar constructs. In languages where there is no performance penalty to using classes, arguing against them is moot in the context of this repo's primary goal, The 3 original languages use the class construct, so that is the premise for the project. 
Maybe we could be accused of being too lenient by allowing similar constructs in languages that don't actually have classes, but excluding all of them would have excluded (too) many languages that make this repo interesting to many people, not in the last place Dave himself.

Concerning avoiding confusion: we're really doing our best and that on a daily basis, I guess we'll have to conclude that with this many people engaging with the project, it is impossible for a very finite group of people to fully anticipate the collective creativity of a much larger, and constantly growing, group of people. As this PR's description states, its aim is to reduce confusion on one specific point where it has clearly arisen.

sglienke:
Another rule that makes no sense imho. If you want to enforce this, then also force that the instance has to be heap-allocated because most languages cannot stack allocate like C++ for example can. To avoid pointless heap allocations structs and alike can be used in other languages.
Some of those rules are way too coupled towards implementation details of the original languages and impose irrational requirements on other implementations. The rules should thus be defined in a more abstract and language agnostic way.

rbergen:
As I explained in #382, the rules are indeed strongly inspired by the original solutions, and that by choice.

marghidanu (approved):

PR #399, opened 2021-07-12
Another improvement to dart solution_1

mmcdon20:
## Description
I was able to improve the times even further on dart solution_1 by initializing the bits list to half the sieve size, and adjusting the indexing accordingly.

## Contributing requirements


rbergen:
Tagging @eagerestwolf and allowing time so they can review and respond to this PR on their original solution, if they want to do so.

mmcdon20:
CI / build (PrimePascal/solution_2) Still hasn't started after about 25 minutes.

mmcdon20:
I have also implemented a 1-bit implementation. It is considerably slower than the 8-bit version on my computer, but apparently 8-bit implementations won't make the leaderboard (https://github.com/PlummersSoftwareLLC/Primes/issues/382#issuecomment-878207639). 

In addition I have also implemented a parallel version. Technically this is just a different main method, the parallel version reuses the existing 8-bit sieve implementation. I could also do a parallel version of the 1-bit implementation. 

Should I add these as a separate pull request, or add it in here? Should it be part of solution_1 or should I create a solution_2?

mmcdon20:
@rbergen should I go ahead and add the additional implementations onto this pull request? I currently have it so that the implementations are all part of solution_1 and the docker file will now run all 4 versions (8-bit single-core, 8bit multi-core, 1bit single-core, 1bit multicore) in the same image.  @eagerestwolf any thoughts?

Edit: Taking a closer look at the rules regarding parallelism, I don't think the solutions I have qualify since they only have logic for parallelization in the main method and not the PrimeSieve class itself. Given dart's somewhat unorthodox approach to parallel programming I don't think it would be very feasible to implement a PrimeSieve that is itself parallel (at least not in a way that is efficient), so I will not include it. I would still however like to include the single-threaded 1-bit implementation.

rbergen:
> should I go ahead and add the additional implementations onto this pull request?

Please do.

> I don't think the solutions I have qualify since they only have logic for parallelization in the main method and not the PrimeSieve class itself.

From a rules perspective, that doesn't matter: either is possible. If an implementation uses more than 1 concurrent thread to calculate the number of sieve runs in 5 seconds it reports in the output, we label it "parallel".

mmcdon20:
@rbergen thanks for the clarification. In that case I have pushed an update that includes all of the implementations.

julemand101:
@mmcdon20 
Don't know if it is allowed but another improvement would be to have the `Uint8List` as a single one-time instance in the main loop and then send it into the `PrimeSieve` as an argument. You can then zero the `Uint8List` with: `list.fillRange(0, list.length, 0)` before use if it has been used before.

When doing this, we can practically prevent memory allocations (and GC) which seems to be something the program are spending a lot of time on. In my tests, it is faster to `fillRange` an existing memory allocation instead of asking for a new one. But I have not tested with the changes in this pull request.

mmcdon20:
@julemand101 I have given this a try, and it does seem to outperform the original when running on the dart vm, but strangely it performs worse when compiled to an exe. I am also uncertain as to weather this would be considered faithful or not.

julemand101:
@mmcdon20 
I can reproduce your observation about the performance after the code is being compiled. Since I guess the speed tests are being executed without the Dart VM I agree it does not really make sense to introduce my suggested improvement. :)

mmcdon20:
@julemand101 maybe it would make sense to include a separate solution that targets the dart vm? 
@rbergen is it worth including versions that target different runtimes even if the characteristics remain the same?

I had also previously messed around a bit with compiling using dart2js. The current implementation doesn't compile to javascript because it imports `dart:io`. So you have to change all the print statements to get it working, but I was surprised at how good the javascript performance was. It performed about even with the exe version (maybe slightly worse).

Maybe it would fit better as a solution_2?

rbergen:
> is it worth including versions that target different runtimes even if the characteristics remain the same?

@mmcdon20 no, we're drag-racing languages, not runtimes. Concerning the runtime, my advice would be to pick the fastest that is generally available.

rbergen:
@mmcdon20 Is this PR now stable?

mmcdon20:
I'm not planning on adding anything else to the PR at this point.

rbergen:
Ok, then let's see if @eagerestwolf wants to comment on the PR as it stands.

eagerestwolf:
While I don't necessarily agree with having multiple versions in the same solution, as long as the admins are okay with it, it looks fine to me.

rbergen:
@eagerestwolf We actually prefer this if all implementations within the one solution (can) run within the same docker image. It saves build time and resources when running the benchmark.

eagerestwolf:
Fair enough, if that's how you want it, then I won't be the one to reinvent the wheel.

mmcdon20:
I caught a bug when setting the size of the Uint8List in the 1-bit version. Everything should be ready to go now.

PrimeDart/solution_1/README.md
```diff
@@ -1,10 +1,30 @@
...
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-1-green)
+
+### PrimeDartParallelOneBit.dart
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
+![Parallelism](https://img.shields.io/badge/Parallel-yes-green)
+![Bit count](https://img.shields.io/badge/Bits-1-green)
```
rbergen: Please add one consolidated line of badges directly underneath the README title that includes one of each of the badges for the combination of implementations in this solution. It's perfectly ok to leave the current per-solution badges underneath it.

PrimeDart/solution_1/bin/PrimeDartOneBit.dart
```diff
@@ -0,0 +1,234 @@
...
+    // print requests to keep the code to under 80 columns (personal preference).
+    stderr.write('Passes: $passes, Time: $duration, ');
+    stderr.write('Avg: ${duration / passes}, Limit: $_sieveSize, ');
+    stderr.write('Count1: $count, Count2: ${countPrimes()}, ');
+    stderr.write('Valid: ${_validateResults()}\n');
+
+    // These 2 lines are for the drag race format
+    stderr.write('\n');
+    stdout.write(
+        'eagerestwolf&mmcdon20;$passes;$duration;1;algorithm=base,faithful=yes,bits=1\n');
```
rbergen: I'm commenting here, but it applies to the combination of implementation output lines: please distinguish between the implementations by giving each a unique label (adding `_<some_keyword>` seems sensible, here).

mmcdon20:
@rbergen updated. Let me know if there is anything else.

rbergen (approved):
LGTM

rbergen:
@mmcdon20 Due to other PRs (#461 in particular) having been merged, there is now a merge conflict in your Dockerfile. I created a Dockerfile that resolves this conflict in a way that combines the merged changes with yours. As I couldn't open a PR on your fork, I'm adding a link to it here: [Dockerfile](https://github.com/rbergen/Primes/blob/dart_dockerfile/PrimeDart/solution_1/Dockerfile)

mmcdon20:
@rbergen your suggested changes to the Dockerfile didn't work. I've submitted my own update to the Dockerfile that should work. Let me know if there is anything else I needed to do to get this merged. This PR has been sitting here a while.

rbergen:
> @rbergen your suggested changes to the Dockerfile didn't work. I've submitted my own update to the Dockerfile that should work.

Could you provide a little more information on how it didn't work? I obviously tested it locally before committing it to my own fork.

mmcdon20:
@rbergen  this is what I get when I use your suggested Dockerfile:
```
PS C:\Users\mmcdo\Desktop\Primes> cd PrimeDart
PS C:\Users\mmcdo\Desktop\Primes\PrimeDart> cd .\solution_1\
PS C:\Users\mmcdo\Desktop\Primes\PrimeDart\solution_1> docker build -t primes-dart .
[+] Building 28.4s (19/19) FINISHED
 => [internal] load build definition from Dockerfile                                     0.0s 
 => => transferring dockerfile: 719B                                                     0.0s 
 => [internal] load .dockerignore                                                        0.0s 
 => => transferring context: 2B                                                          0.0s 
 => [internal] load metadata for docker.io/library/ubuntu:20.04                         10.8s 
 => [internal] load metadata for docker.io/google/dart:2.13.4                           10.8s 
 => [build  1/11] FROM docker.io/google/dart:2.13.4@sha256:a7f4237f214cae4eb16b2f2474fb  0.0s 
 => [internal] load build context                                                        0.0s 
 => => transferring context: 1.36kB                                                      0.0s 
 => [runtime 1/3] FROM docker.io/library/ubuntu:20.04@sha256:b3e2e47d016c08b3396b5ebe06  0.0s 
 => CACHED [build  2/11] WORKDIR /app                                                    0.0s 
 => CACHED [build  3/11] COPY pubspec.* /app/                                            0.0s 
 => CACHED [build  4/11] RUN pub get                                                     0.0s 
 => [build  5/11] COPY . /app                                                            0.1s 
 => [build  6/11] RUN pub get --offline                                                  0.6s 
 => [build  7/11] RUN dart compile exe /app/bin/Runner.dart -o /Runner                   3.3s 
 => [build  8/11] RUN dart compile exe /app/bin/PrimeDart.dart -o /PrimeDart             3.1s 
 => [build  9/11] RUN dart compile exe /app/bin/PrimeDartOneBit.dart -o /PrimeDartOneBi  3.1s 
 => [build 10/11] RUN dart compile exe /app/bin/PrimeDartParallel.dart -o /PrimeDartPar  3.4s 
 => [build 11/11] RUN dart compile exe /app/bin/PrimeDartParallelOneBit.dart -o /PrimeD  3.4s 
 => CACHED [runtime 2/3] WORKDIR /app                                                    0.0s 
 => ERROR [runtime 3/3] COPY --from=build /app/Runner /app/PrimeDart* /app/              0.0s 
------
 > [runtime 3/3] COPY --from=build /app/Runner /app/PrimeDart* /app/:
------
failed to compute cache key: "/app/Runner" not found: not found
```

rbergen:
That's not the same Dockerfile. That log includes slashes in the output files of the compile (`/Runner`, `/PrimeDart`, etc.) that are not in the Dockerfile I included a link to:
image

mmcdon20:
@rbergen ah, I must have missed that you changed that part... You want me to submit your version?

mmcdon20:
@rbergen I just submitted your changes to the Dockerfile.

rbergen:
> @rbergen I just submitted your changes to the Dockerfile.

Thank you, I appreciate that. We've recently reviewed the repo's Dockerfiles and implemented linting as a step towards increasing stability/predictability of CI and benchmark runs. This solution's Dockerfile was one of the ones we modified.

Concerning your other question: there is currently nothing I can point at that you would need to change. My only request is to have a little more patience. As your PR includes (a) new implementation(s), I prefer to have two maintainers review it before I merge. I'd like to allow a few more days for @marghidanu to do his review, if he has the opportunity.

PR #400, opened 2021-07-12
Faster luajit solution

mipett:
## Description
About 2.25 x faster than the previous solution. Also added an accurate timer.

## Contributing requirements


rbergen:
If this indeed does improve the performance of the existing luajit solution, then please change your PR so it applies your performance improvements to that existing solution.

rbergen:
Closing this for reasons indicated. It can be reopened once requested changes have been applied.

PR #402, opened 2021-07-12
Remove hash lookups from perl solution

bbrtj:
## Description
From 26 to 33 passes on my machine (~25% speedup)

## Contributing requirements

### DIFF
--- a/PrimePerl/solution_1/primes.pl
+++ b/PrimePerl/solution_1/primes.pl
@@ -31,23 +31,28 @@ package PrimeSieve {
 
         my $factor = 3;
         my $q      = sqrt $self->{sieve_size};
+        my @bits   = @{$self->{bits}};
+        my $ss     = $self->{sieve_size};
+        $#bits     = $ss;
+
         while ( $factor <= $q ) {
-            my $num = $factor;
-            for ( my $num = $factor ; $num < $self->{sieve_size} ; $num += 2 ) {
-                unless ( $self->{bits}[$num] ) {
+            for ( my $num = $factor ; $num < $ss ; $num += 2 ) {
+                unless ( $bits[$num] ) {
                     $factor = $num;
                     last;
                 }
             }
 
-            my $num2 = $factor * $factor;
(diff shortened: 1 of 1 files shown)


rbergen:
We already have #352 open aiming to improve the Perl solution. Please discuss in that PR how to add your suggestions to it.

bbrtj:
@rbergen that other PR creates a new solution (`solution_2`) which, while being the same algorithm, uses completely different storage. My PR does not aim to create a new solution, but speed up the current one just a bit. Also, the other PR does not need this kind of improvements, as there are no hash lookups and close to no dereferencing - I've already taken a look at it

rbergen:
@brtastic That's obviously a fair point. @marghidanu, as this is a PR aiming to change your original implementation, I'm interested in your opinion about it.

marghidanu:
Let’s keep the solution as is and merge #352 as a second optimized solution.

marghidanu:
@rbergen my opinion seems to be unpopular. We will probably have two identical solutions, but if you want to merge this then do it.

bbrtj:
@marghidanu My pull request does not change how this script works, it only moves some dereferencing + hash lookups into lexical variables, so that those (not free) operations can only be done once, instead of in every loop iteration (which included inner loops as well). Another little bit of improvement is to preallocate the array (with changing the last element index, `$#array`), which also helps perl not to waste time doing stuff over and over again.

That other PR's solution is completely different in terms of how it works, even though the algorithm is the same. The only reason I see for not merging my PR is to leave the current solution as "slow, but readable" which pushes array performance to the limit, while the other one is "fast, but tricky", which uses strings. However, no one gives points for style here, and really I don't think I made the solution that much less readable.

By denying this pull request (without disproving that it does improve performance, as I claim) the whole point of this competition will be defeated. Why should I bother min-maxing the performance and even looking at the results if I know that valid performance improvements are not accepted?

marghidanu:
Yes, you're right. Let's merge this.

PR #403, opened 2021-07-13
IDL solution by KrizTioaN

kriztioan:
## Description
Interactive Data Language ([IDL](https://www.l3harrisgeospatial.com/Software-Technology/IDL)) solution. A free and open-source IDL compiler is available as [GDL](https://gnudatalanguage.github.io/index.html).

## Contributing requirements


rbergen:
@kriztioan  The build of the docker image took almost 4 minutes to complete on my machine. I think that means we will need a base image with GDL already installed as the foundation for this solution's Dockerfile.

kriztioan:
@rbergen Yes, I had the same on my machine. How do I go about creating and making available a base image? I did see there are a number of repositories on PlummersSoftwareLLC that seem to do exactly that for other languages. Is this perhaps something you can do and then I can update the DockerFile to use primeimages/gdl:1.0.0-rc.3? Its generation should be straightforward:

```
FROM fedora

RUN yum update -q -y \
    && yum install -q -y gdl
```


Bij voorbaat dank!

rbergen:
@kriztioan Yes, I can help with that. I was wondering if there is a specific reason for using fedora? I would also prefer to pin down on a specific version of whatever Linux distro serves as the base image of the base image. Would that be possible?

kriztioan:
@rbergen Great! I use Fedora because per the GDL [website](https://github.com/gnudatalanguage/gdl/wiki/GDL-on-Linux) it is the only distribution that maintains and has an up-to-date GDL-package, and has a readily available Docker-image. You can certainly pin down a specific Fedora version; 34 is latest and the one I'm running. The DockerFile would then look like(?):

```
FROM fedora:34

RUN yum update -q -y \
    && yum install -q -y gdl
```

Groetjes!

rbergen:
That was a nice one! The docker buildx build of the arm64 image took a cool 40 minutes. :)

In any case, it's now available at https://hub.docker.com/repository/docker/primeimages/gdl. Do note that I removed the dot (.) between rc and 3 in the version tag.

Doei! :)

kriztioan:
@rbergen Wow! The image also weighs in at a hefty 1+GB. Thank you very much for your help, work, and patience! I've updated the Dockerfile to use the base primeimages/gdl:1.0.0-rc3 image and successfully ran it on my machine:

```
$ docker build -t gdl
...
$ docker run --rm gdl 2>>/dev/null
kriztioan;314;5.012805;6;algorithm=base;faithful=yes;bits=8
```

Nogmaals dank en tot laterz!

rbergen:
@kriztioan Just checking if this PR is now stable?

kriztioan:
> @kriztioan Just checking if this PR is now stable?

@rbergen Yes, I'm done fiddling!  Please go ahead and start the workflow/review/close/etc. the pull request.

Hartstikke bedankt voor alle hulp! 🥇

rbergen:
Please combine both implementations into one solution. It saves docker build time and resources during the benchmark runs. If needed, you can use PrimeBASIC/solution_1 as an illustration of my request or even source of inspiration.

marghidanu:

kriztioan:
> Please combine both implementations into one solution. It saves docker build time and resources during the benchmark runs. If needed, you can use PrimeBASIC/solution_1 as an illustration of my request or even source of inspiration.

@rbergen That makes sense. I've combined both implementations into one solution following PrimeBASIC/solution_1 as an example. Bedankt!

> Please tick the checkboxes using [x] without spaces in between. More info here: https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

@marghidanu Done!

PrimeIDL/solution_1/primeidl_1bit.pro
```diff
@@ -0,0 +1,132 @@
...
+  passes = 0LL
+  time_start = SYSTIME(/SECONDS)
+
+  WHILE 1 DO BEGIN
+    sieve = OBJ_NEW('PRIMESIEVE', 1000000LL)
+    sieve->RUN
+    time_end = SYSTIME(/SECONDS)
+    passes++
+    IF time_end - time_start GE 5D THEN BEGIN
+        PRINT,FORMAT='(A0,";",I0,";",F0,";",I0,";algorithm=",A0,";faithful=",A0,";bits=",I1)', $
```
rbergen: please convert the separator between the tags (algorithm, faithful, bits) from semicolon (;) to comma (,). Semicolons separate mandatory positional fields, the commas are used for optional "named" values.

PrimeIDL/solution_1/primeidl_idlway.pro
```diff
@@ -0,0 +1,116 @@
...
+  passes = 0LL
+  time_start = SYSTIME(/SECONDS)
+
+  WHILE 1 DO BEGIN
+    sieve = OBJ_NEW('PRIMESIEVE', 1000000LL)
+    sieve->RUN
+    time_end = SYSTIME(/SECONDS)
+    passes++
+    IF time_end - time_start GE 5D THEN BEGIN
+        PRINT,FORMAT='(A0,";",I0,";",F0,";",I0,";algorithm=",A0,";faithful=",A0,";bits=",I1)', $
```
rbergen: See previous comment.

kriztioan: done!

PrimeIDL/solution_1/primeidl_1bit.pro
kriztioan: done!

rbergen (approved):
LGTM

marghidanu (approved):

PR #404, opened 2021-07-13
PHP - remove the even number check

B-Sterner:
## Description

Watched the video "EW: What is the FASTEST Computer Language - 45 Tested: Round Two! (E02)" today and I see what he means about the even checks in both gitBit and clearBit methods. added += 2 to skip even numbers in the first for loop and factor is already increased by += 2 so there should be no need to check for even numbers anymore. The second for appears to always be odd and is not needed. This is about 16% faster performance on my older machine.

I added a closing ?> to the end of the file though perhaps that was left off with intent -- I don't know.

Also if I have messed up any formalities here, please let me know. I haven't worked much with git.

## Contributing requirements

Didn't add a few of these as I am only doing a pull request.

### DIFF
--- a/PrimePHP/solution_1/PrimePHP.php
+++ b/PrimePHP/solution_1/PrimePHP.php
@@ -126,3 +126,4 @@ function validateResult($sieveSize): ?int
 {
     return PrimeSieve::$primeCounts[$sieveSize];
 }
+?>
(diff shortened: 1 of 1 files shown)


rbergen:
We already have #307 open to improve the performance of the existing PHP solution. Please discuss in that PR how to add your suggestions to it.

rbergen:
Closing this after merge of #307, as this now causes a merge conflict. A new PR on the new PHP/solution_1 can be opened if you're convinced that solution can be further optimized.

PR #407, opened 2021-07-13
Improve D solution 2

BradleyChatha:
## Description

Solution 2 is actually unfaithful due to the sieve size not being a runtime value, and the bits array not being dynamically allocated.

So this PR adds a second implementation that is faithful, alongside the unfaithful version.

I also took it as an opportunity to shoe-horn in even more of D's features.

And finally, I've added a few people as authors, since I feel even a small review of the code is enough of a contribution.

## Contributing requirements

### DIFF
--- a/PrimeD/solution_2/README.md
+++ b/PrimeD/solution_2/README.md
@@ -1,4 +1,4 @@
-# D solution by BradleyChatha, Paul Backus, Bastiaan Veelo
+# D solution by Bradley Chatha, Paul Backus, Bastiaan Veelo
 
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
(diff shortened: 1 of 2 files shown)


PrimeD/solution_2/README.md
```diff
@@ -1,7 +1,8 @@
-# D solution by BradleyChatha
+# D solution by Bradley Chatha, Paul Backus, Bastiaan Veelo
 
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-no-green)
```
rbergen: For a "not faithful" badge, the color to use is `-yellowgreen` instead of `-green`.

rbergen (commented):
Nice catch on the faithfulness, and thank you for handling it this way.

rbergen (approved):
LGTM

marghidanu (approved):

PR #408, opened 2021-07-14
Scala tail recursive solution

scilari:
## Description

Scala solution to the Prime Sieve Drag Race. 
* Apparently Scala solution_1 is broken*, so this provides the first working implementation on Scala
  * This solution also ~4X faster than solution_1, although comparing to an incorrect solution is fairly pointless
  * Around 4000 passes on my machine (2018 ultrabook, WSL 2 on Win10)
* Used solution_1 as a base for Docker and build files
* Uses tail recursion for the loops (compiled to while loops by the Scala compiler)
* Uses sentinel in the bit (Boolean array) to avoid some out-of-bounds checks
*  EDIT: Also adds trivial fix to solution_1 (see discussion)

Solution_1 output*:
```
2, 3, 5, 7, 11, 13, 17, 19, 23, 25, 29, 31, 35, 37, 41, 43, 47, 49, 53, 55, 59, 61, 65, 67, 71, 73, 77, 79, 83, 85, 89, 91, 95, 97, 
rom1dep;2059635;5.0;1;algorithm=base,faithful=yes
```

## Contributing requirements


rbergen:
> Apparently Scala solution_1 is broken

@scilari What makes this apparent?

scilari:
@rbergen : For example, when you use solution_1 to compute and print out primes up to 100, the sequence looks like below. The output including non-primes, such as  25 makes the incorrectness apparent. I outlined this in README.md, but I will also add this to description above.

```
2, 3, 5, 7, 11, 13, 17, 19, 23, 25, 29, 31, 35, 37, 41, 43, 47, 49, 53, 55, 59, 61, 65, 67, 71, 73, 77, 79, 83, 85, 89, 91, 95, 97, 
rom1dep;2059635;5.0;1;algorithm=base,faithful=yes
``` 

The Scala solution_1 seems to be heavily influenced by Java solution_1, so maybe should check that for correctness as well?

Ping @rom1dep

rbergen:
@scilari That's a good catch! It seems the solution_1 prime list includes the numbers that are the product of 5 and another prime number.

scilari:
> @scilari That's a good catch! It seems the solution_1 prime list includes the numbers that are the product of 5 and another prime number.

@rbergen : I investigated it a little more and there seems to be a bug with prime pairs that are two apart. Seems that it incorrectly skips to the next one before marking the multiples of the first one. After the first special values, the next such pair after (5,7) is (11,13). As I imagined, there's 11x11 = 121 as well as 11x13 = 143 in the output string for the primes up to 200:

```
2, 3, 5, 7, 11, 13, 17, 19, 23, 25, 29, 31, 35, 37, 41, 43, 47, 49, 53, 55, 59, 61, 65, 67, 71, 73, 77, 79, 83, 85, 89, 91, 95, 97, 101, 103, 107, 109, 113, 115, 119, 121, 125, 127, 131, 133, 137, 139, 143, 145, 149, 151, 155, 157, 161, 163, 167, 169, 173, 175, 179, 181, 185, 187, 191, 193, 197, 199, 
rom1dep;832167;5.0;1;algorithm=base,faithful=yes
```

Do you happen to know if this is present also in the Java solution_1? Should the correctness of the algorithms be tested somehow automatically (outside the performance measurement loop)? Note how this is handled in the Scala solution_2 by first marking the multiples and only after that searching for the next prime using the array (starting with current value + 2), so there's no possibility for double advancement (+= 2 and searchForNext).

marghidanu:
I bumped into this problem several times during development, check the conditions for evaluating the bits, you're checking for a false instead of true, or otherwise.

scilari:
> I bumped into this problem several times during development, check the conditions for evaluating the bits, you're checking for a false instead of true, or otherwise.

@marghidanu : What problem are you referring to exactly? We are discussing solution_1 that is not written by me and that most likely does not have any issues with bit evaluation, but the bug is related to incorrect handling of the factor advancement.

Solution_2 (written by me) seems to produce primes correctly and uses false-initialized array on purpose for performance reasons (it is the default for Scala/Java boolean arrays). Here's the output produced by solution_2 after uncommenting the verify line at the end of the main method (and printing primes up to 200 just to be sure):

```
scilari;3210;5.0;1;algorithm=base,faithful=yes
Primes up to 200: 2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97, 101, 103, 107, 109, 113, 127, 131, 137, 139, 149, 151, 157, 163, 167, 173, 179, 181, 191, 193, 197, 199
10: Found prime count (4) matches with expected (4)
100000: Found prime count (9592) matches with expected (9592)
1000: Found prime count (168) matches with expected (168)
1000000: Found prime count (78498) matches with expected (78498)
10000000: Found prime count (664579) matches with expected (664579)
100: Found prime count (25) matches with expected (25)
100000000: Found prime count (5761455) matches with expected (5761455)
10000: Found prime count (1229) matches with expected (1229)
```

EDIT: Good thing I printed this out as I forgot to add "2" to the output primes list manually :) Now updated and committed

marghidanu:
@scilari I was referring to solution_1 obviously :)

rbergen:
@scilari 

> Do you happen to know if this is present also in the Java solution_1? 

Java solution_1 uses the most straightforward implementation of the "seek next factor" and "clear factor multiples" loops, that I have seen over a hundred times by now. I'm confident that approach works.

> Should the correctness of the algorithms be tested somehow automatically (outside the performance measurement loop)?

That's the other reason I'm not concerned about the Java solution. It validates its own results against the map of known prime counts under specific values, which is something Scala solution_1 does not do. It includes the map, but doesn't do anything with it.

For now, let's see if @rom1dep joins the conversation. In the meantime, I will think about how to move forward if solution_1 doesn't get fixed.

marghidanu:
The benchmark tool could check the correctness of the output if the output is modified for all solutions.

rbergen:
@marghidanu I assume you mean by having all solutions output the number of primes they found under 1,000,000? This is mainly my gut feeling talking, but I think it would effectively shift the problem, mostly, to checking if the counting algorithm is valid. 

I think my future reviews are going to include a check if submitted solutions do validate their output after running the main loop.

marghidanu:
> @marghidanu I assume you mean by having all solutions output the number of primes they found under 1,000,000? This is mainly my gut feeling talking, but I think it would effectively shift the problem, mostly, to checking if the counting algorithm is valid.

Yes, that's what I meant. Dave included this check in the original output. I'm thinking of the easiest check, which is to do it ourselves. At this point, I don't trust the programs checking themselves.

rbergen:
But in the context of this idea, you'd still rely on the solution counting its primes correctly, right?

marghidanu:
@rbergen Yes :)

marghidanu:
I'm thinking about how can we get this in the easiest way possible without changing the rules of the game. There are a lot of solutions already out there.

scilari:
@rbergen : Yep, seems like the Java solution_1 is producing the primes correctly, so the mistake is probably done when porting to Scala.

However, I was surprised that the Java algorithm got so many passes with so straighforward implementation, but then noticed it get to run for 10 seconds instead of 5 😄 

```
// from Java solution_1
while ((System.currentTimeMillis() - start) < 10000)
```

Btw, I am new to this kind of community work. How should I report such bugs? Open a new issue and tag the author?

rbergen:
I totally get it. But applying this to all existing solutions (not to mention all open PRs) in sync with the change to the benchmark tool is a massive a job, and I'm just trying to assess for myself how much certainty we would actually buy with that.

marghidanu:
I guess we could also parse the old output format. We could easily evaluate how many of the existing solutions already implement that and make use of that format too, for validation.

marghidanu:
> I totally get it. But applying this to all existing solutions (not to mention all open PRs) in sync with the change to the benchmark tool is a massive a job, and I'm just trying to assess for myself how much certainty we would actually buy with that.

I fully understand that, that's why I'm not suggesting this lightly. Anyway for now we're just debating.

rbergen:
> However, I was surprised that the Java algorithm got so many passes with so straighforward implementation, but then noticed it get to run for 10 seconds instead of 5 😄

This is not a bug, but a grandfathered solution. As in: the Java solution is so "old" that the 5 second rule wasn't in place yet when the solution was added. I guess we could merge a PR that sets all runtimes to 5 seconds in solutions that still run for 10, if someone were to open one.

For the leaderboard, it doesn't really matter, as that is ordered by the number of runs per second, per thread.

scilari:
@rbergen and @marghidanu : 

Maybe it could be handled after measuring for the final results of this "competition" and disqualifying solutions that do not actually produce correct primes? Adding a note about this in the rules (and YouTube video) and emphasizing that incorrect solutions will not be counted in the final results? That is, you would not have to do extra work for solutions in progress, but only the final ones.

rbergen:
> I guess we could also parse the old output format. We could easily evaluate how many of the existing solutions already implement that and make use of that format too, for validation.

Not sure how I feel about that one, as it sort of makes us attach meaning to the old format, which is not as standardized as the "new" one.

> Anyway for now we're just debating.

I know, and as usual I'm enjoying the conversation. :D

scilari:
@rbergen 
> For the leaderboard, it doesn't really matter, as that is ordered by the number of runs per second, per thread.

Oh, ok! This could make a huge difference for JIT-based languages if you can warm up for longer! 🚀

rbergen:
> Adding a note about this in the rules (and YouTube video) and emphasizing that incorrect solutions will not be counted in the final results?

The rules already state that the results need to be correct. This one solution just slipped through the review. I think @marghidanu and I are now discussing how we can decrease the chances of that happening again :)

> Oh, ok! This could make a huge difference for JIT-based languages if you can warm up for longer! 🚀

To whomever that is a source of concern: a PR to fix solutions to 5 seconds where appropriate/necessary can be opened. ;)

marghidanu:
@rbergen yes, that’s a discussion that we can have outside of this thread. I’m enjoying the conversation too but maybe this is the wrong thread to do it. See you on Slack :)

PrimeScala/solution_2/README.md
```diff
@@ -0,0 +1,51 @@
...
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-unknown-yellowgreen)
+
+This tail recursive solution improves the current Scala solution_1 by both performance (~4X) and correctness. Used solution_1 as a base for the build and Docker files.
+
+## Run Instructions
+* JVM
+
+`./mill sieve.run`
```
rom1dep: I think this would run the native version, not the JVM one

rom1dep:
Hello there, long thread already! I plead guilty of leaving _solution1 in a state of abandonment :)

@scilari did a good job at updating things and providing a cooler implementation :)

For the record, my initial attempt wasn't based on the Java version (which didn't even exist then) but on the original Python version by Dave, and wasn't trying to be original. I might have fiddled with things close to submitting the PR, though, and broke things in the process.

Here, https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/PrimeScala/solution_1/sieve/src/Sieve.scala#L38
replacing `.dropWhile(i => rawBits(i / 2))` by `.dropWhile(i => !rawBits(i / 2))` is all it takes to fix it.

> This solution also ~4X faster than solution_1

The dance over an iterator and an Optional is very inefficient, it appears.
This more than helps, but is definitely less idiomatic:

```patch
-      (factor until sieveSize by 2).iterator
-        .dropWhile(i => !rawBits(i / 2))
-        .nextOption() match {
-        case Some(value) => factor = value
-        case _           =>
+      import scala.util.control.Breaks._
+      breakable {
+        for (num <- factor until sieveSize by 2) {
+          if (getBit(num)) {
+            factor = num
+            break
+          }
+        }

```

> > Oh, ok! This could make a huge difference for JIT-based languages if you can warm up for longer! rocket

> To whomever that is a source of concern: a PR to fix solutions to 5 seconds where appropriate/necessary can be opened. ;)

@rbergen That's a legit concern, though! JVM languages should probably use JMH to offset the warm-up/JIT-ing time if all this benchmarking is meant to see how much throughput a "language" can deliver. I might give it a shot later, based on https://github.com/com-lihaoyi/mill/blob/main/integration/test/resources/play-json/jmh.sc eventually.

rbergen:
> > To whomever that is a source of concern: a PR to fix solutions to 5 seconds where appropriate/necessary can be opened. ;)
> 
> @rbergen That's a legit concern, though!

@marghidanu and I have been discussing this offline. I expect we'll open a PR setting all runtimes to 5 seconds, this weekend.

PrimeScala/solution_2/README.md
scilari: You are correct - fixed now. Moved between different namings until decided I do not want to touch the Docker file but forgot to revert README :)

scilari:
> > > To whomever that is a source of concern: a PR to fix solutions to 5 seconds where appropriate/necessary can be opened. ;)
> > 
> > 
> > @rbergen That's a legit concern, though!
> 
> @marghidanu and I have been discussing this offline. I expect we'll open a PR setting all runtimes to 5 seconds, this weekend.

@rbergen @marghidanu  @rom1dep : This is just to make sure, as I do not think the rules are clear about this. Is a warm-up period allowed before the actual 5 sec benchmark run? For example, by letting the JVM to warm up in Scala solution_2 seems to make the throughput much more stable (and in some runs up to 40% higher). However, it seems somewhat random and depending on the system state when the JIT optimizations kick in - this morning I am not able to produce such differences.

EDIT: And now again I am seeing the clear difference (> 50%) in the throughput:
```
scilari;3508;5.003;1;algorithm=base,faithful=yes
scilari;54740;50.001;1;algorithm=base,faithful=yes
```

rbergen:
> This is just to make sure, as I do not think the rules are clear about this. Is a warm-up period allowed before the actual 5 sec benchmark run?

There currently are no rules about this, as we haven't had any excesses to rule out yet. Personally, I would be inclined to allow a warm-up time of a maximum of 5 seconds if it consistently makes a difference; we've effective been allowing up to 10 seconds per execution up to this point. 
@marghidanu I'd also be interested in your view on this.

rom1dep:
for the record, I ran some benchmarks locally:
```
Scala solution_1
rom1dep+imperative+JVM,5s,3142/5.0=628.4/s
rom1dep+imperative+JVM,50s,33361/50.001=667.2066558668827/s
rom1dep+imperative+Native,50s,10373/50.001=207.4558508829823/s
rom1dep+imperative+Native+LTOFull,50s,12275/50.0=245.5/s
rom1dep+functional+JVM,5s,3305/5.0=661.68/s
rom1dep+functional+JVM,50s,33729/50.001=674.5665086698266/s
rom1dep+functional+Native,50s,25492/50.001=509.8298034039319/s
rom1dep+functional+Native+LTOFull,50s,25174/50.0=503.48/s

Scala solution_2
scilari+JVM,5s,4108/5.0=821.6/s
scilari+JVM,50s,48958/50.0=979.16/s
scilari+Native,50s,33828/50.0=676.56/s
scilari+Native+LTOFull,50s,33719/50.0=674.38/s
scilari+JS,50s,8528/50.002=170.5531778728851/s

Java solution_1
MansenC,5s,4627/5.000000=925.4/s
MansenC,10s,9493/10.000000=949.3/s
MansenC,50s,47275/50.000000=945.5/s

C++ original
daveplCpp+clang+OFast,50s,26607/50.000000=532.14/s
daveplCpp+gcc+OFast,50s,46207/50.000000=924.14/s

Rust solution_4
joshallen64,5s,4815/5.000686=962,8678945248712/s
joshallen64,50s,46464/50.0005=929,2707072929271/s
```
(JVM results are obtained with GraalVM 21.1)

The difference between the scalas isn't as drastic as I thought.

I think it also shows how good the JVM is at optimizing/inlining/branch predicting, considering how close to/ahead to native it gets. There is definitely a warm-up penalty but not as bad as I thought (in all fairness there isn't very much for the JIT to compile either…)

rbergen:
@scilari I would like to roll up the fix for solution_1 into this PR; @rom1dep provided the fix in an earlier comment already. Could you thus incorporate this change in your PR?

> Here, https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/PrimeScala/solution_1/sieve/src/Sieve.scala#L38
replacing `.dropWhile(i => rawBits(i / 2))` by `.dropWhile(i => !rawBits(i / 2))` is all it takes to fix it.

scilari:
> @scilari I would like to roll up the fix for solution_1 into this PR; @rom1dep provided the fix in an earlier comment already. Could you thus incorporate this change in your PR?
> 
> > Here, https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/PrimeScala/solution_1/sieve/src/Sieve.scala#L38
> > replacing `.dropWhile(i => rawBits(i / 2))` by `.dropWhile(i => !rawBits(i / 2))` is all it takes to fix it.

Sure thing!

scilari:
@rom1dep 
> for the record, I ran some benchmarks locally:

Nice job! I personally did not have the time or energy to compile and run the C++ implementations. It's really interesting to see how far JVM JIT has progressed in the last 10 years or so. If I read the outputs correctly, ScalaJVM is on par or slightly ahead of C++ original and Rust S4?

scilari:
@marghidanu 
> I bumped into this problem several times during development, check the conditions for evaluating the bits, you're checking for a false instead of true, or otherwise.

You were correct after all. The bug was due to incorrect interpretation of the bit array. (But caused it to skip to the next prime if it was not further than 2!)

marghidanu:
The warm-up time is debatable, you can run the sieve for 5 seconds multiple times but we expect only one output for the evaluation.

scilari:
@marghidanu @rbergen 
> The warm-up time is debatable, you can run the sieve for 5 seconds multiple times but we expect only one output for the evaluation.

I did some quick measurements with a warm-up period as short as 5 secs, i.e. one extra run before actually measuring the throughput. Even with period this short, the throughput was about  25% higher than w/o warm-up.

| Solution      | Passes JVM      | Passes Native | Passes ScalaJS      |
| ----------    | :--------:      | :-----------: | :------------:      |
| This          | 3.0K (std .09K) |     ~2.7K     |     0.6K (std .04K) |
|This (warm 5s) | 3.8K (std .25K) |     n/a       |     n/a             |
| solution_1    |   ~1.8K         |     n/a       |      n/a            |

rbergen:
@scilari Please let me know when you feel this PR is ready for review.

PrimeScala/solution_2/mill
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env sh
+
+# This is a wrapper script, that automatically download mill from GitHub release pages
+# You can give the required mill version with MILL_VERSION env variable
+# If no version is given, it falls back to the value of DEFAULT_MILL_VERSION
+DEFAULT_MILL_VERSION=0.9.6
```
rom1dep: Could be bumpted to 0.9.9

rom1dep:
@scilari out of curiosity, which JVM version did you use? I couldn't reproduce such a drastic difference between Scala solutions, but did notice that the JVM version makes a difference:
image
(results for 50s runs, in runs/s)

Here is an overall breakdown:
image
(so, yayh for the JVM :) )

If you too are getting higher figures on Java16, perhaps you could adjust the Dockerfiles of both solution 1 and 2 to use it instead?

scilari:
@rom1dep 
> out of curiosity, which JVM version did you use?

I used `11.0.10.hs-adpt`, but have now tested with `16.0.1.hs-adpt` too. Quick and dirty runs seem to suggest that `16.0.1.hs-adpt` is noticeably faster, so I can look into how to update those dockerfiles. Here are some representative figures from those runs:

`16.0.1.hs-adpt` => 
`scilari;6236;5.0;1;algorithm=base,faithful=yes?`
`rom1dep;5065;5.0;1;algorithm=base,faithful=yes`

`11.0.10.hs-adpt` =>
`scilari;5481;5.0;1;algorithm=base,faithful=yes`
`rom1dep;4663;5.0;1;algorithm=base,faithful=yes`

Interesting note, though: I also quickly tested with `16.0.1.j9-adpt` and it seems to completely destroy solution_1. No idea why. The performance drops over tenfold! On solution_2 there's no much difference.

`16.0.1.j9-adpt` =>
`scilari;5735;5.0;1;algorithm=base,faithful=yes`
`rom1dep;208;5.024;1;algorithm=base,faithful=yes`

scilari:
@rbergen 
> @scilari Please let me know when you feel this PR is ready for review.

I think the PR is now ready.

rbergen (approved):
LGTM

rbergen:
@scilari Could you please sync your fork with the current state of this branch? It will re-enable our CI to run for your solution.

scilari:
@rbergen Done.

PR #409, opened 2021-07-14
Use JIT optimizations and SplFixedArray instead of raw arrays

HugoSantiagoBecerraAdan:
## Description
Implementation for PHP8 using JIT and `SplFixedArray`

`SplFixedArray` provides a faster aproach to handle large lists with a fixed number of elements in comparison to raw arrays.

In order to improve more the performance, the elements of the list are not initialized with `true` value. They have the default value of `null` and all validation works in inverted form to keep the functionality intact.

This is an inverted validation:

```php
// Inverted validation
if (null === $this->rawBits[$num]) {
```

instead of:

```php
// Regular validation
if ($this->rawBits[$num]) {
```

This is an inverted asignment:

```php
// Inverted value asignment
$this->rawBits[$num] = true;
```

instead of:

```php
// Regular value asignment
$this->rawBits[$num] = false;
```

The Primes passes increase about 20% when compare to same code implementation but using raw array and `array_fill`.

## Contributing requirements

### DIFF
--- a/PrimePHP/solution_2/Dockerfile
+++ b/PrimePHP/solution_2/Dockerfile
@@ -2,6 +2,8 @@ FROM php:8.0-cli-alpine3.13
 
 WORKDIR /opt/app
 
+RUN docker-php-ext-install opcache
+
 COPY *.php .
 
-ENTRYPOINT [ "php", "-dopcache.enable_cli=1", "-dopcache.enable=1", "-dopcache.jit_buffer_size=100M", "-dopcache.jit=1255", "-dxdebug.mode=off", "PrimePHP.php" ]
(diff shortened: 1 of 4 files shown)


rbergen:
This implementation has the same key characteristics as the existing PHP solution_1, which means this PR should be structured as an improvement to that solution. Furthermore, we have #307 already open aiming to improve solution_1. Please discuss in that PR how to add your improvements to it.

rbergen:
As indicated in  my earlier comment, if you feel there are still optimizations to be made to solution_1 after the merge of #307, then please modify your PR to one that improves said solution. We will not add this solution as a second one.

rbergen:
SplFixedArray was implemented with #307. Further improvements can be suggested via a PR on the current PHP/solution_1.

PR #411, opened 2021-07-14
PrimeREXX/solution_1  (just my commits)

jossRRM:
## Contributing requirements

### DIFF
--- a/PrimeREXX/solution_1/Dockerfile
+++ b/PrimeREXX/solution_1/Dockerfile
@@ -1,6 +1,6 @@
 FROM ubuntu:20.04
 RUN apt update && apt upgrade -y \
-	&& apt install -y regina-rexx nano
+	&& apt install -y regina-rexx
 WORKDIR /home/rexx
 COPY ./PrimeREXX.rex .
 # Print the required output
--- a/PrimeREXX/solution_1/Dockerfile
+++ b/PrimeREXX/solution_1/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:latest
+FROM ubuntu:20.04
 RUN apt update && apt upgrade -y \
 	&& apt install -y regina-rexx nano
 WORKDIR /home/rexx
--- a/PrimeREXX/solution_1/Dockerfile
+++ b/PrimeREXX/solution_1/Dockerfile
@@ -1,0 +1,15 @@
+FROM ubuntu:latest
+RUN apt update && apt upgrade -y \
+	&& apt install -y regina-rexx nano
+WORKDIR /home/rexx
+COPY ./PrimeREXX.rex .
+# Print the required output
+ENTRYPOINT ["rexx", "./PrimeREXX", "1000000", "0"]
+# Print the primes <= 1000000
+#ENTRYPOINT ["rexx", "./PrimeREXX", "1000000", "1"]
+
+# Print a summary of the prime counts up to 100, 1000, 10000, 100000 and 1000000
+#ENTRYPOINT ["rexx", "./PrimeREXX", "1000000", "2"]
+
+# Print the primes <= 1000000 AND also a summary as above
+#ENTRYPOINT ["rexx", "./PrimeREXX", "1000000", "3"]
--- a/PrimeREXX/solution_1/PrimeREXX.rex
+++ b/PrimeREXX/solution_1/PrimeREXX.rex
@@ -45,8 +45,6 @@ t = time("E") /* close enough for algorithm run time */
 */
 call lineout 'stdout',"joss_REXX;"count";"t";1;algorithm=base,bits=8,faithful=no"
 select
-  when printSw = 0 then
-    call lineOut 'stdout',"Sieve iterations: "count 
   when printSw = 1 then
     call PrintDetail maxVal,printWidth
   when printSw = 2 then
--- a/PrimeREXX/solution_1/PrimeREXX.rex
+++ b/PrimeREXX/solution_1/PrimeREXX.rex
@@ -1,0 +1,149 @@
+/* rexx */
+/*
+** Determine the prime numbers less than a number passed as a
+** program parameter (maxVal). The default maxVal is 1,000,000
+** Unlike many programming languages  with predefined data structure type,
+** the stem variables used to store those odd numbers that are not prime
+** use more memory.
+**
+*/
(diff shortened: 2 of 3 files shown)


marghidanu:

PrimeREXX/solution_1/README.md
```diff
@@ -0,0 +1,51 @@
+# REXX solution by joss
+Sieve of Erastosthones for the REXX scripting language.
+Uses the Regina REXX interpreter to run the program /home/rexx/PrimeREXX.rex
+A Dockerfile has been supplied. It is based upon an Ubuntu container.
+
+Regina REXX processes standard REXX. Standard REXX does not have classes/objects.
+There is a separate REXX interpreter: oorexx that does support classes.
+
+I set faithful=yes even though the solution does not use a class to run the sieve.
```
rbergen: I'm afraid that without using a class(-like) construct, the solution isn't faithful and can't be marked as such.

PrimeREXX/solution_1/PrimeREXX.rex
```diff
@@ -0,0 +1,149 @@
...
+**   Actual Sieve logic starts here, no comments in Sieve()
+***/
+t = time("R")
+call Sieve
+t = time("E") /* close enough for algorithm run time */
+
+/*
+** Sieve iterations are complete, print the results
+*/
+call lineout 'stdout',"joss_REXX;"count";"t";1;algorithm=base,bits=8,faithful=yes,parallel=no"
```
rbergen: Please remove the "parallel=no" tag. Whether the solution is parallel or not is deduced from the thread count that is also included in the output (1 = no, >1 = yes).

PrimeREXX/solution_1/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM ubuntu:latest
```
rbergen: Please pin the ubuntu base image to a specific version.

PrimeREXX/solution_1/README.md
jossRRM: I changed the README.md & source to correct this.

PrimeREXX/solution_1/PrimeREXX.rex
jossRRM: Done

PrimeREXX/solution_1/Dockerfile
jossRRM: I pinned the base image to ubuntu:20.04

jossRRM:
I've pushed my changes (to fix the issues above) to jossRRM/Primes. Is that the correct procedure for addressing these issues? Sorry, new to github.

rbergen:
@jossRRM No worries, you're doing fine so far :)

I have one question left after running your solution locally: what is the point of the "Sieve iterations: nn" line in the output? The iteration count is already in the line above it.

jossRRM:
Belt & suspenders (braces)? :-)
The output is an artifact I left in to the code that I wrote before I saw the "Contributing guidelines." I can remove it.

rbergen:
> I can remove it.

Well, if it's not too much trouble, then please.

jossRRM:
No trouble at all. It's gone.

PrimeREXX/solution_1/PrimeREXX.rex
```diff
@@ -0,0 +1,147 @@
...
+    if primes.i then
+      do
+        incr = i + i; strt = i * i
+        do j = strt to maxVal by incr
+          primes.j = 0
+        end j
+      end
+  end i
+  count = count + 1
+end
```
rbergen: It looks to me like the full sieve calculation is only done once. After the first run, the array values have been set, which means the outer loop will only find numbers that really are primes.

jossRRM:
I looked at the base requirements again and found an issue with my code: I do not reinitialize the results "array" (**primes.**) after each iteration. I made that change and tested in Windows. For a sieve up to 1,000,000 my results went from 22 iterations to 20.  For sieve up to 10,000, it goes from 5736 iterations down to 5346. Given that the implementation is not faithful to begin with, should I:

- leave the solution alone
- make the change to solution 1
- put this through as solution_2, where solution_2 performs worse than solution_1

Even with the change, the program would not be faithful, as it is implemented without classes. And I realize that interpreted REXX is going to be at the bottom of the pack in performance no matter which version is there. REXX is a Trabant in this drag race.

PrimeREXX/solution_1/PrimeREXX.rex
jossRRM: The first time through the loop, everything is 1 (primes. = 1 sets a default value for any uninitialized entry without taking any memory space for the initial values). In both the first and subsequent iterations, 3 is prime so primes.9, primes.15, ... are set to 0. So that setting is done each time. In both the first and subsequent iterations 5 is prime so primes.15, primes.25, .. are set to 0. So it doesn't matter that I didn't clear primes., the loops execute the same number of total iterations.  I believe my time differences are related to the interpreter (Regina) having to do garbage collection on those values that I set to 0.

In testing I both set "primes. = 1" at the top of the loop and then did a "drop primes." which did the garbage collection at the end of the loop. That was redundantly clearing primes.  When I simply set primes.=1 at the top of the loop, that improved the run time, but it was still slower than the version I submitted.

Note that the even numbers are all at the default value 1, but I never report those as prime, because the print loop only iterates over the odds starting at 3.

jossRRM:
I moved the initialization of the primes. stem variable to inside the timing loop. This will reset the test output for each iteration.

rbergen:
> I looked at the base requirements again and found an issue with my code: I do not reinitialize the results "array" (**primes.**) after each iteration. I made that change and tested in Windows. For a sieve up to 1,000,000 my results went from 22 iterations to 20. For sieve up to 10,000, it goes from 5736 iterations down to 5346. Given that the implementation is not faithful to begin with, should I:
> 
> * leave the solution alone
> * make the change to solution 1
> * put this through as solution_2, where solution_2 performs worse than solution_1

I think you're already acting in line with my answer, but my answer would be to make the change in solution 1. I would say having multiple implementations for a language (which preferably would still be in one solution if possible) only makes sense if there are significant differences in the implementations' characteristics.

rbergen (approved):
LGTM

PR #417, opened 2021-07-15
Swift optimization

yellowcub:
Improves speed of Swift algorithm.

## Description

A number of improvements on the original Swift implementation to make the performance comparable with fortran or rust.
 * Use of memory pointers
 * revised runSieve implementation
 * implemented an argument parser
 * etc.

## Contributing requirements


rbergen:
This solution has the same key characteristics as the existing solution_1. It's faithful, uses the base algorithm and uses `Bool` for prime flag storage (the 1 bit flag size it reports is incorrect, as `MemoryLayout<Bool>.size` reports the storage size in bytes, not bits). I will therefore have to suggest to structure this PR to modify solution_1, including (only) those changes that increase the performance of the benchmarked/timed code.

yellowcub:
Thank you for the feedback rbgen.  I am very much new to contributing on GitHub, and I was reluctant to superseded the previous submission (solution 1) by fl-01.  One notable difference between the two is solution_1 makes use of Swift's Array data structure whereas solution_2 opts for allocated memory and pointer references.  If you are of the opinion this should replace solution_1 (solution_2 is notably faster) I can make the changes, including correcting the bit -> byte size reported, and resubmit.

yellowcub:
PS - I guess I still had an interest in noting the efficiency differences between built-in Arrays and pointer references.

rbergen:
What I would suggest is that you change your PR to _modify_ (not _replace_) the existing solution to switch from Arrays to pointer references if that does increase the performance. We will then ask the author of the original solution to comment on your improvement suggestion(s), and take it from there.

yellowcub:
Will do.

yellowcub:
solution_1 has been modified to reflect changes that significantly increased performance.  Request the originating author review changes.  Thx!

rbergen:
Tagging @j-f1 as they are the original author, and allowing time for them to review and respond if the want to do so.

rbergen:
(@yellowcub While we wait, can you please remove the spaces from your task checkbox ticks (`[x ]` => `[x]`)? That'll make the tick box rendering work.)

yellowcub:
Done.

PrimeSwift/solution_1/Sources/PrimeSieveSwift/main.swift
```diff
@@ -1,75 +1,41 @@
...
-            }
-        }
-        return count
-    }
+    deinit { bits.deallocate() }
 
-    func runSieve() {
-        var factor = 3
-        let q = Int(sqrt(Double(sieveSize)))
+    @inlinable func index(_ num: Int) -> Int { return num >> 1 - 1}
```
j-f1: Style nit:


```suggestion
    @inlinable func index(for num: Int) -> Int { num >> 1 - 1 }
```

Additionally, I think it would be beneficial to keep the `assert` call. Such calls are [completely removed from release builds](https://developer.apple.com/documentation/swift/1541112-assert) but clarify the intent of the code (the similar [`precondition` function](https://developer.apple.com/documentation/swift/1540960-precondition) will still run in release code).

PrimeSwift/solution_1/Sources/PrimeSieveSwift/main.swift
```diff
@@ -1,75 +1,41 @@
...
-        assert(idx % 2 != 0, "If you're setting even bits, you're sub-optimal for some reason!")
-        rawBits[idx / 2] = false
-    }
-}
-
-class PrimeSieveSwift {
+class Sieve {
     let sieveSize: Int
-    var bits: BitArray
+    let q: Int
```
j-f1: Is there a more descriptive name for this variable?

PrimeSwift/solution_1/Sources/PrimeSieveSwift/main.swift
```diff
@@ -1,75 +1,41 @@
...
-    }
+    deinit { bits.deallocate() }
 
-    func runSieve() {
-        var factor = 3
-        let q = Int(sqrt(Double(sieveSize)))
+    @inlinable func index(_ num: Int) -> Int { return num >> 1 - 1}
 
+    /// Does @inlinable help here?
+    @inlinable func runSieve() {
```
j-f1: Since runSieve is only called once per iteration, the cost from calling it directly (rather than inlining) should be minimal.

PrimeSwift/solution_1/Sources/PrimeSieveSwift/main.swift
```diff
@@ -92,40 +70,61 @@ extension PrimeSieveSwift {
...
-        print("j-f1;\(passes);\(duration);1;algorithm=base,faithful=yes\n")
+        print("yellowcub;\(passes);\(duration);1;algorithm=base,faithful=yes,bits=\(8*MemoryLayout<Bool>.size)\n")
     }
 }
 
-let start = Date()
-var passes = 0
-var sieve: PrimeSieveSwift!
+struct StopWatch {
+    private let start: Date
```
j-f1: nit: you can initialize `start` here so that the `init()` method does not need to be explicitly specified:

```suggestion
    private let start = Date()
```

PrimeSwift/solution_1/Sources/PrimeSieveSwift/main.swift
```diff
@@ -92,40 +70,61 @@ extension PrimeSieveSwift {
...
+    init(){ start = Date() }
+
+    mutating func lap() -> Double {
+        stop = start.distance(to: Date())
+        return stop
+    }
 }
 
-sieve.printResults(showingNumbers: false, duration: start.distance(to: Date()), passes: passes)
+struct PrimeSieveSwift: ParsableCommand {
```
j-f1, line 101: Great improvement!

j-f1 (commented):
Nice improvements! I guess all the indirection I was using really wasn't doing the performance any favors :)

PrimeSwift/solution_1/Sources/PrimeSieveSwift/main.swift
yellowcub: changed q to factorLimit

PrimeSwift/solution_1/Sources/PrimeSieveSwift/main.swift
yellowcub: That makes sense.  I had not used @inlinable before, and was borrowing the concept from looking through some of the other codes.  Always learning.

PrimeSwift/solution_1/Sources/PrimeSieveSwift/main.swift
yellowcub: Agreed, that looks cleaner.

yellowcub:
@j-f1, Thanks for the review and great suggestions!  Also thanks @rbergen for guidance through my first contribution on GitHub!  All of the suggested changes were incorporated, and I tested on my end to make sure nothing was incidentally broken.  Let me know if there is any next steps that must be taken before the PR can be completed.

j-f1 (approved):
Awesome, looks good to me!

It looks like you can have a merge conflict, which you should be able to resolve by updating `maxTime` to 5 seconds (per https://github.com/PlummersSoftwareLLC/Primes/commit/0df242ffbe80cbe1aa249d39a2c44ce4c4f1f9fb) and removing the rest of the changes Git proposes.

rbergen:
@yellowcub You're very welcome. As @j-f1 has indicated (and thanks to him for doing so), there is now a merge conflict due to another PR we merged since you opened this one. If you need help with addressing it beyond the suggestions that @j-f1 made, then I'm happy to help.

yellowcub:
Ok, I changed the default for maxTime to be 5 seconds.  It took me a couple seconds to figure out how merge conflicts worked, but I think everything should be ready to go.  👍🏻 @j-f1 !

PrimeSwift/solution_1/Sources/PrimeSieveSwift/main.swift
```diff
@@ -92,40 +69,59 @@ extension PrimeSieveSwift {
...
+    @Option(name: [.customLong("list-results"), .customShort("l")], help: "List all computed primes.")
+    private var listResults = false
+
+    func run() throws {
+        var passes = 0
+        let sieve = Sieve(limit: upperLimit)
+
+        var stopWatch = StopWatch()
+        while stopWatch.lap() < maxTime {
+            sieve.runSieve()
```
rbergen: For a faithful implementation, the sieve object has to be created from scratch for every iteration. That means that `let sieve = Sieve(limit: upperLimit)` needs to be pulled into the timed `while` loop.

yellowcub: Fixed...we are seeking to be faithful.  I will say, while the fix works just fine, I feel someone with a little more experience could make it more elegant.  Since the printResults statement comes after the while closure, the sieve has to be needlessly initialized before the closure.  It shouldn't add any time where it is counted (unless initializing a var is slower than reinitializing a new let each iteration), but it comes across as inefficient nonetheless.  Would implementation of optionals slow the process down?  Thanks for the catch @rbergen!

PrimeSwift/solution_1/Sources/PrimeSieveSwift/main.swift
```diff
@@ -92,40 +69,60 @@ extension PrimeSieveSwift {
...
+        var sieve = Sieve(limit: upperLimit)
+
+        var stopWatch = StopWatch()
+        while stopWatch.lap() < maxTime {
+            sieve = Sieve(limit: upperLimit)
+            sieve.runSieve()
+            passes += 1
+        }
+
+        sieve.printResults(showingNumbers: listResults, duration: stopWatch.stop, passes: passes)
```
j-f1: Here’s one possible solution:

```swift
var stopWatch = StopWatch()
while true {
    let sieve = Sieve(limit: upperLimit)
    sieve.runSieve()
    passes += 1

    if stopWatch.lap() > maxTime {
        sieve.printResults(showingNumbers: listResults, duration: stopWatch.stop, passes: passes)
        break
    }
}
```

yellowcub: Thanks for the feedback @j-f1 .  I thought of that, but that forces the routine to do this check at every iteration during the timed section.

j-f1: How is that any different from the existing code? (note that the `while` loop is now `while true`)

j-f1: How is that any different from the existing code? (note that the `while` loop is now `while true`)

yellowcub: So, the existing while statement has to make a branch decision. In this version you have moved that decision to the if statement. Is while true less hampered than a decision branch because the code always knows it is true?  If that is the case then yeah that would work. If not then we have two branch decisions instead of one at each iteration.  I'm going to try both for several runs. If I don't notice any real difference, I like your version better.

yellowcub: Your suggestions maybe slightly improved the results, so I pushed with it.

rbergen (approved):

rbergen:
As indicated by my approval, your changes to the Swift solution look good to me. However, your PR also includes a change to the `tools` directory (re-adding `tools/src/commands/report.ts`). Please remove this change/file from your PR.

yellowcub:
@rbergen Thanks for your patience.  I had to look back to discover what you were referring to with `tools/src/commands/report.ts` since I had not made changes except to the `PrimeSieveSwift` directory that I was aware of.  At some point in the merge the changes reflect this file was added, so I deleted.  I hope this was what you had intended.  I was confused since the original appears to have `report.ts`.  Let me know if this deletion was incorrect.

rbergen:
I think we're nearly there. :)

You've now deleted two files in your last commits:

- tools/src/**commands**/report.ts, which I indeed asked you to remove. This is related to a change in the benchmark toolchain that was merged in #348. As your PR was opened well after that, I'm not sure how this file ended up still being in your fork, but I do know this one indeed has to go.
- tools/src/**services**/report.ts, which very much has to stay. Without it, the current benchmark toolchain will break, which is something I'd really like to **not** do. :)

So, my request is to now "undelete"/restore tools/src/**services**/report.ts.

yellowcub:
@rbergen I added `report.ts` back to the `services` folder.  If I knew my way around the GitHub interface a little better, I am sure there is some form of undo button.  I could not find one, so I recreated, copied the internals from where GitHub showed the deleted file, then re-uploaded.  I think I am going to take some time and run through some GitHub tutorials 😁.

rbergen:
Thanks. Could you please add a newline to the last line of the report.ts file you re-added? That's the only difference that remains, and I'd really like to limit the changes to "out-of-solution" files to none.

yellowcub:
@rbergen Done.

rbergen:
Thanks! I'm afraid there is one more thing that I previously missed: there is a binary .DS_Store file in your commit. We don't merge binary files into this repo, and I'm pretty sure the file in question is not needed for the solution itself (it's just telling me you're using a Mac of some sort :) ). Please either remove the .DS_Store file, or add a .gitignore file that contains it (which effectively has the same effect once you've removed .DS_Store from git's tracked files).

yellowcub:
@rbergen Deleted.

PR #418, opened 2021-07-15
code cleanup for PrimeJava/solution_2

PratimGhosh86:
## Description

- refactored dictionary as static, available globally
- prime sieve object is re-initialized on every iteration
- use `var` where possible
- removed unnecessary comments, re-ordered utility methods
- removed the cmd/shell script as execution command available in DockerFile

## Contributing requirements

### DIFF
--- a/PrimeJava/solution_2/src/PrimeSieveJavaBitSet.java
+++ b/PrimeJava/solution_2/src/PrimeSieveJavaBitSet.java
@@ -1,7 +1,3 @@
-/*
- * Java Prime Sieve
- */
-
 import static java.lang.System.currentTimeMillis;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import java.util.BitSet;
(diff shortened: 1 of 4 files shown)


PrimeJava/solution_2/src/PrimeSieveJavaBitSet.java
```diff
@@ -12,13 +8,26 @@
...
+      100000000, 5761455 //
+  );
+
   public static void main(final String[] args) {
 
-    long passes = 0;
-    final PrimeSieve sieve = new PrimeSieve(LIMIT);
-    final long tStart = currentTimeMillis();
+    var passes = 0;
+    var sieve = new PrimeSieve(LIMIT);
```
rbergen, line 26: Doesn't this lead to an extraneous creation of the PrimeSieve? It now gets created at the start of the timed loop as well.

PratimGhosh86: this is to avoid a null pointer warning by Sonarlint at line# 35

image

The only scenario where this could be problem is if we initialize to null and the following condition does not execute even once:

`while (MILLISECONDS.toSeconds(currentTimeMillis() - tStart) < TIME)`

So, it seemed better to initialize as it does not impact the result in any way, and if we later parameterize the `TIME` variable, we can be sure to avoid the NPE.

rbergen: I'll accept the "old-fashioned" label if it's issued, but doing runtime work to prevent a compile-time warning does feel like wagging the dog, to me :)

Still, there's no rule against it, so it's no reason not to move forward.

PR #419, opened 2021-07-15
GoPrime solution_1 performance improvement using []uint8

jdemchuk:
## Contributing requirements


rbergen:
This PR changes a key characteristic of solution_1, being the bits per prime (from unknown to 1). That's a change we can't merge as an improvement. However, we do have a single-threaded 1-bit Go solution in place, being solution_2. I would therefore recommend setting up your PR as an improvement to solution_2, instead.

jdemchuk:
Ok, I will withdraw this PR.

Go solution_2 is faster than my own, though I don't fully understand the optimization enough to offer any improvement.
On the surface it appears to be faithful to the base algorithm.  But check out the solution_2 runSieve method:

```
q := uint64(math.Sqrt(float64(s.size)) / 2)
```
Dividing the square root by two means there are only 500 loop iterations, in the loop instead of sqrt(1,000,000) = 1,000.

```
		start = 2 * (factor*factor + factor)
		step = factor*2 + 1

		for ; start < end; start += step {
			s.bits[start/8] |= bits.RotateLeft8(1, int(start))
		}
```
Some really clever inverse logic here, to address and flip bits at the same time.  Seems like this could have been done in other C like languages.

PR #424, opened 2021-07-15
Add Julia solution_3 based on sieve_1of2.c

louie-github:
## Description

This new Julia implementation is a port of [PrimeC/solution_2/sieve_1of2.c](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/PrimeC/solution_2/sieve_1of2.c) by Daniel Spångberg. It runs quite a bit faster than [PrimeJulia/solution_2](https://github.com/PlummersSoftwareLLC/Primes/tree/drag-race/PrimeJulia/solution_2) on my machine:
```
PS D:\Office Files\Programming\PrimesFork\PrimeJulia> julia .\solution_2\Primes.jl
2, 3, 5, 7, 11, …, 999953, 999959, 999961, 999979, 999983
Passes: 6837, Time: 5.0, Avg: 0.0007313149041977475, Limit: 1000000, Count: 78498
epithet-jl;6837;5.0;1;algorithm=base,faithful=yes,bits=1
```
```
PS D:\Office Files\Programming\PrimesFork\PrimeJulia> julia .\solution_3\primes_1of2.jl
Settings: sieve_size = 1000000 | duration = 5
Number of trues: 78498
primes_main.jl: Passes: 8723 | Elapsed: 5.0 | Passes per second: 1744.6 | Average pass duration: 0.0005731972945087699
louie-github_port_1of2;8723;5.0;1;algorithm=base,faithful=yes,bits=1
```

This implementation's biggest differences are that it uses bitwise operations as much as possible to avoid the overhead of arithmetic operations like `div`, and that it manually implements a bit array using `Vector{UInt}` rather than using Julia's native `BitArray` or `BitVector` type. This is very similar to how `sieve_1of2.c` does it.

You could describe this as being as low-level as you can get with Julia to squeeze the most out of its performance.

This implementation is different enough that I don't think it could easily be considered an "improvement" to any of the existing solutions, so I decided to separate it into its own solution.

On another note, this solution uses the `julia:1.6-buster` Docker image instead of `julia:1.6-alpine3.13` for two reasons:
1. Compatibility with ARM64 as well as x86_64 and other architectures
2. Issues that may arise when using `musl` as Julia only has ["Tier 3" support](https://julialang.org/downloads/#currently_supported_platforms) (builds may not be available, tests may fail, considered experimental) for builds linked to musl (e.g. Alpine Linux) rather than glibc.

I appreciate any and all feedback!

## Contributing requirements

### DIFF
--- a/PrimeJulia/solution_3/Dockerfile
+++ b/PrimeJulia/solution_3/Dockerfile
@@ -1,0 +1,7 @@
+FROM julia:1.6-buster
+
+WORKDIR /opt/app
+
+COPY *.jl ./
+
+ENTRYPOINT [ "julia", "primes_1of2.jl" ]
--- a/PrimeJulia/solution_3/README.md
+++ b/PrimeJulia/solution_3/README.md
@@ -30,8 +30,11 @@ simplifies the set_bit operation slightly
 
 If you see any room for improvement in the code or have any
 suggestions, don't hesitate to open an issue, pull request (PR), 
-discussion, or the like. Don't forget to tag me at `@louie-github` so I
+Discussion, or the like. Don't forget to tag me at `@louie-github` so I
 can be notified if my personal input is either wanted or needed.
+I'm open to fixing stylistic issues or discussing cosmetic changes to
+the code, but for those, it might be best if you open a Discussion
+first before opening an issue or pull request.
 
 
 ## Run instructions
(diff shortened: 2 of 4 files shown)


rbergen (approved):
LGTM

marghidanu (approved):

marghidanu (approved):

PR #427, opened 2021-07-15
1-bit Cython solution with some support for multi-threading

ssovest:
## Description
1-bit solution written in Cython, with some interesting optimizaions applied. 
### Key optimization points:
 - Splitting the primes array into blocks, which then are processed one by one. This gives a huge speed boost on high limits. (I know we're counting up to a million, but what's the point of making a memory-efficient 1-bit sieve and then not using it on really big numbers :))
 - Splitting the steps into 32 groups, so the same bitmask can be used for every step within a group. It drastically reduces the number of modulo and division operations.
 - Merging several bitmasks into one if we have more than 1 non-prime number in the current 32-bit section. This works only for numbers that are less or equal to 31, but that are exactly the numbers which require most work.

With this I get about 1500-1600 iterations per second on my (really old) machine.
## Contributing requirements


rbergen:
I'm wondering if it would be possible to add this as an additional implementation to solution_1 and thus #356? Both use the same docker base image and it would save a build during the benchmark runs.

ssovest:
solution_1 is 8-bit, so i've created different subfolders. I hope it does not break the rules somehow.

ssovest:
Well, that's odd. I'm sure i've deleted solution_2 folder, but i still see it in the diff... (_fixed_)

tjol:
> solution_1 is 8-bit, so i've created different subfolders. I hope it does not break the rules somehow.

There's a 1-bit version for solution_1 in PR #356 (which you helped with, quite a bit!).

I have no objection to rolling up the two PRs (and adding this as the multi-threaded variant, I suppose?)

ssovest:
I'm not sure how to resolve merge conflicts. I'm closing this PR, will open a new one.

PR #432, opened 2021-07-15
Set runtime to 5 seconds

rbergen:
This sets the runtime for all solutions to 5 seconds. Some solutions that were grandfathered when the 5 second rule came in effect are now still running for 10.

After this is merged, any runtime other than 5 seconds will be considered a bug and fixed as such.


marghidanu (approved):

marghidanu:
LGTM

PR #436, opened 2021-07-16
More Prolog solutions

jimbxb:
## Description

I have added 2 prolog solutions.

The first ~~(solution_2)~~ makes use of a dynamic predicate to store which numbers are composite.

The second ~~(solution_3)~~ makes use of the C foreign interface in SWI-PL to make use of a more optimised bitvector representation.


rbergen:
@jimbxb Considering the fact that you're also the author of solution_1, I would like to ask if you could convert these 2 solutions to additional implementations in the solution that's already in place. Every additional solution adds a new (docker) build to the benchmark runs, where an additional implementation in an existing solution does not.
If needed, you can look at PrimeBASIC/solution_1 for an example of what I would prefer to do here.

jimbxb:
> @jimbxb Considering the fact that you're also the author of solution_1, I would like to ask if you could convert these 2 solutions to additional implementations in the solution that's already in place. Every additional solution adds a new (docker) build to the benchmark runs, where an additional implementation in an existing solution does not.
> If needed, you can look at PrimeBASIC/solution_1 for an example of what I would prefer to do here.

Sounds good. The reason I wanted to categorise these as separate solutions is because their underlying implementations are quite different, instead of being flavours of the same. Will work on fixing that now.

PrimeProlog/solution_1/primes-c.pl
```diff
@@ -0,0 +1,62 @@
...
+:- load_foreign_library(bitvector).
+
+main(Argv) :-
+  (maplist(term_string, [N], Argv) ; N = 1_000_000), 
+  !,
+  statistics(walltime, [T0|_]),
+  timed_primes(N, 5000, Iters),
+  statistics(walltime, [T1|_]),
+  format(
+    "jimbxb-prolog-c;~d;~f;1;algorithm=base,faithful=yes,bits=1~n",
```
rbergen: As you're using an external dependency (different programming language) for the sieve calculation, the implementation has to be marked unfaithful. Cool to see how this works, though!

PrimeProlog/solution_1/primes-dynamic.pl
```diff
@@ -0,0 +1,73 @@
...
+:- initialization(main, main).
+
+main(Argv) :-
+  (maplist(term_string, [N], Argv) ; N = 1_000_000), 
+  !,
+  statistics(walltime, [T0|_]),
+  timed_primes(N, 5000, Iters),
+  statistics(walltime, [T1|_]),
+  format(
+    "jimbxb-prolog-dynamic;~d;~f;1;algorithm=base,faithful=no,bits=unknown~n",
```
rbergen: Please remove the `bits=unknown` tag. The fact the bit count is unknown is deduced from the tag's absence.

PrimeProlog/solution_1/README.md
```diff
@@ -1,40 +1,61 @@
...
-A solution written in Prolog. This is probably far from optimal for the language, I just thought some Prolog representation would be nice!
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-no-yellowgreen)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-unknown-yellowgreen)
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-1-green)
```
rbergen: In this place, please consolidate the badges in one line that includes of each of the badges that apply to the combination of implementations. You're obviously free to include the applicable badges for each of the individual implementation, below.

rbergen (approved):
LGTM

rbergen (commented):
The docker run fails because run.sh is not executable. Options I see are adding the executable bit to the run.sh file in your commit, adding a `RUN chmod a+x run.sh` to the Dockerfile, or changing the ENTRYPOINT arguments to `[ "bash","./run.sh" ]`

jimbxb:
@rbergen 
Thanks for the heads up. I'm quite new to docker, and it all ran on my machine with docker. It's been fixed now.

jimbxb:
@rbergen 
Thanks for the heads up. I'm quite new to docker, and it all ran on my machine with docker. It's been fixed now.

rbergen (approved):
LGTM, and it runs on my machine. :)

rbergen:
@jimbxb You're welcome!

I don't know anything about your setup of course, but if you're using WSL2 under Windows then it could be more of a filesystem issue than a Docker one. What you're describing is what I experienced when my repo root was on NTFS, I was running docker build and docker run from WSL (which considers every NTFS file as a+x) but used git from Windows (which doesn't "see" any x-bit because it actually isn't there, and thus doesn't add it when committing or pushing.)
If I lost or bored you halfway through that sentence, then please don't mind any of it.

jimbxb:
@rbergen 
Makes perfect sense. I am running WSL2 :) Thanks for the info

marghidanu (approved):

PR #437, opened 2021-07-16
Matlab solutions -> Octave solution_2

Brandon-Johns:
## Description
I have added 2 Matlab solutions.

Unfortunately, Matlab is a propriatary language and I can not provide a license.
Hence, I don't think I can provide a Dockerfile. I have not used Docker before, but from what I can tell, licensing is prohibitive.

Solution 1 uses matlab vectorisation for speed, but uses 8 bits per prime.
Solution 2 enforces 1 bit per prime. It is far slower.

I hope this is acceptible, even if it can't be included in the benchmarking.
Thanks

## Contributing requirements


rbergen:
If I'm honest, I would prefer not to add languages/solutions that cannot be included in the benchmark runs. In this case, I would like to suggest you look into converting these 2 Matlab solutions into one Octave solution with 2 implementations. 
We already have an Octave solution in the codebase, and that is included in the automated benchmark.

Concerning combining multiple implementations in one solution, you could take a look at PrimeBASIC/solution_1 (and by now, quite a few other solutions) to see what I would be aiming for.

Brandon-Johns:
Fair enough. I've only briefly touched Octave once before but trying with this, I'm suprised that it only needed a few changes.
I'll port it and combine into 1 solution.

Thanks.

Brandon-Johns:
Hi, I've now converted them into Octave as solution_2.

Please, I need help with the Dockerfile. I've got the bash command in the readme:
`echo "PrimesRun('1bit',1000000,'stats')" | octave -q -W --norc`

Or I guess, it could be hard coded into a shell script and call that from the Dockerfile?

Thanks

rbergen:
You can look into using a CMD directive instead of ENTRYPOINT. Alternatively, using ENTRYPOINT to kick off a shell script is a tried and tested method, too.

Brandon-Johns:
Thanks for the advice.
Unfortunately I couldn't get it working with CMD or by using ENTRYPOINT to kick off a shell script - so many errors.

Anyway, I got it working in a round about way.
- [x]  I added a Dockerfile that builds and runs my solution.

For comparison to solution 1:
1. PrimeOctave/solution_1: 1244 passes
2. PrimeOctave/solution_2
    * Brandon-Johns_8bit: 370 passes
    * Brandon-Johns_1bit: 1 pass in 198.43148 seconds

But my solution is faithful.
I think there's alot of overhead in OOP because the 8bit implementation is otherwise very similar to solution 1.

Ready to merge now?
Thanks

rbergen (approved):
LGTM

marghidanu (approved):

PR #441, opened 2021-07-16
Work-around for GDL not having BIT_POPULATION

kriztioan:
## Description

GDL doesn't seem to support the BIT_POPULATION function. This pull request implements a work-around, i.e., detects whether GDL is used and, in that case, doing things the hard/slow way in COUNT_PRIMES.


rbergen:
@kriztioan Do you expect further revisions to your solution in the foreseeable future? If so, I would like to kindly ask you to consolidate them into one PR (and squash your commits before opening it). We have quite a few PRs to consider, and you're now effectively triggering a second review cycle on a solution that was merged less than 8 hours ago.

kriztioan:
@rbergen Mea culpa; I understand the frustration. I promise to not touch anything after this ;-)

PR #442, opened 2021-07-16
Cython 1-bit sieve with support for multi-threading

ssovest:
## Description
1-bit solution written in Cython, with some interesting optimizaions applied.

### Key optimization points:
 - Splitting the primes array into blocks, which then are processed one by one. This gives a huge speed boost on high limits. (I know we're counting up to a million, but what's the point of making a memory-efficient 1-bit sieve and then not using it on really big numbers :))
 - Splitting the steps into 32 groups, so the same bitmask can be used for every step within a group. It drastically reduces the number of modulo and division operations.
 - Merging several bitmasks into one if we have more than 1 non-prime number in the current 32-bit section. This works only for numbers that are less or equal to 31, but that are exactly the numbers which require most work.

With this I get about 1500-1600 iterations per second on my (really old) machine.

## Contributing requirements

### DIFF
--- a/PrimeCython/solution_1/Dockerfile
+++ b/PrimeCython/solution_1/Dockerfile
@@ -4,8 +4,8 @@ WORKDIR /root
 
 RUN [ "pip", "install", "cython"]
 
-COPY PrimeCY_bytearray.pyx PrimeCY_bitarray.pyx buildall.sh ./
+COPY PrimeCY_bytearray.pyx PrimeCY_bitarray.pyx PrimeCY_32.pyx buildall.sh ./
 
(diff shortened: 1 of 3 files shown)


rbergen:
The implementation you're adding does not follow the documented structure of the base algorithm. With the division of the sieve space and the sieve run into blocks, using pre-calculated bit masks, I think the algorithm should be marked as "other".
Don't get me wrong: I think this is a very original approach and I'm happy to merge it, but I can't do so as a "base algorithm" implementation.

rbergen (approved):
LGTM

marghidanu (approved):

PR #444, opened 2021-07-17
Added CC dependency to benchmark docs and Makefile

rbergen:
This documents the dependency on cc in the benchmark tool documentation (BENCHMARK.md). It also adds a check on cc being installed in the Makefile, similar to Node.js and docker.


marghidanu (approved):

marghidanu:
Thanks, @rbergen for adjusting the documentation.

PR #447, opened 2021-07-17
Rust solution_1 bit storage variation

mike-barber:
## Description

A small addition to my Rust solution_1 -- added another variation for storing bits in a different order. I think it's `faithful`, because we haven't specified what order we should store bits in. Notably, when using `vector<bool>` in C++, the precise
implementation is up to the compiler vendor, so I think we have a bit of leeway. I'm happy to mark this particular case as `unfaithful` if @rbergen and @marghidanu think so :) It'll be fun to include in some form either way.

## Contributing requirements

### DIFF
--- a/PrimeRust/solution_1/README.md
+++ b/PrimeRust/solution_1/README.md
@@ -12,7 +12,7 @@ Contributors:
 
 This is a somewhat Rust-idiomatic version that has the storage of prime flags abstracted out, with the prime sieve algorithm generic over the storage. Two kinds of storage are implemented:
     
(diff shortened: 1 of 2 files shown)


rbergen:
The `faithful` qualification doesn't actually focus on the way prime candidate flags are stored. The `bits` tag does, but that does indeed not dictate the order in which storage takes place.

rbergen (approved):
LGTM

marghidanu (approved):

mike-barber:
Super, thanks guys!

PR #448, opened 2021-07-17
D solution2 optimise, D-ify, and provide a commentless version.

BradleyChatha:
## Description

I'll eventually run out of ways to stuff more of D's swag in.

Optimises it slightly by only caring about odd numbers, halving the amount of memory used.
Interestingly this only really speeds up Multi-SieveRT, by several thousands of passes! It's *really* sensitive for some reason, since even an extra calculation drops it by thousands (shifting by a constant). Very odd.

I've taken the pleasure of forcing UDAs into the code, because it's not easy to justify a natural way of integrating them into this problem.

FInally, I've created a small script that'll remove the giant blobs of comments from the original code. This script and the resulting file aren't really part of the solution, so aren't in the dockerfile. It's mostly for the benefit of people trying to read the code without 200 comments between each line.

## Contributing requirements


BradleyChatha:
Oh, I didn't actually do the halving thing, only skipping odd numbers.

Welp, there's still an improvement there anyway.

rbergen:
Please drop the commentless version from your commit. It boils down to a lot of duplicated code, that we do need to look at when reviewing. Besides, I see the script to generate it is in as well, so anybody who wants to get rid of the comments can easily generate it themselves.

rbergen:
@BradleyChatha Please merge the recent updates in this repo into your fork. It will merge #457 into your PR, which will allow us to successfully run the CI check.

PR #449, opened 2021-07-17
Add Visual Basic "wheel" implementations

Nukepayload2:
## Description

Add Visual Basic "wheel" implementations. `PrimeBASIC\solution_3` is unfaithful. `PrimeBASIC\solution_4` is faithful.

## Contributing requirements


rbergen:
Please combine both implementations in one solution. It saves docker build time and resources during the benchmark runs.

rbergen:
Maybe I'm jumping the gun by looking at this now, but the output I get when performing a docker run after the docker build of this image, is the following:
```
 Arguments: faithful
-f|--faithful: String
Whether to use the faithful implementation. Valid values: yes or no.
```

One interpretation of this is that your docker image expects arguments to tell it what implementation to run. That's not a setup we can support. Your docker image will have to run all (in this case both) implementations it contains, in sequence.

rbergen:
I'm still getting the same message.

rbergen:
I assume you're marking one implementation as unfaithful due to the use of `ArrayPool`? As `ArrayPool` is a platform-provided array allocation facility that you use to allocate and de-allocate the needed memory with every iteration, I'm actually happy to accept its use as a faithful approach.

rbergen (approved):
LGTM

marghidanu (approved):

PR #450, opened 2021-07-17
add go solution 4

kpym:
## Description

This is go solution using real multithread. Real in a sens that the sieve is build using concurrent workers in contrast to the solutions that runs concurrently multiple single thread algorithms for 5 seconds and count the number of runs.

## Contributing requirements

* [x] I added a README.md (without badges).


PrimeGo/solution_4/main.go
```diff
@@ -0,0 +1,201 @@
...
+
+	// start the calculation
+	passes := 0
+	startClock := time.Now()
+	for {
+		DoSieveMulti(N, routines)
+		passes++
+		if timeSince := time.Since(startClock); timeSince.Seconds() >= 5 {
+			PrintResults(timeSince, passes)
+			fmt.Printf("\nkpym-go-multi;%d;%f;4;algorithm=base,faithful=yes\n", passes, timeSince.Seconds())
```
rbergen, line 209: This solution does not encapsulate sieve state (the prime flags array) in a class(-like) structure. Instead, it tracks them in a global variable. That means it cannot be tagged as faithful.

kpym: Sorry, I was not aware that this is the meaning of `faithful`. What should I put here ?

rbergen: For this approach, you would have to replace `faithful=yes` with `faithful=no`.

kpym: I modified the code by removing the global variable Sieve. I hope it is now `faithful`.

rbergen: The contributing guidelines state that amongst others, a faithful implementation "uses a class to encapsulate the sieve, or a (closest) equivalent feature in your language if a class construct is not available. This class must contain the full state of the sieve. Each iteration should re-create a new instance of this class from scratch."

Your implementation seems to require the prime flags array and the sieve size to be able to run the sieve. Therefore, those two things need to be encapsulated in a class that is recreated from scratch for every iteration.

kpym: @rbergen I think that it was respecting all but was not very clear because the allocation was done inside the creation function. I hope now it is more clear. 

Now given the limit `N` and the number of workers `routines` we can do

```go
sieve := NewSieve(N)
sieve.Build(routines)
```
and then we can count the primes like this 
```go
sieve.CountPrimes()
```

To recover the maximal number that can be checked against the sieve we can write a function that returns `2*len(sieve)+2` but it is not needed in this code. So saving `N` inside the sieve object is useless in this case (but it is necessary for example in the case of go solution 2).

I hope that now it is looking enough like a "class". 

This last solution became 10% faster just by moving `wg.Done()` outside of `PrimeChecker`...strange.

PrimeGo/solution_4/HOW_IT_WORKS.md
```diff
@@ -0,0 +1,16 @@
...
+
+The idea is that if the sieve is ready up to `n` we can use multiple (let say `k`) go routines to fill the sieve between `n` and `n*n`. So we split this space `[n:n*n)` to `k` non overlapping chunks and each of this workers (go routine) reads from the shared memory `[0:n)` and writes to its own chunk. In this way there is no need to use concurrent access to the memory.
+
+# The sieve
+
+In the sieve we encode only the odd numbers bigger that 3. So an odd number `n` is encoded as prime iff `Sieve[(n-3)/2]` is `false`. Because a `boolean` in go occupies 8 bits, we can say that our sieve is a 4 bits per number (because we save only the odd ones).
+
+# Where to start
+
+Because the first three odd numbers (3, 5 and 7) are primes, we can consider that the sieve is ready up to 7 and start to fill the region `[7:121)`. But it is more efficient if the last chunk is as biggest as possible, and so the best solution is to start filling `[7:q)` where q<sup>2r</sup> is slightly bigger then the maximal limit. The right choice of `q` is really important.
```
rbergen: Please note that the algorithm cannot be considered the base algorithm if it uses prior knowledge about what numbers are primes. There is one exception, which is 2 (and, by extension, all even numbers). This means your algorithm cannot use as a premise the fact that 3, 5 and 7 are primes if you want it to be marked as a base algorithm implementation.
Also, and I am not saying that this applies to your solution, a base algorithm can't "project" any prime numbers that were found onto other parts of the sieve.
A more comprehensive description of the structure of the base algorithm can be found in [the relevant section of the contributing guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md#base-algorithm).

PrimeGo/solution_4/main.go
rbergen: I believe the faithfulness requirements have now indeed been met, I have made a note on the requirements for being classified as a base algorithm, above.

PrimeGo/solution_4/HOW_IT_WORKS.md
kpym: @rbergen I do not pre-suppose that 5 and 7 are primes (I never use this in the algorithme). I just explain what the situation is  in a comment. I use that the sieve is ok up to 9 (excluded) because the `base` algorithme for 3 starts at 9 (for 5 at 25, and so on). So the sieve for odd numbers is "ready" up to 7 at its very beginning. That's it.

kpym: @rbergen I have rewritten the comments above [Build](https://github.com/kpym/Primes/blob/7b5cc5990a1541b238f588ac30b95d9e4724e965/PrimeGo/solution_4/main.go#L121-L130) and the contents of [HOW_IT_WORKS.md](https://github.com/kpym/Primes/blob/7b5cc5990a1541b238f588ac30b95d9e4724e965/PrimeGo/solution_4/HOW_IT_WORKS.md) to explain that I have made no assumptions about the first odd primes (not even about 3). What I'm doing is using the way the `base` algorithm works to make it multi-thread by just changing the order in which we fill the sieve.

rbergen: Thank you for clarifying this.

rbergen (commented):
Please remove the empty go.sum file. That will be my last review comment if nothing changes, because beyond this the solution looks good to me.

kpym:
> Please remove the empty go.sum file. That will be my last review comment if nothing changes, because beyond this the solution looks good to me.

Done. Thanks !

rbergen (approved):
LGTM

marghidanu (approved):

PR #452, opened 2021-07-17
Add Scheme Solution

William103:
## Description
Here is a solution in Scheme. I saw #78 is a pull request for a solution in Scheme, but it doesn't look like it's active. My solution is very similar, but differs in a few key ways. #78 used a boolean vector, which isn't necessarily a bit array, while I implemented a bit array. #78 was marked as not being faithful since it lacks a class to encapsulate the state of the sieve. While it is possible to create things like classes in Scheme, it isn't very idiomatic. Instead I created a library to encapsulate the state of the sieve. It accomplishes the same goal but is more idiomatic. If that isn't acceptable, I am happy to rewrite it to use a class.

## Contributing requirements

### DIFF
--- a/PrimeScheme/solution_1/README.md
+++ b/PrimeScheme/solution_1/README.md
@@ -3,6 +3,7 @@
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Parallelism](https://img.shields.io/badge/Parallel-no-green)
 ![Bit count](https://img.shields.io/badge/Bits-1-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
 
 `sieve.ss` creates a library encapsulating the logic and state of the prime
 sieve, as well as implements a bit vector using scheme's `byte-vector`.
--- a/PrimeScheme/solution_1/README.md
+++ b/PrimeScheme/solution_1/README.md
@@ -1,0 +1,33 @@
+# Scheme solution by William103
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-1-green)
+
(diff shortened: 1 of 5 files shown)


PrimeScheme/solution_1/sieve.ss
```diff
@@ -0,0 +1,77 @@
...
+  (define (validate-results n)
+    (= (cdr (assq n results-alist)) (count-primes n)))
+
+  ; convert scheme duration to fractional number of seconds
+  (define (duration-to-seconds duration)
+    (+ (time-second duration) (* 1e-9 (time-nanosecond duration))))
+
+  ; print the results in the expected format
+  (define (print-results duration passes n)
+    (format #t "William103;~a;~a;1;algorithm=base,bits=1" passes (duration-to-seconds duration))
```
rbergen: Is there a particular reason why you don't output a faithful tag? As indicated in the contributing guidelines, that means your solution will be considered to be unfaithful.

William103: I just wanted to make sure it actually was faithful before adding that tag. I changed it to output a faithful tag now.

PrimeScheme/solution_1/README.md
```diff
@@ -0,0 +1,34 @@
...
+    $ docker build -t scheme1 .
+    $ docker run --rm scheme1
+
+## Output
+
+Using chezscheme 9.5.4 on Arch Linux, I get
+
+    compiling sieve.ss with output to sieve.so
+    compiling PrimeScheme.ss with output to PrimeScheme.so
+    William103;141;5.0155461;1;algorithm=base,bits=1
```
rbergen: It looks like you have not updated your example output to include the faithful tag. :) Anyway, I won't withhold approval of the PR on this.

William103: Whoops! Updated that. Thanks!

rbergen (approved):
LGTM

marghidanu (approved):

PR #455, opened 2021-07-18
Add Kos solution

cdragan:
## Description
This solution is written in Kos, which is a scripting language.

It uses a class to encapsulate the sieve, using a `buffer` type, which is an array of bytes.

Flags for subsequent primes are stored in bytes (8 bits), which is much faster than doing bit manipulation in an interpreted language, without using any module specific to handling sequences of bits.

This is a single-threaded solution (a multi-threaded solution is possible, but I haven't explored it yet).

The solutions uses a base faithful algorithm, including the allowed optimizations.

## Contributing requirements

### DIFF
--- a/PrimeKos/solution_1/primes.kos
+++ b/PrimeKos/solution_1/primes.kos
@@ -20,9 +20,6 @@ class Sieve
     {
         const sieve = this.buf
 
-        # Count how many primes we've found
-        var num_primes = 1 # 2 is the first prime, we only count odd primes
-
         # We perform clearing up to sqrt of sieve size
         const max_clear_factor = integer(sqrt(sieve.size * 2)) / 2
 
@@ -33,8 +30,6 @@ class Sieve
             # If this slot has 0, it means it's not a multiple of a prime,
             # therefore it corresponds to a prime number
             if ! sieve[factor] {
-
-                num_primes += 1
 
                 const prime = factor * 2 + 1
 
--- a/PrimeKos/solution_1/primes.kos
+++ b/PrimeKos/solution_1/primes.kos
@@ -39,6 +39,8 @@ class Sieve
                 }
             }
         }
+
+        return sieve
     }
 
     # Generator which yields subsequent primes from the sieve
--- a/PrimeKos/solution_1/primes.kos
+++ b/PrimeKos/solution_1/primes.kos
@@ -44,17 +44,9 @@ class Sieve
                 }
             }
         }
-
-        # Count remaining primes
-        for const factor in range(max_clear_factor + 1, sieve.size) {
-            if ! sieve[factor] {
-                num_primes += 1
-            }
-        }
-
-        return num_primes
     }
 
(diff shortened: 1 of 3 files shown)


PrimeKos/solution_1/primes.kos
```diff
@@ -0,0 +1,124 @@
...
+                }
+            }
+        }
+
+        # Count remaining primes
+        for const factor in range(max_clear_factor + 1, sieve.size) {
+            if ! sieve[factor] {
+                num_primes += 1
+            }
+        }
```
rbergen: It's not necessary to keep track of/count the number of primes found when running the sieve. In fact, doing so could be considered a waste of clock cycles during the timed run. Instead, in a faithful implementation, the function that runs the sieve should not return the number of primes found. but the sieve array itself.

rbergen (approved):

PrimeKos/solution_1/primes.kos
cdragan: Thanks a lot, fantastic, let me fix that. :-)

cdragan: Updated.

PrimeKos/solution_1/primes.kos
```diff
@@ -0,0 +1,120 @@
...
+
+        # Look from primes starting with 3 up to square root of sieve size
+        # Note we divide by 2 since we keep only odd numbers
+        for const factor in range(1, max_clear_factor + 1) {
+
+            # If this slot has 0, it means it's not a multiple of a prime,
+            # therefore it corresponds to a prime number
+            if ! sieve[factor] {
+
+                num_primes += 1
```
rbergen: Is this still necessary?

PrimeKos/solution_1/primes.kos
```diff
@@ -0,0 +1,120 @@
...
+
+                const prime = factor * 2 + 1
+
+                # Mark multiples of this prime, starting at prime squared
+                for const clear_factor in range(prime * prime / 2, sieve.size, prime) {
+                    sieve[clear_factor] = 1
+                }
+            }
+        }
+    }
```
rbergen, line 44: It seems the run function is not returning the sieve array/buffer, yet.

PrimeKos/solution_1/primes.kos
cdragan: Thanks, it is not. I've just noticed this as well. Fix in flight.

PrimeKos/solution_1/primes.kos
cdragan: The sieve buffer/array is the member of the class. I don't think we need to return it from the run function.

cdragan: Updated to return the sieve buffer/array from the run function.

rbergen (approved):
LGTM

marghidanu (approved):

PR #457, opened 2021-07-18
Crystal fix

marghidanu:
Fix for the Crystal build. The new compiler (1.1.0) is failing, switched to the official image at version 1.0.0, for now only available on amd64.


rbergen (approved):
LGTM. Completes docker build locally and I've seen that it also completes in the GitHub CI workflow.

PR #458, opened 2021-07-18
Dockerfile lint

marghidanu:
This one has been sitting for a few days as I was testing it out. This PR introduces various checks for the Dockerfile using Hadolint to make sure we have good practices. Additionally, it changes how the build matrix runs by correctly marking the failed jobs and not stopping the pipeline at the first fail.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -48,8 +48,7 @@ jobs:
         uses: hadolint/hadolint-action@v1.4.0
         with:
           dockerfile: "${{ matrix.solution }}/Dockerfile"
-          failure-threshold: warn
-        # continue-on-error: true
+          failure-threshold: error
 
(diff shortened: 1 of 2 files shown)


rbergen:
I guess we should make addressing the issues in existing Dockerfiles a maintainer action after merging this?

marghidanu:
Yes, that’s the idea. Since we have CI might as well use it :)

rbergen (approved):
Alight then, let's do this! :)

marghidanu:
Maybe I should add a build status badge to the README?

rbergen:
I think that might be a good idea, in fact regardless of this PR.

I'm going to merge this after I submit this comment. Let's try to fix the 10 issues that are now known because of it, quickly. I really don't like big red Xs :)

PR #459, opened 2021-07-19
added sieve implementation using executors

PratimGhosh86:
## Description

Keeping the original `runSieve()` approach in place, this implementation splits the task of calculating sieves to multiple threads without partitioning the sieve itself. Once a thread has completed execution, required data is emitted through Atomic types and the resultant prime count of each individual thread is accumulated globally. For verification, the accumulated count is divided by the number of passes and compared to the predefined dictionary.

```
> Executing task: docker run --rm -it  solution3:latest <

Passes: 10150, Time: 5.000000, Avg: 0.000493, Limit: 1000000, Count: 78498, Valid: true

PratimGhosh86-solution_3;10150;5.000000;212;algorithm=base,faithful=yes,bits=1
```

## Contributing requirements


rbergen:
Please modify this PR to add a second implementation to solution_2, instead of a new solution. It saves docker build time and resources while running the benchmark. Also, please synchronize your fork with the current state of this repo/branch. It will pull in #461, which is needed to fix CI for your PR.

PratimGhosh86:
alright will do that

PratimGhosh86:
@rbergen changes have been pushed

PrimeJava/solution_2/README.md
```diff
@@ -1,23 +1,22 @@
...
 ## Output
 
 ```
 > Executing task: docker run --rm -it  solution2:latest <
 
-Passes: 2504, Time: 5.000000, Avg: 0.001997, Limit: 1000000, Count: 78498, Valid: true
+Passes: 10150, Time: 5.000000, Avg: 0.000493, Limit: 1000000, Count: 78498, Valid: true
 
-PratimGhosh86;2504;5.000000;1;algorithm=base,faithful=yes,bits=1
+PratimGhosh86;10150;5.000000;212;algorithm=base,faithful=yes,bits=1
```
rbergen: You've been running this on a machine with 212 parallel threads?

rbergen (commented):
I see you've now replaced your earlier single-threaded implementation with the executor-based one. That's perfectly fine if that is indeed what you want to do, but it is also okay to have both implementations in the one solution. Both implementations can then be built and run by the same Dockerfile.
An example of a solution using this approach is PrimeBASIC/solution_1; there are multiple other examples in the repo.

PrimeJava/solution_2/README.md
PratimGhosh86: alright, let me check

PrimeJava/solution_2/README.md
```diff
@@ -1,23 +1,25 @@
-# Java solution by PratimGhosh86
+# Java BitSet (with Executors) solution by PratimGhosh86
 
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
-![Parallelism](https://img.shields.io/badge/Parallel-no-green)
```
rbergen, line 5: As the solution now contains both single-threaded and multithreaded implementations, both the parallel-no-green and parallel-yes-green badges should be included.

PrimeJava/solution_2/src/PrimeSieveJavaBitSetMT.java
```diff
@@ -0,0 +1,160 @@
...
+  // cached thread pool to spawn as many threads as possible
+  // alternatively, Executors.newFixedThreadPool(n) can be used to spawn a fixed number of thread
+  private static final ThreadPoolExecutor POOL =
+      (ThreadPoolExecutor) Executors.newCachedThreadPool();
+
+  public static void main(final String[] args) {
+    POOL.setCorePoolSize(Runtime.getRuntime().availableProcessors());
+    final var tStart = System.currentTimeMillis();
+    while (TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis() - tStart) < TIME_IN_SECONDS) {
+      POOL.execute(() -> runSieve(SIEVE_SIZE, getBitArray(SIEVE_SIZE)));
```
rbergen: Could you wrap the sieve size and bit array in a wrapper class that is created for every iteration? Adding that technicality would remove any doubt that it is a faithful implementation.

PrimeJava/solution_2/README.md
PratimGhosh86: Changes have been pushed, updated dockerfile.

```
> Executing task: docker run --rm -it  solution2:latest <

PratimGhosh86-JavaBitSet;1943;5.000000;1;algorithm=base,faithful=yes,bits=1
PratimGhosh86-JavaBitSetMT;10738;5.000000;210;algorithm=base,faithful=yes,bits=1
```

Regarding the thread count, that's the behavior with `Executors.newCachedThreadPool()`. I tracked the active thread count manually, but the difference with the provided `getLargestPoolSize()` is withing 2-3 counts at most.

PrimeJava/solution_2/README.md
```diff
@@ -1,23 +1,25 @@
...
 ## Output
 
 ```
 > Executing task: docker run --rm -it  solution2:latest <
 
-Passes: 2504, Time: 5.000000, Avg: 0.001997, Limit: 1000000, Count: 78498, Valid: true
-
-PratimGhosh86;2504;5.000000;1;algorithm=base,faithful=yes,bits=1
+PratimGhosh86-JavaBitSet;1943;5.000000;1;algorithm=base,faithful=yes,bits=1
+PratimGhosh86-JavaBitSetMT;10738;5.000000;210;algorithm=base,faithful=yes,bits=1
```
rbergen: Where do you find these 210+ thread machines? :)

PrimeJava/solution_2/src/PrimeSieveJavaBitSetMT.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+    SHOULD_RUN.set(false);
+    shutdownPool();
+    final var tEnd = System.currentTimeMillis();
+    printResults(SIEVE_SIZE, PASSES.get(), TimeUnit.MILLISECONDS.toSeconds(tEnd - tStart),
+        COUNT_OF_PRIMES_ACCUMULATOR.longValue(), POOL.getLargestPoolSize());
```
rbergen: Are you sure that `ThreadPoolExecutor.getLargestPoolSize()` returns the maximum number of _parallel threads_ that were used? I believe that not all threads in a pool are necessarily all executing at the same time. As I said before, the example output you include in your README.md basically states that you generated it on a machine that could execute 210 threads _at the same time_.

PrimeJava/solution_2/src/PrimeSieveJavaBitSetMT.java
```diff
@@ -0,0 +1,160 @@
...
+      Thread.currentThread().interrupt();
+    } finally {
+      POOL.shutdownNow(); // Cancel currently executing tasks
+    }
+  }
+
+  // Historical data for validating our results - the number of primes to be found under some
+  // limit, such as 168 primes under 1000
+  private static final Map<Integer, Integer> MY_DICT = Map.of( //
+      10, 1, //
```
rbergen: The reference count of primes under 10 is 4 (2, 3, 5, 7). The value of 1 was part of Dave's original solutions, but incorrect.

PrimeJava/solution_2/src/PrimeSieveJavaBitSetMT.java
PratimGhosh86: For `PrimeSieveJavaBitSet.java` it's already in a separate class `PrimeSieve` and the BitSet is initialized in its constructor `new PrimeSieve(LIMIT)`

```
public PrimeSieve(final int size) {
      // Upper limit, highest prime we'll consider
      this.sieveSize = size;
      // since we filter evens, just half as many bits
      final var bitArrayLength = (this.sieveSize + 1) / 2;
      this.bitArray = new BitSet(bitArrayLength);
      this.bitArray.set(0, bitArrayLength, true);
    }
```

For `PrimeSieveJavaBitSetMT.java`, this method will always return a new BitSet when used in `POOL.execute(() -> runSieve(SIEVE_SIZE, getBitArray(SIEVE_SIZE)))`

```
// generate a clean bit array
  private static BitSet getBitArray(final int sieveSize) {
    // since we filter evens, just half as many bits
    final var bitArrayLength = (sieveSize + 1) / 2;
    final var bitArray = new BitSet(bitArrayLength);
    bitArray.set(0, bitArrayLength, true);
    return bitArray;
  }
```

PrimeJava/solution_2/src/PrimeSieveJavaBitSetMT.java
PratimGhosh86: I added a code to track the current active thread count in the thread pool and this is the result:

my count: 196
getLargestPoolSize: 212

Since the pool state changes while getting the active count, my counter will always be slightly less.

PrimeJava/solution_2/src/PrimeSieveJavaBitSetMT.java
PratimGhosh86: this has been updated for both the files

PrimeJava/solution_2/README.md
PratimGhosh86: this has been updated

PrimeJava/solution_2/README.md
PratimGhosh86: Instead of using a customized thread pool, I used the default implementation of `Executors.newCachedThreadPool()`

PrimeJava/solution_2/src/PrimeSieveJavaBitSetMT.java
rbergen: Ok, now I really am curious. How many processor cores are in the machine that you're using?

PrimeJava/solution_2/src/PrimeSieveJavaBitSetMT.java
rbergen: > For `PrimeSieveJavaBitSet.java` it's already in a separate class `PrimeSieve` and the BitSet is initialized in its constructor new `PrimeSieve(LIMIT)`.

This I had noticed. I was only commenting on `PrimeSieveJavaBitSetMT.java`.

> For `PrimeSieveJavaBitSetMT.java`, this method will always return a new BitSet when used in `POOL.execute(() -> runSieve(SIEVE_SIZE, getBitArray(SIEVE_SIZE)))`

I see that, but the `runSieve` method still requires both the BitSet and the sieve size to operate. This is illustrated by the fact that both the sieve size and the BitSet are passed to `runSieve` as parameters. That means that the "full sieve state" consists of those two components, which means that they do need to be encapsulated by a class.

PrimeJava/solution_2/src/PrimeSieveJavaBitSetMT.java
PratimGhosh86: 6 cores, 12 logical processors

PrimeJava/solution_2/src/PrimeSieveJavaBitSetMT.java
PratimGhosh86: Done, sieveSize and Bitset has been moved to PrimeSieve class, and executed as

`POOL.execute(() -> runSieve(new PrimeSieve(SIEVE_SIZE)))`

PrimeJava/solution_2/src/PrimeSieveJavaBitSetMT.java
rbergen: In that case, the thread count in the output should be 12 if all cores are used for the sieve runs.

PrimeJava/solution_2/README.md
rbergen: I still think we're counting the wrong thing. As I noted in a different comment, on a 6 core, 12 logical processor systems, the output here should be 12.

PrimeJava/solution_2/src/PrimeSieveJavaBitSetMT.java
PratimGhosh86: Done

```
> Executing task: docker run --rm -it  java-solution2:latest <

PratimGhosh86-JavaBitSet;2046;5.000000;1;algorithm=base,faithful=yes,bits=1
PratimGhosh86-JavaBitSetMT;13416;5.000000;12;algorithm=base,faithful=yes,bits=1
```

rbergen (approved):
LGTM

PR #460, opened 2021-07-19
Add Rust solution 5: Multithreading

Kulasko:
## Description

This solution aims to explore the effects of different multithreading approaches to an optimized version of the original C++ algorithm, inspired by the solution of Mike Barber (PrimeRust/solution_1). In contrast to solutions that just run multiple sieving threads, this one uses multiple threads to work on the same sieve.

Currently, there are three algorithms:
- Stock standard single threaded. Mostly useful for a baseline compared to the others.
- Streaming. After finding a new prime number, all threads work together on a flag unset pass. This takes no advantage of data locality whatsoever, also threads can steal memory from each other's caches between passes, so it should be bottlenecked by cache bandwidth and crosstalk latency. Also, this isn't expected to scale well in any way.
- Tiled. After a single thread fetches all the primes up to the square root of the total number, a number of threads on memory blocks receive the list of primes and each apply them on their memory. This has very good data locality, but as threads don't move to where the action happens, they are bound to run dry. If there are no other bottlenecks, this should approach around 50% of CPU core scaling.

Each algorithm is run on a bit-based and a bool-based data set.

## Contributing requirements


mike-barber:
Nice! It's a really cool solution :) I'm a big fan of Rayon too. And it's a really good demo of how important cache invalidation is.

Kulasko:
Thank you Mike!
I am really curious on how it runs on other computers. My only PC's CPU is renown for having a bad cache hierarchy, so it might scale better in practice than I've seen until now.
Your single threaded solution seems to still have a slight edge on mine, so there still may be some room for improvement.

mike-barber:
I gave it a quick go on my machine (Ryzen 3900X), and it seems like it's happiest with threads limited to 2. Everything higher than that, and performance degrades, at least on my machine. I did experiment with larger working set sizes, but it didn't make much of a difference. 

I guess the cost of synchronisation with more threads may be the reason. Or it could be that data that different cores are working on share cache lines, and this becomes more of an issue with more cores in the mix. This is probably a bigger problem with Ryzen due to the separate CCXs. 

Here are the results I got:

### 24 threads (default for Ryzen 3900x)

```
kulasko-rust-serial-bit;10554;5.0002029;1;algorithm=base,faithful=yes,bits=1
kulasko-rust-serial-bool;18454;5.0001539;1;algorithm=base,faithful=yes,bits=8
kulasko-rust-streamed-bit;34;5.0732572;24;algorithm=base,faithful=yes,bits=1
kulasko-rust-streamed-bool;33;5.0459881;24;algorithm=base,faithful=yes,bits=8
kulasko-rust-tiled-bit;2355;5.000949;24;algorithm=base,faithful=yes,bits=1
kulasko-rust-tiled-bool;1845;5.000431;24;algorithm=base,faithful=yes,bits=8
```

### RAYON_NUM_THREADS=2 cargo run --release 2> /dev/null

```
kulasko-rust-serial-bit;10518;5.0004362;1;algorithm=base,faithful=yes,bits=1
kulasko-rust-serial-bool;18338;5.0002495;1;algorithm=base,faithful=yes,bits=8
kulasko-rust-streamed-bit;856;5.0041418;2;algorithm=base,faithful=yes,bits=1
kulasko-rust-streamed-bool;882;5.002631;2;algorithm=base,faithful=yes,bits=8
kulasko-rust-tiled-bit;14570;5.0000863;2;algorithm=base,faithful=yes,bits=1
kulasko-rust-tiled-bool;8825;5.0002056;2;algorithm=base,faithful=yes,bits=8
```

### RAYON_NUM_THREADS=4 cargo run --release 2> /dev/null

```
kulasko-rust-serial-bit;10618;5.0001455;1;algorithm=base,faithful=yes,bits=1
kulasko-rust-serial-bool;18826;5.0002511;1;algorithm=base,faithful=yes,bits=8
kulasko-rust-streamed-bit;205;5.021532;4;algorithm=base,faithful=yes,bits=1
kulasko-rust-streamed-bool;214;5.0148059;4;algorithm=base,faithful=yes,bits=8
kulasko-rust-tiled-bit;11455;5.0002506;4;algorithm=base,faithful=yes,bits=1
kulasko-rust-tiled-bool;6865;5.0005567;4;algorithm=base,faithful=yes,bits=8
```

###  RAYON_NUM_THREADS=8 cargo run --release 2> /dev/null

```
kulasko-rust-serial-bit;10579;5.0001375;1;algorithm=base,faithful=yes,bits=1
kulasko-rust-serial-bool;18643;5.0002538;1;algorithm=base,faithful=yes,bits=8
kulasko-rust-streamed-bit;83;5.0329168;8;algorithm=base,faithful=yes,bits=1
kulasko-rust-streamed-bool;83;5.0512646;8;algorithm=base,faithful=yes,bits=8
kulasko-rust-tiled-bit;5736;5.0008246;8;algorithm=base,faithful=yes,bits=1
kulasko-rust-tiled-bool;4357;5.0006763;8;algorithm=base,faithful=yes,bits=8
```

### RAYON_NUM_THREADS=16 cargo run --release 2> /dev/null

```
kulasko-rust-serial-bit;10605;5.0000339;1;algorithm=base,faithful=yes,bits=1
kulasko-rust-serial-bool;18561;5.0001504;1;algorithm=base,faithful=yes,bits=8
kulasko-rust-streamed-bit;42;5.0355268;16;algorithm=base,faithful=yes,bits=1
kulasko-rust-streamed-bool;42;5.0305261;16;algorithm=base,faithful=yes,bits=8
kulasko-rust-tiled-bit;3098;5.0007853;16;algorithm=base,faithful=yes,bits=1
kulasko-rust-tiled-bool;2497;5.0002631;16;algorithm=base,faithful=yes,bits=8
```

Kulasko:
Thanks for your input, it's nice to see you being so interested in this solution :)

I really didn't expect the results you got. For one, the single threaded results are even closer to your solution (as reported in your README file) than on my machine.
On both multithreaded algorithms, even with limited threads, the scaling on your CPU is much worse than on mine (as seen in the project README). I don't understand the reason at all, I am very curious. My first guess is that Ryzen swings way more towards compute in computer/memory balance.
Maybe the OS scheduler doesn't fire off all threads at the same time, or it keeps moving threads around? As the runtime is extremely short and the working set might fit into L1 cache, moving down the memory hierarchy is bound to have a big effect.
Now that I am thinking about it, are you using Windows? Because some time ago, the Windows scheduler caused performance issues compared to Linux in some programs for this very reason.

I may try to integrate a 64 Byte alignment restriction, as that is what all modern processors I know of use as a cache line size.
Otherwise, I am not sure how to approach this. As I said, the problem size is still too small for meaningful multithreading on the same sieve with higher thread counts. The only algorithm I've seen to scale beyond single threaded as of now was the tiled one with bit storage.

Unfortunately, I haven't implemented a command line sieve size override yet. Maybe the scaling gets better with a bigger problem size.
Also, do you still have a Raspbery Pi available? If so, would you be so kind to test my solution out on it? We then might be able get more insight on this.

mike-barber:
Yeah it's a weird one on the Ryzen -- I ran it after your recent update with similar results. I'm running it in WSL2 under Windows. Usually, this results in no performance issues, but it could be a factor here potentially.  

Yep I still have the raspberry pi -- gave it a quick run on there and the results for the tiled approach are the usual you'd expect: scales nicely. Actually the multithreaded version is quite a bit faster on the pi. 

Raspberry Pi 4 results:

### 4 threads (default) 
```
kulasko-rust-serial-bit;1968;5.001273585;1;algorithm=base,faithful=yes,bits=1
kulasko-rust-serial-bool;2319;5.001624598;1;algorithm=base,faithful=yes,bits=8
kulasko-rust-streamed-bit;1218;5.000006004;4;algorithm=base,faithful=yes,bits=1
kulasko-rust-streamed-bool;1073;5.002411147;4;algorithm=base,faithful=yes,bits=8
kulasko-rust-tiled-bit;7239;5.000118365;4;algorithm=base,faithful=yes,bits=1
kulasko-rust-tiled-bool;1574;5.000242156;4;algorithm=base,faithful=yes,bits=8
```

### 2 threads
```
kulasko-rust-serial-bit;1976;5.000103629;1;algorithm=base,faithful=yes,bits=1
kulasko-rust-serial-bool;1733;5.00161973;1;algorithm=base,faithful=yes,bits=8
kulasko-rust-streamed-bit;1114;5.004093926;2;algorithm=base,faithful=yes,bits=1
kulasko-rust-streamed-bool;1142;5.003198904;2;algorithm=base,faithful=yes,bits=8
kulasko-rust-tiled-bit;3895;5.000005315;2;algorithm=base,faithful=yes,bits=1
kulasko-rust-tiled-bool;1360;5.003268547;2;algorithm=base,faithful=yes,bits=8
```

Kulasko:
Again, thank you for your help so far!

It's reassuring to see the Pi behaving as expected. I was playing with the thought of deleting/not using the multithreaded results save the tiled bit version, but I did see the others catching up (and eventually overtake) the single threaded results with a larger sieve size, so I'll let them stay in, even if just for reference purposes.

mike-barber:
Agreed! It's good to leave them in. Half the fun is how all the generics work together, and comparing the different implementations.

rbergen:
@Kulasko Thank you for adding multithreading to the Rust implementation set. 

I do have to admit that I'm not sure I see the point of adding the single-threaded implementations to this solution. Both the 1 and 8 bit faithful, base algorithm scenarios are already included in @mike-barber's solution.

Kulasko:
@rbergen the point of the single threaded implementation is that it's written in almost the exact same way as the building blocks for the multithreaded ones, thereby serving as a baseline. Although very similar to the solution of @mike-barber, their performance isn't exactly the same.

I also might add a wheel-based implementation if I find the time (probably after this PR), so the solution would present a well-rounded set of all facets.
Nonetheless, if you still say that the single threaded solution is redundant, I will remove it.

rbergen:
@Kulasko I understand. Without saying that the single threaded implementations in your solution are redundant, I would like to ask @mike-barber and yourself if you could maybe find a way to come to one/the best duo of single-threaded implementations, based on your combined efforts?

Kulasko:
@mike-barber would you be okay with a few idiomatic adjustments to your code? For example, using things like `u32::MAX` instead of `0xffffffff`. Since it is the fastest Rust/base solution, I would like it to look as nice as possible.

If we can agree on that, I can remove my single threaded solution.

mike-barber:
@Kulasko I'm definitely happy with cosmetic changes (can't remember _why_ I didn't use u32::MAX before -- there was some reason, or else I just forgot; that happens too). We should definitely try to keep the changes fairly minimal.

One thing we should check before going ahead: I'm not sure if @rbergen will be keen to merge idiomatic/cosmetic changes at this stage that don't improve the solution's performance.

I'll need to re-test performance carefully before approving it, just to make sure there are no regressions. Some of it is non-idiomatic because it was slightly faster that way. And some of it was (originally) to make it slightly more comparable to the C++ solution.

rbergen:
> One thing we should check before going ahead: I'm not sure if @rbergen will be keen to merge idiomatic/cosmetic changes at this stage that don't improve the solution's performance.

@mike-barber I'm not, _unless the original author approves_, with the additional side note that I'm not going to spend time and energy on arranging that approval. :) As you do seem to agree without me making an effort, I'm fine with going ahead with it.

mike-barber:
Thanks @rbergen -- makes sense! 

@Kulasko I've opened https://github.com/mike-barber/daves-garage-primes/issues/1 on my fork -- we can chat about what we want to do there

Kulasko:
@mike-barber just so you're not wondering where I have been, I have been working on it ;)
It's still not finished, though. Also, I have some nasty performance regressions that I still have to figure out and fix in order to move forward.

Kulasko:
As I've seen in other pull requests, since this solution has substantial changes, I'm going to close this PR and open a new one.

PR #462, opened 2021-07-20
Added implementation in Oracle PLSQL

fvbakel:
## Description
Added implementation in Oracle PLSQL. See README.md for the details.

## Contributing requirements


marghidanu:
Please fix the CI errors for your solution in order to move forward.

fvbakel:
@marghidanu I fixed the original issue with a workaround. However the build still fails due to lack of diskspace. I don't know how to solve that.

rbergen:
The technical issue here is that the base image is so big (6.5+GB) that it breaks our CI. I also think that's too much to be able to include it in our automated benchmark run. 

However, there is a more fundamental problem of a different nature. That is that the Dockerfile pulls in Oracle XE, which is commercial software with a proprietary license. I don't think we can pull that into the project without exposing ourselves and those cloning or forking the repo to legal risk I simply can't accept. 

Regretfully, this means that I will have to conclude that we will not be able to merge this.

fvbakel:
@rbergen I don't want to get involved in license issues, so thank you for the warning. I thought it should be fine because the Docker image is build with tools provided by Oracle and the [Oracle-XE](https://www.oracle.com/database/technologies/appdev/xe.html) site reads:

>You get an Oracle Database to use in any environment, plus the ability to embed and redistribute – all completely free!

But just to be on the save side, I will close this PR and remove the pre-backed Dockerfile.

PR #466, opened 2021-07-20
Rust const-fn implementation (solution 6)

Sycration:
## Description

Rust has a const function feature that lets you run code at compile time. I ported the original C++ version to rust and used this feature. This makes it unfaithful as the size of the sieve is determined at compile time. 

## Contributing requirements


rbergen:
@Sycration Please indicate when we can consider this PR to be stable.

Sycration:
@rbergen  I just cleaned up the multithreaded version and it should be all stable & good now.

rbergen (approved):
LGTM

PrimeRust/solution_6/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM rustlang/rust:nightly as build
```
rbergen: Please pin the base docker image to a specific version. We've recently had some spontaneous breakages in our CI due to unpinned version shifts.

Sycration: fixed, with specific digest hash

rbergen: Thank you!

marghidanu (approved):

marghidanu:
Let's wait for the CI to run before merging.

PR #467, opened 2021-07-21
Fix Docker lint warnings

rbergen:
This fixes the Docker lint warnings in existing solutions.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -43,7 +43,7 @@ jobs:
           NAME=$(echo "${{ matrix.solution }}" | sed -r 's/\//_/g' | tr '[:upper:]' '[:lower:]')
           echo "::set-output name=normalized::${NAME}"
       
-      - name: Lint Dockefile
+      - name: Lint Dockerfile
         uses: jbergstroem/hadolint-gh-action@v1
         with:
           dockerfile: "${{ matrix.solution }}/Dockerfile"
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -49,7 +49,7 @@ jobs:
           dockerfile: "${{ matrix.solution }}/Dockerfile"
           output_format: tty
           error_level: 0
-          config_file: ./hadolint.yml
+          config_file: ./config/hadolint.yml
 
       - name: Set up QEMU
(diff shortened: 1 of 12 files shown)


PrimeForth/solution_1/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM mtrute/gforth-container:latest
+FROM mtrute/gforth-container:0.7.9_20190627
```
marghidanu, line 1: I think this warning should be promoted to error. What do you think?

rbergen: I'd say that it's basically the same as not including any tag. So, yes. :)

marghidanu: Feel free to add the rule name in the `hadolint.yml` file under the error section. This will be picked up by Hadolint during CI and re-issued as an error.

rbergen: Would you be ok with moving the hadolint.yml file to a new /config directory?

marghidanu: Sure

rbergen: Then I think I'm done. I'll merge if you agree :)

.github/workflows/CI.yml
```diff
@@ -43,13 +43,13 @@ jobs:
           NAME=$(echo "${{ matrix.solution }}" | sed -r 's/\//_/g' | tr '[:upper:]' '[:lower:]')
           echo "::set-output name=normalized::${NAME}"
       
-      - name: Lint Dockefile
```
marghidanu, line 46: Ooops

.github/workflows/CI.yml
```diff
@@ -43,13 +43,13 @@ jobs:
...
+      - name: Lint Dockerfile
         uses: jbergstroem/hadolint-gh-action@v1
         with:
           dockerfile: "${{ matrix.solution }}/Dockerfile"
           output_format: tty
           error_level: 0
-          config_file: ./hadolint.yml
+          config_file: ./config/hadolint.yml
```
marghidanu, line 52: Cool, will keep this directory for other CI configuration.

marghidanu:
LGTM

PR #468, opened 2021-07-21
Add PrimeNim/solution_3

jamesaorson:
## Description
This solution mostly aims to configure a few small adjustments from Nim's second solution.
Changing the GC is by far the most substantial performance improvement I observed running this locally.
Removing the usage of `method` and changing the backing type from `uint8` to `uint64` on the suggestion of other Nim developers provided no observable speed increase, but are included to give every benefit.

## Contributing requirements


jamesaorson:
solution_2 on my PC runs like so:
```
Passes: 2633, Time: 5.001376390457153, Avg: 0.001899497299831809, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: true
beef331;2633;5.001376390457153;1;algorithm=base,faithful=yes,bits=1
```
solution_3 runs like so:
```
Passes: 5690, Time: 5.000718593597412, Avg: 0.0008788609127587719, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: true
beef331;5690;5.000718593597412;1;algorithm=base,faithful=yes,bits=1
```

rbergen:
@JamesOrson As this represents a performance improvement to solution_2 and doesn't change that solution's implementation characteristics, please structure your PR as an improvement to solution_2 instead of the creation of a new one.

tsoj:
Is just speculation, but maybe it only gives a performance improvement to use `uint64` instead of `uint8` when also calculating the index like this:
```
let 
    byteIndex = i div 64 # instead of 8
    bitInd = i mod 64 # instead of 8
```
(And then also `sieveSize div 64` instead of `sieveSize div 8`)

Maybe it also helps to replace `PrimeSieve = ref object of RootObj` with `PrimeSieve = object`.

jamesaorson:
I will actually be closing this PR in preference of PR #481. There was no real consistent measurable improvement to performance.

PR #471, opened 2021-07-21
100%+ Performance improvement for PrimeCSharp/solution_2

ItalyToast:
## Description
100%+ performance improvement to daves original implementation for C#. (When run under .net core 3.1)

I dont think it needs its own solution so I made the following performance improvements to daves code:

Manual inlining of GetBit and SetBit in the tight loop.
Renaming bitArray to oddNumbersBitArray for clarity.
Marking oddNumbersBitArray readonly to enable the compiler to cache the pointer.
Using BitArray.Count instead of sieveSize so the compiler can more easily remove bound checks.

## Contributing requirements

### DIFF
--- a/PrimeCSharp/solution_2/PrimeCS.cs
+++ b/PrimeCSharp/solution_2/PrimeCS.cs
@@ -13,8 +13,8 @@ namespace PrimeSieveCS
     {
         class prime_sieve
         {
-            private int sieveSize = 0;
-            private BitArray bitArray;
+            private readonly int sieveSize = 0;
+            private readonly BitArray oddNumbersBitArray; //making it readonly so we tell the compiler that the variable reference cant change. around 5% increase in performance
             private Dictionary<int, int> myDict = new Dictionary<int, int> 
             { 
                 { 10 , 4 },                 // Historical data for validating our results - the number of primes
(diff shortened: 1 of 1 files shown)


rbergen:
@ItalyToast Thanks for your submission, I think there are some good improvements in here.
However, I will have to ask you to please revert the rename of bitArray to oddNumbersBitArray. Dave has indicated that he welcomes changes that improve performance, but likes those to be applied with as few "collateral" changes as possible.

ItalyToast:
I wanted to change the name since the inlining changes your mental model of the code. The old GetBit function does the divide by two so you can think of num as a real number but with the inlining it makes more sense to think of num as an "oddnumber index" into the bitarray and thats why it became oddNumbersBitArray. I could revert the name change and/or maybe add some kind of better comment.

rbergen:
I understand your reasoning, but this repo is a mix of a large number of solutions/implementations that in part do, and in part do not divide by two, without making that explicit in the naming of the array. I think you can therefore safely revert the name change, and don't need to worry about commenting to explain.
Most people will look at the code as it is at the time they look at it. It's therefore not necessary to explain the differences between the current state of the code and a previous iteration of it.

ItalyToast:
that makes sense, will revert the name change then.

PR #422, opened 2021-07-15
Added implementation in C with segmented algorithm

fvbakel:
## Description
Added implementation in C with segmented algorithm. See README.md for more details. I assume the algorithm is `other`. However I am not sure, it could also be `base` depending on how flexible you read the rules for the base algorithm.

## Contributing requirements
* [X] I read the contribution guidelines in CONTRIBUTING.md.
* [X] I placed my solution in the correct solution folder.
* [X] I added a README.md with the right badge(s).
* [X] I added a Dockerfile that builds and runs my solution.
* [X] I selected `drag-race` as the target branch.
* [X] All code herein is licensed compatible with BSD-3.

PR #472, opened 2021-07-21
C memcopy algorithm added

rogiervandam:
A new algorithm, almost base which copies memory.
On par or slightly faster than the wheel implementations.

## Description
This implementation uses the base algorithm with a slight extension. It marks all the multiples of a prime in limited range - 2 times the product of all the primes before it (except for 2, which is not stored anyway), multiplied by the size of a word (32 bit, 4 bytes). For example, with prime 7, the range is 2 * (3 * 5 * 7) * 32. This range consists of two blocks of 3 * 5 * 7 * 32. 

So this algorithm is generic and doesn't store any primes up front. It is faithful to the original idea of the sieve, only optimizing CPU and memory performance. However, the rule for base is "the algorithm clears all non-primes individually, increasing the number with 2 * factor on each cycle.". We do clear the non-primes individually, but only up to a certain point. Thereafter, they are cleared simultaneously. So it is a slight alteration of the base model.

I added this because it is the fastest C algorithm on my machine.
I saw a solution_3 in the pull requests, so i went ahead and made this solution_4. 

## Contributing requirements

### DIFF
--- a/PrimeC/solution_4/Dockerfile
+++ b/PrimeC/solution_4/Dockerfile
@@ -1,12 +1,11 @@
-FROM ubuntu
+FROM ubuntu:20.04 as build
 
-# Get GCC
-RUN apt-get update
-RUN apt-get install -y build-essential
+RUN apt-get update -qq && apt-get install -y build-essential --no-install-recommends && apt-get clean && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /home/primes
 
 COPY sieve_memcopy.c run.sh ./
+
 RUN chmod a+x run.sh
 
 ENTRYPOINT ["./run.sh"]
--- a/PrimeC/solution_4/Dockerfile
+++ b/PrimeC/solution_4/Dockerfile
@@ -1,0 +1,12 @@
+FROM ubuntu
+
+# Get GCC
+RUN apt-get update
+RUN apt-get install -y build-essential
+
+WORKDIR /home/primes
+
+COPY sieve_memcopy.c run.sh ./
+RUN chmod a+x run.sh
+
+ENTRYPOINT ["./run.sh"]
--- a/PrimeC/solution_4/README.md
+++ b/PrimeC/solution_4/README.md
@@ -1,6 +1,6 @@
 # C language solution by Rogier van Dam
 
-![Algorithm](https://img.shields.io/badge/Algorithm-other-green)
+![Algorithm](https://img.shields.io/badge/Algorithm-other-yellowgreen)
 ![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
 ![Parallelism](https://img.shields.io/badge/Parallel-no-green)
 ![Bit count](https://img.shields.io/badge/Bits-1-green)
@@ -17,7 +17,7 @@ I added this because it is the fastest C algorithm on my machine.
 ## Run instructions
 You can use the Dockerfile.
 
-First compile the code using the **run** script. Optimization parameters are at the beginning of the script. They are set up to use the gcc compiler.
+First compile the code using the **run.sh** script. Optimization parameters are at the beginning of the script. They are set up to use the gcc compiler.
 
 Then run the code using the run.sh script, with the command:
 ./run.sh
--- a/PrimeC/solution_4/README.md
+++ b/PrimeC/solution_4/README.md
@@ -1,0 +1,29 @@
+# C language solution by Rogier van Dam
+
+![Algorithm](https://img.shields.io/badge/Algorithm-other-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-1-green)
+
+This implementation uses the base algorithm with a slight extension.
+It marks all the multiples of a prime in limited range - 2 times the product of all the primes before it (except for 2, which is not stored anyway), multiplied by the size of a word (32 bit, 4 bytes). For example, with prime 7, the range is 2 * (3 * 5 * 7) * 32. This range consists of two blocks of 3 * 5 * 7 * 32. The first block misses some multiples of 3, 5 and 7, because they where marked in the sieve as prime. The second block has all the multiples of 3, 5 and 7, starting at a 32 bit alignment and with a size of a multiple of 32. This block can be repeated multiple times.  
+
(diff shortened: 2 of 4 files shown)


PrimeC/solution_4/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ubuntu
```
rbergen: Please fix the Dockerfile lint error found.

PrimeC/solution_4/README.md
```diff
@@ -0,0 +1,29 @@
+# C language solution by Rogier van Dam
+
+![Algorithm](https://img.shields.io/badge/Algorithm-other-green)
```
rbergen: The badge colour to use for any algorithm other than `base` is -yellowgreen instead of -green. (I missed this in my review of your NodeJS memcopy implementation, where I've fixed it myself.)

PrimeC/solution_4/Dockerfile
rogiervandam: Fixed

PrimeC/solution_4/README.md
rogiervandam: Fixed

rbergen:
From a conceptual perspective, the algorithm sounds rather similar to that one implemented in #422. @rogiervandam and @fvbakel, could you please lend me your perspectives on this?

rogiervandam:
> From a conceptual perspective, the algorithm sounds rather similar to that one implemented in #422. @rogiervandam and @fvbakel, could you please lend me your perspectives on this?

@fvbakel @rbergen 
I see we each at the same time have come up with conceptually the same (generic) idea of copying a pattern. 
- The "segmented solution" implementation has some hard limits (1% of max product, dont run for limit < 100000 and focus on bits in word). 
- The incremental memcopy solution is generic in the sense that these hard limits are not imposed. In the solution i added some other optimizations (not using sqrt) to see if it was possible to devise the fastest "other" algorithm still being faithful. 
This means that the performance profile across different limits is different.

fvbakel:
@rogiervandam @rbergen I think both solutions are indeed based on the same mathematical concepts. I think the most important differences are:

1. This solution does the memory copy multiple times in a generic way, where solution_3 first attempts to calculate the optimal prime number to do the memory copy and then does the memory copy only once. 
2. This solution uses the C style memory copy. Solution_3 is based on `memcpy`. This can be a tricky difference. In some cases, like on the alphine the use of `memcpy` has has bad consequences, because the implementation of memcpy on alphine is bad for performance. However, I think in some cases `memcpy` can give a performance improvement because the implementation of `memcpy` on the target system is able to use hardware specific instructions if the copy involves multiples of 32 bit. See for example [https://stackoverflow.com/questions/17498743/how-does-the-internal-implementation-of-memcpy-work](https://stackoverflow.com/questions/17498743/how-does-the-internal-implementation-of-memcpy-work)

Small differences:
1. This solution uses different compile options. That does not seem not to matter much to me.
2. This solutions makes it explicit what should be in lined by the compile
3. The code style is different

Now we have both we can see what the difference is in performance between the two strategies. On my hardware, one i5 and one i7 laptop the performance is very close. I wonder what the results are on other hardware. I think both solutions are a nice addition to this project.

rbergen:
@fvbakel Thank you for sharing your perspective so quickly. 

I haven't come to a final conclusion yet, but if I'm honest then I do currently lean towards thinking that the implementations are so similar that we should not have two separate solutions for this approach in the codebase. Would it be possible to get one "best of both worlds" implementation out of the two?

fvbakel:
@rbergen I would really have to dig into the details to think what constitutes a best of both worlds solution, but I guess it would be an interesting exercise. 

One thing to be aware of is that the product of primes has a exponential growth where the grow factor is also growing. So this product is at first very small compared to the sieve size and then it quickly exceeds the sieve size. It is only for the prime product that are marked in yellow in this [sheet](https://drive.google.com/file/d/13QFscZiMkNmsMHWYokQRi0eXSg9AyxWq/view?usp=sharing) where the memory copy matters in my opinion. 

For now I have run both tools in a loop like this:

```sh
#/bin/bash
source ~/node/add-path.sh
OUTPUT_FILE=/home/fvbakel/Documenten/primes/results/output-C-2020-07-23.csv
i=0
while [ $i -lt 303 ]
do
    make DIRECTORY=PrimeC/solution_3 FORMATTER=csv >>$OUTPUT_FILE
    make DIRECTORY=PrimeC/solution_4 FORMATTER=csv >>$OUTPUT_FILE
    i=$[$i+1]
done

```

I am running on native Ubuntu  21.04 on an Intel i7-3520M Details on my hardware and run conditions are in the attachment: [myhardware_json.txt](https://github.com/PlummersSoftwareLLC/Primes/files/6870506/myhardware_json.txt). 

The outcome is of this [measurement](https://drive.google.com/file/d/1LQ7YFg15TcHjbgs-4nLSS26Hks8OzkRa/view?usp=sharing)  is:

|Label|Count - PassesPerSecond|Max - PassesPerSecond|Median - PassesPerSecond|StDev - PassesPerSecond|
|--|--|--|--|--|
fvbakel_Cwords|303|2390,91775|2076,77612|65,2856305591815
rogiervandam|303|2362,82988|2070,93746|66,5451071784725


I also repeated a similar test under windows wsl2 without docker, and and had similar relative results..

**Conclusion**
The performance of both programs is the same in this test.

**Next steps**
Do some more testing, also with higher limits. Determine what is the best of both worlds and come up with an even faster solution :-).

rbergen:
> Determine what is the best of both worlds and come up with an even faster solution :-)

I accept up-front that the following could be seen as a cliché comment coming from a maintainer of a repo like this, but evolving this combination of implementations into an even faster solution would actually be the best outcome :)

fvbakel:
@rbergen I hope you consider PR #517 as "best outcome" :-)

rbergen:
> @rbergen I hope you consider PR #517 as "best outcome" :-)

@fvbakel I would actually prefer it if that wasn't up to me to conclude :) 
@rogiervandam What is your perspective on this?

rogiervandam:
I think fvbakel did a great job with this algorithm. I'll go a different route with some new ideas. As for now, lets close this PR and let fvbakel with PR #517 have the upper hand.

rbergen:
@rogiervandam Thank you for your quick response.
@fvbakel I'll proceed with merging #517 based on this feedback.

fvbakel:
@rbergen Thank you for the coordination.

@rogiervandam Thank you for your feedback.

PR #474, opened 2021-07-22
Go solution_2 variants

ssovest:
## Description
Recently i had a chance to check my Go solution on another, more modern, machine, and what i've noticed is that some decisions like bit storage data type or certain compiler flags have the opposite effect on different machines. So i added a few variations of the original solution:
 - sieve8.go is the original one
 - sieve32.go is like the original, but uses uint32
 - sieve_ptr.go uses uint32 and unsafe pointers

And a couple of fun ones:
 - sieve_other.go implements the same algorithm as PrimeCY_32 solution
 - sieve_cgo.go also implements the same algorithm, but utilizes Go's ability to call C code. I guess it makes it unfaithful. (Well, it isn't exactly an external dependency, but C is not Go and it requires a C compiler)

## Contributing requirements


rbergen:
> sieve_cgo.go also implements the same algorithm, but utilizes Go's ability to call C code. I guess it makes it unfaithful. (Well, it isn't exactly an external dependency, but C is not Go and it requires a C compiler)

This is basically a Go version of the Python and C implementation that #370 aims to add. As my latest comment on that PR indicates:

- We've decided to put these "mixed language" implementation in the PrimeMixed directory, instead of PrimeGo (or PrimePython, ...). That does of course mean that a separate solution would have to be created for it.

- Mixed language implementations are indeed unfaithful by definition. The "external dependency" assessment should not be made from the perspective of the solution/implementation, but from the perspective of the language the solution is filed under. In solutions like these, it doesn't matter which of the two languages you pick, there's always an external dependency. The Go code doesn't work without the C code, and the C code doesn't work without the Go code.

PrimeGo/solution_2/build.sh
```diff
@@ -0,0 +1,12 @@
...
+go build -o ./sieves/sieve8 sieve8.go &&
+go build -o ./sieves/sieve32 sieve32.go &&
+go build -o ./sieves/sieve_ptr sieve_ptr.go &&
+go build -o ./sieves/sieve_cgo sieve_cgo.go &&
+go build -o ./sieves/sieve_other sieve_other.go &&
+go build --gcflags="-B" --ldflags="-X 'main.label=ssovest-go-uint8-B'" -o ./sieves/sieve8_b sieve8.go &&
+go build --gcflags="-B" --ldflags="-X 'main.label=ssovest-go-uint32-B'" -o ./sieves/sieve32_b sieve32.go &&
+go build --gcflags="-B" --ldflags="-X 'main.label=ssovest-go-ptr-B'" -o ./sieves/sieve_ptr_b sieve_ptr.go &&
+go build --gcflags="-B" --ldflags="-X 'main.label=ssovest-cgo-B'" -o ./sieves/sieve_cgo_b sieve_cgo.go &&
+go build --gcflags="-B" --ldflags="-X 'main.label=ssovest-go-other-B'" -o ./sieves/sieve_other_b sieve_other.go
```
jdemchuk: Using -B in gcflags disables array bounds checks.  It could save a little performance, but it might be considered bad practice if this was a real world application.  I still think this is legal in a drag race, where anything goes to get the most performance, but it does strip away one of the benefits of using Go's slices vs C arrays.

ssovest: Yep, definitely not the kind of thing one should use in production. Still, for a small fun program it's okay, i think.

rbergen (approved):
LGTM

marghidanu (approved):

PR #477, opened 2021-07-23
Change conditionals to asserts, since the checks are only useful at the debugging stage.

Kinematics:
Speeds up this pair of sieves by a factor of 2.5x in docker, or about 25% in native.

Run in docker:
bool: 2601 -> 6168
ibool: 3026 -> 8824


Kinematics:
That should be all of the sieves that benefit from updating to an `Assert()`.  Run numbers and benchmark numbers updated as well.

rbergen:
@Kinematics Thanks for this update!

PR #479, opened 2021-07-23
Fix REXX/solution_1 Docker lint warning

rbergen:
Fixes apt => apt-get lint warning in PrimeREXX/solution_1/Dockerfile

### DIFF
--- a/PrimeREXX/solution_1/Dockerfile
+++ b/PrimeREXX/solution_1/Dockerfile
@@ -1,6 +1,6 @@
 FROM ubuntu:20.04
-RUN apt update && apt upgrade -y \
-	&& apt install -y regina-rexx
+RUN apt-get update && apt-get upgrade -y \
+	&& apt-get install -y regina-rexx
 WORKDIR /home/rexx
 COPY ./PrimeREXX.rex .
 # Print the required output
--- a/PrimeREXX/solution_1/Dockerfile
+++ b/PrimeREXX/solution_1/Dockerfile
@@ -1,5 +1,5 @@
 FROM ubuntu:20.04
-RUN apt-get update && apt-get upgrade -y \
+RUN apt-get update \
 	&& apt-get install -y regina-rexx
 WORKDIR /home/rexx
 COPY ./PrimeREXX.rex .


PrimeREXX/solution_1/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM ubuntu:20.04
-RUN apt update && apt upgrade -y \
-	&& apt install -y regina-rexx
+RUN apt-get update && apt-get upgrade -y \
```
marghidanu: Why is the `apt-get upgrade` required here?

rbergen: Good question, to which I don't have the answer. I focused on the lint warning, which required apt to be replaced by apt-get. I think we will need input from @jossRRM to answer your question.

rbergen: I decided to just remove the apt upgrade and see what happened. The solution still builds and runs, so I've committed that change.

jossRRM (commented):
I was under the impression that apt install and apt upgrade are the new (and accepted) commands. Was I wrong? reference: https://phoenixnap.com/kb/apt-vs-apt-get

I have an upcoming pull request for some changes (cosmetic) to the code and one for a solution_2 that uses NetRexx, a compiled language that is based on Rexx, with classes and arrays.  Should I make your changes to my Dockerfile? Also, is the upgrade not needed?

EDIT:
I just saw your emails regarding this. Since replacing the apt with apt-get and removing apt upgrade work, I see no difficulties with these changes. I will make them to my local and repository Dockerfile(s).

rbergen:
> I was under the impression that apt install and apt upgrade are the new (and accepted) commands. Was I wrong?

I don't think it's a black-and-white right vs wrong thing. I believe consensus (still) is that apt-get is the command to use in scripted scenario's, which a Dockerfile obviously really is. In any case, with the consent you've now given I will indeed merge this.

jossRRM:
Hi, Rutger. I tried to create a pull request from my repository. I had changes to solution_1 and a new solution_2, both under the PrimeREXX directory. When I tried to point to my repository both solutions were added. But the solution_2 uses a different REXX dialect and I think it should be in a
separate pull request. Is there a way to separate the two solutions into two separate requests without making a new directory under Primes for solution_2?
Joe Ossino (jossRRM)
---------------------------- Original Message ----------------------------

Subject: Re: [PlummersSoftwareLLC/Primes] Fix REXX/solution_1 Docker lint warning (#479)

From: "Rutger van Bergen" ***@***.***>

Date: Sun, July 25, 2021 6:08 pm

To: "PlummersSoftwareLLC/Primes" ***@***.***>

Cc: "jossRRM" ***@***.***>

"Mention" ***@***.***>

--------------------------------------------------------------------------

PR #482, opened 2021-07-24
Add much faster Nim version as solution_3

GordonBGood:
## Description
Add a Nim version that corrects many of the problems with "solution_1" and "solution_2" including more efficient use of CPU caches and "Extreme Loop Unrolling", which is a technique similar to the "striping" technique used by the fastest Rust implementation, but refined yet further.

## Contributing requirements


rbergen:
I think we're close to merging #309, which adds a number of Kotlin implementations already. I suggest you review that PR, and discuss any improvements you feel you can make to that PR's contents, there.

rbergen:
The `base` algorithm is defined rather specifically in the contributing guidelines. Although I can see that the algorithm in this implementation is inspired by that algorithm, it is not the base algorithm itself. For that reason, I think the algorithm in this PR's implementation will have to be marked as `other`.

GordonBGood:
> The `base` algorithm is defined rather specifically in the contributing guidelines. Although I can see that the algorithm in this implementation is inspired by that algorithm, it is not the `base` algorithm itself. For that reason, I think the algorithm in this PR's implementation will have to be marked as `other`.

With respect, I would argue that you have allowed the Rust "striped" implementation to be marked as `faithful` `base` when it is derived from the `base` specification in a very similar way to the way this implementation is, as follows:

1. The `base` algorithm culls across the array by a span derived as a factor of the base prime value being culled with a bit mask value determined from the bit index to be marked as composite (for the "bit" implementation); this bit mask value could be produced from a Look Up Table (LUT) produced from the bit index instead of a shifted value and I believe you would still consider the algorithm to be `base`?
2. The "striped" algorithm separates the culling loop into eight separate loops in sequence with a fixed bit mask for each, which bit mask can be produced by shifting or by use of the above bit position mask LUT.
3. This algorithm takes the "striped" algorithm and combines the eight "striped" loops into one loop again, with each of the mask values according to the modulo pattern from a `computed goto` as produced by the macro (or could be done from a template), which bit mask values are exactly the same as the "striped" or any of the `base` algorithms would have used.

There is no problem categorizing this algorithm as `other` as long as it is consistent in that other implementations that use similar techniques are also categorized as `other`.   For instance, I see that two of the Julia implementations are allowed to use SIMD instructions and still be called `faithful base`. So what do you see as the characteristics of this algorithm that makes it so different from the Rust "striped" algorithm which is categorized as `base`, when all it does is take the groups of eight loops per base prime with the bit mask calculated only at the start of each of the eight loops and combine them into a single loop with still just one culling loop per base prime value?

Or is it that you have a problem with use of a bit mask LUT from the outset?  If so, I can remove the eight byte `BITMASK` LUT without affecting throughput very much

It may be that I can modify this implementation slightly so that it will qualify as `base` by your definition.

If this algorithm or something similar can't be categorized as `base` in that you don't see the above three points as describing the same algorithm, then I will change the tag to `other`.  If you require me to do this, Is it still `faithful`?

rbergen:
I'm afraid you'll have to help me a little with the pointers to other implementations you include in your comment. As in: for each reference, please make explicit which solution/implementation you're referring to, by mentioning the solution number and, where applicable, implementation source file.

GordonBGood:
> I'm afraid you'll have to help me a little with the pointers to other implementations you include in your comment. As in: for each reference, please make explicit which solution/implementation you're referring to, by mentioning the solution number and, where applicable, implementation source file.

Quoting my applicable comments:
> The Rust "striped" algorithm separates the culling loop into eight separate loops in sequence with a fixed bit mask for each, which bit mask can be produced by shifting or by use of the above bit position mask LUT.

That is from PrimeRust/solution_1/src/main.rs where the "striped" culling in in the `struct` definition starting at line 193 and the "striped" culling trait implementation starts at line 202 with the `fn reset_flags` implementing "striping" at starting at line  211 where the eight loops per base prime are at line 213; it states it is `base` and `faithful` at line 347.

This is clearly an extension of what the `base` algorithm is stated to be, so I thought my implementation that just differs from this Rust "striping" a little bit would also quality.  If you demote the Rust implementation to `other`, than I can certainly see that my implementation should be `other` too.

Like I said:  "It may be that I can modify this implementation slightly so that it will qualify as base by your definition.", but in order to do that, I need to know why you think it doesn't qualify.

Your first qualification for `base` is as follows:

* The algorithm uses an outer loop, in which two operations are performed:
i. searching for the next prime in the sieve, and
ii. clearing this prime's multiples in the sieve.
In order to allow for idiomatic code, **it is permissible to perform these steps in either order, provided the results are correct.**

I thought both the "striped" and my algorithm would still qualify for these conditions as, although the ordering of search and clearing, and the ordering of actual culling in the sieve buffer are not the same as the original implementations, the number of marking operations for both are exactly the same as the original (about 811,000 marking operations per sieve pass); of course, the results are correct.

For your `base` condition as follows:
* When clearing non-primes in the sieve (the second operation), the algorithm clears all non-primes individually, increasing the number with 2 * factor on each cycle.

Both the "striping" algorithm and my algorithm do use a `2 * factor` culling span and mark the composite bits individually; in fact, my algorithm is more true to this condition than the "striping" one in that the culling/marking of composite bits for my algorithm is more in order than the "striping" algorithm, as "striping" skips every eighth cull, with all culls filled in due to the eight inner loops.

Again, I thought that if you accepted Rust "striping" as `base`, then my algorithm should definitely qualify as `base`.

> I see that two of the Julia implementations are allowed to use `SIMD` instructions and still be called `faithful base`.

Sorry, I miss-spoke, I only found one use of Julia @simd as follows:  That's from Julia solution_3/primes_1of2.jl, uses `@simd at line 104; states it is `base` and `faithful` at line 181.

I am not too concerned with this Julia example that does "cheat" a little bit in the use of SIMD as it usually only gives it a little advantage that is dependent on the CPU on which it is run; but just used it to show that there are broader definitions of what `base` actually means in other implementations.

**So in conclusion, if you see the similarities between my algorithm and Rust "striping", it would seem you should accept both or neither as `base`.**  On the other hand, if you see something in my code that is not in Rust "striping" that you don't feel should qualify, let's discuss it to see if I can made it conform.  As I said, the use of the `BITMASK` LUT makes a negligible difference in performance so I have removed it so as not to confuse the real issues.  Other implementation details can also be adjusted.

**The big question you need to answer is if you accept culling/marking/clearing operations that are out of order as is used by both "striping" and my algorithm.**

rbergen:
Thank you for your further elaboration. I think I see the point you're making, and maybe the solution is not that far away. 

I'll start with making a statement on the question you asked in bold: we do not currently have specific limitations on the bit/flag storage algorithm other than what is described as the rules for a faithful implementation. And to be honest, I don't think we can, because in many languages/implementations the actual storage of bits/flags in memory is handled by framework/library-provided types. The inner workings of those should be considered a black box, _and can be_ because they are disconnected from the sieve run algorithm by the interfaces the storage types provide. Those interfaces often resemble some sort of "array-like" indexing. 

I think this brings us to the difference between Rust/solution_1 and the solution this PR aims to add. The striping in the former implementation is isolated from the actual sieve run algorithm, which is therefore still easily recognizable as the base algorithm and thus indeed qualifies as such.

Would it be possible to structure your code in such a way that the storage logic is isolated from the sieve run algorithm, as well? If so, I'm happy to reconsider my earlier statement about the algorithm's classification.

ssovest:
@rbergen  Sorry for interfering with the dicussion, but what interface should such a type provide? Should it be defined by the language's conventions for sequence types or by the rules and be the same for all languages? In other words, if sequence types in some language provide some advanced api beyond getting and setting individual values, would a solution using a custom type which is interchangeable with at least one type from the standard library be considered an implementation of the base algorithm?

rbergen:
@ssovest So far, we've seen no reason to pin down the exact patterns for this. As the range of solutions in this repo illustrates, up to this point we've accepted arrays, indexers, indexer overloads, GetBit/ClearBit function/method variations, etc., captured in separate storage types or embedded in the sieve type, with or without in-index integer divides skipping even numbers and basically any recognizable variations thereof. To summarize that last rather long sentence: I think we've been quite flexible when it comes to this internal interface of the implementations. Obviously, and I say this without any intention of being pedantic, the interface used for a particular language needs to be supported by that language for it to work at all.
All that said and as I've noted in earlier comments, we don't have the illusion that we can cover every possible creative approach with a water-tight ruleset, up-front. That means that we will consider each case individually, spend extra time on approaches that are new to us, and clarify the ruleset when we conclude that is called for.

GordonBGood:
@rbergen:

> I'll start with making a statement on the question you asked in bold: we do not currently have specific limitations on the bit/flag storage algorithm other than what is described as the rules for a faithful implementation. And to be honest, I don't think we can, because in many languages/implementations the actual storage of bits/flags in memory is handled by framework/library-provided types. The inner workings of those should be considered a black box, _and can be_ because they are disconnected from the sieve run algorithm by the interfaces the storage types provide. Those interfaces often resemble some sort of "array-like" indexing.

Okay, I see that, if you try to be too restrictive, then idiomatic solutions in languages like Python, Julia, Fortran, etc. using adjustable-step built-in slicing couldn't be allowed.

> I think this brings us to the difference between Rust/solution_1 and the solution this PR aims to add. The striping in the former implementation is isolated from the actual sieve run algorithm, which is therefore still easily recognizable as the base algorithm and thus indeed qualifies as such.

Ah, so (paraphrasing), if the implementation can make it obvious that the `faithful base` algorithm is being used by making the storage interface separate, then such an implementation will pass, as long as the the storage interface comes from the language's standard library or is part of the implementation (no external libraries).

Then the essential two parameters that define a `faithful base` algorithm are only that storage be twice what it really needs to be in wasting the representations for the even numbers (in order to be same as the original implementation by Dave Plummer) and thus requiring that the testing/marking step size be two times the base prime values, and that the full "odds-only" algorithm be used, which requires about 810K marking operations for a range of a million.  My algorithm qualifies on all counts.

> Would it be possible to structure your code in such a way that the storage logic is isolated from the sieve run algorithm, as well? If so, I'm happy to reconsider my earlier statement about the algorithm's classification.

Of course I can refactor my code to comply by making the `sieveBuffer` field a custom `BitSeq` type/class and implementing all storage accesses through custom functions/methods/operators on that `BitSeq` type/class and am proceeding to do so.  Give me a day or so to test it thoroughly before submitting it.

It is going to take twice as long to compile as it will be written as a general `BitSeq` type/class, meaning it needs to take into account of all step sizes and not just the cases for odd primes so the macro is going to automatically generate about twice as many lines of code, but that isn't too much of a problem, as some other languages take longer to compile even the simple algorithms.

One other thing for your consideration:  the choice of the Docker image to be used is up to the implementer, but there is some discussion in the README.md files for the different solutions that some Docker images are faster at allocating/deallocating memory or in implementations of copying and moving memory.  If so, then that is a variable that makes solution comparisons not really applicable.  **Shouldn't this be standardized such as "Implementations must always use the Alpine Docker image."?**

GordonBGood:
@rbergen:

I've changed my submission in a way that I think will qualify as per your last suggestion.  The memory interface has been moved into "methods"/functions to do with the new `BitSeq` type/class.

The key composite culling/marking is all done at line 178, and is constructed to look like a "slicing" operation as in languages that have step-slicing-by-a-value-other-than-one such as Python, Fortran, Julia, etc.; since Nim does not have that ability (it has "by one" slicing for arrays, sequences, and strings, not does not have a bit array nor stepping by values other than one), I implemented it in the file with enough capabilities to serve the needs of this implementation.

I hope you find the new contribution acceptable to the `faithful base` algorithm.

rbergen:
> Then the essential two parameters that define a `faithful base` algorithm are only that storage be twice what it really needs to be in wasting the representations for the even numbers (in order to be same as the original implementation by Dave Plummer) and thus requiring that the testing/marking step size be two times the base prime values, and that the full "odds-only" algorithm be used, which requires about 810K marking operations for a range of a million.

This is an interesting point to focus on. The fact is that "using twice as much storage as needed for odd numbers" is _not_ a requirement, but indeed accepted. If you want to reserve and use only the memory that is required for the odd numbers then that is also allowed. In relation to that, the step size for testing potential primes and clearing multiples of a prime that's been found can be adapted accordingly (we don't demand the step size for the second operation to be implemented as `factor * 2 / 2` in code, for obvious reasons). As a contributor, I would therefore choose the approach (odds only, or odds and evens) that is fastest in the language you use to write your implementation. 

Concerning processing only odd numbers: without committing to it without having source code to look at, I would actually be inclined to also permit implementations that test and clear even numbers. In Dave's own words, implementations are allowed to skip them because 2 is the only prime of which the algorithm is allowed to use prior knowledge that it is indeed a prime. I haven't actually investigated this, but my guess would be that no one has submitted an algorithm that doesn't skip even numbers because no language has yet been found in which _not_ skipping them is _faster_.

>  Shouldn't this be standardized such as "Implementations must always use the Alpine Docker image."?

For one, we can't do so without excluding a number of languages that simply refuse to work on Alpine, and we don't want to do that. 
In more conceptual terms, the composition of a solution consists of whatever is in the solution's files. That includes the programming language source file(s), the Dockerfile, and any supporting artifacts used/required to build and run the solution('s implementations). There is an, I think relatively limited, number of ground rules, and some more focused rules concerning labelling algorithms, faithfulness and bits-per-prime marking. Personally, I would like to keep them as focused as they currently are.

Part of the issue is that once you start enforcing standardization on a particular Docker base image, then one could also argue the same should be done for runtime environments/VM's (where applicable), compilers (again, where applicable), compiler flags, etc.. I don't think it is possible in any meaningful way, to standardize on that entire chain for a collection of 60+ languages of wildly varying types and ancestries.

One _could_ argue that the current set of rules that apply are somewhat arbitrary both in their scope and their actual composition. As I've said multiple times before, I for one accept that as being part and parcel of this project and projects of a similar nature. So far, it seems that 91 contributors have found a way to deal with them. :)

GordonBGood:
> > Then the essential two parameters that define a `faithful base` algorithm are only that storage be twice what it really needs to be in wasting the representations for the even numbers (in order to be same as the original implementation by Dave Plummer) and thus requiring that the testing/marking step size be two times the base prime values, and that the full "odds-only" algorithm be used, which requires about 810K marking operations for a range of a million.
> 
> This is an interesting point to focus on. The fact is that "using twice as much storage as needed for odd numbers" is _not_ a requirement, but indeed accepted.

Ah, I misunderstood; I saw that many of the contributors seem to understand the restriction as I did so did the "waste the memory" thing.  They were perhaps confused by the statements in the "CONTRIBUTING.md" document that say "The size of the memory buffer must correspond to the size of the sieve." (rather than "The size of the memory buffer must be proportional to the size of the sieve") and "When clearing non-primes in the sieve (the second operation), the algorithm clears all non-primes individually, increasing the number with 2 * factor on each cycle." **(!!!)**.

I don't think that any competent programmer would use twice as much memory as necessary in including the even representations (especially when the representations are larger than single bits) as that will take longer to allocate and especially initialize.  I'll do some experiments reducing the amount of memory used in my implementations since you say that is acceptable. 

> In relation to that, the step size for testing potential primes and clearing multiples of a prime that's been found can be adapted accordingly (we don't demand the step size for the second operation to be implemented as `factor * 2 / 2` in code, for obvious reasons). As a contributor, I would therefore choose the approach (odds only, or odds and evens) that is fastest in the language you use to write your implementation.

That makes sense:  if you don't require that the array have a representation for the full range including the evens, then there is no need to skip over the evens and one can advance the marking by just the base prime value "in play" at the time.

> Concerning processing only odd numbers: without committing to it without having source code to look at, I would actually be inclined to also permit implementations that test and clear even numbers. I haven't actually investigated this, but my guess would be that no one has submitted an algorithm that doesn't skip even numbers because no language has yet been found in which _not_ skipping them is _faster_.

Of course, using an odds-only algorithm will always be faster than "odds and evens" because there are over two and a half times less marking operations, plus the extra time allocating and initializing the larger array as mentioned above.

> > Shouldn't this be standardized such as "Implementations must always use the Alpine Docker image."?
> 
> For one, we can't do so without excluding a number of languages that simply refuse to work on Alpine, and we don't want to do that.
> In more conceptual terms, the composition of a solution consists of whatever is in the solution's files. That includes the programming language source file(s), the Dockerfile, and any supporting artifacts used/required to build and run the solution('s implementations). There is an, I think relatively limited, number of ground rules, and some more focused rules concerning labelling algorithms, faithfulness and bits-per-prime marking. Personally, I would like to keep them as focused as they currently are.
> 
> Part of the issue is that once you start enforcing standardization on a particular Docker base image, then one could also argue the same should be done for runtime environments/VM's (where applicable), compilers (again, where applicable), compiler flags, etc.. I don't think it is possible in any meaningful way, to standardize on that entire chain for a collection of 60+ languages of wildly varying types and ancestries.
> 
> One _could_ argue that the current set of rules that apply are somewhat arbitrary both in their scope and their actual composition. As I've said multiple times before, I for one accept that as being part and parcel of this project and projects of a similar nature. So far, it seems that 91 contributors have found a way to deal with them. :)

My comments about Docker were just an aside in trying to make things fairer if some implementations get better performance from one Docker image or another.  If some contributions don't work at all with some "standard" Docker images such as Alpine, then there is no choice but to accept any Docker image.

Since my implementations tend to use packed-bits and thus less memory, they don't have a problem with any Docker image available, in this case Alpine.

rbergen:
> They were perhaps confused by the statements in the "CONTRIBUTING.md" document that say "The size of the memory buffer must correspond to the size of the sieve." (rather than "The size of the memory buffer must be proportional to the size of the sieve")

I think we're solidly in the area of semantics here :) The use of "correspond" was intentional, in the sense that we don't want people to submit solutions that proportionally reserve 4, 8, 16, etc. times as much memory as required for the sieve's prime candidates (whether memory space is reserved for even numbers, or not).

> "When clearing non-primes in the sieve (the second operation), the algorithm clears all non-primes individually, increasing the number with 2 * factor on each cycle." **(!!!)**.

This phrasing urges implementers to skip even numbers, which is a premise of the entire discussion of the base algorithm. It does _not_ rule out implementations that don't bother with reserving memory for even numbers at all. In many implementations where sieve logic and storage logic are separated at the code level, the skipping of even numbers is in fact implemented in the storage logic. Another way to phrase it is that the rule you quote is a statement in terms of the field of sieve numbers (odd and even), not the projection of that field on memory.

If the rules were written in the level of detail of a technical specification then all of this may well be included in the text of the guidelines. From a general readability perspective, I'm happy that's not the case.

> I don't think that any competent programmer would use twice as much memory as necessary in including the even representations (especially when the representations are larger than single bits) as that will take longer to allocate and especially initialize. 

Based on my own testing, you may well be surprised. I've seen some rather counterintuitive numbers concerning this.

GordonBGood:
@rbergen:

> I think we're solidly in the area of semantics here :) The use of "correspond" was intentional, in the sense that we don't want people to submit solutions that proportionally reserve 4, 8, 16, etc. times as much memory as required for the sieve's prime candidates (whether memory space is reserved for even numbers, or not).

Yes, semantics.  But it would take quite an idiot of a programmer to use more memory than necessary, especially huge factors of whats necessary.  Although that is what programmers that use Fortran's logical type for this are, in that the Fortran logical type is four or eight bytes per number representation ;-)

> > "When clearing non-primes in the sieve (the second operation), the algorithm clears all non-primes individually, increasing the number with 2 * factor on each cycle." **(!!!)**.
> 
> This phrasing urges implementers to skip even numbers, which is a premise of the entire discussion of the base algorithm. It does _not_ rule out implementations that don't bother with reserving memory for even numbers at all. In many implementations where sieve logic and storage logic are separated at the code level, the skipping of even numbers is in fact implemented in the storage logic. Another way to phrase it is that the rule you quote is a statement in terms of the field of sieve numbers (odd and even), not the projection of that field on memory.
> 
> If the rules were written in the level of detail of a technical specification then all of this may well be included in the text of the guidelines. From a general readability perspective, I'm happy that's not the case.

Okay, now that I understand your intention, let's move on.

> > I don't think that any competent programmer would use twice as much memory as necessary in including the even representations (especially when the representations are larger than single bits) as that will take longer to allocate and especially initialize.
> 
> Based on my own testing, you may well be surprised. I've seen some rather counterintuitive numbers concerning this.

Ha, with my implementation, taking out the redundant evens representations means half the sizes of allocation and initialization and does made a difference:  I got about 19150 passes before removing the redundant memory and about 20,500 passes after!

I'll do a new commit to my contribution with my new understanding of the restrictions shortly.

GordonBGood:
@rbergen:

> I'll do a new commit to my contribution with my new understanding of the restrictions shortly.

This is one of those weird Docker issues and I don't know what to do.  Previously, the version with the wasted memory including evens ran about the same speed using Docker as on my local machine.  Now that I have removed using the wasted memory, when run on my local machine, it is about seven percent faster, but **when run on the Docker image (Alpine) it is about six percent slower,** with both fairly repeatable over multiple runs.

I can explain why running on my local machine is faster due to less memory operations, bu II can't explain why it's slower on the Docker image, but I don't know that much about the Docker image.

So should I do a new commit with the new file to get better "real" speed or keep the old one in the interests of getting higher scores in the "drag race" using Docker?

As I said, that's the problem with using Docker - inconsistency.

rbergen:
What is the host OS for Docker? Windows, MacOS, Linux?

GordonBGood:
@rbergen:

> What is the host OS for Docker? Windows, MacOS, Linux?

I've switched to doing all of my development work on Fedora Linux as I find it more stable and consistent than anything else.  The problem I had was fixed by switching from the Alpine Docker image to the Ubuntu Docker image.

GordonBGood:
@rbergen:

With my new understanding (with your help) of what "faithful to base" means, **would it be fair to say that any implementation that uses at least 811,068 marking operations to sieve to a million odds-only would be `base`, no matter the storage representation?**.

Implementations that use Prime Factorial Wheels of course aren't `base` because that is their aim: to reduce the number of marking operations, just as algorithms that are difficult to categorize and called `other` also do.  The difference between `wheel` and `other` algorithms is that the former make it obvious that they are using small prime factorial repeating patterns to reduce the number of marking operations, whereas the later aren't so obvious.

In order to be fast, an implementation then needs to do two things within whatever category it falls, as follows:

1. It needs to reduce CPU clock cycles used per marking operation by using fast CPU cache memory for such marking, being careful not to extend marking operations into slower cache or main memory;  this implementation does that by page buffering by CPU L1 cache at a time and just copying the sieved buffer into the deliverable array.  If you can accept that technique of buffering as `base`, that's the first step.
2. It needs to reduce the number of CPU clock cycles per marking operation to no more than the time it takes to access memory, which is what this implementation does in the "Extreme Loop Unrolling" technique used in the implementation of the `BitSeq` type.  We have discussed this and you have indicated you can accept this if the details are just in the implementation of the storage class, so this should pass.

If you agree with these two points, it would seem to me that this contribution qualifies as `faithful to base`.

rbergen:
The guidelines and definitions have been drafted (and clarified on certain points over time) to capture the spirit of the original implementations, and conversations with Dave have taken place at times to verify certain aspects of it. The key thing here is that the mentioned spirit is respected while allowing for language variations _and_ reasonable creativity, which is what the reviews are supposed to assess and verify. If implementations are proposed that lead to questions about how certain things can be or need to be interpreted, we make a decision based on the case at hand, and translate that decision to a generalized clarification on the rules, if possible. 

I get the feeling that you're trying to extract a higher-level abstraction from the guidelines that aims to capture the purpose of the guidelines. That's not something we're going to do in the context of this project (or hence the review of this solution). The guidelines and definitions as they stand are the guiderails for the project, however arbitrary one may consider them to be. 

All of this means that, for one, I won't commit to "any implementation that uses at least such-and-such qualifies as a this-or-that implementation". Concerning the other points, I'd like to go back to the basics. 

As the guidelines state, for an algorithm to qualify as the base algorithm, the following must be true:

_The algorithm uses an outer loop, in which two operations are performed:_

1. _searching for the next prime in the sieve, and_
2. _clearing this prime's multiples in the sieve._
 
In a base algorithm implementation, the outer loop and both operations (commonly implemented as inner loops) need to be recognizable, and the first and second operations need to be executed in alternation. As in:

1. Find first prime/factor
2. Clear multiples of the first prime/factor in the sieve
3. Find second prime/factor
4. Clear multiples of the second prime/factor in the sieve
5. ...

On top of that, any indexing/memory referencing in the second operation needs to happen "on the spot", as explained in [this comment](https://github.com/PlummersSoftwareLLC/Primes/discussions/454#discussioncomment-1018644).

(As a side note, from a conceptual algorithmic perspective, the main point of the first operation/loop is to prevent "unnecessary" executions of the second operation/loop. The reason for this is that the second operation/loop is very expensive in most (traditional) languages.)

As things stand, I am not (yet) able to identify this basic base algorithm pattern in your implementation. Acknowledging that this may be due to my relative unfamiliarity with Nim, what may be needed is that you "talk me through" the relevant parts of the source code, to explain which parts of the source code correspond with the various parts of the basic algorithm described above.

davepl:
Hey, look at me jumping to the end of a huge thread that I didn't read the entirety of after being warned to stay off the PR anyway!

With those caveats out there, let me just say the INTENT of the rule is that there not be any encoding of prime facts other than the special case of 2, because it's special.  Any other information that is encoded in the program about the primality of other numbers, or even the likelihood of primality, would be a violation of the intent.  If the program can be reverse engineered to deduce any facts about primes, same thing. 

Being well hidden, essential to the algorithm in question, or any other fact isn't a variable here - it either statically encodes knowledge about which numbers >2 are prime or it doesn't.  It has to dynamically generate that information, removing entropy as it were one bit at a time.  If it comes with any of that "baked in", regardless of language, that's what I don't care for.  I have not looked closely at the Rust or Zim impls yet, so have no comment on them.

Other "rules" are harder.  I prefer a single bit used for storage, and use a half bit myself.  Some algorithms use 8 bits.  But what if 64 were even faster, would we allow it?  I'd say no, but there's nothing to prevent it.  So we have to rely on the intent, and I think the intent is clear:  we want to know how long it takes for the program to dynamically calculate the primes, not statically decode or look them up from source.

Subsequently to my beating of this dead horse I now return you to your tangentially related discussion.

GordonBGood:
@rbergen:

> The key thing here is that the mentioned spirit is respected while allowing for language variations _and_ reasonable creativity

I'm just trying to understand the "mentioned spirit" which I suppose harks back to Dave's original implementation in C++, except that you say that you will accept "reasonable creativity and have said you'll accept some differences from that implementation.  I did go back to that implementation but will explain further why this contribution seems to match Dave's code within the extensions that you have already said are acceptable as well as point out the only significant difference.

> I get the feeling that you're trying to extract a higher-level abstraction from the guidelines ...

I am trying to get something more concrete that what the rules state, since my first implementation seemed to qualify by the rules as I understood them at the time.

> back to the basics.
> 
> As the guidelines state, for an algorithm to qualify as the base algorithm, the following must be true:
> 
> _The algorithm uses an outer loop, in which two operations are performed:_
> 
> 1. _searching for the next prime in the sieve, and_
> 2. _clearing this prime's multiples in the sieve._
> 
> In a base algorithm implementation, the outer loop and both operations (commonly implemented as inner loops) need to be recognizable, and the first and second operations need to be executed in alternation. As in:
> 
> 1. Find first prime/factor
> 2. Clear multiples of the first prime/factor in the sieve
> 3. Find second prime/factor
> 4. Clear multiples of the second prime/factor in the sieve

All right, we can stop right there.  I understood from the statement following the above - "In order to allow for idiomatic code, it is permissible to perform these steps in either order, provided the results are correct." - that it would be permissible to find all of the base primes/factors and then do all of the clearing as two separate operations within an outer main loop.  If that is not allowed, then indeed my implementation is not `base`.

> On top of that, any indexing/memory referencing in the second operation needs to happen "on the spot", as explained in "What matters for sure is that all the calculations and indexing related to clearing the prime candidate flags (whether they are booleans, bits or another data type) are performed during the "second operation" within the timed algorithm, and not based on pre-calculated bitmaps or something similar."

Of course, if some of the 811,068 marking operations required for an odds-only SoE to a million weren't don't by base prime spanning by indices within the timeing, it would be using pre-culled wheel values and/or a prime factorial wheel, so would be a `wheel` or `other`implementation.  My program does not do that.

> (As a side note, from a conceptual algorithmic perspective, the main point of the first operation/loop is to prevent "unnecessary" executions of the second operation/loop. The reason for this is that the second operation/loop is very expensive in most (traditional) languages.)

Yes, for every sieve algorithm 99.9% of the time is spent in the marking operations loop. there are then two ways to make it faster, as follows:

1. Use a prime factor wheel to reduce the number of marking operations, which is expressly forbidden or
2. Use clever programming in order to reduce the number of CPU clock cycles per marking operation, which is what my contribution does, just as the Rust "striping" implementations does.  My program achieves this by marking within a cache-sized buffer and by reducing the marking operations to a single read/modify/write CPU instruction.

> As things stand, I am not (yet) able to identify this basic base algorithm pattern in your implementation. Acknowledging that this may be due to my relative unfamiliarity with Nim, what may be needed is that you "talk me through" the relevant parts of the source code, to explain which parts of the source code correspond with the various parts of the basic algorithm described above.

I'll try to show you the key bits as compared to Dave's C++ implementation referenced in "CONTRIBUTING.md" document, as follows:

1.  Since that implementation uses the "waste half the memory used" of creating an array with a bit for both even and odd numbers, I thought that was required, but you have assured me that it is not as per a previous post, so I now allocate only storage bits representing the odd numbers in line 154.
2. I then allocate a CPU L1 cache buffer in which to do the culling in line 157.  Only very naive SoE programs don't consider that fast operations must be done across the CPU L1 cache for the maximum in speed, else suffer the consequences of "cache thrashing".
3. In the following few lines from 158 to 161, I calculate a few constants to be used in subsequent operations.
4. In lines 163 to 178 I find and cull for the base primes and also count the number of base primes as per the usual "odds-only" algorithm starting at the index representing the square of the base primes and spanning by base prime values, but as this loop only finds the base primes, it stops at the square root of the entire range.  This is only used to count the base prime values.
5. In lines 173 to 179, I re-scan for all of the values of the base primes and put them in an array, also putting the initial start indexes representing their squares into another array to be used in the actual full culling operations.
6. In lines 182 to 194 I do the bulk of the culling in which almost all of the time is spent, as follows:
a. Line 182 loops over all of the required CPU L1 cache sizes required to make up the full deliverable sieved bit array (of the defined type `BitSeq` which is the storage interface.
b. In line 184 the `cullSize` limit is calculated, which is always one full CPU L1 cache size except for the last one may be smaller to just sieve to the limit; this is in the unit of bytes.  In this line, the cache sized buffer is zero'ed since it will be used repeatedly.
c. In line 185 the same as the above is done to calculate the bit index limit in units of bits.
d. Line 186 starts the loop for this particular cache-sized buffer over the base primes from the previously initialized base primes array.
e. Line 187 terminates the loop if no more base primes are needed **for this cull buffer** by testing if the start index would exceed the upper limit **of this buffer**.
f. In line 188 the start index for the entire cull sequence is extracted from the array previously initialized with all the initial culling start positions.
g. Line 189 does all the major marking work of the algorithm (other than that used for finding the base prime values, which is minimal) in calling the `BitSeq` bit setting function with the start index, the ending limit index, and the step (which is exactly the base prime value); this is a function rather than an operator because it returns the index of the next cull position past the limit index.
h. Line 190 takes that returned next marking index and stores it back into the same `startIndices` array after preparing it for the next culling cache-sized buffer.
i. After the entire cache-size buffer has been culled, the just-sieved cache-sized buffer is copied into place in the deliverable array of bits.

If you look at the above carefully, you will see that there is no "cheating" in using values calculated outside of the program and that just as many marking operations are used as in Dave's C++ program.  One very important optimization is the culling over cache-sized buffers in order to avoid "cache thrashing" so as to reduce minimum memory access time to one clock cycle.  However, due to the storage unit used being single bits, the program wouldn't be able to take advantage of this due to "bit-twiddling" overheads, so the program implements the `BitSeq` type as the memory access interface, for which there is a method/function that internally uses "Extreme Loop Unrolling" to reduce this to just one CPU instruction per mark.

Thus, the most important function/method of the `BitSeq` interface as the `setRange` function from lines 123 to 135, which emulates "slicing with a step" as some languages have built-in (Python, Julia, Fortran, etc.).  Given a start index, a limit index, and a step size, it sets the specified bits to represent `true`.  The actual implementation of this is inside the `unrollLoops` macro, but it does indeed do a marking operation for every specified bit, for the most part inside an unrolled loop.  The implementation only has a sufficient API as required for this program, but for a more general case could have `clearRange` and a full set of operators as well.

I assure you that this does implement the full odds-only SoE.  **The thing with which you might take exception is only the optimization of culling by CPU cache-sized buffers at a time.**  This is an optimization that any competent programmer using a low-level language such as C/C++/Rust/Nim, etc. should be considering, and I suspect that some compilers may be doing at least part of this optimization "under-the-covers" in order to get the speeds that they do.

I hope that this explanation allows you to follow the flow of my program.  Any further questions, just ask.

rbergen:
@davepl I'd say that you kicking off this whole project entitles you to step in wherever you want to. I think there is no confusion (any more) about prior knowledge about what is a prime and what is not, or about pre-calculated bitmaps being projected in one form or another.
What I think we're actually discussing now is what is the "base" algorithm of the sieve of Erastosthenes, in the context of this project.

GordonBGood:
> What I think we're actually discussing now is what is the "base" algorithm of the sieve of Erastosthenes, in the context of this project.

Specifically, does the `faithful base` algorithm include culling over cache-sized buffers at a time until the entire deliverable has been sieved?

Explanation of the implementation details of how "Extreme Loop Unrolling" works is [in my Nim blog post](https://nim-lang.org/blog/2021/07/28/Nim-Efficient-Expressive-Elegant-Benchmarking.html).

GordonBGood:
@davepl:

> With those caveats out there, let me just say the INTENT of the rule is that there not be any encoding of (base) prime factors other than the special case of 2, because it's special. Any other information (is not acceptable)...

That has been well established further up the thread and also (now) established that this implementation does entirely follow those restrictions.

> I prefer a single bit used for storage, and use a half bit myself. Some algorithms use 8 bits. But what if 64 were even faster, would we allow it?

Again, further up the thread it has been established that the "bittedness" of the number representation doesn't affect how true an implementation is to the `base` algorithm.  In fact, this implementation uses just one bit per odd number and doesn't use any bits for even numbers as they are automatically rejected just for being even (other than two); there are other accepted implementations that do the same.

**The question on which we need an opinion is this:  "Is it true to the `base` algorithm to cull the factors of the base prime values by pages of the CPU cache size and move the resulting sieved pages into place in the returned deliverable array?".**  For culling of odds-only, this requires exactly the same number of marking operations as all the other acceptable implementations, just it can be much faster because memory access can be as little as one CPU clock cycle per marking operations rather than the tens of CPU clock cycles per marking operations when this is not used for larger number representations.  For instance, if 64 bit's/eight bytes represent each number to a million and all representations were used (keeping the unused even representations), then the sieving buffer would be eight Megabytes in size which is well beyond the CPU L1 and L2 cache sizes and larger than the L3 cache sizes of some CPU's if they even have one.  Memory access for this implementation would likely average over ten CPU clock cycles per marking operations due to "cache thrashing".  If cache paging were used, marking operations could be in the order of two CPU clock cycles per marking operation for a speed up of at least five times, even for simple marking operations.

My implementation goes beyond a simple marking operation somewhat (although the method used has been established to still be true to the `base` algorithm) to reduce the culling operation time to average just about one CPU clock cycle per marking operation, but it then also depends on the memory access being fast in order to not bottleneck the operations.  So the opinion on whether using cache paging is allowed is crucial.

Also, the technique used is not limited to the computer programming language used in this implementation, Nim, but is applicable to any compiled language with a somewhat low level memory access; this may include but not be limited to only pointer access, but it also can work to improve execution time for languages without pointer memory access.

rbergen:
> All right, we can stop right there. I understood from the statement following the above - "In order to allow for idiomatic code, it is permissible to perform these steps in either order, provided the results are correct." - that it would be permissible to find all of the base primes/factors and then do all of the clearing as two separate operations within an outer main loop. If that is not allowed, then indeed my implementation is not `base`.

Then I think we have indeed established the precise algorithmic confusion in this case. It is allowed to switch operation 1 and 2, _within the context of the outer loop_. Effectively, the alternating execution of operation 1 and operation 2 I detailed in my previous comment can _start_ with either of them. 
Here, the semantics is in the word "either", which was not meant as "any". 

If your implementation indeed does identify all factors first, and then does all the clearing, then that would be an "other" algorithm by this project's definitions.

GordonBGood:
@rbergen 

> If your implementation indeed does identify all factors first, and then does all the clearing, then that would be an "other" algorithm by this project's definitions.

Okay, then the project's current definitions preclude culling by cache-sized pages, which is one of the prime tenets of writing fast sieving implementations for non-trivial ranges; even this specification of sieving to a trivial range of a million odds-only and even using bit packing to use only one bit per odd number means that there are about a half million bits or almost 64 Kilobytes used, which exceeds the 32 Kilobyte CPU L1 cache that most modern CPU's have.  This will mean there will be some "cache thrashing" with even the best bit-packing that can be done for the "faithful to base" algorithm, which will throttle the maximum speed by limiting the memory access speed for some memory accesses.

This throttling will get worse if maximum bit-packing is not used so that it will be terrible if, for instance, the computer language used can only use logical values of 32 or 64 bits each, but still quite bad for odd representations of eight bit bytes which uses about 500 Kilobytes.  Thus, the definition of the "faithful to base" algorithm puts a cap on the maximum score possible for a given number of bits per representation for a given CPU and computer language.  This seems unfair to some computer languages as if favors computer languages without that limitation.

To the case in point, since Nim can use maximum bit-packing and since most CPU's will have at least 32 Kilobytes of CPU L1 cache, my implementation only overruns the L1 cache by a factor of less than two, which limits the "cache thrashing" to only slow the program when not using marking by cache pages by about twelve percent.

In order to avoid the classification of `other` into which category are placed some implementations which are actually `wheeled` algorithms of a form different enough that you didn't recognize what they were, and thus can't be fairly compared to some `other` algorithms that may not use wheel factorization as my implementation does not, it looks like I should change the implementation to use what you currently find acceptable for a "faithful to base" implementation so that it can be fairly compared.

**I have changed the implementation accordingly (which actually simplifies it) to lose the about twelve percent extra performance, but still leaving it faster than every other `faithful base` submission to date.  The lines that do all of the finding of base prime values and culling of all multiples in the same outer loop are now at lines 151 to 154.**

rbergen:
When running the docker build on your solution locally, I get the following output:
```
------
 > [4/4] RUN nim c -gc:arc -d:release -d:danger -t:-march=native -d:lto primes.nim:
#9 0.373 command line(1, 2) Error: 'on' or 'off' expected, but 'arc' found
------
```

GordonBGood:
@rbergen:

> When running the docker build on your solution locally, I get the following output:
> 
> ```
> ------
>  > [4/4] RUN nim c -gc:arc -d:release -d:danger -t:-march=native -d:lto primes.nim:
> #9 0.373 command line(1, 2) Error: 'on' or 'off' expected, but 'arc' found
> ------
> ```

I don't know how that happened, as I thought I checked it before submitting it.  The problem is that `-gc:arc` should be `--gc:arc` (two dashes), else it gets interpreted as just an option to turn garbage collection on or off.

Also, the Nim designers tell me that `-d:release` isn't necessary when `-d:danger` is used, as the second already turns on the former, but it doesn't do any harm either.  `-d:danger` turns off automatic compiler checking, especially including array bounds checking that slows things up considerably.

Sorry about that.

rbergen:
> Sorry about that.

It's not a problem. It just means you'll need to update the Dockerfile in your PR before we can proceed with bringing this to a merge. :)

GordonBGood:
@rbergen:

> It's not a problem. It just means you'll need to update the Dockerfile in your PR before we can proceed with bringing this to a merge. :)

Done!

rbergen (approved):
LGTM

marghidanu (approved):

louie-github:
> Sorry, I miss-spoke, I only found one use of Julia @simd as follows: That's from Julia solution_3/primes_1of2.jl, uses `@simd at line 104; states it is `base`and`faithful` at line 181.
> 
> I am not too concerned with this Julia example that does "cheat" a little bit in the use of SIMD as it usually only gives it a little advantage that is dependent on the CPU on which it is run; but just used it to show that there are broader definitions of what `base` actually means in other implementations.

For what it's worth, I'd like to clear things up. I didn't read the entire conversation, so correct me if I missed some context.

`@simd` does not actually generate SIMD instructions because of the way the array is accessed (large steps of `factor`). You can see this when we disassemble the function:
*(I've shortened the output for the sake of brevity)*:
```
julia> include("primes_1of2.jl")

julia> code_native(clear_factors!, (Vector{UInt}, UInt, UInt), syntax=:intel)
...
; [OWN COMMENT]: Main loop
L144:
        mov     rsi, rbx
        shr     rsi, 6
; │││└└└
; │││┌ @ primes_1of2.jl:45 within `_get_bit_index_mask'
; ││││┌ @ int.jl:457 within `<<'
        shlx    rdi, rdx, rbx
; │││└└
; │││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
        or      qword ptr [rcx + 8*rsi - 8], rdi
; │└└└
; │┌ @ simdloop.jl:75 within `macro expansion'
; ││┌ @ int.jl:441 within `<'
        add     rbx, r15
        dec     rax
; ││└
        jne     L144
; │└
```

As you can see, there are no vectorized or SIMD instructions to be found, just regular old arithmetic and bitwise operations. This is because the `@simd` macro by itself only transforms the loop into something the LLVM vectorizer is *more likely to*, but is not *guaranteed to* vectorize.

I only used `@simd` since it seemed to give better results on modern x86 machines, but it *is* a quite fragile optimization that can vary from machine to machine. That's why I mentioned in the [README](https://github.com/PlummersSoftwareLLC/Primes/tree/drag-race/PrimeJulia/solution_3) that you can remove `@simd` if you want to. There's an [entire discussion](https://discourse.julialang.org/t/why-is-this-simd-loop-faster-than-a-while-loop-even-if-it-has-longer-assembly/65585) about this tiny change on the Julia Discourse.

Sorry for butting in, but I just wanted to clear this up in case there are any misconceptions! Thanks for your time!

louie-github:
On another note, this seems like an extremely interesting implementation! I might have to see if I can do something similar in Julia :>

GordonBGood:
@louie-github:

> On another note, this seems like an extremely interesting implementation! I might have to see if I can do something similar in Julia :>

You're welcome to try, of course.  I have tried to do it in Julia and ran into some problems: 
 Julia doesn't seem to have a "computed goto" `case`/`match` statement as does Nim/C/C++, which isn't a problem in itself as I've worked around that in other languages by calling from an indexed array of closures, but that didn't work so well in Julia because function signatures can't be typed (or at least than I'm aware) and the run time spent too much time doing meta stuff - it helped speed things up, but not as much as it does in Nim.

OTOH, using the "striping"/"loop unpeeling" technique does work quite well to speeds things up quite a bit as Mike does in Rust solution 1.

louie-github:
I'm going off the top of my head, but I believe their recommendation was to
just create an if-elseif statement as LLVM should optimize that into a
computed goto.

GordonBGood:
@louie-github:

You could try it; didn't seem to work for me, but Julia isn't one of my primary computer languages.

GordonBGood:
@louie-github:

> You could try it (trying to let Julia automatically generate a computed goto from a chain of `if/elseif/end`'s); didn't seem to work for me, but Julia isn't one of my primary computer languages.

I did try it, and couldn't get the desired computed goto to be generated as desired, so while using my "Extreme Loop Unrolling" is slightly faster than "unpeeling", it's not that much faster to be worth the effort.

So I am submitting my "unpeeling by cache-sized blocks" in Julia, which is about 50% faster than your best effort, which effort of your's isn't that bad.

The trouble with Julia is that it depends too much on the excellent LLVM back end, but that can't be guaranteed to do the optimization we desire and currently there isn't any real way to force it.  Also, as mentioned previously, the first-class-function/closure type signature story isn't very good when one wants to program in a mostly pure functional paradigm.  Other than that, it is a very performance-oriented language as per the buzz in the press.

Syntax-wise, I'm not in love with it, but that's another story.

louie-github:
@GordonBGood, that sounds great! I'm already curious to see how you approached this in Julia, especially with the limitations you mentioned, and I hope to learn a lot from it! Your Nim solution is extremely fast and I can't deny that I'm mesmerized by it, but also a little intimidated at the code.

I think that I'll probably pursue a striped variation of my solution and also an "extreme loop unrolled" version similar to your Nim solution. I'll see if I can get around the limitations you mentioned.

By the way, what does "unpeeling" mean, and could you give an example? I apologize if this sounds a little stupid, but I can't seem to understand where "unpeeling" or "loop peeling"(?) would apply to the base faithful algorithm.

Thanks!

GordonBGood:
@louie-github 

> I think that I'll probably pursue a striped variation of my solution and also an "extreme loop unrolled" version similar to your Nim solution. I'll see if I can get around the limitations you mentioned.

Good luck; as mentioned, I got the "Extreme Loop Unrolling" working (and one could make it quite short as for Nim by using a macro), but it wasn't as fast as I wanted due to not being able to emulate or get a computed goto and possibly for some other optimization limitations.

> By the way, what does "unpeeling" mean, and could you give an example? I apologize if this sounds a little stupid, but I can't seem to understand where "unpeeling" or "loop peeling"(?) would apply to the base faithful algorithm.

In order to be `faithful to base algorithm`, one must have the following:
1. There has to be a sieve buffer created and cleared every pass with at least one bit/word/pair of bits/etc. per odd number being sieved.
2. The algorithm needs to have an outer loop searching the sieved buffer for the current state of the sieve buffer for base prime values in sequence to be used in determining the multiples to be marked as composite numbers.
3. The algorithm then needs to have an inner loop or loops the does the actual composite number marking for the given base prime value determined in the outer loop.  Notice that this step is not limited in the order that the marking needs be done or the number of loops used to do the marking, but the end result is that all composite multiples for the given base prime value must be marked for the entire sieve buffer.
4. The final output of the sieve after all base prime values have been determined and all composites marked in the above steps can be any of the sieved buffer, a buffer containing just the determined primes, or a list/iterator over the found primes in the range.  For speed, usually either the first or last will be provided, as a (lazy) list or iterator is generated from the sieved buffer anyway.
5. In order to be marked as `faithful`, it needs to be `base` as above plus it can't use any external libraries other than the language's standard library.  I had considered using the Julia "PointerArithmetic" library, but first, that would be an external library and the solution might not be considered to be faithful, and second, it wasn't any faster using it anyway - that library's main purpose is for interfacing to C/C++ code through a foreign function interface.  Although it wasn't any slower either, which is good.

**So first, what is striped culling?**  This is something that only Mike Berber in his Rust solution_1 uses and is considering the entire bit sieving buffer (allocated in bytes) out of order, so that for his "block" version he considers the bit stream starting at bit zero for all bytes in a block followed by bit one for all bytes in the block, followed by all these bit "stripes" in bit order up to bit seven.  It still uses eight loops per block (one for each "stripe"), but it starts from the first bit to be marked in the block and marks as per usual across each "stripe" where at the end of each "stripe" the byte index of the bit to be marked next is knocked back by the number of bytes in the block.

**How does this compare with what I call "unpeeling"?**  In "unpeeling", I don't change the normal order of sequential bits in the sieve buffer at all, with the successive odd numbers starting in the first byte from bit zero to bit seven, followed by the second byte across bits zero to seven and so on to the end of the buffer.  This has the advantage that addressing a given number representation is easier in using common "bit-twiddling" shifting and masking, but it requires that (for block sieving) an eight entry array must be used to save the next culling/marking position for each of the eight given bit indices.  Then, the culling marking proceeds for each bit index position in order just as for "striping", but as the number representation isn't the same, there is no "knocking back" for the next "stripe" but each "stripe"'s starting culling position obtained from the state saving array.  The work and the speed should be approximately the same other than for individual language exact optimizations.

In the Nim "Extreme Loop Unrolling" implementation, I combine all the eight "unpeeled" loops into one common "unrolled" loop with the eight marking operations inside that loop advancing as for each of the individual "unpeeled" loops.  Since this uses immediate bit masks for each of the marking operations, at the end of any given loop the next bit to be marked can be exactly calculated so that no state saving array is required and the last few (up to seven, average 3.5) marking operations per page can be done by a common loop.  In fact, within the limitations of the `base` algorithm, there are only four different patterns used, as there are only four possible prime values modulo eight (all least significant bits are one for odd base primes) and as all of them start at the beginning of the sieve buffer and proceed to the end, there are thus only four modulo eight starting positions with one for each base prime value modulo eight.  So within the limitations of this "drag race" we don't really need a computed goto, but could just select one of the four patterns with a combination of just two binary branches.  I have done this, and where the "unpeeled by blocks" solution runs about fifteen thousand passes in five seconds, it runs in about seventeen thousand in five seconds (no blocking necessary or even possible within the limitations of being `base`).  However, it likely needs "PointerArithmetic" library operations to work and since that is an imported library, I am afraid it may give up on the `faithful to base` designation and didn't submit it.  If acceptable, it is then faster than the Rust "striped blocks" solution but still a little slower than the Nim "Extreme Loop Unrolling" version, for a combination of reasons of not optimizing the unrolled loops quite as much and due to the overheads of this dual branch to find the right pattern per base prime.

Since "loop unpeeling" is somewhat similar to "Extreme Loop Unrolling" and also somewhat similar to "striping" with the first two doing culling/marking out of order and the latter changing the defined order of bits, then if any one of these passes the `base` algorithm test, then so must they all.  Notice that the contributing specification says nothing about the order of culling/marking, just that all culling/marking for a found base prime value must occur within the secondary phase for the entire sieve buffer for that given base prime value.

Since "unpeeling" and "striping" are so similar, I think you'll find that (when both are use blocks or not) they are about the same speed, and which to use is just a matter of which one finds easier to implement.  "Extreme Loop Unrolling" isn't that hard (once one gets their head around the technique), just it may or may not pay off if one can't use pointer operations due to the way the loops may be optimized;  for the simple cases as required here, it doesn't need a macro - I have done it in various languages just using simple templates or in the worst case where languages don't even have templates, by a code generator.

I hope that helps.

PR #483, opened 2021-07-24
Marking Zig/solution_2 as unfaithful due to compile-time size fix

rbergen:
As pointed out by [a comment](https://github.com/PlummersSoftwareLLC/Primes/discussions/445#discussioncomment-1033651) in discussion #445, the use of a combination of const and comptime in Zig/solution_2 boils down to the sieve (memory) size being set at compile time.
This goes against the rule that _The sieve size and corresponding prime candidate memory buffer (or language equivalent) are set/allocated dynamically at runtime._

As @ManDeJan has indicated that [they can change Zig/solution_2 to make it faithful](https://github.com/PlummersSoftwareLLC/Primes/discussions/445#discussioncomment-1042101), I'll hold on merging this PR for a few days to allow them to do so.

### DIFF
--- a/PrimeZig/solution_2/README.md
+++ b/PrimeZig/solution_2/README.md
@@ -1,3 +1,6 @@
 # Zig solution by ManDeJan
 
-![Category](https://img.shields.io/badge/Category-faithful-green)
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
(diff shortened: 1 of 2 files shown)


rbergen:
Merging this to prevent (further) confusion about benchmark results. Also, the change can be easily reverted once the underlying reason has been addressed.

PR #484, opened 2021-07-24
TeX solution

jfbu:
Benchmarking uses pdftex for its timing facilities

## Contributing requirements


rbergen:
Merging this PR would add TeX as a separate language, next to the LaTeX solution that is already in place. Acknowledging that TeX and LaTeX are not exactly the same thing, they are closely related. I would therefore like to establish if adding both as separate languages is indeed the most logical thing to do. An alternative could be "reclassifying" LaTeX/solution_1 as TeX/solution_1, and the solution in this PR as TeX/solution_2.

@jfbu and @tjol could you please provide me your perspectives on this?

jfbu:
@rbergen I don't have a strong opinion about whether to fuse or keep separate. But if fusing, then as I will explain I have a strong opition that it should be PrimeTeX and I am very willing to rename my pure TeX solution into  `solution_2/` while @tjol's one would be `PrimeTeX/solution_1`.

The solution by @tjol uses LaTeX3 which aims as providing something akin to a programming language, so there is some point into having a LaTeX3 language on the repo. It is however entirely a macro layer on top of TeX. Currently the LaTeX3 has not yet (they plan to do it in future I think) made the TeX primitives completely removed from user access via their original names (such as `\box`, `\rule`, `\vskip` etc...). 

If TeX and LaTeX are to be fused here, then I do have a strong opinion that the language tag should be TeX because LaTeX is entirely built and reduces to TeX (and to the e-TeX extensions from 1999, incorporated in pdftex et al. in TeX installations since at least something like 2004 or 2005), not the opposite.

LaTeX as used by people is a macro layer and there are other ones with wide usage such as ConTeXt, or the more recent OpTeX. So definitely the language to put under its umbrella all potential contributions should be TeX, not LaTeX.

tjol:
Reclassifying the LaTeX solution as TeX solution 1 makes the most sense to me. My solution happens to not use much plain TeX directly, but a different LaTeX solution might. As the distinction isn't all that clear (and they ultimately run on the same e-TeX engine in any case) I think it makes more sense to keep them together.

Of course I absolutely agree that if we're classifying the solutions as using the same language, the name of that language is TeX (or Τεχ? No. TeX.)

jfbu:
I force-pushed a commit because my latest one accidentally used `%` not `#` for comments in Dockerfile sorry.

agreed @tjol, I am ready to move this into `PrimeTeX/solution_2`

rbergen:
@jfbu and @tjol thanks for responding so quickly. Based on your input, I think the best thing to do is to move both solutions under PrimeTeX.
Would both of you be okay with this PR including the rename of PrimeLaTeX/solution_1 to PrimeTeX/solution_1? It'll save all of us a separate PR/review cycle.

jfbu:
> Would both of you be okay with this PR including the rename of PrimeLaTeX/solution_1 to PrimeTeX/solution_1? It'll save all of us a separate PR/review cycle.

@rbergen Sure fine with me. I will start with renaming my solution into solution_2. Then perhaps @tjol could make a PR on my fork with  renaming s/PrimeLaTeX/PrimeTeX/ with target my branch PrimeTeX? (I think this makes sure commit authors stay then respective solution authors.)

rbergen:
> (I think this makes sure commit authors stay then respective solution authors.)

@jfbu I just want to note that, although I understand where you're coming from with this, this is not a requirement in the context of this repo. In fact, we've had some really good commits from other people on original author's solutions, and we have the README.md files and implementation output documenting who authored what.
So, it's fine to do it the way you propose, as far as I'm concerned there's nothing against taking a "shortcut", here.

tjol:
> @jfbu and @tjol thanks for responding so quickly. Based on your input, I think the best thing to do is to move both solutions under PrimeTeX.
> Would both of you be okay with this PR including the rename of PrimeLaTeX/solution_1 to PrimeTeX/solution_1? It'll save all of us a separate PR/review cycle.

Sounds good to me. @jfbu, would you mind just taking care of all the renaming directly?

jfbu:
> Sounds good to me. @jfbu, would you mind just taking care of all the renaming directly?

@tjol yes, sure. Thanks will do it shortly.

jfbu:
@rbergen @tjol renamings done :) (I am not native speaker so perhaps it is "renames"?)

PrimeTeX/solution_2/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM phipsgabler/texlive-minimal:latest
```
rbergen: Please pin your base image to a specific version, instead of `latest`.

PrimeTeX/solution_2/README.md
```diff
@@ -0,0 +1,140 @@
+# TeX solution by jfbu
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
```
rbergen: You've marked your solution as faithful. I can't quite establish how you've met the requirement that _It uses a class to encapsulate the sieve, or a (closest) equivalent feature in your language if a class construct is not available. This class must contain the full state of the sieve. Each iteration should re-create a new instance of this class from scratch._
Could you please explain how this solution implements this?

jfbu: TeX is not object oriented but my solution tries to imitate instantiating a class via dedicated interface.
- `\NewSieve{<user chosen name>}{<given range>}` creates the object in memory.
- Every invocation of `\NewSieve{<some name>}{<some range>}` creates a new object in TeX memory.
- This applies even if the new object has an already used name. Access to previous instance with same name is lost (each instance has a unique id, which can be saved, if this is done then one can still access the "dangling" object).
- Once you have an object say `foo` you can execute some methods: `\Sieve foo.sieve{}` is the core one and runs the sieve. `\Sieve foo.settocount{\macro}` will define `\macro` to hold the prime count (the "sieve" method must have been executed). There is also a "reset" and a "outtofile" method.
- The object has a "range" attribute `\Sieve foo.range(}` which expands to the range parameter. It also has the "instance" attribute which is its unique id. First created object has id `666`, each successive one steps the id by one.

However the big problem is that although TeX does have a scope limiting mechanism (`\begingroup... \endgroup`, or `{...}` ; the former underlies LaTeX's "environments"), here it can not work, because the "font dimension parameter" are purely global object. So you can't for example set-up a macro which will create a Sieving object, use it, and then on exiting the macro it will be purged.

And to be completely honest my poorman code for "objects of class Sieve" should be improved to add the `\global` prefix to all definitions (as stated above, this is automatic for font dimensions which are used for data storage). Things will be pretty much broken if one starts using my poorman interface from inside some scope limiting TeX context (as listed above).

After writing this up I feel I should toggle faithful to unfaithful.

This is what I will do now, possibly reverting if you advice me too.

PrimeTeX/solution_2/Dockerfile
jfbu: Done

PrimeTeX/solution_2/README.md
jfbu: I have now set the badge to `unfaithful` and modified the output to say `faithful=no`

rbergen:
@jfbu I'm curious, what is the "erato" filename prefix a reference to?

jfbu:
@rbergen nothing extraterrestrial :) Erato...Eratosthenes!

I pushed some improvements:
- fix a misleading comment in Dockerfile about some TeXLIve utility
- provide a recipe in README.md about how to produce a `pdftex.fmt` to configure `pdftex` with more font memory, without altering your existing TeXLive configuration files
- add to the benchmark a validation step to check that the correct number of primes was found

jfbu:
Forgot to say that I experimented with another TeX approach using "control sequence names" rather than "font parameters" to mark (non-) primes, but it came out about 40x slower... so not making this a solution_3 (and it also requires significant TeX memory enlargements)

rbergen:
> Erato...Eratosthenes!

I have to admit that that is not a very surprising answer. The reason I asked is that Erato is also a goddess in Greek mythology, with a somewhat different meaning. :)

Could you let me know when I can consider this PR stable and ready for review?

jfbu:
Ah... thanks for telling me, I was not aware of it but it looks fitting also in this sense (at least to me; wasn't Ramanujan inspired by a female divinity?).

About readiness of the PR, I feel it is, but I skipped lunch, so I will have lunch now (CEST) then make a last review of the code base in about one hour from now.

jfbu:
@rbergen My PR is ready for review. I discovered at last minute that some step I was doing to increase the (static) memory of `pdftex` could be simplified.

jfbu:
sorry for again a push; I had copied pasted a comment from Dockerfile to texmf.cnf file but the comment character should be `%` there. At least offically but I had seen no error on testing at my locale so did not detect this. Perhaps the `#` is also comment but not according to [official doc](https://www.tug.org/texlive/doc/texlive-en/texlive-en.html#x1-600007.1.7)

jfbu:
@rbergen sorry that I had to push again... this is all peripheral to core which has not changed since eons... should I squash all those commits?

rbergen:
No, but I am going to wait a day for this to actually settle down :)

jfbu:
@rbergen ok I hope I will actually stop twiddling details... in the mean time I'll try to find competent people tell me if I could have left the `#` in the `texmf.cnf` file after all

jfbu:
@rbergen Now PR really is ready for review... my previous post-"I am ready" commit https://github.com/PlummersSoftwareLLC/Primes/pull/484/commits/737b87bdaa5dd0fbfad8e053c25d69651914ad8c was not needed: TeXLive maintainers have told me since that one can in fact use `#` as comment character (and that this is currently a non-documented feature). But it was good as since then I better understood the TeXLive limitations on pdftex font memory and could document better in README.md the maximal sieving range. So I won't modify anything from now on.

jfbu:
@rbergen I am considering implementing the idea present in many solutions for other languages to sieve only through numbers congruent modulo 30 to the 8 classes which are prime to 2, 3, 5. Am I correct that this would constitute a PrimeTeX/solution_3 ? It would not be improvement of implementation details in the TeX language, but a more sophisticated algorithm. It would necessitate changes to all explanations related to the pdftex memory constraints (which took me time to build up, amend, and polish).

rbergen:
> @rbergen I am considering implementing the idea present in many solutions for other languages to sieve only through numbers congruent modulo 30 to the 8 classes which are prime to 2, 3, 5. Am I correct that this would constitute a PrimeTeX/solution_3

It doesn't have to be a new solution. You can add this as a second implementation to this same solution, which I would actually prefer (it saves the docker build of a separate image). The Dockerfile would then have to build and sequentially run both implementations. PrimeBASIC/solution_1 is an example of a solution that takes this approach.

Do note that the algorithm you refer to is a "wheel" algorithm, and would have to be marked as such both in the implementation's output and the solution's README.md.

rbergen:
I'm again going to hold on reviewing this PR until you declare it stable. And please, then indeed don't change it once you have.

jfbu:
@rbergen My PR is ready for review. I have added a 480 out of 2310 "wheel" approach. I needed to modify the benchmark to use `luatex` because there is a risk that `pdftex` on a fast computer will do `295` passes or more which will exhaust its static font memory, even configured to the max (each pass requiring `500,000` words).

I am not going to change anything from now on as I have passed many hours on matters of documentation, particularly in relation to `pdftex` memory limitations, and in getting a small Docker container to work (and I had to code directly via `\directlua` the needed interface for `luatex` to do timings from within the tex run, as it proved finally too complicated to try to get the needed files from the TL2018 into the Docker image). As documentation has a tendency to be much more fragile than code itself with respect to changes, and I feel the ratio code:documentation is 1:5 regarding time this it for me.

For reasons I do not know, the wheel compared to the base "odd-only" achieves almost 3X gain with `pdftex` but more like `2.3X` with `luatex`.

PrimeTeX/solution_2/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Output
+
+Hardware: 2 GHz Intel Core i7 with 8 Go 1600 MHz DDR3 of memory.
+
+Docker run:
+
+```
+jfbu-tex;10;5.38837s;1;algorithm=base,faithful=no,bits=unknown
+jfbu-tex-480of2310;23;5.04814s;1;algorithm=wheel,faithful=no,bits=unknown
```
rbergen: I'm making this comment once, but it applies to all places where you added a `bits=unknown` tag: please remove the `bits` tag from your output entirely. The number of bits per primes being unknown is indicated by the _absence_ of the tag.

jfbu: ok, sorry for oversight. I have done it locally, waiting for go ahead to push the commit.

I thought this could possibly fix this
```
make DIRECTORY=PrimeTeX/solution_2 FORMATTER=csv
....
....
added 229 packages in 7.472s
info: [PrimeTeX][solution_2] Building...
info: [PrimeTeX][solution_2] Running...
error: No data was collected!
```
which I can't explain as running the Dockerfile succeeds. But no change.

rbergen: I missed this too during my review of just now, but this is caused by the letter "s" following the duration in the output. The number of seconds must be included in the output in a purely numeric form, otherwise the benchmark tool won't recognize it.

jfbu: my bad! It works now. I am pushing the two fixes.
```
info: [PrimeTeX][solution_2] Building...
info: [PrimeTeX][solution_2] Running...
Index,Implementation,Solution,Label,Passes,Duration,Threads,Algorithm,Faithful,Bits,PassesPerSecond
1,PrimeTeX,2,jfbu-tex-480of2310,11,5.28532,1,wheel,no,,2.08124
2,PrimeTeX,2,jfbu-tex,4,5.07770,1,base,no,,0.78776
```

jfbu: however it runs twice slower than directly `docker run`...

jfbu: @rbergen Also with PrimeBasic/solution_2 I get a twice faster direct `docker run` compared to via the Makefile:

with Makefile:
```
info: [PrimeBasic][solution_2] Building...
info: [PrimeBasic][solution_2] Running...
Index,Implementation,Solution,Label,Passes,Duration,Threads,Algorithm,Faithful,Bits,PassesPerSecond
1,PrimeBasic,2,rbergen_vb,875,5.00182,1,base,yes,1,174.93626
```

with Docker direct:
```
$ docker run --rm basicprime:latest
rbergen_vb;1838;5.0001592;1;algorithm=base,faithful=yes,bits=1
```

jfbu: It seems to have been a transient phenomenon for which I don't have real explanation (perhaps system was doing some heavy duty at same time). I still observe some `20%` difference consistently with both Basic/solution_2 and TeX/solution_2.

marghidanu: I will double check this on several systems to make sure the numbers are correct. The difference in speed is quite significant.

marghidanu: I've run the basic solution several times and the difference initially is a bit higher and I'm assuming that has something to do with caching. After several runs, it starts to become more stable and the differences are not that big.

@rbergen I think I mentioned this is in the past, running each solution several times should yield more stable results, obviously, it will take longer but the results would be much more stable.

jfbu: I was wondering whether this could have to do with an almost full disk image. Indeed by default on my mac os, `docker-machine` creates a 20Go image, which is not enough to run the `make`. I had to clone the disk image from vmdk to vmdi then increase size then get the partition to feel the space to 40000Mo and it was barely enough in the end to run the `make`: I have now 40412119040 bytes occupied (37.64Gio out of 39.06Gio according to VirtualBox) Some builds failed for reasons I didn't look at really. For CPP/solution_1 and CPP/solution_2 perhaps it has to do with `FROM ubuntu:21.04 AS build` in Dockerfile because they are the only ones with current PrimeLaTeX (TeX/solution_1 in this PR) doing `FROM ubuntu:21.04` and both failed building. I deactivated the Dockerfile for LaTeX fearing also LaTeX install might take one more Giga. Anyway, I could at last run the `make` almost fully at my locale... great!

edit: I forgot to say that Docker Desktop for Mac does not support my os (high sierra) so I use docker and docker-machine from MacPorts.

rbergen: > @rbergen I think I mentioned this is in the past, running each solution several times should yield more stable results, obviously, it will take longer but the results would be much more stable.

@marghidanu I think you know you've mentioned this in the past, because I do too. I still agree that this is a good idea. We just need to find someone to implement it. :)

marghidanu: I got this :), we can run the solution 5 times, and pick up the output with the best results.

rbergen: I don't mean to start reviewing before the code has been written, but the thought crossed my mind that we may want the benchmark tool user to indicate if they want to execute a 1-time or 5-time run. Certainly if they specify a DIRECTORY, because I think people do use that to test if their solutions work in the benchmark context.

EDIT: I'll leave this comment here because I already made it, but we can discuss further implementation details via other channels :)

jfbu: @rbergen Anything I should do on this PR? At some point I had squashed everything but perhaps I should again separate the commit renaming `PrimeLaTeX/solution_1` into `PrimeTeX/solution_1`? Currently the github interface which gives next to file the last commit modifying it makes it look too much as if I authored  `solution_1`, I fear.

rbergen: The squashed commit are fine. The changes to solution_1 will be marked by the commit message I add when I squash and merge this PR.

rbergen (approved):
LGTM

marghidanu (approved):

PR #486, opened 2021-07-24
add solution in red programming language

mmcdon20:
## Description

Added a faithful 1-bit implementation in the [red programming language](https://www.red-lang.org/). 

This is the first red program I have written, so there is likely room for improvement.

There is no class definitions in red, but I have come pretty close to a class-like structure by returning an `object!` type from a function which takes the constructor arguments. The `object!` type is a bit like javascript objects, where you can define arbitrary key/value pairs which are a bit like fields and methods.

## Contributing requirements


mmcdon20:
Can consider this stable. I don't intend to add/change anything else unless something comes up in review.

rbergen (approved):
LGTM

marghidanu (approved):

PR #493, opened 2021-07-25
Update bitcount badge

joshallen64:
Just a simple update for the storage badge

### DIFF
--- a/PrimeRust/solution_4/README.md
+++ b/PrimeRust/solution_4/README.md
@@ -3,7 +3,7 @@
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
 ![Parallelism](https://img.shields.io/badge/Parallel-no-green)
-![Bit count](https://img.shields.io/badge/Bits-unknown-yellowgreen)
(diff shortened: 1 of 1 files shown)


PrimeRust/solution_4/README.md
```diff
@@ -3,7 +3,7 @@
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
 ![Parallelism](https://img.shields.io/badge/Parallel-no-green)
-![Bit count](https://img.shields.io/badge/Bits-unknown-yellowgreen)
+![Bit count](https://img.shields.io/badge/Bits-1-green)
```
rbergen, line 6: What is this change in bitcount badge based on? I don't see any accompanying changes in the source code.

joshallen64: Since I used Vec<bool> I had naively assumed that rust did bit packing similar to vector<bool> in CPP and therefore single bit flag storage.  However (after going down a educational rabbit hole), I've learned that this is not the case!  Vec<bool> does not do bit packing, and each boolean is instead stored as a byte.  I'll close the pull request.

PR #494, opened 2021-07-25
Fixing undefined behaviour

Michoumichmich:
The CPP version are using a rolling mask which is rotated while settings numbers as not primes.

This mask is sometimes shifted by a value greater than 32. It mostly seems to works, but with some platforms and compilers or optimization settings, it gives wrong results. The right value which is modulo is unused in the code. This PR makes use of that value.

### DIFF
--- a/PrimeCPP/solution_2/run.sh
+++ b/PrimeCPP/solution_2/run.sh
@@ -2,5 +2,5 @@
 # gcc -Ofast -std=c++17 PrimeCPP.cpp -lc++ -oPrimes_gcc.exe
 # clang -Ofast -std=c++17 -lc++ PrimeCPP.cpp -oPrimes_clang.exe
 
-g++ -pthread -Ofast -std=c++17 PrimeCPP_PAR.cpp -oprimes_par.exe
+clang++ -pthread -Ofast -std=c++17 PrimeCPP_PAR.cpp -oprimes_par.exe
 ./primes_par.exe
--- a/PrimeCPP/solution_2/run.sh
+++ b/PrimeCPP/solution_2/run.sh
@@ -2,5 +2,5 @@
 # gcc -Ofast -std=c++17 PrimeCPP.cpp -lc++ -oPrimes_gcc.exe
 # clang -Ofast -std=c++17 -lc++ PrimeCPP.cpp -oPrimes_clang.exe
 
-clang++ -pthread -Ofast -std=c++17 PrimeCPP_PAR.cpp -oprimes_par.exe
+clang++ -march=native -mtune=native -pthread -Ofast -std=c++17 PrimeCPP_PAR.cpp -oprimes_par.exe
 ./primes_par.exe
--- a/PrimeCPP/solution_1/Dockerfile
+++ b/PrimeCPP/solution_1/Dockerfile
@@ -1,11 +1,11 @@
 FROM ubuntu:21.04 AS build
 
 RUN apt-get update -qq \
-    && apt-get install -y g++
+    && apt-get install -y clang
 
 COPY *.cpp /opt/app/
 WORKDIR /opt/app
-RUN g++ -Ofast -std=c++17 PrimeCPP.cpp -o PrimeCPP
+RUN clang++ -march=native -mtune=native -Ofast -std=c++17 PrimeCPP.cpp -o PrimeCPP
 
 FROM ubuntu:21.04
 COPY --from=build /opt/app/PrimeCPP /usr/local/bin
--- a/PrimeCPP/solution_1/PrimeCPP.cpp
+++ b/PrimeCPP/solution_1/PrimeCPP.cpp
@@ -56,7 +56,7 @@ public:
         while (n < arrSize) {
             array[index(n)] &= rolling_mask;
             n += skip;
-            rolling_mask = rol(rolling_mask, skip);
+            rolling_mask = rol(rolling_mask, roll_bits);
         }
     }
 };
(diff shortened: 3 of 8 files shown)


rbergen:
I believe this is a proposed change to code added by @Destroyerrrocket in #290 and #295. I'm tagging them because I'm wondering if they would like to comment on this?

Destroyerrrocket:
Woops, that's indeed a bug. If you don't mind, I'll test this once I'm back from vacations, in a few hours.

Michoumichmich:
Cool, thanks!  It's interesting to note though that it's with `ptxas` and O2 where the "incorrect undefined behaviour" appears, else it works fine on every other platform

Destroyerrrocket:
Ok, this fix by itself seems to take an important hit on my machine, but, some good news: clang seems to now be better again.
Some numbers:
Bugged original: [gcc]
Passes: 13704, Time: 5.000268, Avg: 0.000365, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

davepl_pol;13704;5.000268;1;algorithm=base,faithful=yes,bits=1

With Fix: 
[gcc]
Passes: 12446, Time: 5.000120, Avg: 0.000402, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

davepl_pol;12446;5.000120;1;algorithm=base,faithful=yes,bits=1
[clang]
Passes: 12440, Time: 5.000085, Avg: 0.000402, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

davepl_pol;12440;5.000085;1;algorithm=base,faithful=yes,bits=1

Now, if we added the flags -march and -mtune to native, we can mitigate things by a lot: 
[gcc]
Passes: 12844, Time: 5.000120, Avg: 0.000389, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

davepl_pol;12844;5.000120;1;algorithm=base,faithful=yes,bits=1

[clang]
Passes: 13705, Time: 5.000316, Avg: 0.000365, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

davepl_pol;13705;5.000316;1;algorithm=base,faithful=yes,bits=1

As we can see, clang is now pretty much as capable as the bugged solution, so if we were to change the compilation command to "clang++ -march=native -mtune=native -Ofast -std=c++17 PrimeCPP.cpp -o PrimeCPP", I would not mind this change at all.

Michoumichmich:
@Destroyerrrocket can you try the new version I updated ? Turns out we don't need to carry the initial mask around. Carrying it created a data dependency which serialises the loop. With this little change, I was able to get a great speedup.  Without `-march=native -mtune=native`, I'm going from 3889 to 5700

Michoumichmich:
As we can only have 32 different rolling masks, can't we go further and create a constexpr array containing these 32 masks ? Removing thus any runtime call to rot. 

EDIT: Tried it, it went to 7515 (from 3889)

Destroyerrrocket:
The constexpr idea looks promising! Given that this would not execute the algorithm at compile time, it should be faithful. I'd love to see the implementation! With your latest one, I unfortunately experience a decrease from 13705 to 12506 with march=native (this time gcc wins), but I'd love to see if your suggestion is able to flip this!

Michoumichmich:
@Destroyerrrocket I opened another PR (sorry for PR spamming) with the constexpr thing! Of course no, the idea is not to execute it at compile time ahah. I'm also experiencing a decrease with march/mtune=native, but I still had a speedup over the previous version

rbergen:
@Michoumichmich and @Destroyerrrocket I'm sorry, but I'm really not happy with having two PRs open on one solution at the same time, under authorship of the same contributors. Trust me when I say that managing this project really is challenging enough without adding this particular dimension.

I see two options here:
1. Both PRs are merged into one (I would then suggest keeping this open and closing #497).
2. #497 is closed for the time being, this PR is brought to a conclusion (hopefully a merge), and then #497 is reopened for whatever you aim to change next.

If you really do want to work on multiple PRs targeting one solution in parallel, then I suggest you do so in either of your individual forks, and open a PR here when you feel that an improvement you have agreed upon is ready to be merged.

Destroyerrrocket:
I'm sorry to hear that, I don't want to bug you @rbergen. I'd suggest 2, but I'll let @Michoumichmich decide. Fixing the undefined behavior should be as simple as 
[fixUndefinedBehaivor.txt](https://github.com/PlummersSoftwareLLC/Primes/files/6874424/fixUndefinedBehaivor.txt)

Michoumichmich:
Sorry, I closed the other PR temporarily

rbergen:
Thank you to both, Please let me know when this PR is ready for review.

Michoumichmich:
I applied the same edits to the parallel version, with the fix, the results were slightly better with clang++ as for the single threaded.

Destroyerrrocket:
It has (almost, missed a bureaucratic detail :P) my blessing :D

PrimeCPP/solution_2/run.sh
```diff
@@ -2,5 +2,5 @@
 # gcc -Ofast -std=c++17 PrimeCPP.cpp -lc++ -oPrimes_gcc.exe
 # clang -Ofast -std=c++17 -lc++ PrimeCPP.cpp -oPrimes_clang.exe
 
-g++ -pthread -Ofast -std=c++17 PrimeCPP_PAR.cpp -oprimes_par.exe
+clang++ -pthread -Ofast -std=c++17 PrimeCPP_PAR.cpp -oprimes_par.exe
```
Destroyerrrocket: missed the -march flags!

Michoumichmich:
> It has (almost, missed a bureaucratic detail :P) my blessing :D

It was so sad to see you edit the comment in live. What's missing?

Destroyerrrocket:
missed the -march flags thing in solution_2/run.sh (I also left you a message in the "files changed" tab)

Destroyerrrocket:
And now it is ready!

Destroyerrrocket (approved):

PR #495, opened 2021-07-25
PHP Speed Improvements (~5-10%)

thomax-thomasj:
## Description
A couple of small changes that have improved performance ~5-10% on my machine
 * Changing to multiply by 0.5 instead of divide by 2
 * Copying the rawBitsSize into a variable within the function

### DIFF
--- a/PrimePHP/solution_1/PrimePHP.php
+++ b/PrimePHP/solution_1/PrimePHP.php
@@ -41,7 +41,8 @@ class PrimeSieve
         $factor = 3;
         $sieveSize = $this->sieveSize;
         $q = sqrt($sieveSize);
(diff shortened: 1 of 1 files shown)


rbergen:
Tagging @DennisdeBest and allowing time for them to review and respond to this PR on their original solution, if they want to do so.

DennisdeBest:
Hi  @thomax-thomasj thanks for your input.

I ran these changes compared to the code without the changes 20 times inside the docker container

With the changes : 230 passes
Without the changes : 218 passes

It seems *0.5 is indeed faster than /2

@rbergen it is worth merging this pull request in my opinion

PR #497, opened 2021-07-25
Optimisation of the CPP

Michoumichmich:
There can be only 32 different rolling masks and we can get the index of the rolling mask at any point in the computing. This commit proposes to remove the false data dependency created and to pre-compute the rolling masks. I was able to almost get twice the performance on a laptop from these small changes.


Michoumichmich:
Before:
```
Passes: 3571, Time: 5.000568, Avg: 0.001400, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

davepl_pol;3571;5.000568;1;algorithm=base,faithful=yes,bits=1
```

After: 
```
Passes: 7522, Time: 5.000221, Avg: 0.000665, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

davepl_pol;7522;5.000221;1;algorithm=base,faithful=yes,bits=1
```
(tested on a I5-6360U with macOS)

Destroyerrrocket:
Are these the results from docker? Or natively?

Michoumichmich:
Just running `./run.sh` with Apple's clang-13, didn't have time to check on another machine

Destroyerrrocket:
I'm really not getting anywere close to the performance gains you report, which is incredibly odd. I'm doing a bit of research on the resulting assembly, because your proposal should be better.

Michoumichmich:
Do you still get some performance improvement? Maybe it's related to my hardware, the original version is really slow compared to the values I see posted everywhere

Michoumichmich:
On a server with rhel8 and clang13, I go from:
```
Passes: 4289, Time: 5.001130, Avg: 0.001166, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1
```
to 
```
Passes: 5196, Time: 5.000946, Avg: 0.000962, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1
```

With gcc8 the improvement is negligible though

Destroyerrrocket:
Nope, I in fact get closer to the 11000s. Given that I don't have that many setups where to test, I cannot report in a generalized sense. The difference in assembly does not seem to warrant that much of a difference, where the difference is of only two instructions. As such, maybe we could propose both solutions, that way we can cover as much room as possible.

Destroyerrrocket:
[bothSolutions.txt](https://github.com/PlummersSoftwareLLC/Primes/files/6874409/bothSolutions.txt)
For instance, this would do the trick

rbergen:
@Michoumichmich This PR now causes a merge conflict. Please solve that before declaring it ready for review.

Michoumichmich:
> [bothSolutions.txt](https://github.com/PlummersSoftwareLLC/Primes/files/6874409/bothSolutions.txt)
> For instance, this would do the trick

    It should all be fixed by this evening, sorry!

Michoumichmich:
> @Michoumichmich This PR now causes a merge conflict. Please solve that before declaring it ready for review.

It should all be fixed by this evening, sorry!

rbergen:
@Michoumichmich Do you still want to move forward towards getting this PR merged? If so, the merge conflict still needs to be resolved.

Michoumichmich:
> @Michoumichmich Do you still want to move forward towards getting this PR merged? If so, the merge conflict still needs to be resolved.

Hello @rbergen, it should be ready 🎉 (really sorry for the huge delay)

rbergen:
@davepl As this is a proposal to further update your original C++ implementation, I would be very interested to hear what you think about it.

@Destroyerrrocket Considering that you submitted the first update in #290, I would also like to know how you feel about the "current final version" of this PR?

Destroyerrrocket:
I'll be reviewing this once I'm out of work, at a first glance it's what I expected.

Destroyerrrocket:
Yep, I see no problems with it, it performs as expected. Sorry for not answering, IRL things happened, and I forgot.

rbergen:
@Destroyerrrocket That's perfectly okay. Thanks for your input.

PR #499, opened 2021-07-25
Added a native java example

jagodevreede:
## Description
This PR will add a native compiled version for java. The solution is a copy-past of solution 1 but is run completely differently. This will run a machine code version instead of the default interpreted bytecode.

## Contributing requirements


rbergen:
I'm happy to merge this because it illustrates a relevant development in the language the solution is written in, and it clearly is does not improve the performance of the existing solutions. If and when native compilation does get adopted by one of the existing solutions, this solution may become moot and subject to removal.

rbergen (approved):
LGTM

marghidanu (approved):

PR #501, opened 2021-07-25
Save file to disk

marghidanu:
Adds the ability to save the output as files on disk.

```
make DIRECTORY=PrimePerl/solution_1 FORMATTER=json OUTPUT_FILE=$PWD/output.json
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,9 @@ all: benchmark
 
 .PHONY: benchmark
 benchmark: check-env
-	@REALPATH=$$(cd "$${DIRECTORY}" && pwd); cd tools/; \
-	npm ci --silent && npm start --silent -- benchmark -d "$${REALPATH}" -f "$(FORMATTER)"
+	@REALPATH=$$(cd "$${DIRECTORY}" && pwd); \
+	ARGS=""; [ ! -z $${OUTPUT_FILE} ] && ARGS+="-o $${OUTPUT_FILE}"; \
+	cd ./tools; npm ci --silent && npm start --silent -- benchmark -d "$${REALPATH}" -f "$(FORMATTER)" $${ARGS}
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -10,8 +10,9 @@ all: benchmark
 
 .PHONY: benchmark
 benchmark: check-env
-	@REALPATH=$$(cd "$${DIRECTORY}" && pwd); cd tools/; \
-	npm ci --silent && npm start --silent -- benchmark -d "$${REALPATH}" -f "$(FORMATTER)"
+	@REALPATH=$$(cd "$${DIRECTORY}" && pwd); \
+	ARGS=""; [ ! -z $${OUTPUT_FILE} ] && ARGS+="-o $${OUTPUT_FILE}"; \
+	cd ./tools; npm ci --silent && npm start --silent -- benchmark -d "$${REALPATH}" -f "$(FORMATTER)" $${ARGS}
```
rbergen: This being the first optional argument it doesn't matter much yet, but if/when another optional argument is added, consistently managing the whitespaces between arguments would become something to consider. To prepare for that scenario, one approach you could consider is this:

```bash
	ARGS=""; [ ! -z $${OUTPUT_FILE} ] && ARGS+=" -o $${OUTPUT_FILE}"; \
	cd ./tools; npm ci --silent && npm start --silent -- benchmark -d "$${REALPATH}" -f "$(FORMATTER)"$${ARGS}
```

marghidanu: I think it’s better if I make ARGS a bash array, that way I get the spaces for free :)

marghidanu: @rbergen Done. I think it's better with a bash array.

rbergen: I agree. :)

rbergen (approved):
LGTM

PR #502, opened 2021-07-25
Improve performance and reduce stack usage. (CPP solution 3)

Destroyerrrocket:
## Description
This highly improves performance and reduces stack usage, making it hopefully run in arm. Regardless, seeing 9 digit numbers should be a good reason on its own to consider this PR. 

In order to respect the preference for clang, I had to put the sieve in a separate consteval execute() function, but gcc should be able to execute it at compile time without it.


BlackMark29A:
I also have implemented a compile-time only version, but I've arbitrarily decided for myself that making the `runSieve`, or `execute` in this case, a no-op is going too far. All your solution is measuring is how high you can increment `passes` in 5 seconds, the compiler optimizes out everything related to the sieve. So yes, you're getting insanely high numbers, but you're also not doing any work and don't even have a sieve in every iteration.
What I've decided for myself is the minimum amount of work you can do for it to still make sense in the context of this drag-race, is to create a new sieve allocated on the heap in each iteration, but fill it with compile-time generated data. So that way you still adhere to some of the faithfulness rules and actually do some amount of work related to the sieve.

Destroyerrrocket:
(sorry for the open/close, I miss-clicked)
@BlackMark29A Yeah, it's kinda pointless from the perspective of this competition, I think that this is quite unfair. That said, my improvements are not really in the constexpr-ness of this solution, but instead on the thread allocation-ness. The improvements are indeed on the way I can increase a number, but that was already what this solution did.

Destroyerrrocket:
@BlackMark29A btw, your solution is great, had a look yesterday :)

BlackMark29A:
Right, looking at how it was before your changes, it seems like the solution was measuring how many empty threads can be spawned in 5 seconds, which is also already kind of missing the point. Although did the compiler optimize out the sieve completely? Because you say you had to wrap it in `execute` for that to happen with clang?

> 
> 
> @BlackMark29A btw, your solution is great, had a look yesterday :)

Thank you :)

Destroyerrrocket:
I am pretty sure it did, but for some reason I had to push my clang version a bit, at least with my changes. That said, since I could be wrong, I'll test if with the old threading model I still get about the same results, that way we can confirm that; Expect the results this afternoon once I get back from work!

If we decided to restrict things to your standard, I think this solution would be redundant, as now it does not quite do what the name suggests and would be duplicated from your solution, as such, even if it's pointless because we're just running an empty function, I think there is some justified amount of interest. That's not saying that it is fair or any fun though, once there's a constexpr solution it's already as good as it gets.

Destroyerrrocket:
Diff from latest commit: 
[slowThreading.txt](https://github.com/PlummersSoftwareLLC/Primes/files/6879815/slowThreading.txt)

I revert back to 6 digits: 
Computing primes to 1000000 on 24 threads for 5 seconds.
Passes: 290232, Time: 5.000117, Avg: 0.000017, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

flo80_pol_constexpr;290232;5.000117;24;algorithm=base,faithful=no,bits=1
Computing primes to 1000000 on 1 thread for 5 seconds.
Passes: 189437, Time: 5.000022, Avg: 0.000026, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

flo80_pol_constexpr;189437;5.000022;1;algorithm=base,faithful=no,bits=1

BlackMark29A:
Right, so again just measuring the thread spawning overhead, and the sieve is optimized out completely. Which is also why you see abysmal scaling with the number of threads, because the thread spawning overhead increases with the number of threads, making everything slower (relatively) than the single threaded version.

@flo80 Were you aware that the sieve was optimized out completely and you were only measuring the thread spawning overhead?

@Destroyerrrocket I'm not sure what's the best approach here. As I said, I feel like letting the compiler optimize out the sieve completely and doing 0 work related to the sieve feels very wrong and against the spirit of the competition to me. But you're right, solution 3 is already doing that, so improving it by not measuring the thread spawning overhead, and instead just the counting of `passes` does make sense.

flo80:
@BlackMark29A well it was a bit the point of the constexpr solution to optimize most away (I think actually in my submission, the compiler did not optimize it completely, for sure there was still a stack allocation in each pass with some copying of results since I did not optimize down to only keeping the count; runSieve though should have been largely gone -- depending on your compiler, flags, etc). 

Having said that: I was surprised, that some of the other solutions (iirc nim, zig) still got similiar / even higher numbers (zig e.g. also by doing it in compile time and just counting in run time)

I feel in the spirit of the competition, **it is not faithful, it should not be on the leaderboard** -- but it can  give an indication of what is a maximum a solution can achieve when the language & compiler really optimize stuff (in the end, it still uses base algorithm, so the work happens at some point). 

And I also see it as a learning opportunity for developers to see, what can be done with more or less modern language constructs. I would actually be also very interested in some secondary leaderboards of memory usage, executable size, etc. to also better see the tradeoffs (without checking, I would expect a constexpr solution has a larger binary if it still keeps the whole sieve around and not just the count)

Destroyerrrocket:
I just wanted to make sure that indeed the previous solution was optimizing all of runSieve in the benchmark. I've written a smaller patch that only tackles the performance improvements of the threading. 
[fastThreadingBadStack.txt](https://github.com/PlummersSoftwareLLC/Primes/files/6881001/fastThreadingBadStack.txt)
(patch against original solution)
As you will see, it indeed is optimizing all the way, and I get back the massive results of:
Computing primes to 1000000 on 24 threads for 5 seconds.
Passes: 741834527, Time: 5.000520, Avg: 0.000000, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

flo80_constexpr;741834527;5.000520;24;algorithm=base,faithful=no,bits=1
Computing primes to 1000000 on 1 thread for 5 seconds.
Passes: 232676405, Time: 5.000046, Avg: 0.000000, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

flo80_constexpr;232676405;5.000046;1;algorithm=base,faithful=no,bits=1

So, no doubt I'm not optimizing constexpr any more than it already is.

About whether it is worth it, I'll side with @flo80. It is a clear and simple, it's easy to read, and could be useful to someone. As long as everyone is aware that this is not "kinda not faithful" (like preallocating memory) but "absolutely not faithful" (like literally having done all the work at compile time), it is fine by me.

BlackMark29A:
So, I think it's very important for our discussion that I make something clear. Consider this diff where I remove every `constexpr` (except some constants, which doesn't matter for the point I'm trying to make):
<details>
<summary>Removing constexpr diff</summary>

```diff
diff --git a/PrimeCPP/solution_3/PrimeCPP_CONSTEXPR.cpp b/PrimeCPP/solution_3/PrimeCPP_CONSTEXPR.cpp
index a13ac11..3a45e89 100644
--- a/PrimeCPP/solution_3/PrimeCPP_CONSTEXPR.cpp
+++ b/PrimeCPP/solution_3/PrimeCPP_CONSTEXPR.cpp
@@ -20,7 +20,7 @@ struct RESULTS
     int count;
 };
 
-constexpr RESULTS resultsDictionary[] = {
+RESULTS resultsDictionary[] = {
     {10LL, 4},   // Historical data for validating our results - the number of primes
     {100LL, 25}, // to be found under some limit, such as 168 primes under 1000
     {1000LL, 168},
@@ -30,7 +30,7 @@ constexpr RESULTS resultsDictionary[] = {
     {10000000LL, 664579},
 };
 
-constexpr int find(const unsigned long long val)
+int find(const unsigned long long val)
 {
     for (const auto d : resultsDictionary)
     {
@@ -40,12 +40,12 @@ constexpr int find(const unsigned long long val)
     return -1;
 }
 
-constexpr int countPrimes(const Sieve &sieve)
+int countPrimes(const Sieve &sieve)
 {
     return sieve.count();
 }
 
-constexpr bool validateResults(const Sieve &sieve)
+bool validateResults(const Sieve &sieve)
 {
     const auto sieveSize = sieve.size();
     const auto result = find(sieveSize);
@@ -85,7 +85,7 @@ void printResults(const Sieve &sieve, bool showResults, double duration, int pas
            validateResults(sieve));
 
     printf("\n");
-    printf("flo80_constexpr;%d;%f;%d;algorithm=base,faithful=no,bits=1\n", passes, duration, threads);
+    printf("flo80_no-constexpr;%d;%f;%d;algorithm=base,faithful=no,bits=1\n", passes, duration, threads);
 }
 
 void run(uint64_t upperLimit, const Sieve &checkSieve, int numberOfThreads = 1, int runtimeSeconds = 5)
@@ -125,7 +125,7 @@ void run(uint64_t upperLimit, const Sieve &checkSieve, int numberOfThreads = 1,
 int main(int argc, char **argv)
 {
     uint64_t upperLimit = 1'000'000L;
-    constexpr int runtimeSeconds = 5;
+    int runtimeSeconds = 5;
 
     if (argc > 1)
     {
diff --git a/PrimeCPP/solution_3/Sieve.h b/PrimeCPP/solution_3/Sieve.h
index e3c2a5c..725a962 100644
--- a/PrimeCPP/solution_3/Sieve.h
+++ b/PrimeCPP/solution_3/Sieve.h
@@ -4,7 +4,7 @@
 #include <cstdint>
 
 // from https://baptiste-wicht.com/posts/2014/07/compile-integer-square-roots-at-compile-time-in-cpp.html
-static constexpr uint64_t ct_sqrt(uint64_t res, uint64_t l, uint64_t r)
+static uint64_t ct_sqrt(uint64_t res, uint64_t l, uint64_t r)
 {
 	if (l == r)
 	{
@@ -25,7 +25,7 @@ static constexpr uint64_t ct_sqrt(uint64_t res, uint64_t l, uint64_t r)
 	}
 }
 
-static constexpr uint64_t ct_sqrt(uint64_t res)
+static uint64_t ct_sqrt(uint64_t res)
 {
 	return ct_sqrt(res, 1, res);
 }
@@ -34,23 +34,23 @@ class Sieve
 {
 public:
 	using U = uint64_t;
-	static constexpr uint64_t maxSize = 50000000L; // stack size limited
+	constexpr static uint64_t maxSize = 50000000L; // stack size limited
 
 private:
-	static constexpr auto wordsize = sizeof(U) * 8;
-	static constexpr auto words = (maxSize + wordsize - 1) / wordsize;
+	constexpr static auto wordsize = sizeof(U) * 8;
+	constexpr static auto words = (maxSize + wordsize - 1) / wordsize;
 
 	U bitmap[words] = {0}; // 0 means contains, 1 means missing
 	uint64_t sieveSize = maxSize;
 
-	constexpr void insert(uint64_t index)
+	void insert(uint64_t index)
 	{
 		const auto bi = index / wordsize;
 		const auto off = index % wordsize;
 		bitmap[bi] &= ~(((U)1) << off);
 	}
 
-	constexpr void remove(uint64_t index)
+	void remove(uint64_t index)
 	{
 		const auto bi = index / wordsize;
 		const auto off = index % wordsize;
@@ -58,23 +58,23 @@ private:
 	}
 
 public:
-	constexpr bool contains(uint64_t index) const
+	bool contains(uint64_t index) const
 	{
 		const auto bi = index / wordsize;
 		const auto off = index % wordsize;
 		return 0 == (bitmap[bi] & (((U)1) << off));
 	}
 
-	constexpr Sieve() {}
+	Sieve() {}
 
-	constexpr Sieve(const uint64_t size) : sieveSize(size) {}
+	Sieve(const uint64_t size) : sieveSize(size) {}
 
-	constexpr uint64_t size() const
+	uint64_t size() const
 	{
 		return sieveSize;
 	}
 
-	constexpr uint64_t count() const
+	uint64_t count() const
 	{
 		uint64_t count = sieveSize > 1 ? 1 : 0;
 		for (uint64_t num = 3; num <= sieveSize; num += 2)
@@ -88,7 +88,7 @@ public:
 		return count;
 	}
 
-	constexpr void runSieve()
+	void runSieve()
 	{
 		uint64_t factor = 3;
 		const auto q = (int)ct_sqrt(sieveSize);
```

</details>

Running this yields the following results:

```
Computing primes to 1000000 on 16 threads for 5 seconds.
Passes: 144736, Time: 5.000040, Avg: 0.000035, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

flo80_no-constexpr;144736;5.000040;16;algorithm=base,faithful=no,bits=1
Computing primes to 1000000 on 1 thread for 5 seconds.
Passes: 110369, Time: 5.000025, Avg: 0.000045, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

flo80_no-constexpr;110369;5.000025;1;algorithm=base,faithful=no,bits=1
```

Compared to the original solution 3:
```
Computing primes to 1000000 on 16 threads for 5 seconds.
Passes: 138496, Time: 5.000145, Avg: 0.000036, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

flo80_constexpr;138496;5.000145;16;algorithm=base,faithful=no,bits=1
Computing primes to 1000000 on 1 thread for 5 seconds.
Passes: 109540, Time: 5.000021, Avg: 0.000046, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

flo80_constexpr;109540;5.000021;1;algorithm=base,faithful=no,bits=1
```

In other words, the reason why this solution performes the way it does has absolutely nothing to do with `constexpr`. The compiler optimizes out the sieve in the loop, because it has no side effects, and not because it's `constexpr`. The original C++ solution prevented this by validating the result of the sieve in the loop, creating a dependency that the compiler cannot optimize out, whereas this solution generates a sieve beforehand that is then used for validation, allowing the compiler to completely remove the sieve in the loop.

Which is why I keep saying that I think it's a problem that there is no work being done related to the sieve in the loop. Because if you introduce a dependency on the sieve in the loop, so that the compiler cannot optimize it out, the solution will be very slow and not actually do anything at compile-time:

```diff
diff --git a/PrimeCPP/solution_3/PrimeCPP_CONSTEXPR.cpp b/PrimeCPP/solution_3/PrimeCPP_CONSTEXPR.cpp
index a13ac11..b982997 100644
--- a/PrimeCPP/solution_3/PrimeCPP_CONSTEXPR.cpp
+++ b/PrimeCPP/solution_3/PrimeCPP_CONSTEXPR.cpp
@@ -90,7 +90,7 @@ void printResults(const Sieve &sieve, bool showResults, double duration, int pas
 
 void run(uint64_t upperLimit, const Sieve &checkSieve, int numberOfThreads = 1, int runtimeSeconds = 5)
 {
-
+    unsigned long long dummy = 0;
     unsigned int passes = 0;
 
     printf("Computing primes to %llu on %d thread%s for %d second%s.\n", upperLimit,
@@ -106,10 +106,11 @@ void run(uint64_t upperLimit, const Sieve &checkSieve, int numberOfThreads = 1,
 
         for (unsigned int i = 0; i < numberOfThreads; i++)
         {
-            threadPool.push_back(thread([upperLimit]
+            threadPool.push_back(thread([upperLimit, &dummy]
                                         {
                                             auto sieve = Sieve(upperLimit);
                                             sieve.runSieve();
+                                            dummy += sieve.size();
                                         }));
         }
         for (auto &th : threadPool)
```

```
Computing primes to 1000000 on 16 threads for 5 seconds.
Passes: 12080, Time: 5.003280, Avg: 0.000414, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

flo80_constexpr;12080;5.003280;16;algorithm=base,faithful=no,bits=1
Computing primes to 1000000 on 1 thread for 5 seconds.
Passes: 3531, Time: 5.001343, Avg: 0.001416, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

flo80_constexpr;3531;5.001343;1;algorithm=base,faithful=no,bits=1
```

 So I think the best way to solve this would be to introduce this dependency (not like I did, but like the original C++ solution) and enforce compile-time evaluation (`consteval`, for example), which should then actually show the power of compile-time evaluation.

Destroyerrrocket:
That's a great idea! I think that introducing this dependency is the best compromise. I'll work on that.

Destroyerrrocket:
This has obviously reduced the performance a bit, but by halving the amount of data necessary, I'm able to get ahead of the amazing solution of @BlackMark29A :D

BlackMark29A:
Nice, but this could still be improved. The dependency you introduced with `shadowFunc` will incur a function call overhead in every iteration, which is why you are not getting the same performance. If you change the dependency to be like the original C++ solution by Dave, where the result of the last iteration is used for validation, you should be able to get the same performance as you had before, where the sieve was completely optimized out. Because the sieve will be needed for the validation it cannot be optimized out completely, but it also has no run-time cost, because it was computed at compile-time.

Also, please don't do this:
```cpp
static constexpr auto words = (upperLimit >> 7) + (((upperLimit >> 1) & 63) > 0);
```
This is impossible to read, for no reason. It's not going to be any faster, because it's a `constexpr` anyway, so writing it in a way that makes it clear what is happening and why would be much better.

Destroyerrrocket:
If you don't mind, could you provide such a change? I did my best to do that myself but clang outsmarted me and only allocated the stack on the last iteration. 

I'll simplify the formula, I just used what I already used in my previous solutions so I did not have to think much.

BlackMark29A:
> If you don't mind, could you provide such a change? I did my best to do that myself but clang outsmarted me and only allocated the stack on the last iteration.

Why is that a problem though? That's what I would want to happen anyway, only loading the sieve in the last iteration. So I think you already had the implementation I had in mind. I think it's fine for the compiler to optimize out the sieve for every iteration except the last one, I just don't think it's fine to get rid of the sieve entirely.

> I'll simplify the formula, I just used what I already used in my previous solutions so I did not have to think much.

Thanks.

Destroyerrrocket:
Oh, ok then. The performance is about the same as my first proposal, though. Man, what a shame I did not stash the changes.

BlackMark29A:
Yes, the performance should be the same. But the important difference is that your initial solution and the current state of solution 3 does not depend on `constexpr` to achieve this performance. The reason why it's fast is because nothing depends on the sieve and the compiler just throws it out completely. This could be done with C as well, which doesn't even have `constexpr`. But the point of solution 3 should be to showcase the power of `constexpr` and so its performance must entirely depend on `constexpr` which it will, if you introduce a dependency on the sieve that the compiler cannot optimize out.

So in somewhat pseudo-code, that's what I had in mind:
```cpp
while(true) {
  constexpr auto sieve = execute(sieveSize);
  if(elapsedTime >= 5s) {
    validate(sieve);
    break;
  }
}
```

but I guess that's what you had as well. Sorry for not making myself more clear.

Destroyerrrocket:
Ok, here you go!

Destroyerrrocket:
I accually found a stash with my changes, I've moved things a bit but is mostly what I had already implemented. The randomness with the decided thread was to try to dissuade clang from doing what we wanted it to do, but since I find it a nice addition, I've left it in.

BlackMark29A:
Looks good to me, now it actually shows the power of compile-time evaluation. Well done.

@flo80 What do you think, it's a PR for your solution after all?

rbergen:
@Destroyerrrocket, @BlackMark29A, @flo80 If there are no last-minute changes or insights, I will move forward with merging this, tomorrow.

PR #503, opened 2021-07-25
	modified:   solution_1/PrimeREXX.rex

jossRRM:
Modified variable names to match the sample C# names in
	solution_1 (for example sieveSize).
	Moved the Sieve() function to the middle of the code. It was at
	the top because the interpreter starts at the top looking for a
	subroutine when it is called. But since it is only called once
	and since the timer has been moved to the subroutine, there is
	no reason for Sieve() to sit alone at the top of the program.
	corrected the "docker run" comments to change the directory
	separator character to "/".
	New Dockerfile to build the NetRexx solution
	NetRexx sieve of Erasthosthenes program
	How to build and run the Sieve.

## Contributing requirements

### DIFF
--- a/PrimeREXX/solution_1/PrimeREXX.rex
+++ b/PrimeREXX/solution_1/PrimeREXX.rex
@@ -1,149 +1,206 @@
 /* rexx */
-signal Main
-
-Sieve:
-count = 0
-do while time("E") <= 5
-  primes. = 1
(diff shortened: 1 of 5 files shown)


jossRRM:
I think I need to split the Solution_1 and solution_2 into separate pull requests.

PR #504, opened 2021-07-26
Pascal solution 3

olivierbrun:
## Description

Implemented an optimized implemention of the base algorithm in freepascal. The optimization in the constructor should provide some boost. This implementation is quite close to Dave's original source which should provide some fair comparison with other language.

## Contributing requirements

### DIFF
--- a/PrimePascal/solution_3/Dockerfile
+++ b/PrimePascal/solution_3/Dockerfile
@@ -1,0 +1,11 @@
+FROM primeimages/freepascal:3.2.0 AS build
+
+WORKDIR /opt/app
+COPY *.pas .
+
+RUN fpc PrimePas -O2
+
+FROM ubuntu:20.04
+COPY --from=build /opt/app/PrimePas /opt/app/
+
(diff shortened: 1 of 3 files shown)


rbergen:
This solution has the same key characteristics (base algorithm, faithful, 1 bit per prime) as PrimePascal/solution_2 does. That means that in principle, we won't merge this as a new solution.
If you believe you can make changes to solution_2 that improve its performance, then please structure your PR as one that modifies said solution accordingly.

olivierbrun:
hello Rutger, thanks for the quick feedback, I agree with your comment. 
I'll update my code to propose a different and hopefully faster algorithm, also providing ability to optionally handle multithreading.

rbergen:
@olivierbrun Thank you, and that's perfectly fine. Is it okay with you if we close this PR while you work on that, and have you open a new one when you feel it's ready to merge?

olivierbrun:
Let's close this PR, and I'll push a new one.

PR #506, opened 2021-07-26
Added a Standard ML solution

NotMatthewGriffin:
## Description

Standard ML of NJ implementation.

## Contributing requirements


rbergen:
@NotMatthewGriffin Thank you for submitting this solution. I always love it when we can add a language to the collection that one tends to come across less often.

I do have a question about the naming of the language directory. As I understand it, SML/NJ is actually a compiler for the StandardML language. If that is correct, I think it would be better to name the language directory `PrimeStandardML` or `PrimeSML` instead of `PrimeSMLofNJ`. What are your thoughts on this?

NotMatthewGriffin:
Hi @rbergen . Thanks for taking a look at this. I would be happy to have this renamed to `PrimeSML` or `PrimeStandardML` but do you think that would be okay if I haven't tested it with other SML compilers?

rbergen:
I can vouch for the fact that most of the languages here had only been compiled with one compiler when they were originally added, and I expect many still are. If you want, you can mention the use of SML/NJ in the README.md (which you do) and include it in the label in your drag-race output line (`NotMatthewGriffin_SMLofNJ` or something similar). 

Then, please rename the root folder of your solution to whichever mentioned alternative you prefer.

PrimeStandardML/solution_1/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu:20.04
+RUN apt-get update -y && \
+    apt-get install smlnj -y
+COPY sml_primes.sml sml_primes.sml
+COPY build.sml build.sml
```
rbergen, line 6: Please set a WORKDIR before performing COPYs to relative destinations. CI now fails due to it missing.

rbergen (approved):
LGTM

marghidanu (approved):

PR #510, opened 2021-07-26
Rust idiomatic improvements for `solution_1`

mike-barber:
## Description

@Kulasko and I have worked through the improvements to make the code more idiomatic. There are a couple of performance improvements in there too.

Ideally, we would also like to refactor the single `main.rs` into separate files. However, this will make review more painful as it's hard to see what lines were changed when doing this, so we haven't done it yet. 

It might be easier if this is in a completely separate refactor-only PR. @rbergen would it be easier doing the file split separately? Or would you prefer us to add it to this PR? It's also an option to just not do it at all -- I know you guys have a lot on your plates.

## Contributing requirements

### DIFF
--- a/PrimeRust/solution_1/Dockerfile
+++ b/PrimeRust/solution_1/Dockerfile
@@ -6,7 +6,7 @@
 # Note: Debian based, not Alpine; for some reason, the 
 # Alpine malloc call is slower than it should be.
 #
-FROM rust:1.52.1 AS build
+FROM rust:1.53 AS build
 
 WORKDIR /app
 COPY . .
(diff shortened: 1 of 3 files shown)


rbergen:
> It's also an option to just not do it at all -- I know you guys have a lot on your plates.

I appreciate you offering that option, because I really would prefer if you didn't. In a future video, Dave will make the point that with PRs (in number and scope), the "less is more" principle really does apply. 
Simply put, if you were to restructure your code along the lines of what you described, we would basically have to review the whole set of implementations as if they were part of a brand new solution. That is a significant amount of work, which I hope you understand I would rather not do unless really necessary.

mike-barber:
Thanks @rbergen -- I understand completely. It's certainly more of a nice-to-have than a necessity, and we left it as an optional last step for this reason. Thanks for the quick review and merge of the other changes :)

PR #515, opened 2021-07-26
[x86_64 Assembly] Use C wrapper function for clock_gettime (possible performance improvement)

weirddan455:
On my system, in a loop simply calling clock_gettime as fast as possible, I am able to make 221 million iterations in 5 seconds using the C wrapper function while making a syscall slows this number down to 11 million.  It doesn't seem to make a huge difference in the overall result once it starts doing real work in the loop but who knows, it may make a bigger difference on some systems.

The reason for this performance increase is that this function is available via the vDSO.  The C wrapper function takes advantage of this and avoids making a context switch into kernel space (instead executing in userspace.)  There's an article that details this issue here (it's about EC2 but the principle is the same):  https://blog.packagecloud.io/eng/2017/03/08/system-calls-are-much-slower-on-ec2/

For some reason, primes_uff_byte and primes_uff_bitbtr are segfaulting with this change and I'm not sure why so I've left those out of this PR.

### DIFF
--- a/PrimeAssembly/solution_1/primes_ff_bitbtr.asm
+++ b/PrimeAssembly/solution_1/primes_ff_bitbtr.asm
@@ -7,6 +7,7 @@ global main
 extern printf
 extern malloc
 extern free
+extern clock_gettime
 
 default rel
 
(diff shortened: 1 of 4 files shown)


rbergen:
@weirddan455 Thanks for proposing this. Based on my own testing:
- I'm happy to merge this for the ff_bitshift and uff_bitshift implementations. The change does seem to make a positive impact on their performance.
- I see a negative impact on the performance on the ff_byte and ff_bitbtr implementations. I would therefore like to ask you to take those two implementations out of your commit.

weirddan455:
Done

PR #516, opened 2021-07-27
add solution in yoix scripting language

mmcdon20:
## Description

What? You've never heard of [yoix scripting language](https://en.wikipedia.org/wiki/Yoix)? It's a scripting language created by AT&T, [which is no longer supported according to their README on github](https://github.com/att/yoix).

I have never written a yoix program before, and I'm not really a fan after trying it out.

This is a faithful 1-bit implementation.

There are no class definitions, so I took a similar approach to my solution in the red programming language. The PrimeSieve function returns a yoix `Dictionary` which acts like a class/object and contains the fields and methods. 

There doesn't appear be anything like a bitset in the standard library. The only numeric types seem to be `int` and `double`. The `int` is a 32 bit integer, so I used an `Array` of `int` where each `int` stores 32 prime flags. 

## Contributing requirements


mmcdon20:
Consider this stable. I don't intend to add/change anything else unless something comes up in review.

rbergen (approved):
LGTM

marghidanu (approved):

PR #521, opened 2021-07-27
[C++ Solution 1/2] Fixed incorrect array size

weirddan455:
These AND masks need to be in decimal rather than hex.  The 0x31 mask happens to work for 1 million sieve size but you get a compiler warning if you try to compile for 1 million + 2 for example (complains that you're trying to fill 1 too many bytes with memset.)

0x07 is obviously the same in hex as it is in decimal.  Just changed it for consistency.

### DIFF
--- a/PrimeCPP/solution_1/PrimeCPP.cpp
+++ b/PrimeCPP/solution_1/PrimeCPP.cpp
@@ -20,7 +20,7 @@ class BitArray {
     size_t arrSize;
 
     inline static size_t arraySize(size_t size) {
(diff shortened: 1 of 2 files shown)


rbergen:
@Destroyerrrocket Would you like to comment on this?

Destroyerrrocket:
Yep, that's true. I think it can be merged, it should not affect performance.

Destroyerrrocket:
Accually, solution 2 should have the same bug. If you don't mind, you could change that as well.

weirddan455:
Added solution 2 fix to this PR.

PR #522, opened 2021-07-27
Rust `solution_1` bit storage improvements

mike-barber:
## Description

This PR adds a variant of `striped-bits` that addresses the locality problem with the original one. It was always pretty fast on CPUs with plenty of cache, but performed pretty poorly on smaller CPUs. This solves the locality problem by dealing with smaller blocks that fit better in cache.

Changes: 
- corrected extra iteration bug on main sieve
- added more cache-friendly `FlagStorageBitVectorStripedBlocks` version of the striped implementation
- simplified stripe/word accounting on `FlagStorageBitVectorStriped` based on learnings from the above
- added more tests to validate storage directly

@rbergen - apologies for the recent frequency of PRs for this solution. I'm hoping this is the last one for a while - this was the last big thing to address. 

## Contributing requirements

### DIFF
--- a/PrimeRust/solution_1/README.md
+++ b/PrimeRust/solution_1/README.md
@@ -79,7 +79,7 @@ bit-storage     Passes: 151710, Threads: 24, Time: 5.0007929802, Average: 0.0000
 bit-storage     Passes: 151644, Threads: 24, Time: 5.0008230209, Average: 0.0000329774, Limit: 1000000, Counts: 78498, Valid: Pass
 ```
 
-## Performance - **Raspberri Pi 4**
+## Performance - **Raspberry Pi 4**
 
 Tested on a **Raspberry Pi 4**, standard clock with active cooling, running 64-bit Ubuntu Server 20.04.
 - Examples are run with `cargo run --release -- --repetitions=3` to get 3 repetitions.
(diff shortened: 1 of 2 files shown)


rbergen:
@mike-barber I was wondering if these latest implementations rely on some form of one-time calculation of bitmaps (blocks) that are then projected on the sieve?

mike-barber:
@rbergen no luckily not :) It's the same algorithm as before, but I've just broken the work set up into smaller blocks that fit in the cache better. 

Before, I was running over the entire sieve 8 times -- one for each bit. Now, I run over a smaller part of the sieve (a block) 8 times (one for each bit), then proceed to the next block. 

The way the bits are masked and applied is still calculated on the fly. I think that's what you're asking, but I may have misunderstood :) 
```rust
// get mask for this bit position
let mask = !(1 << bit_idx);

// unrolled loop
while word_idx < effective_len.saturating_sub(skip * 3) {
    // Safety: we have ensured that (word_idx + skip*N) < length
    unsafe {
        *block.get_unchecked_mut(word_idx) &= mask;
        *block.get_unchecked_mut(word_idx + skip) &= mask;
        *block.get_unchecked_mut(word_idx + skip * 2) &= mask;
        *block.get_unchecked_mut(word_idx + skip * 3) &= mask;
    }
    word_idx += skip * 4;
}
...
```

rbergen:
@mike-barber Your phrasing of the question would indeed have boiled down to the same thing; mine just felt a little easier to type in. :)

In any case, thank you for your explanation!

mike-barber:
You're welcome! I'm glad I answered the right question - I'm really good at answering completely the wrong one at times.

PR #524, opened 2021-07-27
Java solutions using raw integer storage and bit manipulation

chrvanorle:
## Description

These are faithful Java implementations of the base algorithm using boolean/byte/int/long arrays. The first being an 8 bit implementations and the others being 1 bit. Multithreading is supported for all implementations. The Dockerfile will run all implementations both single threaded and multithreaded.

The I32 and I64 versions are about twice as fast as the BitSet 1-bit solution_2 (710/s vs 364/s). Even though i thought the 64 bit implementation would be faster, the 32 bit one outperforms it on every machine i tried it on.

## Contributing requirements


rbergen:
Would it be fair to say that the "boolean" implementation is conceptually the same as Java/solution_1? Because in that case I would like to pull that particular implementation out and have the solution focus on the ones that are indeed different.

chrvanorle:
The bool solution indeed ended up very similar to solution_1, so i removed it.

chrvanorle:
I've added solutions that cache the masks used. I get mixed performance results, for some CPU's it seems the caching performs worse then just doing the operations, so i keep both. I'll update the readme with results when i get the time.

rbergen:
> I've added solutions that cache the masks used. I get mixed performance results, for some CPU's it seems the caching performs worse then just doing the operations, so i keep both. I'll update the readme with results when i get the time.

That's fine, but then I'll hold on reviewing and merging this, until then.

chrvanorle:
Added I64PatternCalc, which i believe is an 'other' algorithm due to using masks to clear multiple prime candidates per write. The results have also been added.

rbergen (approved):
LGTM

marghidanu (approved):

PR #525, opened 2021-07-27
add solution in umple language

mmcdon20:
## Description

[Umple](https://cruise.umple.org/umple/) is a programming language / UML modelling language created by The University of Ottawa. 

This is a faithful 1-bit implementation.

Umple apparently can compile to source code in a variety of languages. This implementation targets java.  Since I am targeting java, I was able to import the `BitSet` from the java standard library, which is how I am storing the prime flags. 

## Contributing requirements

### DIFF
--- a/PrimeUmple/solution_1/.gitignore
+++ b/PrimeUmple/solution_1/.gitignore
@@ -1,0 +1,4 @@
+*.jar
+*.java
+logs/
+build/
--- a/PrimeUmple/solution_1/Dockerfile
+++ b/PrimeUmple/solution_1/Dockerfile
@@ -1,0 +1,9 @@
+FROM openjdk:11
+
+WORKDIR /opt/app
+COPY PrimeUmple.ump .
+
+RUN curl -sL https://try.umple.org/scripts/umple.jar -o umple.jar
+RUN java -jar umple.jar PrimeUmple.ump
+
+ENTRYPOINT [ "java", "PrimeUmple.java" ]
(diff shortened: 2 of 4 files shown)


mmcdon20:
Consider this stable. I don't intend to add/change anything else unless something comes up in review.

rbergen:
Wow, this language gives me flashbacks to when I was using Rational Rose (before IBM bought it) to draw UML that the tool would translate to Java source files, heavily annotated with comments that tried to keep UML model and Java source connected. That sort of worked, but not really. This is an interesting more direct approach.

EDIT: after a first look at the code you submitted, I guess a one-class association- and state machine-less application does not illustrate too much of the specific capabilities of Umple over Java. :)

mmcdon20:
> Wow, this language gives me flashbacks to when I was using Rational Rose (before IBM bought it) to draw UML that the tool would translate to Java source files, heavily annotated with comments that tried to keep UML model and Java source connected. That sort of worked, but not really. This is an interesting more direct approach.

I thought this was a pretty interesting approach too. 

> EDIT: after a first look at the code you submitted, I guess a one-class association- and state machine-less application does not illustrate too much of the specific capabilities of Umple over Java. :)

You would be right about that. This is similar to a java implementation, with a few differences. You have "depend" statements inside the classes instead of "import" statements at the top of the files. The compiler automatically generates a constructor, so there is no constructor that I have defined explicitly. I used an [after statement](https://cruise.umple.org/umple/BeforeandAfterStatements.html) to set all of the bits in the `BitSet` to 1 in the constructor of the generated code. I only just thought of this now, 
 but I could also add before statements on `getBit` and `clearBit` for assertions. I'll go ahead and add that in just to show a little bit more umple capabilities.

mmcdon20:
I just noticed a mistake in the README file, which I corrected. 
This PR is now stable.

rbergen (approved):
LGTM

marghidanu (approved):

PR #530, opened 2021-07-28
Java: Added SimpleBitSet which is faster than the boolean[] and uses far less data

squir3l:
Added SimpleBitSet which is faster than the boolean[] and uses far less data.
Fixed bug where q was being checked for less than not less than or equal.
Reduced the amount of factors checked (using num +=2 instead of num++)

## Contributing requirements

* [ ] I placed my solution in the correct solution folder. (Existing solution)
* [ ] I added a README.md with the right badge(s). (Existing solution)
* [ ] I added a Dockerfile that builds and runs my solution. (Existing solution)


squir3l:
Didn't update the README.md to indicate bitset is green

PR #531, opened 2021-07-28
Java solution_1 implemented a SimpleBitSet that is faster than the previous boolean[]

squir3l:
## Description
Java solution_1
- Added SimpleBitSet which is faster than the boolean[] and uses far less data
- Update to indicate Java solution_1 is using bits
- Fixed bug where sqrt of initial number was using LT instead of LTE

## Contributing requirements

* [ ] I placed my solution in the correct solution folder. (Existing solution)
* [ ] I added a Dockerfile that builds and runs my solution. (Existing solution)

### DIFF
--- a/PrimeJava/solution_1/PrimeSieveJava.java
+++ b/PrimeJava/solution_1/PrimeSieveJava.java
@@ -5,24 +5,22 @@ public class PrimeSieveJava
 {
 	private static final Map<Integer, Integer> VALIDATION_DATA;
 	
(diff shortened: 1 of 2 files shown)


rbergen:
This PR changes the key attributes of the solution it's targeting. That means we can't merge this is a change to that solution. As we already have a faithful, 1-bit, base algorithm solution in place, you can open a PR on that one to improve it, if you think that is possible.

PR #533, opened 2021-07-28
[C#]A solution using pointers, intrinsics and rolling mask marking.

ItalyToast:
## Description
This solution is using pointers to dodge boundchecks, hardware intrinsics and mask marking. I submitted it as a new solution since it is not as readable as any of the existing ones.

Im not sure if this solution is considered faithful since:
- it does mark with a mask 
- it does a couple of warmup rounds

(I have marked it faithful for now since i did not find any rule that it breaks).

## Contributing requirements


rbergen:
> Im not sure if this solution is considered faithful since:
> 
> * it does mark with a mask
> * it does a couple of warmup rounds

It's faithful, but not the base algorithm due to the use of masking. Masking algorithms should be marked as "other". 

For now, we've decided that a warm-up of a maximum of 5 seconds is allowed. If your warm-up routine exceeds this, then please shorten (or remove) it.

ItalyToast:
@rbergen The solution have been updated to use a rolling mask to remove the factors one at a time (similar to the C++ solution). The warmup is now limited to 3 seconds. 

I was a little bit bummed out that the mask marking wasnt considered a base algorithm (sad performance noises). But I get the reasons why. Comparability of the algorithm against bits:8 and bits:32 is not really useful since they dont have access to those kinds of tricks.

mike-barber:
@ItalyToast this is still a really smart solution, even if you can't apply the whole masking approach. I was going to propose my own C# solution that outperforms all the other ones by a fair margin. Yours is faster than mine, so I'm not going to waste reviewers' time with mine :)

Having said this, I pulled your code, and applied some of the same optimisations and fixes I have in my solution. It adds another 17% throughput on my machine (Ryzen 3900X)-- ~12800 -> ~15000.

I'll open a PR into your branch with the changes. If you're interested in merging them, I thought it'd be a good time to do it before this is merged -- save reviewers some time ;)

EDIT: I've opened that PR here: https://github.com/ItalyToast/Primes/pull/1

ItalyToast:
To the reviewer, I'll write an update here when its ready for merging. I maybe also have some more tricks.

ItalyToast:
@rbergen I think its ready for a merge. @mike-barber pr has been merged and credited. I did a reversed version of his unrolled loop for a couple of extra % increase. 

One of the few places left to squezze out the last bit of performance is if you can remove the subtraction in the inner loop in ClearBitsDense but the ways I tried did not increase performance. Or if you can write the code in such a way that it would optimize it to a masking operation but semantically it removes one bit at a time but I'm not sure its even possible.

Its been very fun working on this solution. Thanks for making this a thing. 👏

PrimeCSharp/solution_4/PrimeCS.cs
```diff
@@ -0,0 +1,257 @@
...
+                    i2 += factor4;
+                    i3 += factor4;
+                }
+
+                while (i0 < limit)
+                {
+                    ptr[i0 / 64] |= 1ul << (int)(i0 % 64);
+                    i0 += factor;
+                }
+            }
```
rbergen: I can see the point of leaving the otherwise unused `ClearBitsSparse` method in to document the foundation of the unrolled approach. However, if the reverse unrolled version is the one used in the implementation (which it is), I think we can drop the unrolled, non-reversed version from the code. For one, I found out I could ignore it from a review perspective after I'd already reviewed it.

rbergen:
> Its been very fun working on this solution. Thanks for making this a thing. 👏

Glad to hear that. Credits for making this a thing go to @davepl, as far as I'm concerned. He started it, so it's his fault we're all here. :)

PrimeCSharp/solution_4/PrimeCS.cs
```diff
@@ -0,0 +1,257 @@
...
+                    i1 += factor4;
+                    i2 += factor4;
+                    i3 += factor4;
+                }
+
+                for (iter += 4; iter >= 0; iter--)
+                {
+                    ptr[i0 / 64] |= 1ul << (int)(i0 % 64);
+                    i0 += factor;
+                }
```
mike-barber, line 133: This is faster, but I couldn't understand why initially. Did a bit of digging, and could be because we've hoisted the limit somehow. 

As a result, this is even faster as it generates less instructions than maintaining a separate `iter` variable. I'm not 100% sure why, but might be worth a try. Small improvement on my machine -- 15,550 -> 15,750 or so.
```C#
int lim = (int)limit; // hoist and convert
var i0 = start;
var i1 = start + factor;
var i2 = start + factor * 2;
var i3 = start + factor * 3;

var factor4 = factor * 4;
while (i3 < lim)
{
    ptr[i0 / 64] |= 1ul << (int)(i0 % 64);
    ptr[i1 / 64] |= 1ul << (int)(i1 % 64);
    ptr[i2 / 64] |= 1ul << (int)(i2 % 64);
    ptr[i3 / 64] |= 1ul << (int)(i3 % 64);

    i0 += factor4;
    i1 += factor4;
    i2 += factor4;
    i3 += factor4;
}

while (i0 < lim)
{
    ptr[i0 / 64] |= 1ul << (int)(i0 % 64);
    i0 += factor;
}
```

PrimeCSharp/solution_4/PrimeCS.cs
ItalyToast: I will move the Unrolled version that is not used to the ClearBits.cs file where I have experimented with different versions. I also added a comment to ClearBitsSparse that it is a reference.

PrimeCSharp/solution_4/PrimeCS.cs
ItalyToast: I tried this code on my laptop (CPU is a I5 3230M) and this code ran significantly worse: ~6700 -> ~5600. Im not sure what went wrong. If I remeber correctly the normal unrolled loop didnt perform that bad. Its getting the same kind of numbers that it got when it wasnt inlined 😒.

I wont bother with it right now so I'll let the current version stay. The next step is to reverse the whole sieve algorithm so that would make this small fix not matter anyway. 

I added your code in ClearBits.ClearBitsSparseUnrolled4V2 as an alternative

ItalyToast:
The more I think about the reversed algorithm shouldnt take to long, you can probably hold of on the review @rbergen

PrimeCSharp/solution_4/PrimeCS.cs
mike-barber: That's a strange one. Might be an Intel / AMD thing, and not worth diagnosing in too much detail: my new variation isn't a huge gain.

Happy for you to remove all of the stuff you're not using: we've got the code on branches if we want to work on them more in the future.

ItalyToast: Okay, its removed. I made my own branch for that stuff.

rbergen (approved):
LGTM

marghidanu (approved):

PR #537, opened 2021-07-28
New x86_64 Assembly Solution

weirddan455:
## Description
This is an x86_64 assembly single bit solution.  There is both a faithful and an unfaithful version.  The only difference is that faithful calls malloc while unfaithful stores the bit array in the bss segment.  I've added a number of optimizations that result in significant performance improvement over the existing single bit assembly solutions.

The most significant optimization is implementing rolling mask (idea taken from C++ solution 1.)  In assembly, this can be done with a single "rol" instruction reducing the number of instructions needed in the inner clear bit loop.

Other optimizations include using 128 bit SIMD instructions to clear the bit array and using the C wrapper for clock_gettime (the latter has been merged into 2 of the existing solutions.)

On my system, performance results are:

weirddan455;5757;5.000453;1;algorithm=base,faithful=yes,bits=1
weirddan455;6046;5.000629;1;algorithm=base,faithful=no,bits=1

Compared to existing assembly solutions:
rbergen_x64uff_byte;5867;5.000;1;algorithm=base,faithful=no,bits=8
rbergen_x64ff_byte;5848;5.000;1;algorithm=base,faithful=yes,bits=8
rbergen_x64uff_bitbtr;2745;5.001;1;algorithm=base,faithful=no,bits=1
rbergen_x64ff_bitbtr;2738;5.000;1;algorithm=base,faithful=yes,bits=1
rbergen_x64uff_bitshift;4781;5.000;1;algorithm=base,faithful=no,bits=1
rbergen_x64ff_bitshift;4582;5.000;1;algorithm=base,faithful=yes,bits=1

As you can see, my unfaithful implementation beats all existing solutions while my faithful is slightly worse than the byte implementations while both are only using single bit.  Both have significant improvement over the single bit implementations.

If you'd prefer, I can see about merging my optimizations into the existing solutions but my initial thought is that it's different enough that it warrants a separate solution.


PrimeAssembly/solution_3/primes_ff_1bit_rol.asm
```diff
@@ -0,0 +1,234 @@
...
+mov     rdi, 4
+lea     rsi, [startingTime]
+call    clock_gettime
+test    rax, rax
+jnz     clockError
+
+mov     r12, [startingTime+time.sec]
+mov     r13, [startingTime+time.nsec]
+
+mov     rdi, 125000
```
rbergen, line 44: Sieve size is hardcoded into memory buffer size as 1.000.000 / 8.

PrimeAssembly/solution_3/primes_ff_1bit_rol.asm
```diff
@@ -0,0 +1,234 @@
...
+rol     eax, cl
+
+add     rdi, r8
+cmp     rdi, SIEVE_SIZE
+jl      clearBitLoop
+
+clearBitEnd:
+
+add     rbp, 2
+cmp     rbp, 1000
```
rbergen, line 127: Sieve size is hardcoded as square root of 1,000,000.

PrimeAssembly/solution_3/primes_ff_1bit_rol.asm
```diff
@@ -0,0 +1,234 @@
...
+lea     rsi, [startingTime]
+call    clock_gettime
+test    rax, rax
+jnz     clockError
+
+mov     r12, [startingTime+time.sec]
+mov     r13, [startingTime+time.nsec]
+
+mov     rdi, 125000
+call    malloc
```
rbergen, line 45: Memory allocation takes place once, outside of the timed loop.

PrimeAssembly/solution_3/primes_ff_1bit_rol.asm
```diff
@@ -0,0 +1,234 @@
...
+loopStart:
+
+inc     rbx
+xor     rbp, rbp
+
+resetBitArray:
+
+movdqa  [r15 + rbp], xmm0
+add     rbp, 16
+cmp     rbp, 124992
```
rbergen, line 68: Sieve size is hardcoded in bit clearing limit.

rbergen (commented):
From a faithfulness perspective, I have to conclude that the malloc implementation cannot be marked as faithful, at this time. The reasons for this are as follows:

- The sieve (memory) size is effectively hard-coded into the program's logic in a number of places, at least those pointed out in specific comments. A faithful implementation must have the sieve size passed to it as a parameter, that is used for memory allocation/release and the sieve run logic itself. In the C++ implementation that you say inspired you, this is reflected in the line where the sieve class is created:
  ```c++
  prime_sieve sieve(1000000L);
  ```

- A faithful implementation must capture the full sieve state (usually consisting of the sieve size and the array/vector/bitset/buffer/... holding the prime flags) in a "class-like" construct. 
  Although Assembly clearly does not include an intrinsic class concept, it is possible to encapsulate the sieve state; this is what the `struct sieve` in Assembly/solution_1 is for.

- In a faithful implementation, the sieve state construct needs to be created and torn down for every iteration of the sieve run pass loop. This implementation creates the buffer once, and reuses it for all main loop iterations.

From a PR structure perspective, it aims to improve the performance of an Assembly solution/implementation that has the same key characteristics as this one (faithful, base algorithm, 1 bit). In that case, you should structure your PR to improve the original solution (in this case, Assembly/solution_1).

weirddan455:
Fair enough.  I can try to merge my perf improvements into the existing solution then.  Should I just close this and put in a new PR when I get that done?

rbergen:
> Fair enough. I can try to merge my perf improvements into the existing solution then. Should I just close this and put in a new PR when I get that done?

You can either open a new PR (which is cleaner) or update this one when you're ready (which keeps the history of your proposed changes together). In either case, if you do not pursue having this PR merged in it current form, then please close it (for now).

PR #539, opened 2021-07-29
Tweaked java impl

addertheblack:
## Description

Added a Java solution 3 which is a slightly tweaked Java solution 2 that uses 1 bit per number instead of a byte (disguised as a boolean). Also contains an additional tweak that uses a pre computed set of masks for an extra 20%.

## Contributing requirements
I skimmed over CONTRIBUTING.md and I think it's got everything. 


rbergen:
Java/solution_2 is actually a 1-bit solution. If you feel you can improve it, then please open a PR that improves it. On top of that, this solution causes a merge conflict. In relation to the previous. I'm closing this.

PR #540, opened 2021-07-29
Lua performance improvement and bug fixes

greatwolf:
## Description
This PR contains some important performance improvement and bug fixes for Lua `solution_1` implementation.

I plan to improve this solution further by refactoring it into a more object-oriented design so it satisfies the `faithful` status. Will do it later as a separate pull request though.

## Contributing requirements

### DIFF
--- a/PrimeLua/solution_1/Primes.lua
+++ b/PrimeLua/solution_1/Primes.lua
@@ -45,7 +45,7 @@ function getBit(index)
 	if index % 2 == 0 then
 		return false
 	end
-	return bitArray[index/2]
+  return bitArray[math.floor(index / 2)]
 end
 
 function clearBit(index)
@@ -69,7 +69,7 @@ function runSieve()
 		for num = factor, sieveSize do
 			if getBit(num) then
 				factor = num
-				--continue
+        break
 			end
 		end
 
--- a/PrimeLua/solution_1/Primes.lua
+++ b/PrimeLua/solution_1/Primes.lua
@@ -37,10 +37,9 @@ function countPrimes()
 end
 
 function validateResults()
-	if myDict["k" .. tostring(sieveSize)] then
-		return myDict["k" .. tostring(sieveSize)] == countPrimes()
-	end
-	return nil
+  local k = 'k' .. sieveSize
+
+  return myDict[k] == countPrimes()
 end
 
 function getBit(index)
--- a/PrimeLua/solution_1/Primes.lua
+++ b/PrimeLua/solution_1/Primes.lua
@@ -15,11 +15,13 @@ local myDict =
 	k100000000 = 5761455 
 }
 
+local floor, sqrt = math.floor, math.sqrt
+
 function setup(size)
 	sieveSize = size
 	bitArray = {}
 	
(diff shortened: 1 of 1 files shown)


rbergen:
Please undo all whitespacing changes. The diff should only contain the actual changes to the logic.

greatwolf:
Redid the branch and omitted the commit with the whitespace changes.

rbergen:
Could you provide me with some numbers on the performance improvements you're seeing with this? I admit I don't have that much hands-on experience with Lua, but the changes you're proposing basically look cosmetic to me. I will check the numbers you provide on my machine, but with your numbers and mine we have at least two data points.

greatwolf:
Sure thing. With the old code I get:

    lua53.exe Primes.lua  10000
    Passes: 49, Time: 5, Avg: 0.10204081632653, Limit: 10000, Count: 1229, Valid: true

    lua;49;5;1;algorithm=base,faithful=no,bits=64

After the changes:

    lua53.exe Primes.lua  10000
    Passes: 3298, Time: 5, Avg: 0.001516070345664, Limit: 10000, Count: 1229, Valid: true

    lua;3298;5;1;algorithm=base,faithful=no,bits=64

When run with drag-race default of 1000000, problem is even more pronounce:

    -- 1 million iterations
    lua53.exe Primes.lua
    Passes: 1, Time: 96, Avg: 96.0, Limit: 1000000, Count: 78498, Valid: true

    lua;1;96;1;algorithm=base,faithful=no,bits=64

That `96` seconds is not a fluke. It takes that long just to run a single pass and the timing code can't stop execution in the middle of a pass so it has to wait till completion even if it exceeds `> 5` seconds.

After the fixup:

    lua53.exe Primes.lua
    Passes: 22, Time: 5, Avg: 0.22727272727273, Limit: 1000000, Count: 78498, Valid: true

    lua;22;5;1;algorithm=base,faithful=no,bits=64

greatwolf:
Let me know if you need me to rebase first.

rbergen:
> Let me know if you need me to rebase first.

No, that won't be necessary at this time.

What you're basically saying then is that the performance improvement is achieved by making a couple of functions local?

greatwolf:
Making things `local` in lua helps performance in general since otherwise they're global by default and going through a table adds an extra indirection. You can confirm this by inspecting the lua bytecode generated with `luac.exe`.

In this particular case the performance problem is from improper implementation of the sieving algorithm, (maybe during translation by the original author from another language?) which is causing the quadratic behavior. See 
744a8a6.

Were you able to reproduce the results btw?

rbergen:
Yes, I have just now. In my case it was 99 seconds for 1 pass without your changes and 21 passes in 5 seconds with them, which is pretty close to your findings. It's clearly an impressive improvement.

In any case, I somehow overlooked the missing break in the loop; I now realize there's a commented-out `continue` in the original source. Which is an interesting concept in itself, because I understand that Lua doesn't have a `continue` statement. Maybe my brain conceptually converted the `--continue` into a `break` or something. Who knows.

With that said, I'm going to merge this.

PR #541, opened 2021-07-29
Further misc implementations for Dlang solution_2

BradleyChatha:
## Description

Adds six new solutions:

* SieveRT_8 - An 8-bit version of SieveRT. (Faithful)
* SieveRT_16 - A 16-bit version of SieveRT. (Faithful)
* SieveRT_32 - A 32-bit version of SieveRT. (Faithful)
* SieveRT_64 - A 64-bit version of SieveRT. (Faithful)
* SieveRTCT_Cheatiness - A cheaty solution that generates the entire sieve at compile-time (C++ has it, so I thought D should as well) (unfaithful)
* SieveRT_LookupTable - Uses a compile-time generated lookup table. (unfaithful)

Adds thread execution modes, alongside a new one called `staticThreads` which is a speed increase.

On the side there's also mdgen.d which generates some information for the README.

## Contributing requirements

* [x] I read the contribution guidelines in CONTRIBUTING.md.
* [x] I placed my solution in the correct solution folder.
* [x] I added a README.md with the right badge(s).
* [x] I added a Dockerfile that builds and runs my solution.
* [x] I selected `drag-race` as the target branch.
* [x] All code herein is licensed compatible with BSD-3.

PR #542, opened 2021-07-29
D Solution 2 additions

BradleyChatha:
## Description
My apologies.

#541 

## Contributing requirements

### DIFF
--- a/PrimeD/solution_2/Dockerfile
+++ b/PrimeD/solution_2/Dockerfile
@@ -4,4 +4,4 @@ WORKDIR /app
 RUN apk update \ 
     && apk add curl bash xz build-base ldc dub libgcc --no-cache \
     && dub build -b release --compiler=ldc2
-ENTRYPOINT [ "./solution_2" ]
+ENTRYPOINT [ "./solution_2", "-m=leaderboard" ]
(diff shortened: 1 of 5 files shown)


rbergen:
- Please limit your algorithm qualifications to "base", "wheel" and "other". 
- The set of implementations in this solution now looks like the full range of permutations of a number of implementation variations you have implemented. I'd like to ask you to please consider "cherry picking" a subset of them, being those that illustrate something that is specifically interesting, individually or as (sub)groups. Of course, "interesting" should primarily be considered from a perspective of performance.

BradleyChatha:
Added a --mode argument.

The Dockerfile defaults to using `-m leaderboard` which runs the runtime 1-bit, runtime 8-bit, and compile time impls (even though it's unfaithful). These are ran in single threaded and static threaded modes, so 3 single threaded and 3 multithreaded impls.

Hope that's ok now :)

rbergen:
> Hope that's ok now :)

Yes, and I can say that I think this is actually a very elegant way to implement this. Thank you for it!

rbergen (approved):
LGTM

marghidanu (approved):

PR #543, opened 2021-07-29
First Visual Basic .NET solution

DoctorDalek1963:
## Description
An unfaithful solution in Visual Basic .NET. Only unfaithful because I didn't use a class. This solution uses the base algorithm and a BitArray for storing the primes. I believe that makes this solution 1-bit, although I'm not sure how BitArray is implemented in VB.NET, so I can't be sure.

## Contributing requirements

### DIFF
--- a/PrimeVisualBasic/solution_1/Dockerfile
+++ b/PrimeVisualBasic/solution_1/Dockerfile
@@ -1,0 +1,9 @@
+FROM ubuntu:20.04
+
+RUN apt-get update && apt-get install -y mono-vbnc
+
+COPY Primes.vb /opt
+WORKDIR /opt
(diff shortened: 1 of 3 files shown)


rbergen:
PrimeBASIC/solution_2 and PrimeBASIC/solution_3 are both Visual Basic solutions. The first of those is a faithful, base algorithm implementation using a BitSet for storage. 

As such, I'm inclined to think this PR will not be able to serve the purpose it was opened for, and should therefore be closed.

DoctorDalek1963:
Thanks for pointing that out. I didn't bother to check PrimeBASIC because I assumed that VB was different enough to have its own directory.

PR #544, opened 2021-07-29
Some improvements in Go solution_2

ssovest:
## Description
For some reason, moving the bit reset logic from the sieve loop to a method of a custom type gave a huge speed boost in sieve_other.go.
But now the sieve loop looks almost identically to that of Python [solution_2](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/PrimePython/solution_2/PrimePY.py#L40) and some other solutions implementing the base algorithm. And the Bitarray type doesn't know anything about primes, it just sets every n-th bit in a given range. But the logic in the Bitarray was ported from the [PrimeCY_32](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/PrimeCython/solution_1/PrimeCY_32.pyx#L63) implementation which is marked as implementing "other" algorithm. So now i'm not sure how to classify this implementation. Should i mark it as "base" (and remove other implementations in the solution folder, as they are far less performant) or leave it "other"?

## Contributing requirements

### DIFF
--- a/PrimeGo/solution_2/sieve_other.go
+++ b/PrimeGo/solution_2/sieve_other.go
@@ -3,7 +3,6 @@ package main
 import (
 	"flag"
 	"fmt"
-	"math"
 	"math/bits"
 	"time"
 	"unsafe"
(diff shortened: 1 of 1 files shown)


rbergen:
I think that in this case, the key thing for the `base` vs `other` determination is whether the implementation uses calculated masks that it projects onto the sieve. A solution that does this is an `other` implementation. That's what I think I see happening in this implementation and in PrimeCY_32, but not in Python/solution_2. 
This is explained in different phrasing in [this answer to a Q&A topic](https://github.com/PlummersSoftwareLLC/Primes/discussions/454#discussioncomment-1018644). 

Which implementations you include in your solution and which you remove is largely up to you. I only urge people to remove implementations when I feel that, conceptually speaking, they are duplications of existing implementations in the same language.

ssovest:
Hmm, so should this rule be read as "the algorithm makes exactly one write operation to the underlying buffer per cleared bit"? If so, then this implementation is indeed not a `base` one and i won't be commiting any other changes to this PR.

By the way, doesn't this reading of the rule mean that if some programming language would happen to have a bitarray/bitvector/bitset class and a method/function for setting every n-th element in it, then the author of a solution using it will need to research the language's source code, inspect the bytecode or disassemble the resulting binaries to prove it's not using a similar set of optimizations?

rbergen:
> Hmm, so should this rule be read as "the algorithm makes exactly one write operation to the underlying buffer per cleared bit"?

I would say it makes one indexing/write operation per cleared bit (in)to something. The main thing behind that is preventing reuse of calculated bitmask, as that is a little close to `wheel` algorithm implementation.

> By the way, doesn't this reading of the rule mean that if some programming language would happen to have a bitarray/bitvector/bitset class and a method/function for setting every n-th element in it, then the author of a solution using it will need to research the language's source code, inspect the bytecode or disassemble the resulting binaries to prove it's not using a similar set of optimizations?

No, it doesn't mean that. Just like we don't demand that people inspect the guts of their language or platform to discover if the allocation mechanism it provides reuses buffers under the hood. Which of course, most of them do.

As I've said before, I accept there are gray areas and edge cases, but that's always the case no matter where one draws the line. We'll have to keep taking those one case at a time.

ssovest:
Ok, i think i get it. Just one more question. Does a bit resetting function like this conform to the `base` definition?
```go
func (b Bitarray) SetSliceTrue(start, stop, step uintptr) {
	var index, end uintptr
	var mask uint64
	begin := unsafe.Pointer(&b[0])
	end = (stop + 63) / 64
	for i := 0; i < 64; i++ {
		mask = bits.RotateLeft64(1, int(start))
		index = start / 64
		start += step
		for ; index < end; index += step {
			*(*uint64)(unsafe.Pointer(uintptr(begin) + index*8)) |= mask
		}
	}
}
```
The mask can still be called "precalculated", but there is exactly one write operation per bit. It does provide a slight performance bonus even without using one mask to clear several bits, so i'd want to update my other implementations later, if such a function can be used in `base`.

rbergen:
There needs to be a loop that runs over the sieve, to clear the multiples of the factor, one bit at a time; this is the "second operation" in the discussion of the base algorithm in the contributing guidelines. 
In that loop, the indexing and bit setting needs to be performed for each of those operations. What is allowed within the clearing loop is using a rolling single bit mask, where the mask is rolled for every iteration of the loop. What's not allowed is enveloping the actual clearing loop in another one, that calculates a bitmap and projects it onto the sieve. The latter is what I believe I see in your last code fragment.

ssovest:
> What's not allowed is enveloping the actual clearing loop in another one, that calculates a bitmap and projects it onto the sieve.

Isn't exactly this optimization being used in Rust [solution_1](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/PrimeRust/solution_1/src/main.rs#L269)? (And, possibly, in some others as well: at least Nim solution_3 states it's using it)

rbergen:
> Isn't exactly this optimization being used in Rust [solution_1](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/PrimeRust/solution_1/src/main.rs#L269)? (And, possibly, in some others as well: at least Nim solution_3 states it's using it)

The key thing in this discussion is mask projection. Rust/solution_1 does distribute the bits across bytes "at 90 degrees" over the sieve's memory buffer, but does the required indexing for each bit individually. In my opinion, that falls within the boundaries of indexing and writing taking place for each individual sieve item that needs to be cleared.

Maybe the issue is that I don't fully understand the similarity between that approach and yours. Maybe you can talk me through your approach to clear up the confusion at my end?

ssovest:
Well, if by required indexing you mean that on every iteration the index of the single cleared bit is known, then it's `index*64 + start%64`. The main difference with the Rust solution is that here bit positions are inrepreted in a "classic" way rather than "90 degrees rotated", so the logical order of the bits cleared differs but physically the same process happens in both solutions.

rbergen:
My main concern is this section of the code (line 40 to 44):
```go
        for next == index {
		mask |= bits.RotateLeft64(1, int(start))
		start += step
		next = start / 64
	}
```

It looks to me like this code sets multiple bits in the mask at once, if the step is smaller than 64. If I read that right, that would mean that there is a pre-calculated mask covering "multiple multiples" for smaller steps, that is projected on the sieve in the sense that multiple bits are set with a single indexing and write operation.

ssovest:
Oh, you're referring to the code in the PR. Of course it writes multiple bits at once, this is the most fruitful optimization here (Which makes it `other`, which, among other things, mean this PR doesn't require any other changes in other files and can be considered final). But i was speaking about the code snippet i've posted in one of my comments above. Here it is:

```go
func (b Bitarray) SetSliceTrue(start, stop, step uintptr) {
	var index, end uintptr
	var mask uint64
	begin := unsafe.Pointer(&b[0])
	end = (stop + 63) / 64
	for i := 0; i < 64; i++ {
		mask = bits.RotateLeft64(1, int(start))
		index = start / 64
		start += step
		for ; index < end; index += step {
			*(*uint64)(unsafe.Pointer(uintptr(begin) + index*8)) |= mask
		}
	}
}
```

And you answered:

> What's not allowed is enveloping the actual clearing loop in another one, that calculates a bitmap and projects it onto the sieve.

Which is exactly what striped storage in Rust solution_1 does: calculates a mask in the outer loop, then applies it in the inner loop (well, in 2 inner loops). So i just wanted to clarify if it's not allowed at all, allowed always or allowed in some cases.

rbergen:
I'll take ownership of the confusion about which code we were discussing. My apologies for it. 

Concerning what is and what's not allowed in the "second operation": the key requirement in the context of this discussion is that the indexing and bit-flipping is done individually for each multiple of the factor. In the case of the striping approach, a single bit is indeed selected in the loop enveloping the actual clearing loop, but it does not clear multiple multiples in one go through mask projection.

PR #546, opened 2021-07-30
Lazy wheel

nmerriam:
## Description
This is a wheel algorithm but without a statically pre-computed prime set. Rather, the wheel is computed dynamically (lazily) after a concept by Colin Runciman, see https://eprints.whiterose.ac.uk/3784/1/runcimanc1.pdf

## Contributing requirements
As far as I can tell, this algorithm is significantly different from any current implementation.

### DIFF
--- a/PrimeC/solution_4/Dockerfile
+++ b/PrimeC/solution_4/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu
+FROM ubuntu:20.04
 
 # Get GCC
 RUN apt-get update
--- a/PrimeC/solution_4/Dockerfile
+++ b/PrimeC/solution_4/Dockerfile
@@ -1,0 +1,12 @@
+FROM ubuntu
+
+# Get GCC
+RUN apt-get update
+RUN apt-get install -y build-essential
+
+WORKDIR /home/primes
+
+COPY sieve.c prime-check.h run ./
+RUN chmod a+x run
+
+ENTRYPOINT ["./run"]
--- a/PrimeC/solution_4/README.md
+++ b/PrimeC/solution_4/README.md
@@ -1,0 +1,34 @@
+# C solution by Nick Merriam and GLIWA team
+
+![Algorithm](https://img.shields.io/badge/Algorithm-wheel-yellowgreen)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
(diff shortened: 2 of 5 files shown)


PrimeC/solution_4/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ubuntu
```
rbergen: Please pin the base image down to a specific version.

nmerriam:
Hi Rutger,

Thank you for getting in touch. I am afraid I have never done anything with
Docker or GitHub before and I don't know what you mean. I got a friend to
coach me through the steps as far as here and I don't really understand
what's going on.

I have put Hans on cc. Maybe you can straighten this out?

Sorry to be so helpless. One of my motivations for getting involved was to
learn more.

Best wishes
Nick

rbergen:
@nmerriam That's ok, it's one of the learning curves everybody needs to go through.

Your Dockerfile indicates it wants to use an Ubuntu docker image as a base image with this directive:
```dockerfile
FROM ubuntu
```

What's needed is that you make explicit what version of Ubuntu base image you want to use. We prefer you use a Long Time Support image like 20.04, which would be specified like this:
```dockerfile
FROM ubuntu:20.04
```

If there is a valid reason to choose another Ubuntu version than 20.04 then that is also fine, as long as a specific version is indeed selected.

nmerriam:
Hi Rutger,

Can I ask you to add the Ubuntu version? To be honest, I just copied the
Docker file from another solution. I used WSL with Ubuntu 20.04, so I'm
sure that will be fine.

I would do it myself but I'm visiting friends and don't have access to
anything more than my smartphone and Hans is on the wrong end of a time
difference that means it's his night.

Many thanks
Nick

rbergen:
I'm afraid that I can't do that. The correct Dockerfile needs to be part of the PR, and reviewed and merged as such. 
I think we'll just have to wait until it is possible for the Dockerfile in your PR to be updated so that it passes our CI.

nmerriam:
Added missing ubuntu version to Dockerfile

nmerriam:
@rbergen this looks better. Thank you for your guidance. Are you able to review and merge now?

rbergen:
Thank you, with this I am able to review.

rbergen (approved):

marghidanu (approved):

PR #547, opened 2021-07-30
Primepascal olivierbrun solution 4

olivierbrun:
## Description

New approach using a different mechanism capable of truly handling any 64 bits sieve size.

The mechanism proposed in this implementation improves the solution I proposed in solution 3 and brings the following benefits:

- it passes the test with a sieve size of 10.000.000.000 (including on a Raspberry PI) or any range up to 2^64 - 1 (18.446.744.073.709.551.615), as long as the machine has sufficient memory (memory requirement for each thread to store the bits is sieveSize/16 as bits are used for odd numbers only)
- performance is significantly better on linux machines (about 33% in average) than with solution 3
- this solution is using an array of uint32 to store the bits (instead of a bitpacked array as proposed in solution 3, which was limiting the sieve size to 2^32 - 1), each bit is used for odd numbers only, which reduces the memory footpring by 2

I've inlined the GetBit and ClearBits functions, providing some boost during the execution.
I've noticed that using GetMem to allocate memory on the heap for the bits array, provides better performances than using a setlength instruction.

## Contributing requirements

The mechanism proposed in this implementation is different from the ones proposed in solutions 1 to 3 and provides a way to truly handle any 64 bits sieve size, with the benefit of adding a bit of performance boost (33% better on linux machines).


olivierbrun:
The pull request does not contains the commits I intended to put in it, so I'm closing it

PR #549, opened 2021-07-30
APL solution

Garklein:
## Contributing requirements


rbergen:
Please:
- Squash your commits
- Fix the Dockerfile lint errors found by CI (visible under "Files changed")

Garklein:
I hope I fixed the Dockerfile lint errors.  
I squashed as much as I could, I was getting an error since I had a merge.

rbergen:
Thank you.

Your solution indicates it uses one bit per prime candidate. Could you maybe talk me through how this is setup? I am rather unfamiliar with APL, but/and I don't see anything that looks like the "bit twiddling" that is required in many (if not most) other languages that don't use a specific bit-level container.

Garklein:
Oops, this was a bit array in an earlier version, but I changed some operators and it was actually using an 8-bit signed integer array. I'll fix that. Anyways, with some operators, if it is creating an array of just 0s and 1s it will convert it to a bit array (this is also implementation specific). You can then use all the normal array operations on the array and as long as it only has 0s and 1s it will stay a bit array.

Garklein:
Never mind, I logged the array type and it is all a bit array.

PrimeAPL/solution_1/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM dyalog/dyalog:18.0
```
rbergen, line 1: I'm concerned about the use of a commercial product as the foundation for any solution in this project. I wonder if it would be possible to use a free software version of APL (like GNU APL) instead?

Garklein: My impression is that Dyalog is the main APL used. I've never tried GNU APL (I'm on Windows and the setup is a pain because I'm not familiar with Linux at all).

rbergen: I understand your point about Dyalog and I would absolutely love to add APL to the collection of languages. However, after looking at the licensing information that's available on the Dyalog website, I don't think we can incorporate it in this project, even if it is only through the use of a docker image.

Garklein: Dyalog is free for non-commercial use. Is using it through a docker image breaking one of the distribution license rules?

rbergen: There is mention of run-time licenses, and whoever clones this repo and runs the benchmark would be running Dyalog via the docker image. In any case, the actual answer to your question is:

- I don't know, and we don't have a lawyer at hand to get a reliable answer
- We can't accept Dyalog's license terms on behalf of all the (potential) users of this repo's contents. We also can't enforce (or even facilitate) those users accepting the license terms in question before they run the Dyalog docker image after they clone. Nor do we want to. 

In my view, to be able to use the Dyalog docker image, Dyalog would have to provide a blanket lift of their licensing requirements for all code in this repo _and all its derivatives_. I don't expect them to do this, and I for one don't even have the time to ask them if they would.

Garklein: Ok.

rbergen:
Would it be possible to keep your fork available for some time? When I have the time, I'd like to see if I can get your solution to work with a free software version of APL. If I can get that to work, I'll open a PR on your fork so you can see if you're happy with what I've put together.

What's behind this is that I really would like to add APL to the language collection.

Garklein:
Yep, I'll just leave it as it is.  
When using another version of APL, you'll have to change the timings. Also, I don't believe GNU APL supports classes. You will also probably have to change the main file (it probably has a different way to load other files).

dzaima:
Note that GNU APL is ~10-100x slower than Dyalog APL, which makes it much less suited for a performance competition. I don't think there are any free APL implementations that really focus on speed.

rbergen:
@dzaima I understand that. Unfortunately, that doesn't change what we can and cannot responsibly add to this project. 

In any case, looking at @Garklein's last comment, it very much looks like getting their APL submission to work on GNU APL requires way more knowledge of (GNU's) APL (implementation) than I could ever dream of having. I sadly think that means that APL will not be part of the collection of languages until someone else submits an APL solution based on GNU APL, if and when that happens.

rbergen:
@Garklein As discussed in the context of #722, it is now possible to add this solution, provided that the automated build/benchmark is disabled for it. This is described in [the relevant section of the contributing guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md#disabling-build-and-benchmark) and concerns a CI/benchmark capability that we added after you closed this PR. 

You can reopen your PR and add the `build-no` flag file if you would still like your solution to be added to the project.

Garklein:
Thanks for the heads up, but since Arcfide's solution is so much better, I won't open my PR.

rbergen:
That's entirely your prerogative, of course. Thanks for letting us know!

PR #551, opened 2021-07-30
Pascal solution 4

olivierbrun:
## Description

This solution implements a multithreaded version written in freepascal (Object Pascal) using an array of 32 bits values to store flags for prime numbers.  
As an array index is 32 bits max in FreePascal, with this approach the addressable bit range on an array storing 32 bits values is actually 2^64.
Bits are used to flag odd numbers only, reducing the memory footprint by half.

Benefits from this approach:

- it passes the test with a sieve size of 10.000.000.000 (including on a Raspberry PI) or any range up to 2^64 - 1 (18.446.744.073.709.551.615), as long as the machine has sufficient memory space,
- performance is significantly better on linux machines (about 33% in average) compared to what I was obtaining with solution 3. However, no performance change on Windows machines.
- I've optimized and inlined the GetBit and ClearBits functions to provide some performance boost.

(I've noticed that using GetMem to allocate memory on the heap for the array, provides better performances than using a setlength instruction in the constructor).

### Solution:

The following 3 simple principles are what has made this solution capable of truly handling 64 bits sieve sizes: 
- bits are stored in an array of 32 bits values, each bit stores a flag for an odd number.
- the Getbit(n) function retrieves a bit value using (pseudo code) ```bits[n >> 6] & (1 << ((n >> 1) & 31)))```, where  
index=n >> 6 (divide by 64 = 2 * 32 as only odd numbers are flagged, and each array element is 32 bits), the bit for the ```&``` operation is shifted left by the modulo 32 of n/2
- the ClearBits(n, skip) method loops from i=n/2 to sizeSize/2 and clears each bit using (pseudo code) ```bits[i>>5]&=~(1<<(i & 31))``` followed by ```i+=skip/2``` to take into account the fact that the bits array stores flags for odd numbers only.
## Contributing requirements


olivierbrun:
cancelling this PR as it does not contain only the latest 2 commits (f312a79 and c1bdcd0)

PR #556, opened 2021-07-31
Performance improvements to PrimeAssembly/solution_1/primes_ff_bitshift.asm

weirddan455:
I've moved my performance optimizations from PR #537 into the existing solution as discussed in the now closed original PR.  I'm seeing about the same performance here as in my separate solution.

Without optimizations:  rbergen_x64ff_bitshift;4582;5.000;1;algorithm=base,faithful=yes,bits=1

With optimizations:  rbergen_x64ff_bitshift;5767;5.000;1;algorithm=base,faithful=yes,bits=1

The two optimizations implemented are:

- Use the 128bit SIMD movdqa in initLoop (small perf boost, but measurable)
- Use a rolling mask, similar to C++ solution 1 (large perf boost)

### DIFF
--- a/PrimeAssembly/solution_1/primes_ff_bitshift.asm
+++ b/PrimeAssembly/solution_1/primes_ff_bitshift.asm
@@ -17,7 +17,7 @@ struc time
 endstruc
 
 struc sieve
-    .bitSize: resd    1
+    .sieveSize: resd    1
     .primes:    resq    1
 endstruc
 
@@ -200,8 +200,7 @@ newSieve:
     mov         rdi, sieve_size             ; ask for sieve_size bytes
     call        malloc wrt ..plt            ; rax = &sieve
 
-    inc         r12d                        ; array_size = sieve limit + 1
-    mov         [rax+sieve.bitSize], r12d   ; sieve.bitSize = array_size
+    mov         [rax+sieve.sieveSize], r12d ; store sieve size parameter for faithfulness
 
 ; registers:
 ; * rax = primesPtr (&sieve.primes[0])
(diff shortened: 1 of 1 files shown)


rbergen:
Thank you for updating your earlier PR to an update to the existing implementation. I look forward to merging an improvement to my layman's implementation, soon. :)

Before we can do that, there is at least one thing that will need changing. In the sieve run logic, you rely on the `SIEVE_SIZE` constant, which goes against one of the requirements of a faithful implementation, as I explained in a comment on #537. 

As I see it, if you need both the `bitSize` and sieve size when you run the sieve, then you can either add the sieve size to the `sieve` struct when the struct is created in `newSieve`, or store either of them and calculate the other when you need it.

weirddan455:
Updated.  I really only need sieveSize for this implementation.  Previous implementation was only allocating enough memory to store odd numbers.  Here, I'm allocating sieveSize (in bits) rounded up to a 16 byte multiple (which is calculated at run time in newSieve.)  After the memory is allocated, we only really care about what the sieveSize is (bytes after that are just filler so that it plays well with SIMD.)

As such, I've just renamed bitSize to sieveSize and then retrieve that value at run time in runSieve.  This hurt performance a bit but it's still much faster than the original implementation.  Now I'm getting around:

rbergen_x64ff_bitshift;5492;5.000;1;algorithm=base,faithful=yes,bits=1

PrimeAssembly/solution_1/primes_ff_bitshift.asm
```diff
@@ -331,16 +343,16 @@ countPrimes:
...
 countLoop:
     bt          [rbx], ecx                  ; if !sieve.primes[0][bitIndex]...
     jnc         nextItem                    ; ...move on to next array member
     inc         eax                         ; ...else primeCount++
 
 nextItem:
-    inc         ecx                         ; bitIndex++
-    cmp         ecx, [rdi+sieve.bitSize]    ; if bitIndex < sieve.bitSize...
+    add         ecx, 2                      ; bitIndex += 2
+    cmp         ecx, SIEVE_SIZE             ; if bitIndex < SIEVE_SIZE
```
rbergen: This seems to be the last direct reference to SIEVE_SIZE in the implementation's logic. The good news is that replacing this with a reference to `sieve.sieveSize` won't impact the sieve run performance. :)

weirddan455: Fixed

rbergen:
Thanks again for this improvement to my implementation, it looks good to me now!

PR #557, opened 2021-08-01
GordonBGood FSharp "Loop Unpeeling" solution_4...

GordonBGood:
## Description
An F# "unpeeling" implementation that is almost exactly the same as to structure of the code as the Rust "striped" implementation, but does not implement a separate storage class as the details are so simple.

## Contributing requirements


rbergen (approved):
LGTM

marghidanu (approved):

PR #559, opened 2021-08-01
Haskell "Loop Unpeeling" solution_2...

GordonBGood:
## Description
Add a new solution to implement "Loop Unpeeling" which is like the Rust implementation of "striped" marking of composite number representations.

## Contributing requirements


rbergen (approved):
LGTM

marghidanu (approved):

PR #259, opened 2021-07-03
Add Cython Implementation

rpkak:
## Description
I copied the second implementation of python and changed it to a cython implementation.

View: [Cython](https://cython.org/)

## Contributing requirements

* [x] I read the contribution guidelines in CONTRIBUTING.md.
* [x] I placed my solution in the correct solution folder.
* [x] I added a README.md with the right badge(s).
* [x] I added a Dockerfile that builds and runs my solution.
* [x] I selected `drag-race` as the target branch.
* [x] All code herein is licensed compatible with BSD-3.

PR #564, opened 2021-08-01
Cython compile flags

ssovest:
## Description
Added -march=native and -mtune=native flags for Cython solution_1, it seems a little faster after that

## Contributing requirements

### DIFF
--- a/PrimeCython/solution_1/buildall.sh
+++ b/PrimeCython/solution_1/buildall.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-CFLAGS="-march=native -mtune=native `python3-config --embed --includes`"
+CFLAGS="-march=native `python3-config --embed --includes`"
 LDFLAGS=`python3-config --embed --ldflags`
 CC=gcc
 
--- a/PrimeCython/solution_1/buildall.sh
+++ b/PrimeCython/solution_1/buildall.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-CFLAGS=`python3-config --embed --includes`
+CFLAGS="-march=native -mtune=native `python3-config --embed --includes`"
 LDFLAGS=`python3-config --embed --ldflags`
 CC=gcc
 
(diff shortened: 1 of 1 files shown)


rbergen:
Tagging @rpkak and allowing time for them to review and respond to this PR on their original implementation, if they want to do so.

rpkak:
In my implementation there only had been one solution and the `buildall.sh` file didn't even existed (View #259). But is was completely overwritten in #356, so better tag @tjol.

rbergen:
Thanks @rpkak for coming back with this, so quickly. 

@tjol Are you then ok with merging this?

tjol:
If this leads to a performance benefit, absolutely, works for me.

Just one note: from my understanding, `-mtune=native` shouldn't make a difference when `-march=native` is set. Please double-check whether `-march=native -mtune=native` is faster than `-march=native`, and if it's not maybe remove the superfluous option.

You could also check whether adding `-Ofast` after the `python3-config` call makes a difference. (I doubt it)

ssovest:
@tjol thanks for pointing that out! You're right, `-mtune` flag doesn't make any difference at all. But, oddly, now i see that your implementation suffers a minor performance hit with `-march=native` (with or withour `-mtune`) on my machine... But i haven't had a chance to test it on other hardware, so maybe we should see how it works in a couple of "official" benchmark runs, what do you think?

tjol:
Hard to say. I noticed that `-march=native` slowed down Fortran solution 2, but I'm not 100% sure that wasn't a fluke. It certainly didn't help for my carefully hand-optimised Fortran code, and it might have done a little damage. Could be the same here.

I don't know nearly enough about amd64 extensions to understand what the compiler might is attempting and why it might be counterproductive.

ssovest:
Well, as far as i understand it, the benchmark runs on a daily basis, so in case of any performance hit i could just remove the flag.

tjol:
I've tested it out on my machine with docker, and for `rpkak+bit-array` I'm seeing a significant performance hit from the change. For `rpkak+byte-array`, any change is within the margin of error.

If that matches up with what you're seeing, I'd recommend changing the build script such that `ssovest-cy` gets `-march=native`, and `rpkak+bit-array` doesn't.

ssovest:
Yes, this is close to what i see on my machine. It looks like, unfortunately, my implementation is the only one which benefits from the `-march=native`. I've edited the buildall.sh accordingly.

tjol:
Excellent. Looks good to me!

PR #565, opened 2021-08-02
amd64 handwritten 8-bit faithful

dacvs:
## Description

I am submitting an 8-bit faithful implementation of the base algorithm for the Prime Sieve Drag Race. The language of my submission is amd64 (machine code).

I do not see any submissions written in machine code already, so this is new. This submission is not written in assembly. No assembler is needed. Rather, the source file is a heavily commented sequence of numbers in a text format. The binary is built by "undumping" it with command xxd.

This method of programming is described in my YouTube series "Handmade Linux x86 executables" (https://www.youtube.com/playlist?list=PLZCIHSjpQ12woLj0sjsnqDH8yVuXwTy3p). The series involves a compiler (DMQ compiler), but no DMQ compiler is used in the present submission.

I believe this submission complies faithfully with the rules for the base algorithm. Please let me know if I missed something. The binary should work on most (all?) amd64 Linux machines. I have tried it only on my machine, a Skylake laptop, where the 8-bit program does around 1300 passes per second.

## Contributing requirements

I constructed a Dockerfile, but I am not sure yet that it works.


rbergen:
Thank you for this contribution. 

There are a few things I'd like to note in the first instance:

- Your PR currently includes a binary file. Please remove it, as we don't accept binaries in the repo. 

- The docker build currently fails with the following message:
  ```
  > [5/5] RUN ./make.sh:
  ./make.sh: not found
  ------
  Dockerfile:5
  --------------------
     3 |     COPY *.sh *.dmp ./
     4 |     RUN chmod +x *.sh
     5 | >>> RUN ./make.sh
     6 |     ENTRYPOINT [ "./run.sh" ]
     7 |     
  --------------------
  ```

- I'll assume for now that your implementation does indeed implement the base algorithm; I'll review this later. However, your solution is also marked as `faithful`, which comes with its own specific set of requirements concerning the implementation, as detailed in [the respective section of the contributing guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md#faithfulness). Please review these requirements to check if your implementation indeed complies with them.

Macleykun:
Happen to see this issue, and out of curiousity i saw there was a docker issue mentioned.
It seems that dacvs uses the alpine image, but uses in his make file #!/bin/bash. Alpine images don't have bash by default unless it's installed.

The solution to the problem is to use the latest alpine:3.14 image and using the correct shebang. I assume sh is preferd over bash.

More so the run.sh file is used an entrypoint but only a run8.sh is provided. I suggest to change the entrypoint to run8.sh and also change the shebang from the run8.sh to #!/bin/sh aswell.

More so, the run8.sh would then run dacvs8 which isn't copied by default. I think to keep things simple using the COPY . . will just copy all the files from the solution to the workfolder.

I believed that multiple RUN command's should be summarized into one RUN command by using "\" to improve building/caching?

A pull request is on it's way to update dacvs fork, so that this pull request will work :).

dacvs:
Thank you, Rutger!
I have removed the binary. I will reissue the pull request, including some
fixes from Macleykun. Docker and GitHub are new to me. I don't have a
working Docker system yet. Sorry for the errors. I added to README.md an
explanation of how the program complies with the rules, especially
faithfulness.
David

dacvs:
Ok, I see now that the pull request is still open. I suppose there is no need for me to make another request. Now I am waiting for approval from a maintainer.

rbergen:
@dacvs I just opened a PR on your fork to set the executable bit on the scripts in your solution. Without it, the docker build and run still won't work. I will proceed with reviewing this once this has been pulled into this PR.

rbergen:
I can confirm that docker build and docker run now work. However, now that they do I can see that your solution outputs the full list of identified primes. That generates a lot of output bloat in the context of the benchmark runs. Therefore, please suppress that part of the output at least when running the application within the docker image and without parameters.

dacvs:
Thank you very much, Rutger, for trying it out. I committed the change you mention. Now there is no list of primes.

I am a bit confused, as the Rules say "Your benchmarked code returns either a list of primes or the is_prime array, containing the result of the sieve," and the Output description says that output other than the single benchmark line should go to stderr if possible. I was sending my primes list to stderr. I guess the idea is that there should be a way to see the primes list, but only if the user "opts in." I created a new script run8p.sh for this purpose. There does not appear to be a way in the GitHub web interface to set the executable bit on this new file, but this should not affect the benchmark runs.

rbergen:
> I am a bit confused, as the Rules say "Your benchmarked code returns either a list of primes or the is_prime array, containing the result of the sieve,"

I understand the confusion. What the rule is aiming to ensure is that the list of primes (possibly in the shape of an array) is either returned _or retained in the encapsulating class-like construct_ so that its prime count can be verified. I'll give some thought to rephrasing this rule to better express this.

> the Output description says that output other than the single benchmark line should go to stderr if possible.

That's true, but we have urged other contributors to also limit their non-drag race output to no more than a few lines. We considered including a guideline to "limit output to the drag-race line only", but considered that a little too restrictive. For one, the original 3 solutions that started this project also show a few additional lines of diagnostic output. 
In general, I think that sending 78498 numbers to output, whether that is stdout or stderr, is something worthy of a command-line argument. :) That's also how all other solutions that offer the option of listing all found primes, do it.

>  There does not appear to be a way in the GitHub web interface to set the executable bit on this new file, but this should not affect the benchmark runs.

No, there is no way to set this using the GitHub web interface. The fact of the matter is that git does work best in a Unix-like context, or actually more accurately on a Unix-oriented file system. I guess it's visible that the guy who initially developed git did so after first writing the Linux kernel. :)

rbergen (approved):
LGTM

marghidanu (approved):

PR #567, opened 2021-08-02
Add Brainfuck solution 1

ThatAquarel:
## Description
This is a submission of the prime sieve written in the [BrainFuck language](https://en.wikipedia.org/wiki/Brainfuck). It accomplishes this task given only 8 commands, that each consist of a character: `<>+-[],.` It's placed under the `PrimeBrainFuck/solution_1/` folder since there aren't any other implementations yet.

Also, this submission does not comply to the original prime sieve since it does not comply to the "Faithfulness" due to these reasons:
- An external C++ file is used for timing and printing the results, since no such API exist for this language.
- It runs for at least 5 seconds, but stops 48 seconds later because one single run of the sieve takes ~53 seconds.
- It does not calculate the primes up to 1,000,000, it does so until 1,000, because my implementation and the language is extremely slow.

Please let me know if I've missed some important criteria.

## Contributing requirements


PrimeBrainFuck/solution_1/PrimeBrainFuck.cpp
```diff
@@ -0,0 +1,110 @@
...
+        printf("Passes: %d, Time: %lf, Avg: %lf, Limit: %ld, Count1: %d, Valid: %d\n",
+               passes,
+               duration,
+               duration / passes,
+               sieve_size,
+               count,
+               validateResults());
+
+        printf("\n");
+        printf("aquarel;%d;%f;1;algorithm=base,faithful=no,parallel=no,bits=32\n", passes, duration);
```
rbergen: There is no need (or actually, support) for a `parallel` tag. Parallelism is deduced from the number of threads used: 1 means it's not a parallel implementation.

PrimeBrainFuck/solution_1/README.md
```diff
@@ -0,0 +1,64 @@
+# BrainFuck solution by ThatAquarel
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-no-yellowgreen)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-32-yellowgreen)
```
rbergen, line 6: Please add a deviation badge to indicate that your solution does not run a 1,000,000 entry sieve. The README.md of MIXAL/solution_1 has one that you can copy, if you want.

rbergen (commented):
Thanks for this submission. I can tell you that at least two maintainers were waiting for this language to show up in a PR. :)

ThatAquarel:
Thanks a lot for pointing that out :D

I've applied these changes in the latest commit.

rbergen:
Please address the Dockerfile lint warnings shown under "Files changed" (replace `apt install` with `apt-get install`).

rbergen (approved):
LGTM

marghidanu (approved):

PR #568, opened 2021-08-02
C solution 3 improve block size parameter

fvbakel:
## Description
C solution 3 improve block size parameter. It turns out that the current solution does not correctly determine the L1 size on different hardware. So replaced the method with a new one. In this new solution we just try a few parameters and then decide on the best value to use in the given situation.

## Contributing requirements

### DIFF
--- a/PrimeC/solution_3/README.md
+++ b/PrimeC/solution_3/README.md
@@ -39,7 +39,7 @@ The algorithm consists of the following steps:
 
 ## L1 cache optimization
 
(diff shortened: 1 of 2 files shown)


rbergen:
@fvbakel How much time would you say is spent during the actual benchmark run on establishing the "correct" block size?

fvbakel:
@rbergen 1.2 second. I test for 6 values, 0.2 second each.

PR #570, opened 2021-08-02
Docker: disable compsec (distorts benchmarks)

olliemath:
## Description

Disables seccomp when running docker.

For unknown reasons, running python under docker can be over 50% slower than native usage. This is a performance bug in docker, which has something to do with its use of seccomp. See https://pythonspeed.com/articles/docker-performance-overhead/
and the corresponding docker issue https://github.com/moby/moby/issues/41389

Practically, the performance difference is clear - locally when running `make DIRECTORY=PrimePython`, on a linux system the performance of `ssovtest`'s solution goes from **1998** passes to **4110**. It seems seccomp doesn't affect the usage of compiled languages (in particular the non-faithful numpy solution seems unaffected), but you *may* see a speed boost for some other interpreted languages with this fix.

### Other considerations

Should we disable seccomp if it's providing security? Well, if all code in this repo has been reviewed and verified, it should be absolutely fine (note all the containers are running as root user anyway, so we better hope this is the case).

## Contributing requirements

~

### DIFF
--- a/BENCHMARK.md
+++ b/BENCHMARK.md
@@ -267,7 +267,7 @@ For the installation of Docker follow the instructions as described in <https://
 Below is an example of the benchmark output.
 
 ```bash
-                                                                   Single-threaded                                                                    
+                                                                   Single-threaded
 ┌───────┬────────────────┬──────────┬───────────────────────────────────┬────────┬───────────┬─────────┬───────────┬──────────┬──────┬───────────────┐
 │ Index │ Implementation │ Solution │ Label                             │ Passes │ Duration  │ Threads │ Algorithm │ Faithful │ Bits │ Passes/Second │
 ├───────┼────────────────┼──────────┼───────────────────────────────────┼────────┼───────────┼─────────┼───────────┼──────────┼──────┼───────────────┤
@@ -282,7 +282,7 @@ etc
 │  74   │ powershell     │ 1        │ crowbar27_ps1                     │   1    │ 150.03900 │    1    │   base    │   yes    │ 1    │    0.00666    │
 └───────┴────────────────┴──────────┴───────────────────────────────────┴────────┴───────────┴─────────┴───────────┴──────────┴──────┴───────────────┘
 
-                                                                   Multi-threaded                                                                    
+                                                                   Multi-threaded
 ┌───────┬────────────────┬──────────┬───────────────────────────────────┬────────┬──────────┬─────────┬───────────┬──────────┬──────┬───────────────┐
 │ Index │ Implementation │ Solution │ Label                             │ Passes │ Duration │ Threads │ Algorithm │ Faithful │ Bits │ Passes/Second │
 ├───────┼────────────────┼──────────┼───────────────────────────────────┼────────┼──────────┼─────────┼───────────┼──────────┼──────┼───────────────┤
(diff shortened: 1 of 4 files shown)


rbergen:
@marghidanu I think you're better suited to review this PR than I am, so I'll leave it to you in first instance

marghidanu:
I would make this feature optional, to give the user the option to disable it on demand.

olliemath:
OK, I've added an `UNCONFINED` option, which I've set to `true` by default.

If you'd like it to be `false` by default, revert the final commit.

BENCHMARK.md
```diff
@@ -335,9 +335,20 @@ Make the `DIRECTORY` variable point to the directory that contains the solutions
 make DIRECTORY=PrimeCPP
 ```
 
+## Running with seccomp sandboxing
+
+For some interpreted languages (Python, Ruby, NodeJS), docker has a non-zero affect slowing on CPU-intensive code
```
tjol: `s/affect slowing on/effect slowing down/`?

Definitely effect, not affect.

rbergen:
> If you'd like it to be `false` by default, revert the final commit.

Please do make `false` the default. In line with the security by default principle, seccomp should be _enabled_ unless a user chooses to _disable_ it through deliberate action.

Makefile
```diff
@@ -4,14 +4,15 @@ SHELL := /bin/bash
...
+UNCONFINED := "true"
 
 .PHONY: all
 all: benchmark
 
 .PHONY: benchmark
 benchmark: check-env
 	@REALPATH=$$(cd "$${DIRECTORY}" && pwd); \
-	ARGS=("-d $${REALPATH}" "-f $(FORMATTER)"); \
+	ARGS=("-d $${REALPATH}" "-f $(FORMATTER)" "-u $(UNCONFINED)"); \
```
marghidanu: Please make this a flag without having to specify true/false values.

tools/src/commands/benchmark.ts
```diff
@@ -29,8 +29,12 @@ export const command = new Command('benchmark')
   .requiredOption('-d, --directory <directory>', 'Implementation directory')
   .option('-f, --formatter <type>', 'Output formatter', 'table')
   .option('-o, --output-file <file>', 'Write output to given file')
+  .option('-u, --unconfined <bool>', 'Run with seccomp:unconfined (native performance for interpreted languages)')
   .action(async (args) => {
     const directory = path.resolve(args.directory as string);
+    const unconfined = args.unconfined === 'true';
```
marghidanu: Just the presence of the flag should be enough.

tools/src/services/docker.ts
```diff
@@ -7,8 +7,13 @@ export default class DockerService {
     });
   }
 
-  public runContainer(imageName: string): string {
-    const output = child_process.execSync(`docker run --rm ${imageName}`, {
+  public runContainer(imageName: string, unconfined: boolean): string {
```
marghidanu: Rather than having `unconfined` as a single parameter please add a list of options.

Makefile
olliemath: I'm afraid I don't know enough makefile voodoo for that - my googling is not sufficient to reveal how to do
```
make something --flag
```

marghidanu: There is no need for voodoo; please look at the OUTPUT_FILE variable one line below and how it's set conditionally. ;)

olliemath: OK, I've done it that way - it's now enabled only if `UNCONFINED` is set. Of course, doing it this way means that
```
make UNCONFINED=false
```
will also turn **on** unconfined mode

marghidanu: It just needs to be set, that’s fine

marghidanu:
LGTM

marghidanu (approved):

PR #572, opened 2021-08-02
Performance improvement to Assembly/solution_1/primes_uff_bitshift

weirddan455:
This PR is for merging the same set of performance improvements in PR #556 to the unfaithful implementation.  Performance results are as follows:

rbergen_x64uff_byte;6056;5.000;1;algorithm=base,faithful=no,bits=8
rbergen_x64ff_byte;6060;5.000;1;algorithm=base,faithful=yes,bits=8
rbergen_x64uff_bitbtr;2826;5.000;1;algorithm=base,faithful=no,bits=1
rbergen_x64ff_bitbtr;2820;5.001;1;algorithm=base,faithful=yes,bits=1
rbergen_x64uff_bitshift;6065;5.000;1;algorithm=base,faithful=no,bits=1
rbergen_x64ff_bitshift;5498;5.000;1;algorithm=base,faithful=yes,bits=1

This puts this implementation performing better than any of the others, including byte implementations.  Number of runs before optimizations were around 4800 on my machine and are now consistently over 6000.

### DIFF
--- a/PrimeAssembly/solution_1/primes_uff_bitshift.asm
+++ b/PrimeAssembly/solution_1/primes_uff_bitshift.asm
@@ -18,9 +18,8 @@ section .data
 
 SIEVE_SIZE      equ     1000000             ; sieve size
 RUNTIME         equ     5                   ; target run time in seconds
-BIT_SIZE        equ     (SIEVE_SIZE+1)/2    ; prime candidate bit count
-BLOCK_COUNT     equ     (BIT_SIZE/64)+1     ; 8-byte block size
-BYTE_SIZE       equ     BLOCK_COUNT*8       ; bytes needed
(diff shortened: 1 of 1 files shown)


rbergen:
Are you planning to apply this fix to yet other implementations? If so, I'd prefer it if all of them were rolled up in one PR.

weirddan455:
I wasn't planning on modifying other implementations (and I'm not sure if it's possible without re-working the algorithm.)  bitbtr is using a CPU instruction to clear the bits so I think by definition it can't be changed to use rolling mask (which relies on bitshifts and AND'ing the mask to clear bits.)

I haven't looked at the byte implementations but I would think you won't need a mask at all if you're storing 1 byte per number (I assume you would just clear the byte to 0 with a single mov instruction.)

PR #578, opened 2021-08-03
Modified solution_2 (netrexx) to use native java types

jossRRM:
## Description

After reading the fine manual(s) I realized that the netrexx solution used more Rexx string types than were necessary. By marking the classes binary the netrexx compiler attempts to make all undeclared types native java types.

I then modified the Eratosthenes object, changing the primes array to from Rexx[] to byte[]. This improved performance from 60+ passes in five seconds to about 1400+. But then, by simply changing the sievesize from a Rexx type to an int, the performance bumped up to 4363 passes in five seconds.

Lesson learned: know the language options better before making your ignorance public.

## Contributing requirements


PrimeREXX/solution_2/PrimeNetRexx.nrx
```diff
@@ -3,32 +3,8 @@
 ** program parameter (sieveSize). The default sieveSize is 1,000,000
 */
 
-class Eratosthenes private
```
rbergen, line 6: Was there a particular reason for moving this class within the source file? It makes identifying the differences notoriously more difficult.

jossRRM: Reason: uh...because I could?

scan_004

Sorry. I moved it back to the top of the source.

rbergen: > Reason: uh...because I could?

Was this something you had to test to be sure of? :)

Anyway, thank you very much for restoring its former position.

PR #580, opened 2021-08-04
Small changes to Julia solution_1

louie-github:
## Description

This PR changes the following:
1. Use `div(limit+1,2)` instead of `floor(Int, (limit+1)/2)`
2. Truncate `q` to an Int
3. Remove the even number check on line 30 (`num % 2 == 0`)
4. Use `@inbounds` on line 30 (`if @inbounds sieve.rawbits[div(num+1,2)]`)
5. Start factor clearing loop at `factor * factor` instead of `factor * 3` (this provides the biggest speedup)
6. Redirect auxillary output to stderr

Output:
```
❯ julia PrimeSieveJulia.jl && julia new.jl
true
Passes: 6067, Time: 5.000481128692627, Avg: 0.0008242098448479688, Limit: 1000000, Count: 78498, Valid: true

dcbi;6067;5.000481128692627;1;algorithm=base,faithful=yes,bits=1
true
Passes: 6390, Time: 5.000395059585571, Avg: 0.0007825344381198077, Limit: 1000000, Count: 78498, Valid: true

dcbi;6390;5.000395059585571;1;algorithm=base,faithful=yes,bits=1
```

## Contributing requirements

### DIFF
--- a/PrimeJulia/solution_1/PrimeSieveJulia.jl
+++ b/PrimeJulia/solution_1/PrimeSieveJulia.jl
@@ -11,7 +11,7 @@ struct prime_sieve
 	sieveSize::Int
 	rawbits::BitVector # was BitArray before
 
-	prime_sieve(limit::Int) = new(  limit, trues( floor(Int,(limit+1)/2) )  )
+	prime_sieve(limit::Int) = new(  limit, trues( div(limit+1,2) )  )
 end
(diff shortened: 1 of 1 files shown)


louie-github:
I did not add `@inbounds` to the other functions since they were not being benchmarked.

PR #585, opened 2021-08-04
made CI.yml more consistent

mandar1jn:
## Description
Made the workflow more consistent. At some points quotation marks were used and at some points they weren't. I also changed the on defenition. It does exactly the same. Not defining a branch is the same as using a double wildcard (**)

## Contributing requirements


rbergen:
These are purely cosmetic changes with no change in functionality. As such, I'm closing this.

PR #588, opened 2021-08-05
C solution 3 Added striped blocks

fvbakel:
## Description
C solution 3 Added striped blocks, based on [Rust solution 1](../../PrimeRust/solution_1) by @mike-barber. It does not match that performance but still I would like to add it because this technique is not implemented in C yet.

## Contributing requirements

### DIFF
--- a/PrimeC/solution_3/.gitignore
+++ b/PrimeC/solution_3/.gitignore
@@ -1,1 +1,2 @@
 primes_words
+primes_stripped-block
--- a/PrimeC/solution_3/.gitignore
+++ b/PrimeC/solution_3/.gitignore
@@ -1,2 +1,2 @@
 primes_words
-primes_stripped-block
+primes_striped-block
--- a/PrimeC/solution_3/README.md
+++ b/PrimeC/solution_3/README.md
@@ -1,11 +1,15 @@
 # Implementation in C
 
 ![Algorithm](https://img.shields.io/badge/Algorithm-other-yellowgreen)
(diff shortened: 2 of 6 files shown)


PrimeC/solution_3/.gitignore
```diff
@@ -1 +1,2 @@
 primes_words
+primes_stripped-block
```
rbergen: I'm only going to point it out here, but you may have intended this to be `primes_striped-block` (minus one p) throughout your solution.

rbergen (approved):
There's one small textual thing you may want to change (see below), but even in its current for this LGTM.

PrimeC/solution_3/.gitignore
fvbakel: @rbergen Thank you for noticing. Al least it was a consistent copy paste typo in al places :-)

rbergen: @fvbakel Yes, I did notice that :D

fvbakel:
@rbergen please wait with the merge, I just discovered that there is a issue with the performance when I run from Docker. I want to double check that for the merge.

fvbakel:
@rbergen I make one small change, I changed the storage to uint32_t instead of char because that seems to make a big difference in the performance in docker. The docker performance was 1500 passes in 5 seconds with this small update it becomes 9832 passes in 5 seconds!. 

It is now ready for the merge :-)

rbergen (approved):

marghidanu (approved):

PR #589, opened 2021-08-05
dart 1-bit optimization

mmcdon20:
## Description

I was looking at the solutions for node and typescript and noticed an optimization that could also be applied to the dart solution. Those solutions took advantage of the fact that the values in the integer lists were initialized to 0, and use a 0 bit to indicate a prime value rather than 1.

## Contributing requirements

### DIFF
--- a/PrimeDart/solution_1/bin/PrimeDartOneBit.dart
+++ b/PrimeDart/solution_1/bin/PrimeDartOneBit.dart
@@ -51,7 +51,7 @@ class PrimeSieve {
   /// Also, as stated previously, Dart doesn't have a way to specify an integer
   /// size, but it does allow you to have lists of specific integer sizes.
   /// Values are truncated when placed into the list.
-  final Uint64List _bits;
+  final Int64List _bits;
 
   /// This field contains the results we would expect to find for any given
   /// [sieveSize].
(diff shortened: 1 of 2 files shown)


PrimeDart/solution_1/bin/PrimeDartOneBit.dart
```diff
@@ -51,7 +51,7 @@ class PrimeSieve {
   /// Also, as stated previously, Dart doesn't have a way to specify an integer
   /// size, but it does allow you to have lists of specific integer sizes.
   /// Values are truncated when placed into the list.
-  final Uint64List _bits;
+  final Int64List _bits;
```
rbergen, line 54: Why the switch from unsigned int to signed int?

mmcdon20: The `int` type in dart is a 64 bit signed int. It actually didn't appear to make any difference in performance when using `Uint64List` vs `Int64List` that I could tell. I tried both because I saw a comment in the nodejs solution that said the `Int32Array` was chosen over the `UInt32Array` for a similar reason.

mmcdon20: If you want I can switch it back to `Uint64List`, since it doesn't seem to matter either way.

rbergen: Generally speaking, if a change doesn't have a positive effect on benchmarked performance, it shouldn't be included. However, this is a change to your own solution. In that case I'm more inclined to leave it up to you, unless you open a PR that leads to a significant amount of unnecessary review effort. I can't say that's the case here.

So, it's up to you in this case.

mmcdon20: @rbergen It doesn't make a difference to me either way, so I guess I'll leave the PR as is.

PR #590, opened 2021-08-05
Go/solution_2 a few new cache-friendly implementations 

ssovest:
## Description
Some new implementations which (well, in theory) solve the problem with low performance on CPUs with small cache.

 - *sieve32_block.go* is a modification of sieve32.go. The sieve loop is untouched, but the bit clearing loop now processes the buffer in blocks.
 - *sieve_other_block.go* is a modification of sieve_other.go. The bit clearing loop is modified in a similar manner.
 - *sieve_other_segmented.go* is a segmented version of sieve_other.go. It's not a `base` implementation anyway, so why not :)

Also, now all versions are only compiled with `-B` flag, so, despite of adding 3 more files, the total number of actual running versions is 1 less. (By the way, is there a maximum number of implementations a solution can have? I'm not planning to add more, just curious.)

## Contributing requirements

### DIFF
--- a/PrimeGo/solution_2/README.md
+++ b/PrimeGo/solution_2/README.md
@@ -2,15 +2,21 @@
 
 Collection of 1-bit single-threaded Go solutions.
 
-*sieve8.go* stores bits in []uint8. Implements the base algorithm.
+*sieve8.go* stores bits in []uint8. Implements the base algorithm. The slowest one, it's here just for comparison.
 
-*sieve32.go* stores bits in []uint32. Implements the base algorithm.
+*sieve32.go* stores bits in []uint32. Implements the base algorithm, but uses a datatype optimized for setting ranges of bits.
+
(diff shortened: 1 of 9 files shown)


rbergen:
> By the way, is there a maximum number of implementations a solution can have?

In line with most other things, the answer is: it's not limited to a specific number (and technically, a Docker container can run for years), but we do aim to apply common sense to what we merge and don't. 
Concretely, there have been PRs aiming to add a solution that demonstrates a number of implementation aspects in a particular language, and originally ran every possible permutation of those aspects. In certain cases, we have requested contributors to pick a subset of them that illustrates the key/most interesting points of the implementation aspects they explored.

ssovest:
Sorry for fixing things after opening the PR. Now it should be stable (at least stable enough to do what it is intended to do :)).

rbergen (approved):
LGTM

marghidanu (approved):

PR #593, opened 2021-08-05
amd64 handwritten 1-bit and 8-bit faithful base

dacvs:
## Description
Adding a 1-bit solution otherwise same as the already-accepted 8-bit solution.
I do not have a Docker environment. I would be grateful if a reviewer would verify that the Dockerfile still runs.

## Contributing requirements


rbergen:
> I do not have a Docker environment. I would be grateful if a reviewer would verify that the Dockerfile still runs.

We verify that more or less automatically, because we don't merge without running CI, which includes a verification of docker build working, by definition. In the case of a new solution or the addition of new implementations, I tend to run the solution locally as part of my review.
However, I would urge you (and anybody else submitting solutions or implementations) to test everything locally before submitting, including the Dockerfile. A review is not intended to be or replace testing, and you may end up adding a solution to the repo that you've spent time on and just doesn't work.

rbergen (approved):

Macleykun:
@dacvs https://labs.play-with-docker.com/
This is an online platform
I somehow can’t ssh into it, but overall it works well if you don’t have/want too get a local docker environment.

marghidanu (approved):

PR #595, opened 2021-08-06
Pascal solution_3 single_threaded and multi-threaded

olivierbrun:
## Description

I have provided the ability to run the solution both single-threaded and multi-threaded.
- a run.sh file has been added to run the application once with a -t1 parameter (1 thread) and once without parameter (the number of threads corresponds to the number of CPU cores available on the system)
- the Dockerfile has been updated to make use of the run.sh file
- the program has been updated to display a label containing the number of threads to differentiate both runs

## Contributing requirements

### DIFF
--- a/PrimePascal/solution_3/Dockerfile
+++ b/PrimePascal/solution_3/Dockerfile
@@ -1,14 +1,15 @@
 FROM primeimages/freepascal:3.2.0 AS build
 
-COPY *.pas /opt/app/
-COPY run.sh /opt/app/
 WORKDIR /opt/app
+COPY *.pas ./
+COPY run.sh ./
 
+RUN chmod a+x run.sh
 RUN fpc PrimePas -O3 -v0
-RUN chmod +x run.sh
 
 FROM ubuntu:20.04
-COPY --from=build /opt/app/PrimePas ./
-COPY --from=build /opt/app/run.sh ./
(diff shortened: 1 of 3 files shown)


rbergen:
The distinguishing characteristic between Pascal/solution_2 and yours, is that yours is multi-threaded. The single-threaded version of your solution is an implementation with the same key characteristics as solution_2 (base algorithm, faithful, 1 bit per prime). I therefore don't see the point of adding this?

olivierbrun:
hello Rutger, thanks for the feedback. 
- actually the single-threaded run uses the same multi-threaded program but spawns only one thread (runs at least 1.5 times faster per thread, on a single thread than on multiple threads)
- it uses a different mechanism than solution_2 and supports 64 bits primes (solution_2 uses a packedbit array which limits primes to 32 bits primes). In one of his videos, Dave mentioned that ```vector<bool>``` supports 64 bits integers in CPP, I wanted to implement a mechanism in Pascal that also supports 64 bits primes.
- solution_3 runs 1.9 times faster than solution_2 (5812 passes i.e 1162.4 passes/s/thread on a single thread vs 2959 passes i.e. 591.8 passes/s/thread for solution_2 in the VM on my local machine).

That said, I understand your point, should you want me to close this pull request, please let me know.

rbergen:
@olivierbrun Thank you for your explanation. I think that's fair enough,

PR #599, opened 2021-08-06
Ignore some hadolint info messages, fix SC2035 and DL3003

rbergen:
Updates hadolint configuration to ignore a few "info" rules:

- ~~DL3009: Delete the apt-get lists after installing something.~~
- DL3015: Avoid additional packages by specifying --no-install-recommends.
- ~~DL3019: Use the `--no-cache` switch to avoid the need to use `--update` and remove `/var/cache/apk/*` when done installing packages.~~
- DL3047: `wget` without flag `--progress` will result in excessively bloated build logs when downloading larger files.
- ~~DL3059: Multiple consecutive `RUN` instructions. Consider consolidation.~~

Also fixed SC2035 in three Dockerfiles and DL3003 in one.


marghidanu:
DL3009: Smaller Docker images are always welcome! I believe this is a good practice.
DL3019: Same as above only for Alpine, for as much as possible we should have smaller Docker images.
DL3059: Unless you have strict rules about Docker layers, then this should be excluded, otherwise no. 

The rest are not that important, I think those are marked as warnings as default

rbergen:
All of these are info messages, which is below warning level, and the lowest notification level that hadolint supports. They were never visible in the Files changed tab in PRs (warnings and errors were), but they started showing up today and are causing a lot of bloat on the aforementioned tab. 

If you want to keep 3009, 3019 and 3059 in place then that's fine, but I'll consistently ignore them while reviewing.

rbergen:
@marghidanu Could you take another look?

marghidanu (approved):

marghidanu:
LGTM

PR #605, opened 2021-08-07
Clojure solution 2

axvr:
## Description

Fast and faithful (but not quite idiomatic) Clojure solution.

## Contributing requirements

## Sample output

```
Passes: 6518, Time: 5.000845743, Avg: 7.672362E-4, Limit: 1000000, Count: 78498, Valid: True
axvr_clj-sln-2_8-bit;6518;5.000845743;1;algorithm=base,faithful=yes,bits=8
Passes: 3040, Time: 5.001040736, Avg: 0.0016450792, Limit: 1000000, Count: 78498, Valid: True
axvr_clj-sln-2_1-bit;3040;5.001040736;1;algorithm=base,faithful=yes,bits=1
```

(On an Intel Core i7-8550U @1.80GHz x8 CPU with 16GB of RAM using warm-up.)

### DIFF
--- a/PrimeClojure/solution_2/.gitignore
+++ b/PrimeClojure/solution_2/.gitignore
@@ -1,0 +1,1 @@
+.cpcache/
--- a/PrimeClojure/solution_2/Dockerfile
+++ b/PrimeClojure/solution_2/Dockerfile
@@ -1,4 +1,4 @@
 FROM clojure:openjdk-17-tools-deps-alpine
 WORKDIR /primes
-COPY deps.edn primes.clj ./
-ENTRYPOINT ["clojure", "-X:run", ":warm-up?", "true"]
+COPY deps.edn sieve_1_bit.clj sieve_8_bit.clj run.sh ./
+ENTRYPOINT ["./run.sh"]
--- a/PrimeClojure/solution_2/Dockerfile
+++ b/PrimeClojure/solution_2/Dockerfile
@@ -1,0 +1,4 @@
+FROM clojure:openjdk-17-tools-deps-alpine
+WORKDIR /primes
+COPY deps.edn primes.clj ./
+ENTRYPOINT ["clojure", "-X:run"]
--- a/PrimeClojure/solution_2/Dockerfile
+++ b/PrimeClojure/solution_2/Dockerfile
@@ -1,4 +1,4 @@
 FROM clojure:openjdk-17-tools-deps-alpine
 WORKDIR /primes
 COPY deps.edn primes.clj ./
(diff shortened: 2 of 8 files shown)


axvr:
I've force pushed an update which improved variable naming and added some explanatory comments.

axvr:
Pushed change confirming boolean array bit usage as 8 bits/boolean.  Unfortunately this disqualifies it from the main leaderboard.  :(

axvr:
Pushed final change to slightly improve performance further.  Should now be ready for review.

rbergen (approved):
LGTM

axvr:
@rbergen, I've re-requested review as I've pushed some new changes.
- Greatly improved the performance of solution 2
- Added solution 3 using a 1-bit data structure (but is half the speed of solution 2).

axvr:
Also these solutions use a "warm-up" like the Java solution 4 does to reduce the variability of the results.  If this is not OK, I'm happy to disable/remove it.

rbergen:
If you want to add a second approach, then please restructure this PR to one solution with two implementations.

rbergen (commented):
If you want to add a second approach, then please restructure this PR to one solution with two implementations.

rbergen:
> Also these solutions use a "warm-up" like the Java solution 4 does to reduce the variability of the results. If this is not OK, I'm happy to disable/remove it.

We've basically settled on condoning a warmup of a maximum of 5 seconds. In the case of multiple implementations within a solution, it's nice if the warmup time can be limited to a total of 5 seconds for all implementations combined. Also, we may revisit this decision when we implement a "best of n" benchmark setup, meaning we run every solution n times (5 is what we're now considering) and then report the result of the best-performing run.

axvr:
Merged both solutions into the `solution_2` folder.   Only the 8-bit solution is using a warm-up.

I like the idea of a "best-of-n-runs" as a way to reduce results variability, although I'd imagine that it'll take n times longer to run all the benchmarks, so the 5 seconds maximum warm-up seems fair.

rbergen:
> I like the idea of a "best-of-n-runs" as a way to reduce results variability, although I'd imagine that it'll take n times longer to run all the benchmarks.

It will indeed multiply the benchmark runtime with something close to n. One of the things we're considering is how to make this configurable (and what to make the default). Longer runtimes is less of a problem when it concerns scheduled jobs on more or less dedicated hardware, which is how the PrimeView reports are/will be generated. That is also where a reduction of variability is most relevant.

rbergen (approved):
LGTM

marghidanu (approved):

PR #607, opened 2021-08-07
Minor tweaks to compilation

nmerriam:
## Description
Minor tweaks to compilation: compiler options and static/inline keywords. I still do not understand why this code performs so well relative to other solutions when we run benchmarks on various computer and so badly in the official league table but there you go. I would love to learn what I am doing wrong :)

## Contributing requirements
I realize that optimizing these parameters for a native Windows build was not the best configuration.


rbergen:
> I still do not understand why this code performs so well relative to other solutions when we run benchmarks on various computer and so badly in the official league table but there you go.

It may be the age/nature of the benchmark hardware; other people have also voiced their surprise at the order of results the leaderboard now shows.
We're working on running the benchmarks on other machines and posting those results, too.

PR #498, opened 2021-07-25
Common Lisp solution 2 patch3

mayerrobert:
## Description
This PR adds another method for setting the bits in the inner loop of the base algorithm to Common Lisp solution 2. Also all variants are now marked as faithful.

Plus some minor speedups, tests and fixes for odd sieve sizes.

## Contributing requirements

* [x] I read the contribution guidelines in CONTRIBUTING.md.
* [x] I placed my solution in the correct solution folder.
* [x] I added a README.md with the right badge(s).
* [x] I added a Dockerfile that builds and runs my solution.
* [x] I selected `drag-race` as the target branch.
* [x] All code herein is licensed compatible with BSD-3.

PR #609, opened 2021-08-07
Lisp: unroll loops, bitvector set from 2.1.8, hashdot: eval at readtime

mayerrobert:
## Description

The biggest performance improvement (and it could be controversial regarding faithfulness) is: I nagged the SBCL maintainers into improving bitvector-set-operations, this will be included in the upcoming SBCL 2.1.8. The file bitvector-set-2.1.7-2.1.8-snap.lisp backpatches this improvement on the fly into the currently used SBCL 2.1.7.

Further improvements:
- loop unrolling in PrimeSieve.lisp and PrimeSievebitops.lisp
- PrimeSieve.lisp uses the Lisp function `position` instead of the first loop
- PrimeSieve.lisp now contains a second (unfaithful) solution that uses `#.`
- PrimeSieveWheelBitvector.lisp is based on PrimeSieveWheelBitOpt.lisp but uses the builtin bitvector instead on manual bit-fiddling
- `(1+ (floor...` was replaced by `(ceiling...` at several places
- some stricter type declarations, some single-use variables were inlined (sbcl's optimizer doesn't like assignments)

Sorry that the diff is fairly large, but I would't know what to leave out.

## Contributing requirements

### DIFF
--- a/PrimeLisp/solution_2/Dockerfile
+++ b/PrimeLisp/solution_2/Dockerfile
@@ -1,4 +1,4 @@
-FROM daewok/sbcl:2.1.6-alpine3.13
+FROM daewok/sbcl:2.1.7-alpine3.14
 
 WORKDIR /opt/app
 
--- a/PrimeLisp/solution_2/PrimeSieve.lisp
+++ b/PrimeLisp/solution_2/PrimeSieve.lisp
@@ -1,8 +1,16 @@
-;;;; based on mikehw's solution, approx. 120x speedup
+;;;; based on mikehw's solution, approx. 220x speedup
 ;;;
 ;;; run as:
 ;;;     sbcl --script PrimeSieve.lisp
 ;;;
+
+
+#+(and :sbcl :x86-64)
+(progn
+  (when (equalp "2.0.0" (lisp-implementation-version))
+    (load "bitvector-set-2.0.0-2.1.8-snap.lisp")) ; teach sbcl 2.0.0 the new bitvector-set from sbcl 2.1.8
+  (when (equalp "2.1.7" (lisp-implementation-version))
+    (load "bitvector-set-2.1.7-2.1.8-snap.lisp")))  ; teach sbcl 2.1.7 the new bitvector-set from sbcl 2.1.8
 
 
 (declaim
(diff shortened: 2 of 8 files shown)


rbergen (commented):
Please mark the implementation in PrimeSievewordops.lisp as faithful, but an "other" algorithm. I know this is a change from what was discussed in #498, but since then we've settled on the use of pre-calculated bitmasks not representing a deviation from faithfulness, but instead a deviation from the base algorithm. Specifically, it goes against the point that _"When clearing non-primes in the sieve (the second operation), **the algorithm clears all non-primes individually**, increasing the number with 2 * factor on each cycle."_

With regards to the use of the SBCL 2.1.8 bitvector, why could that be controversial from a faithfulness perspective? In my personal opinion it makes the solution somewhat brittle (the developments in SBCL may turn out differently than you now expect), but that's not a faithfulness aspect, in itself.

mayerrobert:
Re: PrimeSievewordops.lisp; will gladly do that, I actually prefer it that way and am happy that you settled that way. I'd like to add that I am perfectly ok with the way you sometimes amend/ change previous decisions and/ or amend rules to address newly found understandings.

Re: SBCL 2.1.8 bitvector; I don't change or add to the programming language Common Lisp, the actual timed program doesn't use anything that is not Common Lisp, and I'm glad you seem to see it the same way. I was afraid you'd say "changing the compiler is not allowed". All solutions should run unchanged on any conforming Common Lisp implementation, I did some tests with ABCL which is a Common Lisp implementation that runs on the JVM. The patch to the compiler will only be applied on SBCL 2.1.7.

rbergen:
> Re: PrimeSievewordops.lisp; will gladly do that, I actually prefer it that way and am happy that you settled that way. 

I'm happy to hear that. I just saw the commit updating the output line. Could you please also update the badges in your readme? (Add "algorithm: other", remove "faithful: no"). 

> I'd like to add that I am perfectly ok with the way you sometimes amend/ change previous decisions and/ or amend rules to address newly found understandings.

Thank you, I appreciate you saying that.

mayerrobert:
I have updated the README but we both missed the second "algorithm" in PrimeSieve.lisp which is not really faithful. My second commit addresses this, the badges should now be accurate.

PR #610, opened 2021-08-07
[C#-4]Adding striped solution

ItalyToast:
## Description
A simplified version of the striped algorithm used by rust solution_1. Instead of having the sieve oriented "vertically", we keep the normal bit array representaion. What we do instead is create the mask and push the pointer by factor. That way we effectively mark every 8th factor. Instead of walking the stripes 1 to 8, we walk the stripe that the first factor to be marked is located in. The next stripe to clear is calculated by rotating the mask by factor.

There is currently some duplicated code because of my implementation of the multiple sieve implementations. I kept it that way since I didnt want create an interface that runs the sieve because it might affect performance. If I used generics we could dodge the indirect call but that was more work than its worth and it reduces clarity of the code.

## Contributing requirements

### DIFF
--- a/PrimeCSharp/solution_4/Dockerfile
+++ b/PrimeCSharp/solution_4/Dockerfile
@@ -10,7 +10,7 @@ COPY *.cs /source
 RUN dotnet publish -c release -o /app --no-restore
 
 # final stage/image
-FROM mcr.microsoft.com/dotnet/runtime:3.1-focal
+FROM mcr.microsoft.com/dotnet/runtime:5.0-focal
 WORKDIR /app
 COPY --from=build /app .
 ENTRYPOINT ["/app/PrimeSieveCS"]
--- a/PrimeCSharp/solution_4/PrimeCS.cs
+++ b/PrimeCSharp/solution_4/PrimeCS.cs
@@ -1,196 +1,53 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
-using System.Numerics;
-using System.Runtime.CompilerServices;
 
 namespace PrimeSieveCS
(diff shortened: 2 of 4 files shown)


rbergen:
When trying to run this version of your solution in Docker, I get the following message:
```
It was not possible to find any compatible framework version
The framework 'Microsoft.NETCore.App', version '5.0.0' was not found.
  - The following frameworks were found:
      3.1.17 at [/usr/share/dotnet/shared/Microsoft.NETCore.App]

You can resolve the problem by installing the specified framework and/or SDK.

The specified framework can be found at:
  - https://aka.ms/dotnet-core-applaunch?framework=Microsoft.NETCore.App&framework_version=5.0.0&arch=x64&rid=ubuntu.20.04-x64
```

This is caused by #603 changing the version of the "build" base image to .NET 5.0, but leaving the "runtime" base image set to .NET Core 3.1. Changing [line 13](https://github.com/PlummersSoftwareLLC/Primes/blob/831dbedbe460d544f5260f7db51065f87cafbaae/PrimeCSharp/solution_4/Dockerfile#L13) in your Dockerfile to 
```dockerfile
FROM mcr.microsoft.com/dotnet/runtime:5.0-focal
```
fixes this.

ItalyToast:
Docker file has been fixed, I guess it built for me since I already got the 3.1 image.

rbergen:
I think it builds for everybody, but it doesn't run. Today's benchmark report doesn't include CSharp/solution_4 because of this.

ItalyToast:
Ahh that makes sense. My bad for not actually running the solution, first time using docker for me with this project 😅.

rbergen (approved):
LGTM

marghidanu (approved):

PR #611, opened 2021-08-08
Haskell solution_2 add loop unrolling by block...

GordonBGood:
## Description

Increases speed about twenty percent by unrolling most of the marking loops by four marking operations, and also running the eight composite bit marking operation loops by CPU L1 cache sizes in the sieve buffer at a time for better cache associativity (cache "thrashing" is reduced although not eliminated for CPU L1 cache sizes less than 64 Kilobytes).

The `faithful to base` algorithm is still preserved as it still alternately searches for the next base prime value to be used to determine composite number multiples with actually marking the composite representation bits that are multiples of that base prime value in the sieve buffer, just that within the latter operation, the marking is done out of order for better loop and cache efficiency.  This is similar although not the same as the Rust "striping_block" implementation, with is categorized as `faithful to base` algorithm.

## Contributing requirements


PrimeHaskell/solution_2/README.md
```diff
@@ -2,9 +2,11 @@
...
...sociativity (less cache "thrashing").  To do this easily, it uses a small eight-integer array created at line 90 just below where the two views into the sieving buffer array were created.  Then, for each base prime value to be used for marking of composite representation bits, it fills this small array with the initial start byte indices at lines 91 and 92, followed by the rest of the loops, first by buffer "page" starting at line 93, then by the up to eight "bit-loops" starting at line 99, and then the marking loops.  For this implementation, marking is broken up into two phases, with the first unrolled loop starting at line 184 and the second single mark per loop starting at line 118 doing the marking up to the end of the page when there are less than four markings to be done.  This implementation uses `foldM` as necessary to thread the current marking value through the loops so that it can be saves back to the current bit position location in the small eight value array for use in the next page pass, which saving is done at line 123.  As before, the boolean view is returned ti be easy to scan for unmarked values and convert to prime values using the `assocs` function in line 74.
```
rbergen: In the last sentence, I imagine "...is returned **ti** be easy..." is meant to read "...is returned **to** be easy..."

GordonBGood: Corrected the typo just now.

PR #613, opened 2021-08-08
[Do not merge] Mark C/solution_4 as not faithful

### LINKED ISSUE #612: C/solution_4 is not faithful but marked as such
As [pointed out](https://github.com/PlummersSoftwareLLC/Primes/commit/75d954f76160b03d694979cfa945dc218a63b4e4#commitcomment-54561309) by @rogiervandam, C/solution_4 should be marked as not faithful, due to the fact that the construct that contains the sieve is created only once, outside of the timed loop. This goes against the rule that _"[The solution] uses a class to encapsulate the sieve, or a (closest) equivalent feature in your language if a class construct is not available. This class must contain the full state of the sieve. **Each iteration should re-create a new instance of this class from scratch.**"_

I will open a PR to change the classification of this solution as not faithful. @nmerriam I'll hold on merging it for a few days so you can open an alternative PR to modify the implementation to one that is faithful, if you want to do so.

rbergen:
This resolves #612 by marking C/solution_4 as not faithful. 

Marked as "do not merge" to allow @nmerriam to submit an alternative PR that changes the implementation to a faithful one, instead.

### DIFF
--- a/PrimeC/solution_4/README.md
+++ b/PrimeC/solution_4/README.md
@@ -1,7 +1,7 @@
 # C solution by Nick Merriam and GLIWA team
 
 ![Algorithm](https://img.shields.io/badge/Algorithm-wheel-yellowgreen)
-![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-no-yellowgreen)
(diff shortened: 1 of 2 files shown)


rbergen:
Closing due to #614 making the solution faithful.

PR #614, opened 2021-08-08
Destroy and reallocate bit array on each solution

nmerriam:
## Description
Destroy and reallocate bit array on each solution. And improve support for 32-bit targets.

## Contributing requirements
PR kindly requested by @rogiervandam / @rbergen 


rbergen:
This is definitely a step in the right direction, but for the implementation to qualify as a faithful one, the whole SIEVE struct will have to be created in every iteration of the timed loop.

nmerriam:
1. You're completely right. I get it now and it's kind of embarrassing.
Suffice to say that I don't use environments where I get to call "new"
explicitly.
2. [edit] I can't open a new Pull Request because you kept this one open to give me a second chance. Thank you!

Best wishes,
Nick

rbergen:
@nmerriam Thanks for your efforts to address this. We don't have that many rules to follow, but the ones that exist we have to apply as consistently as possible.

And you're welcome.

PR #615, opened 2021-08-08
D solution 2 fixes, graphs, and tweaks.

BradleyChatha:
## Description

In short:

* Consolidate the aux tool scripts into a single script.
* Add tool to generate CSV files based on run results.
* Fixed that some of the solutions weren't reporting their bits properly in single threaded runs.
* Fixed that the "SieveRTB1_XX" solutions were still assuming that their storage was a byte, not their actual unit size. This has provided significant speed changes.
* Updated README with latest results, and added some graphs.
* A few more tweaks and misc fixes.
* Added `SieveRTBX` series of solutions, but they're not on the leaderboard since the performance isn't exactly surprising nor interesting.

Furthermore I enabled some more results into the leaderboard:

* One variant of the cheaty solution was added because C++ made me jeleous.
* Two variants (single and staticThreads) of the "SieveRTB1_XX" solutions. This is because the new fixes to them show an interesting performance difference. If needed I'm fine with cutting these down to one variant each instead of two.

## Contributing requirements


rbergen:
@BradleyChatha When running your solution in Docker, I happened to notice the following line in the output:
```
BradleyChatha-MultistaticThreads-SieveRTCT_Cheatiness;375313931;5.0001;8;algorithm=other,bits=0,faithful=no
```
It seems this implementation is claiming to use no bits at all, which I think isn't technically possible. 

Concerning the number of implementations in your solution: you're free to include as many as you want in your codebase, but we do prefer the set of implementations that is run during the automated benchmark (through `docker run -ti --rm <solution>`) to be a "clever subset" that illustrates the most important/interesting differences from a performance standpoint, without including every permutation of implementation aspects.

BradleyChatha:
> It seems this implementation is claiming to use no bits at all, which I think isn't technically possible.

In this case yea, I think it should technically be 1 bits: It computes and runs the entire sieve at compile time, then just reuses that over and over, so it's still using up that 1 bit per prime technically.

But also technically, if I were to do (which I won't, just an example):

```d
enum PRIME_COUNT(size_t sieveSize) = (){
    auto sieve = new SieveRT(sieveSize);
    sieve.runSieve();
    return sieve.countPrimes();
}();
```

Neither the compile-time sieve and even the return value would take up any space (enum in D in this case is essentially a `# define`) >:3.

Anyway, that was just a for fun thing to think about, nothing serious.

As for the subsets to show, I think I'll drop some things to 1 variant each, since that still shows the performance differences without it bloating the leaderboard.

PR #617, opened 2021-08-08
PrimeSwift/solution_1 fine tuning.

yellowcub:
## Description

Yet another performance bump to the `PrimeSwift/solution_1` single bit implementation.  This update found significant speed increases with restructuring the main while loop in `Sieve.runSieve()` to a `repeat { } while`, as well as replacing the modulus call (`%`) in `BooleanBitArray.maskIndex(...)` with a method that takes advantage that the divisor is a power of two.  Lastly, a restructure of the main clocked loop resulted in a notable speed increase.

Other changes reflect opportunities to make the code more readable (in the contributor's eye at least).

## Contributing requirements

### DIFF
--- a/PrimeAssembly/solution_1/Dockerfile
+++ b/PrimeAssembly/solution_1/Dockerfile
@@ -1,15 +1,14 @@
 FROM alpine:3.13
 
-RUN apk add --update build-base nasm
+RUN apk add --no-cache build-base nasm
 
 WORKDIR /opt/app
 
 COPY *.sh ./
 COPY *.asm ./
 
-RUN chmod a+x *.sh
-
-RUN ./build.sh
+RUN chmod a+x ./*.sh && \
+    ./build.sh
 
 ENTRYPOINT [ "./run.sh" ]
 
--- a/PrimeAssembly/solution_2/Dockerfile
+++ b/PrimeAssembly/solution_2/Dockerfile
@@ -1,15 +1,14 @@
 FROM alpine:3.13
 
-RUN apk add --update build-base
+RUN apk add --no-cache build-base
 
 WORKDIR /opt/app
 
 COPY *.sh ./
 COPY *.s ./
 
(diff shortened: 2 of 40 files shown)


yellowcub:
@rbergen This is an update that resulted in a 25-35% performance increase over the last commit (as witnessed on my machine).  I was careful, I think, to push/pull in the right order so as not to have a mess of other peoples commits attached to mine.  

As I may have mentioned before, I am not a professional programmer.  Each of these commits have involved countless trials and errors.  I can't say I truly understand why some of them seem to impact performance, but I have learned a number of fundamentals, such as don't introduce another branch/loop unless you absolutely have to.  Each time I have initiated a PR, I was under the belief that my abilities to improve the solution had run its course, only to have a spark of inspiration after resting my eyes on it.  My hope is that each of these incrimental updates to this solution branch has not been overly burdensome for you.  I would not be requesting another PR unless I thought it was significant enough to compete higher in the leaderboard.  Thanks again for your patience.

rbergen:
> Each time I have initiated a PR, I was under the belief that my abilities to improve the solution had run its course, only to have a spark of inspiration after resting my eyes on it.

After having developed software for 35 years, I can tell you that this is how it works for many, if not most, developers. And that's perfectly fine; if I categorically didn't like handling PRs I shouldn't be a maintainer of a collaborative repo. 
That said, I do ask people to consolidate changes where reasonably possible, and try to limit changes to those that are meaningful in the context of this project's goals. The former is often triggered by an author opening multiple PRs within 24 hours that aim to change the same solution. The latter comes from the fact that in principle, every line in the diff of a PR needs to be looked at. Purely cosmetic changes thus cause review work that can really add up, while not really contributing to adding more languages to the collection and/or squeezing the most out them from a performance perspective. (Full transparency: some of the changes in this PR do seem to fall in the latter category, but are within boundaries because they're quite easy to judge.)

PR #618, opened 2021-08-09
PowerShell solution_2 - Fine tuning references and type conversion

RobCannon:
## Description
This is some fine tuning of solution_2 for PowerShell by @crowbar27.  I dereferenced some variables and ensured that type conversions where kept to a minimum.  On my computer this changed from an an average of around 27 passes to 50 passes.  I also removed the use of pipelines during the measured code as direct references are faster.

## Contributing requirements


rbergen:
@crowbar27 If you want to, please let us know if you're okay with applying these changes to PowerShell/solution_2.

RobCannon:
I have added a few more changes that add a couple of more passes.  The main change is to use a bit array with all possible numbers and not just odd numbers, which prevents the shift operation on the index in every loop of runSieve.  This presents the choice made in the original cpp version of the code.
I also made the file more linux friendly by adding a hashbang so it can from run from the shell command line and set the line ending to LF.

Finally, I would like to modify the code that touches the BitArray to all be function in the Sieve class to provide better encapsulation and it is more in keeping with the design of the original.  The code that calls the Sieve can still be more like traditional PowerShell.  @crowbar27, are you OK with that change?  If so, I will update this PR.

crowbar27:
First of all, thank you for your patience.

Second, @rbergen, I get the idea why you want to limit the number of solutions per language, but for me, one of the interesting aspects of this project is not only to compare different languages, but also seeing the cost of certain ways of implementing it in the same language. Therefore, I would consider it beneficial if a change that removes the pipeline and puts everything in the class would stay beside the original implementation (I do not know whether the original C# implementation has been stripped of the division/bit shifts, too, but if not, I would argue to keep one PS implementation that is in line with the C# one).

Having said all of that, I would dedicate this solution_2 to maximally optimise it (it has already "inlined" some stuff and therefore deviates structurally from the original) if we keep solution_1 the more straightforward one using pipelines and staying closer to the C# implementation. @RobCannon I guess the moving the manipulation into the class does not change the performance, so please go ahead.

Finally, I am not a fan of reformatting everything for style reasons, particularly after the disaster last time when I did not see that the algorithm was broken. Still embarrassed about that, but as basically everything lit up, I did not closely look at it thinking it was all only formatting ... Also, as I Windows user, I prefer Windows-friendly more 😉

rbergen:
> @rbergen, I get the idea why you want to limit the number of solutions per language, but for me, one of the interesting aspects of this project is not only to compare different languages, but also seeing the cost of certain ways of implementing it in the same language. 

Limiting the number of _solutions_ does not mean that multiple _implementations_ aren't possible; they can easily be combined in one solution (as quite some solutions do). The reason we want to apply a reasonable limit to the number of solutions is that each solution (directory) triggers a docker build (and docker run) when benchmarking.

> Therefore, I would consider it beneficial if a change that removes the pipeline and puts everything in the class would stay beside the original implementation (I do not know whether the original C# implementation has been stripped of the division/bit shifts, too, but if not, I would argue to keep one PS implementation that is in line with the C# one).

Again, it's perfectly okay to retain the current implementation and add the version suggested by this PR as a second one, if that's what you prefer.

> Finally, I am not a fan of reformatting everything for style reasons

This I agree with. @RobCannon I do have to ask you to revert this change, as it makes it almost impossible to identify and review the changes in logic you're suggesting.

RobCannon:
I didn't reformat everything for style reasons.  This file was checked in the CRLF line endings, which make the file un-executable on linux or WSL systems.  CRLF endings should only be used if everyone using the repo is a Windows user and that is not the case here.
https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings  

I think the PowerShell pipeline feature is a great feature when you have lots of items or fields going through the pipeline, but here there is only one sieve at a time and it just adds overhead and the idea here is to be the fastest implementation.  IMO, the class feature of PowerShell is very unlike the rest of the language, but that was one of the requirements so we should embrace it to fullest.

But I can create this as a new solution.  I think it is probably the right idea since I am also going to try to implement a parallel option.

rbergen:
> I didn't reformat everything for style reasons. This file was checked in the CRLF line endings, which make the file un-executable on linux or WSL systems. CRLF endings should only be used if everyone using the repo is a Windows user and that is not the case here.

The script can be executed on Linux and WSL with CRLF line endings, provided that the PowerShell interpreter is explicitly invoked (`pwsh PrimePowerShell.ps1`); this is also what the Ubuntu-based Docker image does. Using LF as line endings is indeed required if you want to use the hashbang approach.

PR #621, opened 2021-08-09
Refactor Julia solution_3 into "package" structure

louie-github:
## Description

This is the "major refactor" of Julia solution_3 I mentioned in #584. It separates the implementation from the benchmarking code and wraps it into a Julia "package" structure so that it would be easier to add new implementations in the future.

I apologize for the rather large diff, but I think this is a good tradeoff for less code duplication and an easier time reviewing new implementations in the future.

## Contributing requirements


rbergen:
@louie-github As I said in the discussion that was part of #584, I can understand a refactor like this, in the context of a solution that indeed contains multiple implementations. This PR seems to apply a significant restructure to the solution to facilitate the management of multiple implementation, while there still is only one implementation in place. That does not make sense to me. 

I'll have to suggest that this refactor is applied when there is an actual reason to do so, which would be when at least one other implementation is indeed added to the solution.

louie-github:
You make a good point.

I've been working on a striped version for a while now, so should I close this PR for now and then reopen it once I add the new implementation(s)? Or should I leave it open? Or should I just make a new PR when the time comes?

rbergen:
My preference would be to close this PR, and open a new one when you have your striped implementation ready. That PR will then introduce the package structure and the second implementation. The benefit of that is that the ability of the new structure to manage multiple implementations has then also been put to the test, by definition. As such, we will then be reviewing an approach that has been proven to work.

louie-github:
Understood. Thanks for your time once again, and I hope to finalize the PR soon!

PR #623, opened 2021-08-09
[CPP] SeiveBitOrs 1-2 pass solution

schmide:
This is my first contribution. I may need help. Thanks in advance. 

The formatting for all the docker files and readmes are copied from PrimeCPP solution 1 and 2

## Description

SeiveBitOrs is a 2 pass implementation of the base algorithm.

## Contributing requirements

## Output

Passes: 18434, Time: 5.000000, Avg: 0.000271, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: 1

schmide;18434;5.000000;1;algorithm=base,faithful=yes,bits=1

### DIFF
--- a/PrimeCPP/solution_5/Dockerfile
+++ b/PrimeCPP/solution_5/Dockerfile
@@ -1,0 +1,14 @@
+FROM ubuntu:21.04 AS build
+
+RUN apt-get update -qq \
+    && apt-get install -y clang
+
+COPY *.cpp /opt/app/
+COPY *.h /opt/app/
+WORKDIR /opt/app
+RUN clang++ -march=native -mtune=native -Ofast -std=c++17 SeiveBitOrs.cpp -o SeiveBitOrs
+
+FROM ubuntu:21.04
+COPY --from=build /opt/app/SeiveBitOrs /usr/local/bin
+
+ENTRYPOINT [ "SeiveBitOrs" ]
--- a/PrimeCPP/solution_5/Helpor.h
+++ b/PrimeCPP/solution_5/Helpor.h
@@ -1,0 +1,35 @@
+#pragma once
+
+// note below must match size
+#define BITORDER 6 // ... 5 = 2^5 = 32, 6 = 2^6 = 64 ...
+#define WORDTYPE int64_t 
+#define UWORDTYPE uint64_t 
+
+#define ONETYPE ( (UWORDTYPE)(1) )
+template<typename T>
+constexpr auto ONESHIFT(T a) { return (ONETYPE << (UWORDTYPE)(a)); }
+#define MASK(a) ( ONESHIFT(a) - ONETYPE )
+#define BITS(a) ( (a) & MASK(BITORDER))
+#define WORDS(a) ( (a) >> BITORDER)  // byte word dword qword lane etc
+#define MAPOUT(a) ( ( (UWORDTYPE)(a) << ONETYPE ) + ONETYPE)
+#define MAPIN(a) ( ( (a) - ONETYPE ) >> ONETYPE)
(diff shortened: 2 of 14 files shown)


rbergen:
With four existing C++ solutions in the collection, we really need a good reason to add another one. I'm not sure I understand what this solution adds to the ones we have. Could you please explain this?

schmide:
It's unique fast and I put time into it. 

You have a solution called BrainFuck. Does mine really fall below that ?

schmide:
There is some detail in the readme.

Mine does a lot of unique things. It builds a list of masks and replicates them over the bitfield. In addition it groups the updates of primes greater than bit 32. Against the single thread solutions in CPP it preforms close to 50% faster. Although it seem docker is not executing perfectly.

rbergen:
> I put time into it.

It would be a mistake to think that that's a reason to get your solution merged. We've not merged other solutions that people clearly have put time into.

> You have a solution called BrainFuck. Does mine really fall below that ?

That's not how this works. Brainfuck was a language we did not have a solution for until recently, and now we do have one. For C++, there are already 4 solutions in place, exploring a very decent range of approaches.

> Mine does a lot of unique things. It builds a list of masks and replicates them over the bitfield. In addition it groups the updates of primes greater than bit 32. 

For one, that means the algorithm used is not the base algorithm. We may still be able to add the solution as an "other" algorithm implementation, but then I think the README should explain the algorithm in some more detail than it currently does.

schmide:
> For one, that means the algorithm used is not the base algorithm. We may still be able to add the solution as an "other" algorithm implementation, but then I think the README should explain the algorithm in some more detail than it currently does.

I believe this is an honest implementation of the base algorithm. It is nothing more than grouping writes. There are discussions on the segmenting of writes to improve cache access in other algorithms. I believe the single threaded solution uses a rolling mask. Mine is just delays the writing of those rolling masks and allows for seeking to a word boundary.

schmide:
adding I kept a lot of details out of my readme because the other solutions have very sparse information.

I am willing to go into detail. Is there a existing solution that I should follow to classify that information.

rbergen:
> I believe this is an honest implementation of the base algorithm. It is nothing more than grouping writes. There are discussions on the segmenting of writes to improve cache access in other algorithms. I believe the single threaded solution uses a rolling mask. Mine is just delays the writing of those rolling masks and allows for seeking to a word boundary.

All base algorithm implementations perform an indexing and write operation for each individual bit that needs to be set or cleared. This requirement is explicitly mentioned in the description of the base algorithm as included in the contributing guidelines.  That means that any mask used in a base algorithm implementation is a one-bit mask; in most languages that is in fact the only way to do any bit twiddling. 
Your solution "saves up" bits to be able to reduce the number of writes. That is mask projection (in fact, you described it as such in an earlier comment, yourself), and that is an "other" algorithm. 

> adding I kept a lot of details out of my readme because the other solutions have very sparse information.

Most solutions use a common approach that is easy to recognize and/or are themselves easy to read (as in, the code is self-documenting).

> I am willing to go into detail. Is there a existing solution that I should follow to classify that information.

You could take a look at the Notes section of the [PrimeNim/solution_3 README](https://github.com/PlummersSoftwareLLC/Primes/tree/drag-race/PrimeNim/solution_3#readme) as an example.

schmide:
I will rethink a bit of this.

Before I regress away from masks

Question. If a unique mask greater than one byte is created one bit at a time, ORed with memory, then discarded, would that sill violate the rules? Or is it basically one bit at a time to memory in order, no exceptions?

Thanks

schmide:
I read though a few other solutions. There are some things I see as quasi-unfaithful. Such as multiple masks in unrolled loops. 

I was thinking about that too. Oh well

Closing till some reevaluating is done.

Thanks

PR #624, opened 2021-08-09
Pascal solution_3 25% performance boost

olivierbrun:
## Description

Updating few bits in the code giving 25% performance boost:
- prime flags are now set to 1 in the constructor
- the prime bit test is now directly included in the  RunSieve procedure instead of inlined via a function
- the ClearBits procedure has been slightly simplified to reduce the number of CPU cycles

Those 3 simple changes bring a surprising 25% performance boost when compiling using freepascal.
Do not hesitate to reject this PR if you will, as I know I have raised to many of them over the past days.

## Contributing requirements

### DIFF
--- a/PrimePascal/solution_3/PrimePas.pas
+++ b/PrimePascal/solution_3/PrimePas.pas
@@ -34,10 +34,10 @@ type
   prime_sieve = class
     private
       sieveSize: uint64;
-      bits: PBitsArr;
-      mask: array [0..31] of uint32;
+      bits: PbitsArr;
+      mask: array [0..31] of uint32; // masks used for the IsPrime and ClearBits methods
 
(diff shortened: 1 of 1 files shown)


rbergen:
> Do not hesitate to reject this PR if you will, as I know I have raised to many of them over the past days.

I think rejecting a PR over that would be a little too drastic. However, I will ask again if there are other things you plan to investigate or improve. :)

olivierbrun:
Hello Rutger, thanks so much for your kind indulgence, I very much appreciate. I have no plan to investigate further, this is the latest bit that I'm going to change on this program.

rbergen:
@olivierbrun Haha, we'll see. I know how hard it is to shut down those background optimization threads in one's head once they get started. :)

PR #636, opened 2021-08-11
Add Batch file solution

sxxov:
## Description
This solution implements the sieve algorithm in Windows Batch. It attempts to be faithful by encapsulating all pass logic into "workers" & instantiating them per pass. It uses Wine to run in Unix based environments.

All things said though, I'm not sure of the viability of this PR. The main reason is the sheer speed (or lack thereof) of this implementation. Refer to the readme for the full details but it took a depressing 10 and a half hours to complete 1 pass. Everything else seems to work though, & I'm open for any help getting that number down.

## Contributing requirements


rbergen:
@Sxxov I've been wondering for a while if anyone was going to give this a shot. Very nice to see that you did! :)

I love the effort you've put into this, and even more so that you've even provided a Dockerfile to go with it. I do think you're correct in assuming that a runtime of 10+ hours is a problem when it comes to including this in automated benchmark runs.

To address this, I'd like to suggest you rename the Dockerfile to ManualDockerfile. That way it will not be picked up by our benchmark tooling, but people can still choose to run it in Docker manually. The `docker build` command in your README.md could then be modified accordingly:
```bash
docker build --build-arg workers=1 --build-arg sieveSize=1000000 -t primebatch -f ManualDockerfile . && docker run -d primebatch
```

As this no longer concerns an "official" benchmark run, you could also consider using a different sieveSize in your README.md (and I mean this across the board), to decrease the time the solution runs.

Obviously, we can reconsider this setup if and when a way is found to very significantly decrease the runtime of your solution.

PrimeBatch/solution_1/Dockerfile
```diff
@@ -0,0 +1,9 @@
+# not ideal to use wine, but no other choice on unix
+ARG workers=1
+ARG sieveSize=1000000
+FROM scottyhardy/docker-wine:latest
```
rbergen: For Dockerfiles that are included in our benchmark runs, it is a requirement that the base image is pinned to a specific tag. As we are not going to be able to include this solution in those runs at this time, this is of less importance, now.

PrimeBatch/solution_1/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+It takes advantage of the worker model to imitate classes, spawning a new `cmd` window per pass. This also means multithreading is supported, but is disabled by default.
+
+> Note: This implementation runs really *really* slowly. I'd recommend skipping this one if anyone intended to run all benchmarks at once. Check the [output](#output) section for reference.
+
+
+
+## Run instructions
+
+These instructions assumes that you're in the directory of `PrimeBatch` & have all the dependencies installed as stated in `BENCHMARK.md`.
```
rbergen: I think one needs to be in `PrimeBatch/solution_1` for the commands in this README.md to work.

rbergen (commented):
I think it would be good to add an arch-amd64 flag file to this solution, as the Docker base image you're using is not available for the arm64 architecture.

sxxov:
@rbergen 

I thought I'd be the only one bothering with batch, really cool to see others interested too xd. I've made all the changes you pointed out.

Regarding speed, it's outside of my knowledge at this point. It's limited to binaries available in both Wine & Windows, meaning no cheating with `powershell` or `wmic` calls (I had to swap out simple things like `timeout` & `break` too xd). We could probably shave a few hours by doing super undefined-behaviour things, like messing with the command buffer itself, but somehow I doubt a 756000% speed-up is likely xd.

rbergen (approved):
LGTM

rbergen:
> I thought I'd be the only one bothering with batch, really cool to see others interested too xd. I've made all the changes you pointed out.

Just for the record:
- You did not add an arch-amd64 flag file. With that I mean an empty file named `arch-amd64`, which signals to our CI and benchmark tool that it is an amd64-only solution. You added a --platform flag to docker build, instead.
- You also didn't change the sieve size in the examples in your README.md

Neither of these things are blocking, although the first will become that if we're ever able to include this solution in the automated benchmark. As things stand, I've approved this PR as it is.

sxxov:
> * You did not add an arch-amd64 flag file. With that I mean an empty file named `arch-amd64`, which signals to our CI and benchmark tool that it is an amd64-only solution. You added a --platform flag to docker build, instead.

Ah, this is my bad. Looked around but couldn't find a precise match on the internet so I did the closest, thanks for the pointer.

> * You also didn't change the sieve size in the examples in your README.md

Felt like that took away from the spirit a little, so instead I added a note above the run instructions examples telling the user that ~1000 sieve size is optimal.

rbergen:
> Ah, this is my bad. Looked around but couldn't find a precise match on the internet so I did the closest, thanks for the pointer.

No worries. I would have been very surprised of you _had_ found information about "our" arch flag files on the broader web. :)

> Felt like that took away from the spirit a little, so instead I added a note above the run instructions examples telling the user that ~1000 sieve size is optimal.

Ah, you're relying on people to read the instructions instead of skipping straight to the examples. Yes, you can indeed do that. :)

marghidanu (approved):

marghidanu:
This is very interesting!

marghidanu:
What's with the ManualDockerfile? @rbergen

PrimeBatch/solution_1/ManualDockerfile
```diff
@@ -0,0 +1,9 @@
+# not ideal to use wine, but no other choice on unix
```
marghidanu, line 1: I believe this file needs to be renamed

rbergen: No. We can't include this in the benchmark, because the runtime for a single 1000000 sieve is more than 10 hours.

rbergen: (And that's a native run on Windows, without Wine)

rbergen:
> What's with the ManualDockerfile? @rbergen

@marghidanu As already commented, we can't include this in the benchmark runs because the runtime _on native Windows, so without Wine overhead_ is more than 10 hours for a 1,000,000 sieve size benchmark run (we'd need a 750,000%-ish performance improvement, as @Sxxov commented earlier). 
As I do like the opportunity for people to still see and run this solution on their own (non-Windows) systems if they want to, I'd like to include the Dockerfile in a way that doesn't get picked up by CI or the benchmark tool. The README provides instructions on how to use the Dockerfile in a manually-triggered run.

The alternative I see would be using a "flag-file like" approach for skipping solutions altogether, but we have nothing like that in place.

marghidanu:
At this point, having the ManualDockerfile in place will exclude the solution from the benchmark suite. It's ok for now, but we definitely need to introduce a skip/ignore flag in the future. 

Otherwise, this LGTM.

PR #637, opened 2021-08-11
PowerShell Solution 3

RobCannon:
## Description
This PowerShell implementation of the sieve of Erastosthenes is derived from crowbar27's PowerShell solution #2. It improves
the speed by dereferencing object references used inside the calculation loop and other optimzations.  I also removed the 
pipelining aspect that, while a prime feature of PowerShell, do help the speed out.  In keeping with the original algorithm,
I used a BitArray that hold a bit for all number (and not just odds) which removes the need for the bit shift operation inside
the cacluation loop.  I also embraced the class, since it was a requirement of the implementation and used where appropriate. The
end result is about double the number of passes.

I also made sure this script was linux friendly and can be run from a shell prompt on a linux system as well as a Windows system.

The implementation maintains the used of the `PsObject` to the pipeline from crowbar27's implementation, which can be discarded by piping to `Out-Null`.

## Contributing requirements

### DIFF
--- a/PrimePowerShell/solution_3/Dockerfile
+++ b/PrimePowerShell/solution_3/Dockerfile
@@ -1,0 +1,7 @@
+FROM mcr.microsoft.com/powershell:7.1.3-ubuntu-20.04
+
+WORKDIR /opt/primes
+
+COPY PrimePowerShell.ps1 ./
+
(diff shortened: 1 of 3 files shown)


rbergen (approved):
LGTM

marghidanu (approved):

PR #638, opened 2021-08-13
Add Rust solution 5: Multithreading, version 2

Kulasko:
## Notes
This is a replacement for #460 as it basically is a new solution. It turned out to be a bit more code, but the structure makes things reasonable to understand, so I hope you'll bear with me. A serial algorithm is still included for the purpose of developing and testing new flag data types, but it is not used for the bench, so I hope that's okay. If not, I can remove it.

## Description
This solution aims to extend `solution_01` of Michael Barker by multiple multithreaded algorithms. Contrary to most approaches that just run independent sieves in parallel, this solution provides algorithms that work on the same sieve. It really is a shame there is no differentiation for this competition.

Currently, there are two algorithms:
- Streaming. After finding a new prime number, all threads work together on a flag unset pass. This takes no advantage of data locality whatsoever, also threads can steal memory from each other's caches between passes, so it should be bottlenecked by cache bandwidth and crosstalk latency. Also, this isn't expected to scale well in any way.
- Tiled. After a single thread fetches all the primes up to the square root of the total number, a number of threads receive the list of primes and each apply them on their part of the sieve. This has very good data locality, but as threads don't move to where the action happens, they are bound to run dry. If there are no other bottlenecks, this should approach around 50% of CPU core scaling.

Each algorithm runs on each combination of flag handling and internal data primitive that makes sense to test. Others (such as boolean with u32 elements) are tested, but not run.

## Contributing requirements


rbergen:
> A serial algorithm is still included for the purpose of developing and testing new flag data types, but it is not used for the bench, so I hope that's okay. If not, I can remove it.

That's okay. I expect this to be a stable file, so it won't contribute (much) to review workload on updates to the solution.

> It really is a shame there is no differentiation for this competition.

It's a collaborative project and we accept PRs on anything in it, in principle. So, if you or anyone else would open a PR that

- adds a clean distinction between multi-threaded approaches to the contributing guidelines, implementation output and README badges, and
- retroactively applies this distinction to all existing multi-threaded solutions

then we'll consider it as we do every other PR.

Of course, you can also open a discussion (I think an Idea would be appropriate in this case) to discuss the matter, first.

rbergen (approved):
LGTM

marghidanu (approved):

PR #640, opened 2021-08-13
PrimeSwift/solution_1 - replace ops with overflow ops & added 32bit

yellowcub:
## Description

**Operation Overflow:**
This update replaces several of the integer operations with overflow equivalents.  Swift by default performs overflow checking on operations such as `+`, `*`, `<<`, etc.  Switching to `&+`, `&*`, `&<<`, etc. permits the compiled routine to execute these operations in a manner consistent with C.

**Added UInt32 to Solution Mix:**
In previous commits, only `UInt` (`UInt64`)  as the Boolean array base type was included since the other `Uint` types resulted in fewer passes.  With the inclusion of the overflow operators, `UInt32` is notably faster than the original `UInt`; however, `UInt16` and `UInt8` were still slower.  Added `UInt32` as a separate dedicated run appears to be consistent with solutions in other languages.

## Contributing requirements


rbergen (commented):
Please remove .DS_Store from your commit. We don't merge binary files.

yellowcub:
@rbergen I think I fixed it.  Learning to use CLI.  I thought I already had `.DS_Store` in `.gitignore` when I used `git add .`.  It seems when I happen to traverse into these folders on the file browser it generates these files.  There was one in the main `Primes/` folder that was created, but I don't know that the `.gitignore` file under `PrimeSwift` can be modified to also ignore files at a higher folder level.

rbergen:
> but I don't know that the `.gitignore` file under `PrimeSwift` can be modified to also ignore files at a higher folder level.

No, it can't.

rbergen (approved):
LGTM

marghidanu (approved):

PR #641, opened 2021-08-14
[Julia] Add "unrolled modulo" implementation

louie-github:
## Description

Hello! I've decided to give up on the proposed refactor of Julia/solution_3 since I can't really see the benefits of it in retrospect, but I do still wish to add another implementation.

This "unrolled_modulo" implementation is based off [PrimeNim/solution_3](https://github.com/PlummersSoftwareLLC/Primes/tree/drag-race/PrimeNim/solution_3) by @GordonBGood, but I don't really think that I implemented exactly the same algorithm.

Rather, I read the README and noticed the line about "modulo patterns", and so I decided to try to write my own solution based on modulo patterns. I'd appreciate if they could clarify if what I'm doing is even remotely close to the "extreme loop unrolling" that they've written, but either way, I think that this is an interesting variation on the base faithful algorithm.

In short, this implementation creates 64 unrolled loop function based on every possible modulo pattern for an 8-bit array (8 for start index offset and 8 for skip value patterns), and decides which function to use through a large generated if-elseif statement inside the main factor clearing function. This speeds things up by avoiding bit shift operations, but only on CPUs which have enough speed to overcome the overhead of having to go to the correct function based on the start index and skip value.

I'm not sure whether string manipulation in this case would have been easier than mangling Julia Expr objects. It's a bit of a mess, honestly. Sorry about that.

This implementation is faster than [primes_1of2.jl](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/PrimeJulia/solution_3/primes_1of2.jl), but not as fast as Nim/solution_3. The README contains an example output.

I wanted to write a striped solution, but I need some more time to finalize it before submitting it. I wrote this "modulo unrolled" implementation in the meantime since I felt it was a fun challenge and a fun variation, even if it's not necessarily the fastest Julia implementation. That would be the "unpeeled" [PrimeJulia/solution_4](https://github.com/PlummersSoftwareLLC/Primes/tree/drag-race/PrimeJulia/solution_4) by none other than @GordonBGood. Amazing!

On another note, I wanted to release this under the MIT License, but I don't know if this is different enough from PrimeNim/solution_3 that I could consider doing that without prior permission. I think that I could probably release primes_1of2.jl under the MIT License since each implementation is simple enough that any developer could reasonably come up with the optimizations on their own? I really don't know, but I just wanted to say that if I could, I would apply the MIT License to my code.

Thank you for your time!

## Contributing requirements

### DIFF
--- a/PrimeJulia/solution_3/Dockerfile
+++ b/PrimeJulia/solution_3/Dockerfile
@@ -4,4 +4,4 @@ WORKDIR /opt/app
 
 COPY *.jl ./
 
-ENTRYPOINT [ "julia", "primes_1of2.jl" ]
+ENTRYPOINT [ "sh", "run.sh" ]
--- a/PrimeJulia/solution_3/Dockerfile
+++ b/PrimeJulia/solution_3/Dockerfile
@@ -2,6 +2,7 @@ FROM julia:1.6-buster
 
 WORKDIR /opt/app
 
-COPY *.jl ./
+COPY primes_*.jl ./
+COPY run.sh ./
 
 ENTRYPOINT [ "sh", "run.sh" ]
--- a/PrimeJulia/solution_3/README.md
+++ b/PrimeJulia/solution_3/README.md
@@ -77,8 +77,11 @@ machines because of the large code size from having to compile and
 inline 64 (8 \* 8) unrolled loop functions, as well as having to decide
 which one to jump to for each start index and skip value.
 
-To illustrate this, you can check the generated `unsafe_clear_factors!`
-function Expr object:
+It might also take some time to precompile the functions before running
+the benchmark proper, especially on older and/or slower machines.
+
+To better explain this implementation, we can take a look at the
+generated `unsafe_clear_factors!` function Expr object:
 
 ```
 julia> generate_final_loop_clearing_function() |> Base.remove_linenums!
--- a/PrimeJulia/solution_3/README.md
+++ b/PrimeJulia/solution_3/README.md
@@ -1,27 +1,50 @@
 # Julia solution by louie-github
+
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
 ![Bit count](https://img.shields.io/badge/Bits-1-green)
 ![Parallelism](https://img.shields.io/badge/Parallel-no-green)
 
-This solution is a port of [PrimeC/solution_2/sieve_1of2.c](../../PrimeC/solution_2/sieve_1of2.c)
-by Daniel Spångberg, with a few small tweaks and Julia-specific
-optimizations. This is a sort-of "low-level" style implementation in
-Julia to get as much as speed as possible out of the language. It is
-*not* designed to be idiomatic Julia code.
+This solution contains mutiple implementations:
 
-This solution requires at least **Julia 1.5** to run. Julia 1.6 is
-recommended and is used in the Docker image.
+1.  [primes_1of2.jl](primes_1of2.jl)
 
-## Description
+- This is a port of [PrimeC/solution_2/sieve_1of2.c](../../PrimeC/solution_2/sieve_1of2.c)
+  by Daniel Spångberg, with a few small tweaks and Julia-specific
+  optimizations. This is a sort-of "low-level" style implementation in
+  Julia to get as much as speed as possible out of the language. It is
+  _not_ designed to be idiomatic Julia code.
+
+2. [primes_unrolled_modulo.jl](primes_unrolled_modulo.jl)
+
+- This implementation was inspired by [PrimeNim/solution_3](../../PrimeNim/solution_3)
+  by GordonBGood. It follows a similar method of setting bits using
+  generated bit masks based on modulo patterns, but does not follow the
+  original implementation entirely, since I can't yet fully understand
+  what's going on in the macro. I simply based this solution off the
+  solution's [README.md](../../PrimeNim/solution_3/README.md), where
+  it is mentioned that there are "eight culling operations per loop"
+  "based on modulo patterns that result from stepping within a (even
+  bits) byte array".
+
+This solution has only been tested with Julia 1.6. Lower versions of
+Julia may work, but are not guaranteed to.
+
+## Descriptions
+
+### primes_1of2.jl
+
+This is currently the simplest implementation.
 
 Instead of using Julia's native `BitVector` type, this solution
 manually implements a bit array using a `Vector{UInt32}` since 32-bit
 integers seem to give the best performance. You can switch which
 unsigned integer type you want to use by changing the line:
+
 ```
 const MainUInt = UInt32
 ```
+
 Change `UInt32` to whichever unsigned integer type you wish to use.
 
 Manual bit shifting and bit masking is employed to both read and set
@@ -35,41 +58,159 @@ does with its native `BitVector` type (see
 [bitarray.jl](https://github.com/JuliaLang/julia/blob/master/base/bitarray.jl)).
 
 This implementation only stores bits for the odd numbers above 3 in an
-"inverted" bit array, i.e., bits are set when the number is *not prime*
-and bits are unset when the number is *prime*. This simplifies the
+"inverted" bit array, i.e., bits are set when the number is _not prime_
+and bits are unset when the number is _prime_. This simplifies the
 set_bit operation slightly (`arr[i] |= mask vs. arr[i] &= ~mask`).
 
+### primes_unrolled_modulo.jl
+
+This implementation is similar to primes_1of2.jl, except that it
+pre-generates every possible bit mask for all the starting indices and
+skip values using modulo patterns.
+
+It then uses an unrolled loop to set the corresponding bits using the
+calculated bit masks. This saves some time since we don't have to do
+a modulo and bit shift each time we set a bit.
+
+To illustrate this, you can check the generated `unsafe_clear_factors!`
+function Expr object:
+
+```
+julia> generate_final_loop_clearing_function() |> Base.remove_linenums!
+quote
+    function unsafe_clear_factors!(arr::Vector{MainUInt}, start::Integer, skip::Integer, stop::Integer)
+        start_modulo = start & 7
+        skip_modulo = skip & 7
+        index = start
+        if stop > skip * 7
+            unrolled_stop = stop - skip * 7
+            if start_modulo == 0
+                if skip_modulo == 0
+                    index = _unrolled_clear_bits_start_0_skip_0(arr, start, skip, unrolled_stop)
+                elseif skip_modulo == 1
+                    index = _unrolled_clear_bits_start_0_skip_1(arr, start, skip, unrolled_stop)
+                elseif skip_modulo == 2
+                    index = _unrolled_clear_bits_start_0_skip_2(arr, start, skip, unrolled_stop)
+                elseif skip_modulo == 3
+                    index = _unrolled_clear_bits_start_0_skip_3(arr, start, skip, unrolled_stop)
+                elseif skip_modulo == 4
+                    index = _unrolled_clear_bits_start_0_skip_4(arr, start, skip, unrolled_stop)
+                elseif skip_modulo == 5
+                    index = _unrolled_clear_bits_start_0_skip_5(arr, start, skip, unrolled_stop)
+                elseif skip_modulo == 6
+                    index = _unrolled_clear_bits_start_0_skip_6(arr, start, skip, unrolled_stop)
+                elseif skip_modulo == 7
+                    index = _unrolled_clear_bits_start_0_skip_7(arr, start, skip, unrolled_stop)
+                end
+            elseif start_modulo == 1
+                if skip_modulo == 0
+                    index = _unrolled_clear_bits_start_1_skip_0(arr, start, skip, unrolled_stop)
+                ...
+```
+
+Looking at an example unrolled bit clearing function, we get:
+
+```
+julia> generate_loop_clearing_function(0, 3, get_modulo_pattern(0, 3, UINT_BIT_LENGTH)) |> Base.remove_linenums!
+quote
+    #= /home/louie/dev/PrimesTemp/PrimeJulia/solution_3/primes_unrolled_modulo.jl:91 =# @inline function _unrolled_clear_bits_start_0_skip_3(arr::Vector{MainUInt}, start::Integer, skip::Integer, unrolled_stop::Integer)
+            index = start
+            #= /home/louie/dev/PrimesTemp/PrimeJulia/solution_3/primes_unrolled_modulo.jl:95 =# @inbounds while index < unrolled_stop
+                    begin
+                        arr[(index + skip * 0) >> 3 + 1] |= 1
+                        arr[(index + skip * 1) >> 3 + 1] |= 8
+                        arr[(index + skip * 2) >> 3 + 1] |= 64
+                        arr[(index + skip * 3) >> 3 + 1] |= 2
+                        arr[(index + skip * 4) >> 3 + 1] |= 16
+                        arr[(index + skip * 5) >> 3 + 1] |= 128
+                        arr[(index + skip * 6) >> 3 + 1] |= 4
+                        arr[(index + skip * 7) >> 3 + 1] |= 32
+                    end
+                    index += skip * 8
+                end
+            return index
+        end
+end
+```
+
+As you can see, the generated functions simply end up creating an
+unrolled loop with pre-calculated bit masks based on the start index
+and the skip value.
+
+These bit masks are pre-calculated based on every possible modulo value
+of both the start index and the skip value. For a bit array using
+integers of bit length _N_, we will have to generate _N_ \* _N_
+unrolled loop functions (_N_ possible start index offsets with _N_
+possible skip value bit mask patterns for each offset).
+
+To decide which unrolled loop function to use, the generated
+`unsafe_clear_factors!` function uses a large if-elseif statement
+(since Julia does not have a switch-case statement) to calculate
+the modulo of the start index and the skip value, and then finds the
+corresponding function. This _should_ be optimized into a computed
+goto by the compiler, but I'm not entirely sure.
+
+This is why the implementation uses a `Vector{UInt8}` so we only have
+to calculate 8 \* 8 or 64 functions. If we used a `Vector{UInt32}`, for
+example, we would have to calculate 32 \* 32 or 1024 functions, each of
+which takes time for Julia to precompile. Plus, the generated code size
+from the the inlined functions would be enormous.
+
+The other parts of running the sieve, such as reading the bits in the
+array and finding the next factor index, are mostly unchanged from
+[primes_1of2.jl](primes_1of2.jl).
+
+## Suggestions and improvements
+
 If you see any room for improvement in the code or have any
-suggestions, don't hesitate to open an issue, pull request (PR), 
+suggestions, don't hesitate to open an issue, pull request (PR),
 Discussion, or the like. Don't forget to tag me at `@louie-github` so I
 can be notified if my personal input is either wanted or needed.
 I'm open to fixing stylistic issues or discussing cosmetic changes to
 the code, but for those, it might be best if you open a Discussion
 first before opening an issue or pull request.
 
-
 ## Run instructions
 
 ### Running locally
 
-First, make sure that you have installed
-[Julia 1.5 or newer](https://julialang.org/downloads/) and have
-verified that your installation works.
+First, make sure that you have installed [Julia](https://julialang.org/downloads/)
+and have verified that your installation works.
 
-To build and run the solution locally, run the following command:
+To build and run a single implementation locally, run the following
+command, replacing `primes_1of2.jl` with the filename of the
+implementation you want to run:
+
 ```
 julia primes_1of2.jl
 ```
 
 Optionally, you can specify a sieve size as an additional command line argument:
+
 ```
 julia primes_1of2.jl 1000
 ```
 
 You can also specify the duration to run the benchmark for after the
 sieve size:
+
 ```
 julia primes_1of2.jl 1000000 10
+```
+
+### Running all implementations at once locally
+
+If you want to run all implementations at once, run:
+
+```
+./run.sh
+```
+
+You can also pass in a sieve size and/or duration as additional
+arguments. These will be passed through to all implementations.
+
+```
+./run.sh 1000000 10
 ```
 
 ### Running in Docker
@@ -84,17 +225,23 @@ docker build --pull -t primejulia-3 .
 ```
 
 2. Run the `primejulia-3` image as a container.
+
 ```
 docker run -it --rm primejulia-3
 ```
+
+This will run all implementations, as it uses [run.sh](run.sh).
+
 You can also pass in a sieve size and/or a duration as additional
 arguments similar to when you run the solution locally:
+
 ```
 docker run -it --rm primejulia-3 1000
 docker run -it --rm primejulia-3 1000000 10
 ```
 
 3. If you want to remove the built Docker image, run:
+
 ```
 docker image rm primejulia-3
 ```
@@ -105,7 +252,6 @@ importantly ARM64 and x86_64. This also helps avoid any possible issues
 caused by Julia's [Tier 3 support for musl](https://julialang.org/downloads/#currently_supported_platforms),
 including Alpine Linux.
 
-
 ## Validating results
 
 The validation code has been moved to a separate file:
@@ -115,32 +261,43 @@ Since 32-bit systems cannot represent the final element of the results
 dictionary (10^10 or 10000000000) in a single `UInt32`, this number is
 skipped when running validate.jl on 32 bit systems.
 
-To validate the results given by [primes_1of2.jl](primes_1of2.jl), run:
+To validate the results given by an implementation, run the following
+command, replacing `primes_1of2.jl` with the implementation you wish
+to test:
+
 ```
-julia validate.jl
+julia validate.jl primes_1of2.jl
 ```
+
 Note that this may take a few minutes to complete.
 
-If you want to verify that primes_1of2.jl does not go out-of-bounds
-when accessing memory since it uses `@inbounds` quite liberally, run
-either primes_1of2.jl or validate.jl with `julia --check-bounds=yes`.
+If you want to verify that an implementation does not go out-of-bounds
+when accessing memory since all of them use the `@inbounds` macro quite
+liberally, run either the implementation `.jl` file or validate.jl with
+`julia --check-bounds=yes`.
+
 ```
 julia --check-bounds=yes primes_1of2.jl
 julia --check-bounds=yes validate.jl
 ```
-Note that this will make the code run a fair bit slower.
 
+Note that this will make the code run a fair bit slower.
 
 ## Output
 
-This is the output from running the solution on a machine with an Intel
-Core i5-9300H CPU and 24 GB of RAM running Windows 10 Home:
+This is an example output from running the solution on a machine with an
+Intel Core i5-9300H CPU and 24 GB of RAM running Ubuntu 20.04 LTS under
+WSL 2 on Windows 10 Home:
+
 ```
-PS D:\Office Files\Programming\Primes> julia primes_1of2.jl
+❯ ./run.sh
 Settings: sieve_size = 1000000 | duration = 5
 Number of trues: 78498
-primes_1of2.jl: Passes: 9146 | Elapsed: 5.0 | Passes per second: 1829.2 | Average pass duration: 0.0005466870763175158
-louie-github_port_1of2;9146;5.0;1;algorithm=base,faithful=yes,bits=1
+primes_1of2.jl: Passes: 10276 | Elapsed: 5.00010085105896 | Passes per second: 2055.1585470168407 | Average pass duration: 0.00048658046429145193
+louie-github_port_1of2;10276;5.00010085105896;1;algorithm=base,faithful=yes,bits=1
+
+Settings: sieve_size = 1000000 | duration = 5
+Number of trues: 78498
+primes_unrolled_modulo.jl: Passes: 15787 | Elapsed: 5.000248908996582 | Passes per second: 3157.2428267712044 | Average pass duration: 0.0003167320522579706
+louie-github_unrolled_modulo;15787;5.000248908996582;1;algorithm=base,faithful=yes,bits=1
 ```
-On said machine, when no heavy tasks are running, the number of passes
-usually ranges between 8900 and 9300.
--- a/PrimeJulia/solution_3/README.md
+++ b/PrimeJulia/solution_3/README.md
@@ -5,7 +5,7 @@
 ![Bit count](https://img.shields.io/badge/Bits-1-green)
 ![Parallelism](https://img.shields.io/badge/Parallel-no-green)
 
-This solution contains mutiple implementations:
+This solution contains multiple implementations:
 
 1.  [primes_1of2.jl](primes_1of2.jl)
 
@@ -68,9 +68,14 @@ This implementation is similar to primes_1of2.jl, except that it
 pre-generates every possible bit mask for all the starting indices and
 skip values using modulo patterns.
 
-It then uses an unrolled loop to set the corresponding bits using the
-calculated bit masks. This saves some time since we don't have to do
-a modulo and bit shift each time we set a bit.
+It then uses an unrolled loop wrapped in a function to set the
+corresponding bits using the calculated bit masks. This saves some time
+since we don't have to do a modulo and bit shift each time we set a bit.
+
+This implementation will probably run a lot slower, though, on older
+machines because of the large code size from having to compile and
+inline 64 (8 \* 8) unrolled loop functions, as well as having to decide
+which one to jump to for each start index and skip value.
 
 To illustrate this, you can check the generated `unsafe_clear_factors!`
 function Expr object:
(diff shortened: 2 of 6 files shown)


rbergen:
@louie-github Could you please take me through the if and how of your implementation clears the composites (multiples of factors) individually, and doesn't clear multiple composites with a single write? I ask this because the individual clearing is a requirement for a base implementation.

louie-github:
I understand. If you look at the PrimeSieve constructor, we see that it has almost the same code as primes_1of2.jl:
```
# This is more accurate than _div_uint_size(_div2(i)) + 1
@inline _get_num_uints(i::Integer) = _div_uint_length(
    # We only store odd numbers starting from 3.
    _div2(i + (2 * UINT_BIT_LENGTH - 1))
)

function PrimeSieve(sieve_size::UInt)
    return PrimeSieve(sieve_size, zeros(MainUInt, _get_num_uints(sieve_size)))
end
```

This should give a hint that the bit vector still stores bits in the same way, it's just that the way we set the bits is different.

Later on in the `run_sieve!` function, again, we see almost the same pattern as primes_1of2.jl, where we clear factors by setting a start index (factor * factor, mapped to index), a skip value (factor, which maps to 2 * factor), and a stop index (sieve_size mapped to index):
```
@inline function unsafe_clear_factors!(arr::Vector{MainUInt}, factor_index::Integer, max_index::Integer)
    factor = _map_to_factor(factor_index)
    start_index = _div2(factor * factor)
    unsafe_clear_factors!(arr, start_index, factor, max_index)
end

function run_sieve!(sieve::PrimeSieve)
    is_not_prime = sieve.is_not_prime
    sieve_size = sieve.sieve_size
    max_bits_index = _map_to_index(sieve_size)
    max_factor_index = _map_to_index(unsafe_trunc(UInt, sqrt(sieve_size)))
    factor_index = UInt(1) # 0 => 1, 1 => 3, 1 => 5, 2 => 7, ...
    @inbounds while factor_index <= max_factor_index
        if iszero(
            is_not_prime[factor_index >> UINT_BIT_SHIFT + 1] &
            (1 << (factor_index & (UINT_BIT_LENGTH - 1)))
        )
            unsafe_clear_factors!(is_not_prime, factor_index, max_bits_index)
        end
        factor_index += 1
    end
    return is_not_prime
end
```

So, the main problem is how `unsafe_clear_factors!` actually clears the bits. Now, to be fair, this is rather difficult to see with the code, since we have to look at the Expr objects generated on the fly.

As said in the README, if we attempt to inspect the `unsafe_clear_factors!` function generated by `generate_final_loop_clearing_function()`, we get, and this is the entire output:

<details open>
<summary>
Output of generate_final_loop_clearing_function():
</summary>

```
julia> include("primes_unrolled_modulo.jl")

julia> generate_final_loop_clearing_function() |> Base.remove_linenums!
quote
    function unsafe_clear_factors!(arr::Vector{MainUInt}, start::Integer, skip::Integer, stop::Integer)
        start_modulo = start & 7
        skip_modulo = skip & 7
        index = start
        if stop > skip * 7
            unrolled_stop = stop - skip * 7
            if start_modulo == 0
                if skip_modulo == 0
                    index = _unrolled_clear_bits_start_0_skip_0(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 1
                    index = _unrolled_clear_bits_start_0_skip_1(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 2
                    index = _unrolled_clear_bits_start_0_skip_2(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 3
                    index = _unrolled_clear_bits_start_0_skip_3(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 4
                    index = _unrolled_clear_bits_start_0_skip_4(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 5
                    index = _unrolled_clear_bits_start_0_skip_5(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 6
                    index = _unrolled_clear_bits_start_0_skip_6(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 7
                    index = _unrolled_clear_bits_start_0_skip_7(arr, start, skip, unrolled_stop)
                end
            elseif start_modulo == 1
                if skip_modulo == 0
                    index = _unrolled_clear_bits_start_1_skip_0(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 1
                    index = _unrolled_clear_bits_start_1_skip_1(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 2
                    index = _unrolled_clear_bits_start_1_skip_2(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 3
                    index = _unrolled_clear_bits_start_1_skip_3(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 4
                    index = _unrolled_clear_bits_start_1_skip_4(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 5
                    index = _unrolled_clear_bits_start_1_skip_5(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 6
                    index = _unrolled_clear_bits_start_1_skip_6(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 7
                    index = _unrolled_clear_bits_start_1_skip_7(arr, start, skip, unrolled_stop)
                end
            elseif start_modulo == 2
                if skip_modulo == 0
                    index = _unrolled_clear_bits_start_2_skip_0(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 1
                    index = _unrolled_clear_bits_start_2_skip_1(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 2
                    index = _unrolled_clear_bits_start_2_skip_2(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 3
                    index = _unrolled_clear_bits_start_2_skip_3(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 4
                    index = _unrolled_clear_bits_start_2_skip_4(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 5
                    index = _unrolled_clear_bits_start_2_skip_5(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 6
                    index = _unrolled_clear_bits_start_2_skip_6(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 7
                    index = _unrolled_clear_bits_start_2_skip_7(arr, start, skip, unrolled_stop)
                end
            elseif start_modulo == 3
                if skip_modulo == 0
                    index = _unrolled_clear_bits_start_3_skip_0(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 1
                    index = _unrolled_clear_bits_start_3_skip_1(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 2
                    index = _unrolled_clear_bits_start_3_skip_2(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 3
                    index = _unrolled_clear_bits_start_3_skip_3(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 4
                    index = _unrolled_clear_bits_start_3_skip_4(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 5
                    index = _unrolled_clear_bits_start_3_skip_5(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 6
                    index = _unrolled_clear_bits_start_3_skip_6(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 7
                    index = _unrolled_clear_bits_start_3_skip_7(arr, start, skip, unrolled_stop)
                end
            elseif start_modulo == 4
                if skip_modulo == 0
                    index = _unrolled_clear_bits_start_4_skip_0(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 1
                    index = _unrolled_clear_bits_start_4_skip_1(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 2
                    index = _unrolled_clear_bits_start_4_skip_2(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 3
                    index = _unrolled_clear_bits_start_4_skip_3(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 4
                    index = _unrolled_clear_bits_start_4_skip_4(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 5
                    index = _unrolled_clear_bits_start_4_skip_5(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 6
                    index = _unrolled_clear_bits_start_4_skip_6(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 7
                    index = _unrolled_clear_bits_start_4_skip_7(arr, start, skip, unrolled_stop)
                end
            elseif start_modulo == 5
                if skip_modulo == 0
                    index = _unrolled_clear_bits_start_5_skip_0(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 1
                    index = _unrolled_clear_bits_start_5_skip_1(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 2
                    index = _unrolled_clear_bits_start_5_skip_2(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 3
                    index = _unrolled_clear_bits_start_5_skip_3(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 4
                    index = _unrolled_clear_bits_start_5_skip_4(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 5
                    index = _unrolled_clear_bits_start_5_skip_5(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 6
                    index = _unrolled_clear_bits_start_5_skip_6(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 7
                    index = _unrolled_clear_bits_start_5_skip_7(arr, start, skip, unrolled_stop)
                end
            elseif start_modulo == 6
                if skip_modulo == 0
                    index = _unrolled_clear_bits_start_6_skip_0(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 1
                    index = _unrolled_clear_bits_start_6_skip_1(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 2
                    index = _unrolled_clear_bits_start_6_skip_2(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 3
                    index = _unrolled_clear_bits_start_6_skip_3(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 4
                    index = _unrolled_clear_bits_start_6_skip_4(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 5
                    index = _unrolled_clear_bits_start_6_skip_5(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 6
                    index = _unrolled_clear_bits_start_6_skip_6(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 7
                    index = _unrolled_clear_bits_start_6_skip_7(arr, start, skip, unrolled_stop)
                end
            elseif start_modulo == 7
                if skip_modulo == 0
                    index = _unrolled_clear_bits_start_7_skip_0(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 1
                    index = _unrolled_clear_bits_start_7_skip_1(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 2
                    index = _unrolled_clear_bits_start_7_skip_2(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 3
                    index = _unrolled_clear_bits_start_7_skip_3(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 4
                    index = _unrolled_clear_bits_start_7_skip_4(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 5
                    index = _unrolled_clear_bits_start_7_skip_5(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 6
                    index = _unrolled_clear_bits_start_7_skip_6(arr, start, skip, unrolled_stop)
                elseif skip_modulo == 7
                    index = _unrolled_clear_bits_start_7_skip_7(arr, start, skip, unrolled_stop)
                end
            end
        end
        _unrolled_clear_bits_finalizer(arr, index, skip, stop)
    end
end
```

</details>


This is what I was referring to when I meant that the code had to calculate the right function to call from 64 different unrolled loop clearing functions.

We can look at the generated output of `generate_loop_clearing_function` with certain arguments to check the content of each `_unrolled_clear_bits_start_N_skip_N` function, since that is how we generate each one:

```
julia> generate_loop_clearing_function(0, 3, get_modulo_pattern(0, 3, UINT_BIT_LENGTH)) |> Base.remove_linenums!
quote
    #= /home/louie/dev/PrimesTemp/PrimeJulia/solution_3/primes_unrolled_modulo.jl:87 =# @inline function _unrolled_clear_bits_start_0_skip_3(arr::Vector{MainUInt}, start::Integer, skip::Integer, unrolled_stop::Integer)
            index = start
            #= /home/louie/dev/PrimesTemp/PrimeJulia/solution_3/primes_unrolled_modulo.jl:91 =# @inbounds while index < unrolled_stop
                    begin
                        arr[(index + skip * 0) >> 3 + 1] |= 1
                        arr[(index + skip * 1) >> 3 + 1] |= 8
                        arr[(index + skip * 2) >> 3 + 1] |= 64
                        arr[(index + skip * 3) >> 3 + 1] |= 2
                        arr[(index + skip * 4) >> 3 + 1] |= 16
                        arr[(index + skip * 5) >> 3 + 1] |= 128
                        arr[(index + skip * 6) >> 3 + 1] |= 4
                        arr[(index + skip * 7) >> 3 + 1] |= 32
                    end
                    index += skip * 8
                end
            return index
        end
end
```

As you can see, we're just setting each bit like normal, just with a pre-calculated bit mask since we know the modulo patterns of the start and skip values. It's just an unrolled loop, except without having to calculate the bit mask.

This is the snippet of code from `generate_loop_clearing_function` that actually generates each of these lines:
```
    for (skip_iter, bit_index) in enumerate(modulo_pattern)
        push!(body_expr.args, :(
            arr[(index + skip * $(skip_iter - 1)) >> $UINT_BIT_SHIFT + 1] |=
            $(1 << bit_index)
        ))
    end
```

GordonBGood:
@louie-github:

I was automatically notified about this PR, I suppose because my "handle" is used in the text.  You have the right idea, as this PR appears to use the principles of "Extreme Loop Unrolling".  I have been thinking about doing something similar in Julia, but likely would have used a macro to do the "Expr mangling" in order to get it all in one small area of the code.

I expect this is slower than expected because of the overheads as you mention, made worse by Julia not handling the type signatures for functions very well - if all the optimizations worked as they should, this should come close to my Nim solution_3 contribution, probably not quite there because Julia does some miss-optimizations due to the one-based rather than zero-based indices and misses a few other things.

One of the reasons that I wanted to implement this is to see if Julia would optimize away the chain of if/elseif/... conditionals into a single computed goto as we discussed before, but so far I can't see that happening.

A hint that @rbergen gave me in order to make it obvious that your implementation is not cheating:  organize the code so that the second culling/marking loop based on the single base prime value found in the outer loop calls a macro/function something like `setbits(sievebuffer, startindex, stepsize, limitindex)` which then makes it obvious that it is marking composite representations bit by bit; that is what I ended up doing for the Nim implementation.

As to the speed, for this trivial cull over the entire sieve buffer per base prime value, we don't actually need 64 patterns as only 4 of them are being used, which can greatly shorten the time to find the right pattern and make the solution much faster for Julia; however it is a kind of special case adapted only for this trivial benchmark where page segmentation is not allowed so I don't know if it will be accepted or not.

If Julia would only properly optimize the selection of the proper pattern without going though an extended series of conditions (ie. by a computed goto), then constructing functions for each pattern would not be necessary and the resulting code would be almost, but not quite, as fast as my Nim solution_3 contribution due to Julia not optimizing quite to that level.  If I cut the patterns down to the actually only used four, then my Julia code implementing this is only about 5 per cent slower than my Nim equivalent.

However, I don't think your code will get there, as I don't know that you are helping the compiler enough to get to that level of optimization:

```julia
                        arr[(index + skip * 0) >> 3 + 1] |= 1
                        arr[(index + skip * 1) >> 3 + 1] |= 8
                        arr[(index + skip * 2) >> 3 + 1] |= 64
                        arr[(index + skip * 3) >> 3 + 1] |= 2
                        arr[(index + skip * 4) >> 3 + 1] |= 16
                        arr[(index + skip * 5) >> 3 + 1] |= 128
                        arr[(index + skip * 6) >> 3 + 1] |= 4
                        arr[(index + skip * 7) >> 3 + 1] |= 32
```

You've got the constant mask values worked out, but the "sum of indexes" being based on the base `index` still has multiplying, shifting, and summing operations that theoretically could be optimized away but probably aren't.  In my Nim macro, the index value used in the marking is a byte index instead of a bit index and each of the offsets from that base byte index are constants that are pre-calculated, which in the case of use with Julia with one based indices can have the offset of one already applied.  One of the missed optimizations I then saw was that it still applies the one offset before checking for loop completion at a cost of a cycle or two per loop.

given the byte index has already been offset by one, then the above snippet then becomes as follows:

```julia
                        arr[(byteIndex] |= 1
                        arr[byteIndex + r1] |= 8
                        arr[byteIndex + r2] |= 64
                        arr[byteIndex + r3] |= 2
                        arr[byteIndex + r4] |= 16
                        arr[byteIndex + r5] |= 128
                        arr[byteIndex + r6] |= 4
                        arr[byteIndex + r7] |= 32
```
where r1 .. r7 are pre-computed constant offsets for the give base prime value and pattern start index the the loop terminating value is when `byteIndex` gets to the sieve buffer size minus `r7`, which when reached just leaves the average of 3.5 culls left over that won't fit into an entire eight-cull loop to be culled/marked in a "clean up" loop.

rbergen:
> however it is a kind of special case adapted only for this trivial benchmark where page segmentation is not allowed so I don't know if it will be accepted or not.

@GordonBGood What's not accepted is the algorithm using the following prior knowledge:

- that any number in the sieve is a prime, with the exception of 2
- what the sieve size is, as that should be considered a run-time parameter

The fact that it concerns a trivial benchmark, as you call it, is inherent to the project. I would say that optimizing against that is allowed. Of course, that's true as long as the arbitrary rules, as multiple people have labeled them, are respected.

rbergen:
@louie-github I'm wondering if you would like to use any of @GordonBGood's comments to modify your implementation? If so, I'll hold on my actual review until further notice.

louie-github:
@rbergen You're right, I'd probably want to use the new information to improve the solution first. Thanks!

@GordonBGood: Thanks for replying!
> I was automatically notified about this PR, I suppose because my "handle" is used in the text.

I meant to do that, as I did want to notify you. I do apologize if this did inconvenience you, though.

> One of the reasons that I wanted to implement this is to see if Julia would optimize away the chain of if/elseif/... conditionals into a single computed goto as we discussed before, but so far I can't see that happening.

I don't really understand Assembly all to well, but it looks to me like Julia's doing *something* here with that `jmpq` instruction which also appears in many of the inlined functions:
```
julia> code_native(unsafe_clear_factors!, (Vector{UInt8}, UInt, UInt, UInt))
        .text
; ┌ @ primes_unrolled_modulo.jl:153 within `unsafe_clear_factors!'
; │┌ @ int.jl:923 within `*' @ primes_unrolled_modulo.jl:147
        pushq   %r15
        pushq   %r14
        pushq   %r13
        pushq   %r12
        pushq   %rbx
; ││ @ int.jl:923 within `*' @ int.jl:88
        leaq    (,%rdx,8), %r8
        movq    %r8, %r10
        subq    %rdx, %r10
; │└
; │┌ @ operators.jl:305 within `>'
; ││┌ @ int.jl:441 within `<'
        cmpq    %rcx, %r10
; │└└
        jae     L6886
; │ @ primes_unrolled_modulo.jl within `unsafe_clear_factors!'
        movl    %edx, %r11d
        andl    $7, %r11d
; │ @ primes_unrolled_modulo.jl:150 within `unsafe_clear_factors!'
; │┌ @ int.jl:923 within `&' @ int.jl:309
        movl    %esi, %ebx
        andl    $7, %ebx
; │└
; │ @ primes_unrolled_modulo.jl:154 within `unsafe_clear_factors!'
; │┌ @ int.jl:923 within `*' @ int.jl:88
        movq    %rdx, %r9
        subq    %r8, %r9
; │└
; │┌ @ int.jl:86 within `-'
        addq    %rcx, %r9
        movabsq $.rodata, %rax
        jmpq    *(%rax,%rbx,8)
; │└
; │ @ int.jl within `unsafe_clear_factors!'
        movabsq $140203258448752, %rax          # imm = 0x7F839D6A8B70
        jmpq    *(%rax,%r11,8)
; │ @ primes_unrolled_modulo.jl:128 within `unsafe_clear_factors!'
; │┌ @ primes_unrolled_modulo.jl:91 within `_unrolled_clear_bits_start_0_skip_0'
; ││┌ @ int.jl:441 within `<'
        cmpq    %rsi, %r9
; ││└
        jbe     L6886
; │└
; │┌ @ primes_unrolled_modulo.jl:92 within `_unrolled_clear_bits_start_0_skip_0'
; ││┌ @ abstractarray.jl:1173 within `getindex' @ array.jl:0
        movq    (%rdi), %rax
; ││└
; ││┌ @ int.jl:462 within `>>' @ int.jl:456
; ...
```

This `code_native` output is so long I have to press Ctrl+C as fast as possible to get this to stop and for the scrollback to reach the start. It's a lot.

> A hint that @rbergen gave me in order to make it obvious that your implementation is not cheating: organize the code so that the second culling/marking loop based on the single base prime value found in the outer loop calls a macro/function something like setbits(sievebuffer, startindex, stepsize, limitindex) which then makes it obvious that it is marking composite representations bit by bit; that is what I ended up doing for the Nim implementation.

I did do that, I think, here:
```
@inline function unsafe_clear_factors!(arr::Vector{MainUInt}, factor_index::Integer, max_index::Integer)
    factor = _map_to_factor(factor_index)
    start_index = _div2(factor * factor)
    unsafe_clear_factors!(arr, start_index, factor, max_index)
end
```

> You've got the constant mask values worked out, but the "sum of indexes" being based on the base index still has multiplying, shifting, and summing operations that theoretically could be optimized away but probably aren't.

I believe you're right. Reading this, I think Julia has optimized the loop to be adding `(skip * N)` constants, shifting right, then accessing the array (without having to `inc` the index first):
```
L2538:
        movq    %rsi, %rbx
        shrq    $3, %rbx
; ││└
; ││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
        orb     $4, (%rax,%rbx)
; ││└
; ││┌ @ int.jl:462 within `>>' @ int.jl:456
        leaq    (%rdx,%rsi), %rbx
        shrq    $3, %rbx
; ││└
; ││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
        orb     $-128, (%rax,%rbx)
; ││└
; ││┌ @ int.jl:462 within `>>' @ int.jl:456
        leaq    (%r11,%rsi), %rbx
        shrq    $3, %rbx
; ││└
; ││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
        orb     $16, (%rax,%rbx)
; ││└
; ││┌ @ int.jl:462 within `>>' @ int.jl:456
        leaq    (%r14,%rsi), %rbx
        shrq    $3, %rbx
; ││└
; ││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
        orb     $2, (%rax,%rbx)
; ││└
; ││┌ @ int.jl:462 within `>>' @ int.jl:456
        leaq    (%r15,%rsi), %rbx
        shrq    $3, %rbx
; ││└
; ││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
        orb     $64, (%rax,%rbx)
; ││└
; ││┌ @ int.jl:462 within `>>' @ int.jl:456
        leaq    (%r12,%rsi), %rbx
        shrq    $3, %rbx
; ││└
; ││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
        orb     $8, (%rax,%rbx)
; ││└
; ││┌ @ int.jl:462 within `>>' @ int.jl:456
        leaq    (%rsi,%r13), %rbx
        shrq    $3, %rbx
; ││└
; ││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
        orb     $1, (%rax,%rbx)
; ││└
; ││┌ @ int.jl:462 within `>>' @ int.jl:456
        leaq    (%r10,%rsi), %rbx
        shrq    $3, %rbx
; ││└
; ││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
        orb     $32, (%rax,%rbx)
; │└└
; │┌ @ primes_unrolled_modulo.jl:93 within `_unrolled_clear_bits_start_2_skip_5'
; ││┌ @ int.jl:87 within `+'
        addq    %r8, %rsi
; │└└
; │┌ @ primes_unrolled_modulo.jl:91 within `_unrolled_clear_bits_start_2_skip_5'
; ││┌ @ int.jl:441 within `<'
        cmpq    %r9, %rsi
; ││└
        jb      L2538
        jmp     L6886
; │└
; │┌ @ primes_unrolled_modulo.jl:91 within `_unrolled_clear_bits_start_2_skip_6'
; ││┌ @ int.jl:441 within `<'
        cmpq    %rsi, %r9
; ││└
        jbe     L6886
; │└
; │┌ @ primes_unrolled_modulo.jl:92 within `_unrolled_clear_bits_start_2_skip_6'
; ││┌ @ abstractarray.jl:1173 within `getindex' @ array.jl:0
        movq    (%rdi), %rax
        leaq    (%rdx,%rdx), %r8
        leaq    (%rdx,%rdx,2), %r10
        leaq    (,%rdx,4), %r11
; ││└
```

It could probably be made faster using `byteIndex` and constant adds, as you said. I'll look into that.

Thanks for the insights!

GordonBGood:
@louie-github:

> > One of the reasons that I wanted to implement this is to see if Julia would optimize away the chain of if/elseif/... conditionals into a single computed goto as we discussed before, but so far I can't see that happening.
> 
> I don't really understand Assembly all to well, but it looks to me like Julia's doing _something_ here with that `jmpq` instruction which also appears in many of the inlined functions...
> 
>         jmpq    *(%rax,%r11,8)

That's what we are looking for!  In that case, you have a computed goto, so then you don't have to generate functions and can just inline the code, but I suppose inline functions will do as well.

> > You've got the constant mask values worked out, but the "sum of indexes" being based on the base index still has multiplying, shifting, and summing operations that theoretically could be optimized away but probably aren't.
> 
> I believe you're right. Reading this, I think Julia has optimized the loop to be adding `(skip * N)` constants, shifting right, then accessing the array (without having to `inc` the index first):
> 
> ```
> L2538:
>         movq    %rsi, %rbx
>         shrq    $3, %rbx
> ; ││└
> ; ││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
>         orb     $4, (%rax,%rbx)
> ; ││└
> ; ││┌ @ int.jl:462 within `>>' @ int.jl:456
>         leaq    (%rdx,%rsi), %rbx
>         shrq    $3, %rbx
> ; ││└
> ; ││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
>         orb     $-128, (%rax,%rbx)
> ; ││└
> ; ││┌ @ int.jl:462 within `>>' @ int.jl:456
>         leaq    (%r11,%rsi), %rbx
>         shrq    $3, %rbx
> ; ││└
> ; ││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
>         orb     $16, (%rax,%rbx)
> ; ││└
> ; ││┌ @ int.jl:462 within `>>' @ int.jl:456
>         leaq    (%r14,%rsi), %rbx
>         shrq    $3, %rbx
> ; ││└
> ; ││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
>         orb     $2, (%rax,%rbx)
> ; ││└
> ; ││┌ @ int.jl:462 within `>>' @ int.jl:456
>         leaq    (%r15,%rsi), %rbx
>         shrq    $3, %rbx
> ; ││└
> ; ││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
>         orb     $64, (%rax,%rbx)
> ; ││└
> ; ││┌ @ int.jl:462 within `>>' @ int.jl:456
>         leaq    (%r12,%rsi), %rbx
>         shrq    $3, %rbx
> ; ││└
> ; ││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
>         orb     $8, (%rax,%rbx)
> ; ││└
> ; ││┌ @ int.jl:462 within `>>' @ int.jl:456
>         leaq    (%rsi,%r13), %rbx
>         shrq    $3, %rbx
> ; ││└
> ; ││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
>         orb     $1, (%rax,%rbx)
> ; ││└
> ; ││┌ @ int.jl:462 within `>>' @ int.jl:456
>         leaq    (%r10,%rsi), %rbx
>         shrq    $3, %rbx
> ; ││└
> ; ││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
>         orb     $32, (%rax,%rbx)
> ; │└└
> ; │┌ @ primes_unrolled_modulo.jl:93 within `_unrolled_clear_bits_start_2_skip_5'
> ; ││┌ @ int.jl:87 within `+'
>         addq    %r8, %rsi
> ; │└└
> ; │┌ @ primes_unrolled_modulo.jl:91 within `_unrolled_clear_bits_start_2_skip_5'
> ; ││┌ @ int.jl:441 within `<'
>         cmpq    %r9, %rsi
> ; ││└
>         jb      L2538
>         jmp     L6886
> ; │└
> ; │┌ @ primes_unrolled_modulo.jl:91 within `_unrolled_clear_bits_start_2_skip_6'
> ; ││┌ @ int.jl:441 within `<'
>         cmpq    %rsi, %r9
> ; ││└
>         jbe     L6886
> ; │└
> ; │┌ @ primes_unrolled_modulo.jl:92 within `_unrolled_clear_bits_start_2_skip_6'
> ; ││┌ @ abstractarray.jl:1173 within `getindex' @ array.jl:0
>         movq    (%rdi), %rax
>         leaq    (%rdx,%rdx), %r8
>         leaq    (%rdx,%rdx,2), %r10
>         leaq    (,%rdx,4), %r11
> ; ││└
> ```
> 
> It could probably be made faster using `byteIndex` and constant adds, as you said. I'll look into that.

Yes, if you get it right each loop should consist of nothing but a list of eight `orb` instructions with addressing mode by just a pair of registers, the base one and the offset one and minimum loop code to just add the step value to the base byte index, check if it has reached the loop limit, and if not branch back to the loop label.  I don't know that the Julia optimizer well get quite to the minimum instructions, but it can come close.

> Thanks for the insights!

You're welcome.  If you get this right, it means I won't have to get far enough into Julia to complete it myself.  I would highly recommend looking into doing the whole thing with a macro or two as I think that the code can then be cut down to a few dozen lines that are reasonably clear, something like the Nim macro I wrote.  Julia's handling of Abstract Syntax Tree code isn't all that much different than Nim.

rbergen:
@louie-github I'm wondering if this PR is ready for review? If so, could you please sync your fork with this branch? It will update the CI workflow so it works for your PR.

louie-github:
Sorry about the delay, but I'm pretty sure that I'm going to have a really busy week, so I don't think I'll be able to get this PR ready for review within the week. Life is catching up with me :>

Once again, should I close this PR or just leave it as is?

If left as is, once I finish the changes, I'll sync the fork to finalize everything.

rbergen:
@louie-github It's okay to leave it as it is. Now I know the most recent commits don't constitute the final sprint :)

GordonBGood:
@louie-github:

> Sorry about the delay...

I've been hoping for some code to review, but expected that you were busy.

Thinking about your past comment:

> This code_native output is so long I have to press Ctrl+C as fast as possible to get this to stop and for the scrollback to reach the start.

Why don't you just pipe the text output to a text file so then you can open it in an editor and do searches, etc. and/or clip out relevant snippets of code?

In your Julia compiler generated code:

```julia
; ││└
; ││┌ @ multidimensional.jl:639 within `setindex!' @ array.jl:839
        orb     $8, (%rax,%rbx)
; ││└
; ││┌ @ int.jl:462 within `>>' @ int.jl:456
        leaq    (%rsi,%r13), %rbx
        shrq    $3, %rbx
```
This explains why your code is slower than expected:  the `orb` instruction takes as little as one CPU clock cycle, but the `leaq` takes a half a cycle and the `shrq` takes a third of a cycle so minimum clock cycles per marking operation will be 1 5/6 CPU cycles; then there will be some "cache thrashing" lost time due to the sieving taking place over buffer of almost 64 Kilobytes where the CPU L1 data cache size is only 32 Kilobytes, although in this case that may be hidden in the extra lost time due to the extra CPU cycles.

So the fastest that you could run with this with your 4.4 GHz clock (boosted single threaded) on your AMD 3700X for the required about 810K marking operations would be a little less than 3000 passes per second or 15,000 in five seconds, plus the time do the sieve buffer allocations per pass, probably about 13,000 to 14,000 passes in five seconds all told.

If the computed goto is working effectively and the functions are being compile time typed and inlined (or you inline the function body code without messing with a function) then it should run about as fast as my Nim "Extreme Loop Unrolling" contribution.

On the PrimeView results web page, I do note that the "Extreme Loop Unrolling" and the "Loop Unpeeling" solutions are lower in ranking than equivalent "Striped" solutions for Dave Plummer's Celeron machine as compared to using @rbergen's newer CPU;  This is due to the older lower level CPU's having lesser branch prediction capabilities than newer ones and these algorithms (especially "Extreme Loop Unrolling") greatly depending on the efficiency of branch prediction as there are at least two branches per base prime value and especially the first one is pseudo random/unpredictable, meaning missed branch predictions.  The older CPU architectures often used a much longer micro-instruction queue than newer ones, which queue must be dumped on a missed branch prediction, so a longer time to recover.

rbergen:
@louie-github Just checking in whether you have an expectation on when we can take this PR forward?

louie-github:
@rbergen Once again, sorry for the delay, but I think I can finish this either by tomorrow or Monday (UTC +8), since Monday is a non-working holiday over here. If not, it's up to you as to what you want to do with this. Thank you!

rbergen:
@louie-github Thanks for the update. Concerning what to do with this, in first instance it really is up to you. If, let's say by Monday, you think that the PR is ready to be merged, then you can say so and we'll start the "final" review. On the other hand, if you feel it still needs more work but you don't currently have the time or don't expect to get it soon, then I'd recommend closing it for now.

(On a side note and in a work context, I have a team member who's off on Monday for Araw ng mga Bayani. If you are too, then I hope you do have a chance to enjoy the celebrations.)

PR #642, opened 2021-08-14
CI publish Docker images

marghidanu:
Publishing Docker images in Docker Hub from the default branch. Also makes use of caching to speed up the builds in GitHub Actions.

Remove Docker scanning, will delegate this to Docker Hub later.


rbergen (approved):
LGTM!

PR #643, opened 2021-08-15
[D Solution 2] Use inverted array, other small improvements

BradleyChatha:
## Description

~~So I just noticed that SieveB1_64 isn't actually reporting the right prime count, so I've removed it from the leaderboard set because I really, really can't be bothered to figure out why it's 64-bits specifically that doesn't work.~~

I've also changed the code to use an inverted array to help with performance.

This is *probably* the last PR from me. I can't get my head around the faster solutions, and all my own attempts haven't shown any speed increases, so this is about as fast as I can get things in D. (I tried to make it use only half the memory needed and halving the index, but that made things *slower*.

Regardless, this was quite a fun thing to be a part of. 

## Contributing requirements


louie-github:
Do the lines below still apply when saying that 64 has the best performance, since you did change `s8` to 32 bits?

```d
alias s8 = SieveRTB1_32;
runMultiThreaded!(s8, st)(IsFaithful.yes, "base", 1); // 64 has the best performance on my machine, so I'll use that for the multithreaded leaderboard.
```

I couldn't follow the program flow, I just noticed this while looking through the diff.  Sorry if I missed something.

rbergen:
Could it be that the bit shifting/mapping operations yield a 32-bit integer unless specifically made to operate on/return 64 bits? If so, that would explain why you'd effectively lose about half the sieve bit field, which is what seems to be happening. I say this because the reported sieve count when using ulongs (41603) is close-ish to half of the correct prime count (78498).

yellowcub:
@BradleyChatha It sounds like you have gone through as many personal iterations of the code as I have with Swift.  I do not know D, but I will share some lessons learned from the Swift solution that resulted in notable speed boosts.

1. The organization of the loops makes a big difference.  Even if using embedded loops jumps through the steps the same - if one loop can be used it should run faster.  I am thinking or your step through of factor in `runSieve` where you have two loops and then a branch (`if`) statement.  If D has a continue statement, you could eliminate the inner loop altogether.
2. Anywhere calculations are made inside a loop where they could be done before will make the loop faster.  These are not always easy to spot, but there are a number of embedded calculations made each time you navigate between the number you are interested in and the index in the array it is located.
3. Some compilers use overflow checking by default for math operations, which is a disadvantage (speed wise) to languages such as C.  If this is true for D, see if there is an alternative operation to allow overflow.  You should of course be confident your code is not going to actually overflow anyplace you don't want it to.
4. Lastly, certain math/logic operations are more costly than others.  Division and modulus are much slower than addition or multiplication.  Most of these involve division by the number `bitSize` that should be known to be a power of two.  If these were constants compiled at run-time, some compiler might do some equivalent math optimizations.  But, if the compiler doesn't know how to optimize them, you could implement some faster logic to do the same task.  For example `a % b` (where `b` is of 2^n) should be faster using `a & (b - 1)`, and `a / b` could be implemented as `a << n` provided you can compute (or define) `n` once for `bitSize` before any of the loops.

Good luck, and I hope you don't give up just yet!

BradleyChatha:
> Do the lines below still apply when saying that 64 has the best performance, since you did change `s8` to 32 bits?

I forgot to change it to mention that 64-bits was faster, but wrong, so I went with 32-bits. But that doesn't matter now.

> I couldn't follow the program flow, I just noticed this while looking through the diff. Sorry if I missed something.

Yea, it's definitely turned into a giant mess as I forced more and more D features to be shown off in the code.

> Could it be that the bit shifting/mapping operations yield a 32-bit integer unless specifically made to operate on/return 64 bits? 

Very good catch, yes it was because of that. A few "UL"s later and it works properly, thank you!

>  I do not know D, but I will share some lessons learned from the Swift solution that resulted in notable speed boosts.

Thank you for the insights, I'll see what I can do.

BradleyChatha:
I think that'll be it now for this PR.

For @yellowcub 's reply:

`bitSize` is a compiler-known constant, and the compiler is outputting the correct optimised versions of division and modulus.

Overflow checking is opt-in which I haven't used, and I think I'm using all the right compiler flags for things like turning off bounds checking.

I've eliminated the inner loop but I think the compiler was producing the same ASM with or without it, since there was no speed difference.

I did however see that it was doing a lot of needless calcs when doing the square root as a result of me using `std.conv.to` instead of raw casts, so I've changed that for a small but decent speed boost.

BradleyChatha:
Seems `iota`, even though it was being fully inlined, wasn't quite as efficient as a normal for loop

yellowcub:
https://github.com/PlummersSoftwareLLC/Primes/blob/591d2cf216c454d940bac8d893362b4743aa48c3/PrimeD/solution_2/source/app.d#L192

Change `factor * 3` to `factor * factor`.  This will save unnecessary setBit runs that should already get cleared from earlier primes, and is permitted for the leaderboard rules.

https://github.com/PlummersSoftwareLLC/Primes/blob/591d2cf216c454d940bac8d893362b4743aa48c3/PrimeD/solution_2/source/app.d#L181-L187

I personally found in Swift that placing only the continue inside the if statement made the routine faster.  @rbergen or @louie-github could probably explain more precisely why this is the case, but I suspect the compiled ASM here should just create a single branch instruction; whereas, multiple instructions might create a couple - I can only say that my efforts to keep the incremented inside the `if` statement slowed it down.  You still need to increment `factor` somehow, of course.  I worked around this (see below) with initializing the `factorIndex` below the starting point, and incrementing between the loop and the if statement (the increment at the end is not needed then either).  Note: the `repeat` it part of a trailing `while` loop, which may also have benefited the Swift routine.

https://github.com/PlummersSoftwareLLC/Primes/blob/6792b3ffce380f36401a09c8aa0212ac2b32c950/PrimeSwift/solution_1/PrimeSwift_1bitBool_64bit/Sources/PrimeSieveSwift/main.swift#L89-L92

The last thing I can think of would be to somehow work primarily in the array index space rather than the odd integer space to reduce the amount of index calculations that need to be done.

Your mileage with these considerations might vary, I only offer them up to collaborate what has worked for me and maybe inspire a different thought approach.  If you happen to come up with something that works out even better, I would love to test it in Swift also.

BradleyChatha:
`factor * factor`, subtle, but you're right!

> but I suspect the compiled ASM here should just create a single branch instruction; whereas, multiple instructions might create a couple 

I think this might just be the case of the compiler outsmarting us.

>  If you happen to come up with something that works out even better, I would love to test it in Swift also

Having a branch into an unrolled loop has helped quite a bit for me. Seems 64 is the magic number for how many unrolls will work before it starts to slow it down. (see latest commit to PR).

In D this is simply done by `static foreach`, a compile-time foreach that duplicates its body for every iteration, which may not have a Swift equivalent.

BradleyChatha:
I also tried to vectorise the unrolled loop, but I couldn't manage to get any speed increase with D's built-in SIMD stuff.

Although the major issue is that after the vector operations I still did this:

```d
static foreach(j; 0..16)
    this.setBit(vect[j]);
```

Which I suspect didn't help much at all.

Buuuut my brain goes to mush with any amount of maths let alone SIMD, so I couldn't figure out if it's possible to do a vectorised `setBit`.

louie-github:
> Buuuut my brain goes to mush with any amount of maths let alone SIMD, so I couldn't figure out if it's possible to do a vectorised setBit.

I've tried using Julia's `@simd` macro when setting bits, and even tried some external packages to help speed it up, but it didn't really help.

This was probably because we use large steps when gathering values from the array, so it takes time to gather and then scatter thr values if we tried to SIMD things.

Julia's `@simd` macro didn't even generate any vectorized assembly from what I can tell. Using an external library, `LoopVectorization`, it did, but it was even slower due to overhead.

At least, that's my experience.

yellowcub:
@BradleyChatha Unrolling loops is something I have tried multiple times without success. Perhaps someone might come along and make it work in Swift so I can see what I've been doing wrong. ... Until then I will probably try a few dozen more times 🤣. Just like the loop organization, the unrolled loop probably works best in a very specific way.

louie-github:
I think the line
```d
while(factor < q)
```

actually needs to be
```
while(factor <= q)
```

since `q` *is* the max factor, so you should include it.

And do you really need to `round` the result? Maybe you could just truncate it.

BradleyChatha:
@louie-github Honestly I can't even remember why I was rounding it in the first place, so I kept it in because I assumed past me was correct. I likely tried it to see if it'd fix some issue I was having, but then forgot to remove it, if I know myself well.

And yeah, the while loop needs to be fixed.

rbergen:
@BradleyChatha Happy to see that you got the 64-bit implementation working again. 

Please let me know when you've finished working on this PR. I'm going to hold on my review until then.

BradleyChatha:
@rbergen It works again thanks to your observation, thank you.

I think this is finished now.

rbergen:
@BradleyChatha Could you please sync your fork with this branch? It should enable CI for your PR.

PR #644, opened 2021-08-15
corrected Dockerfile problem

mike-barber:
## Description
Small correction to Dockerfile for PrimeCSharp/solution_4; would not build on a Linux box correctly, and it's missing from the results as a consequence. This resolves the problem on my machine, and I think @ItalyToast will be happy with the change.

## Contributing requirements


ItalyToast:
I approve of the change. This problem should also apply to solution_2 @davepl

rbergen:
@mike-barber Could you please sync your fork with this branch? It should make the CI work again for your PR.

mike-barber:
@rbergen done :) 
It's all looking good this time, except for what looks like a transient failure on `PrimeBallerina/solution_1` 
I don't have the authority to retry that job, but I suspect that'd fix it.

PR #645, opened 2021-08-15
Prime power shell solution 4

RobCannon:
## Description
This is a PowerShell implementation of davepl's sieve of Erastosthenes for
benchmarking programming languages and CPUs. This implementation uses the
capability of PowerShell to build a dynamic assembly from C# and run that.
The C# code is based on the current fastest C# implementation by tannergooding, 
solution_3.  While the core parts of this solution are implemented in another 
language, I think it is a faithful implementation since it has long been a capability of 
PowerShell to allow a script to access .NET directly when it is needed for 
speed or accessing some feature not available in PowerShell.  Since this 
capability is built into PowerShell and does not need and external dependencies, 
it should count as faithful, IMO.

## Contributing requirements


rbergen:
Considering that the core of this implementation is written in C#, it will have to be submitted as a "mixed" language solution (akin to the PrimeMixed/solution_1 that is already place), and marked as unfaithful. Regardless of how intrinsic a quality of a language it is that it can use code written in a different language, the solution does mix languages.

Also, please only open PRs on this repo when they are ready to be reviewed and merged.

RobCannon:
OK, I will move this solution. 

I created this PR as a draft so I could review everything.  That is the mechanism that GitHub uses to allow you to validate everything yourself before you hit go.  We use this extensively at work and it a must when you use GitHub actions to automate checks.

PR #648, opened 2021-08-16
Introduction of build-no flag file

rbergen:
This adds the ability to disable automated builds and benchmarks for a solution by adding a `build-no` flag file to the solution directory. 

Also:
- The flag file is added to the PrimeBatch/solution_1 directory
- The ManualDockerfile in that directory is renamed to Dockerfile
- Documentation is updated


marghidanu (approved):

marghidanu:
LGTM

PR #654, opened 2021-08-17
Made Lua/solution_1 faithful

greatwolf:
## Description
This PR fixes up Lua solution_1 to satisfy the faithful requirement by making it more OOP's. The changes are mostly straightforward. 

#### Quick rundown of changes:

 - The sieve state is now encapsulated into a lua table to represent it as an 'object'.
 - Created `PrimeSieve` table and placed the free functions into it so it acts like a 'class' definition.
 - The various free functions were also updated to accept an implicit `self` as first parameter effectively turning them into methods that operate on a given object.
 - Lua's metatable mechanism is used to tie the methods to a given object(table) so it can be operated on with object-like syntax. 
 - Renamed `setup` function to `new` so it's more appropriate. New instantiations can be created with `PrimeSieve:new(size)`.
 - Updated `printResults` function to just print the output string directly so it more closely resembles Dave's original implementation.

## Contributing requirements

### DIFF
--- a/PrimeLua/solution_1/README.md
+++ b/PrimeLua/solution_1/README.md
@@ -1,7 +1,7 @@
 # Lua solution based on sources from seanbouk
 
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
-![Faithfulness](https://img.shields.io/badge/Faithful-yes-yellowgreen)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
 ![Parallelism](https://img.shields.io/badge/Parallel-no-green)
 ![Bit count](https://img.shields.io/badge/Bits-64-yellowgreen)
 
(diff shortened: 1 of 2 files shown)


rbergen (commented):
I believe the "Faithful" badge in the README.md also needs to be updated (to "yes-green")

greatwolf:
whoops forgot that one. ok done.

PrimeLua/solution_1/README.md
```diff
@@ -1,7 +1,7 @@
 # Lua solution based on sources from seanbouk
 
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
-![Faithfulness](https://img.shields.io/badge/Faithful-no-yellowgreen)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-yellowgreen)
```
rbergen: We're nearly there. :) Please change -yellowgreen to -green.

greatwolf: that should do it.

PR #655, opened 2021-08-17
Add Chapel solution_1 to race...

GordonBGood:
## Description

There has been no contribution for the Cray language Chapel yet.

This solution includes the techniques of bit-packed odds-only, unpeeled bit-packed odds-only, unpeeled by blocks bit-packed odds-only, and extreme loop unrolled bit-packed odds-only in the Sieve of Eratosthenes.

## Contributing requirements


PrimeChapel/solution_1/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+    bash \
+    make \
+    mawk \
+    file \
+    pkg-config \
+    git \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.7 20
+RUN update-alternatives --set python /usr/bin/python3.7
```
rbergen: Please consolidate these RUNs. It decreases the number of image layers, and these three all concern updating/installing prerequisites.

PrimeChapel/solution_1/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+
+# Configure locale
+RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
+    echo 'LANG="en_US.UTF-8"'>/etc/default/locale && \
+    dpkg-reconfigure --frontend=noninteractive locales && \
+    update-locale LANG=en_US.UTF-8
+
+# Configure dummy git user
+RUN git config --global user.email "noreply@example.com" && \
+    git config --global user.name  "Chapel user"
```
rbergen: I'm not sure I understand the point of this RUN. That is to say, the docker build and run seem to work fine without it.

rbergen (commented):
The docker build of this solution took just over 6 minutes on my machine, mainly caused by the compiler build executed by the RUN that starts at line 34 of the Dockerfile. This means we'll have to create a Chapel Docker base image that serves as the foundation for the Dockerfile in this solution.

You can do this yourself, but I'm also happy to help out with this. I would effectively:

1. Create a base image that incorporates your Dockerfile up to and including the last ENV directive
2. Push it to a repo under this organization and to Docker Hub
3. Try to make it an amd64 and arm64 image; this one seems to support amd64 only

The Dockerfile in this solution could then use that base image in its FROM directive.

If "3" doesn't turn out to be possible, we'd have to mark this solution as an amd64-only solution, by adding an `arch-amd64` flag file.

GordonBGood:
@rbergen:

> The docker build of this solution took just over 6 minutes on my machine, mainly caused by the compiler build executed by the RUN that starts at line 34 of the Dockerfile. This means we'll have to create a Chapel Docker base image that serves as the foundation for the Dockerfile in this solution.
> 
> You can do this yourself, but I'm also happy to help out with this. I would effectively:
> 
> 1. Create a base image that incorporates your Dockerfile up to and including the last `ENV` directive
> 2. Push it to a repo under this organization and to Docker Hub
> 3. Try to make it an amd64 and arm64 image; this one seems to support amd64 only
> 
> The Dockerfile in this solution could then use that base image in its `FROM` directive.
> 
> If "3" doesn't turn out to be possible, we'd have to mark this solution as an amd64-only solution, by adding an `arch-amd64` flag file.

The Chapel documentation indicates that they have done some work cross-compiling to ARM which doesn't do us any good as we need to both build and run on an ARM machine (at least ARM64) and they think the Chapel compiler should be able to be compiled on a ARM64 machine but it looks like they haven't worked on that much.  I don't have an ARM64 machine to play with this so as to verify that it works or not, so I am dependent on you and marghidanu (who seems to be doing testing/comparisons on a Raspberry Pi with ARM64) to do this.

> Please consolidate these RUNs. It decreases the number of image layers, and these three all concern updating/installing prerequisites.

I'm fairly new to using Docker and just copied and pasted the compiler Dockerfile from the Chapel website, so didn't look at it too closely.  I guess the `RUN`'s can be consolitated by the following:
```docker
RUN apt-get update && apt-get install -y --no-install-recommends \
    wget \
    ca-certificates \
    ...
    git \
    && rm -rf /var/lib/apt/lists/* \
    && update-alternatives --install /usr/bin/python python /usr/bin/python3.7 20 \
    && update-alternatives --set python /usr/bin/python3.7
```
but that is kind of pointless in doing a commit to my Dockerfile as you suggest moving all of this up to the last `ENV` directive out to build a docker image to be made available both in a Github repo and on Docker Hub and having me then just change my Dockerfile to use/`FROM` that docker image.

I could likely do it, but if you can help it would be most appreciated, as you seem to have more experience with Docker.

> I'm not sure I understand the point of this `RUN`. That is to say, the docker build and run seem to work fine without it.
```docker
RUN git config --global user.email "noreply@example.com" && \
    git config --global user.name  "Chapel user"
```
As I said, I just copy and pasted the Dockerfile to build the Chapel compiler from the Chapel website and this `RUN` was included; I don't see the purpose and it may as well be deleted if the benchmark can be built and run without it.

I guess you know that the Chapel compiler is an open source project sponsored by Cray Computing (now owned by HP and renamed to HPE) to produce a computer language whose purpose competes with Fortran, but is a more modern OOP language.  It often scores quite well in performance as compared to many of the other compiled languages and is almost always faster than Fortran and often has close or better performance than C/C++, so I wanted to include it in the "Software Drag Race".  The only way to run it as a Docker image that they document is the Dockerfile I copy and pasted, which leads us to where we are now.

One of the Chapel language's downsides is long compile times, as it takes about 10 seconds to compile a "Hello World" program in Chapel on my machine, which is comparable in speed to yours, at least for single threading.

rbergen:
Ok, I will work towards putting a Chapel base image together. I will be using docker buildx, which means that my approach will be to perform a cross-platform build for arm64. My experience with cross-compiled compilers is that they work if they compile, but I can and will test any arm64 build outcome on a Raspberry Pi that I have laying around as well.

> I guess the RUN's can be consolidated by the following but that is kind of pointless in doing a commit to my Dockerfile as you suggest moving all of this

That's true if you're okay with me picking up the task. I now know that is the case, but I didn't want to make assumptions about this, earlier.

> One of the Chapel language's downsides is long compile times

Starting with stating the obvious that "long" is a very flexible term, we do have other languages in the repository that also take their time compiling. In itself, that's not a problem. 

So, I'll get going on this when I can, and I'll report progress in (a) future comment(s) on this PR.

GordonBGood:
@rbergen:

> Ok, I will work towards putting a Chapel base image together. I will be using docker buildx, which means that my approach will be to perform a cross-platform build for arm64. My experience with cross-compiled compilers is that they work if they compile, but I can and will test any arm64 build outcome on a Raspberry Pi that I have laying around as well.

Thanks for this; you have some experience (with buildx) and equipment that I don't have.

> > I guess the `RUN`'s can be consolidated by the following but that is kind of pointless in doing a commit to my Dockerfile as you suggest moving all of this
> 
> That's true if you're okay with me picking up the task. I now know that is the case, but I didn't want to make assumptions about this, earlier.

Yes, it would be advantageous to have a Chapel docker image that runs on ARM64 as well as the Intel processors.  Once you have the image, I'll commit a new Dockerfile for this contribution to use it.

> > One of the Chapel language's downsides is long compile times
> 
> Starting with stating the obvious that "long" is a very flexible term, we do have other languages in the repository that also take their time compiling. In itself, that's not a problem.

Just it's noticeably longer than most of the other compilers I habitually use, longer than some by almost ten times for re-compiling when debugging.

> So, I'll get going on this when I can, and I'll report progress in (a) future comment(s) on this PR.

Again, thank you very much for helping out.  Although Chapel isn't one of my favorite languages, I think it has its place and should be properly representing in the "race".

I would also like to provide new Docker images for Haskell and Nim, neither of which current images support ARM64, and may learn enough from seeing what you do to be able to do these myself.  Nim is fairly easy as it is known to compile for ARM (it uses gcc or clang as either C or C++ intermediate source) and compile times to produce a compiler from source are relatively short at only a minute or so; for GHC Haskell there are downloadable binaries already available for most CPU architectures so one just has to pick the right one for the detected CPU, decompress it, and install it - the current Haskell Docker file builds the LLVM back end from source which takes a fair while, but if one picks the right Linux distro image file to base on, the LLVM back end can be installed from the distro package repos.  I am thinking the Fedora Docker image because then one can easily obtain the latest supporting programs such as gcc and/or clang.

rbergen:
@GordonBGood:

The "final" base image Dockerfile has just completed its build for amd64. Based on an earlier run I expect the arm64 build to take another 45 minutes. All in all, I expect the primeimages/chapel images for amd64 and arm64 to be on Docker Hub within about an hour.
I will then test the base image with your solution on both architectures. Once I've done that, I'll open a PR on your branch to modify the Dockerfile to use primeimages/chapel as a basis; I have to make that modification anyway to be able to test.

> I would also like to provide new Docker images for Haskell and Nim, neither of which current images support ARM64

That would be highly appreciated. We would like to extend the number of solutions that run on both architectures.
When creating and using base images, compile times for those base images are effectively irrelevant; you perform the build only once per architecture once the base image Dockerfile is final.
 
I'll share the link to the repo in which I've put the Chapel base image Dockerfile once I've confirmed it works. If you do create the base images for Nim and Haskell and need help with doing the actual multi-arch build, you can let me know. It's also possible to publish your images under the primeimages Docker Hub organization. In that case I'll create repositories for them under this GitHub organization (as they now already exist for a number of other languages) and build and publish the images for you.

>  I am thinking the Fedora Docker image because then one can easily obtain the latest supporting programs such as gcc and/or clang.

Personally, I have good experience with Ubuntu. 21.04 is still fairly recent and is also pretty up-to-date with the supporting programs I've been needing.

> Again, thank you very much for helping out.

You're very welcome. I like expanding the collection with new (or old) languages, and there's no reason to exclude Chapel, as you state.

rbergen:
@GordonBGood I've just opened GordonBGood/Primes#3, after confirming that everything works on amd64 and arm64. The base image Dockerfile can be found [here](https://github.com/PlummersSoftwareLLC/Chapel-docker/blob/master/Dockerfile).

rbergen (approved):
LGTM

GordonBGood:
@rbergen:

I've merged your PR into my Chapel branch after review.  Looks good to go.

Your Docker image primeimages/chapel is up on DockerHub; although the internal tags are complete as to showing support for amd64 and arm64, you might consider adding tags to the "front" page for the image to show what it supports.

> When creating and using base images, compile times for those base images are effectively irrelevant; you perform the build only once per architecture once the base image Dockerfile is final.

Yes, I appreciate the time to compile the compiler is irrelevant once there is a base Docker image that includes that compiler.  My comment was more related to that there must be some kind of a trigger point where the compile time to create the compiler becomes unacceptably slow.  For Chapel you mentioned that it took six minutes (I assume on your desktop machine, which would mean something like twenty minutes to a half hour on marghidanu's Raspberry Pi) as being too slow.  if a compiler took only one minute to compile (perhaps meaning eight minutes on Raspberry Pi), would that require creation of a base image including that compiler?

Do you find the time to compile LLVM as usied in the Haskell Dockerfile(s) acceptable (I can't remember how long that takes)?

> If you do create the base images for Nim and Haskell and need help with doing the actual multi-arch build, you can let me know. It's also possible to publish your images under the primeimages Docker Hub organization. In that case I'll create repositories for them under this GitHub organization (as they now already exist for a number of other languages) and build and publish the images for you.

Thank you, I probably will need help with the actual multi-arch build.

> Personally, I have good experience with Ubuntu. 21.04 is still fairly recent and is also pretty up-to-date with the supporting programs I've been needing.

I (along with Linus Torvalds himself) use Fedora on my personal desktop and have found it to be generally very solid while also very up-to-date; I know that Ubuntu 21.04 is also very solid.  It doesn't really matter, but it would be nice to use new versions of (especially) gcc as that can affect the speed of Nim which uses it as a backend (Fedora default is gcc 11.2.1) and the latest version of LLVM is 12.0.1 which might make GHC Haskell faster and would also likely avoid having to compile LLVM from scratch.

rbergen:
@GordonBGood 

> Your Docker image primeimages/chapel is up on DockerHub; although the internal tags are complete as to showing support for amd64 and arm64, you might consider adding tags to the "front" page for the image to show what it supports.

We could do that. For us, whatever we put in Docker Hub is only there to support this project. Of course, people are free to use the images we publish for whatever they like, and it seems some images have indeed found broader use. However, we (the maintainers) have decided not to spend time on decorating the core content of our images. Maybe we'll get around to it at a later time, maybe we never will.

> My comment was more related to that there must be some kind of a trigger point where the compile time to create the compiler becomes unacceptably slow. 

There is, but it is in no way an exact science. I'd say that I start looking more closely at the compile time of an image when it takes longer than 2 minutes, and I start having opinions when it takes more than 3. By the way, my frame of reference for this is not the i7 I use to generate benchmark reports, but a laptop with an 8th gen i5. 

> Do you find the time to compile LLVM as used in the Haskell Dockerfile(s) acceptable (I can't remember how long that takes)?

If the compile time had crossed the threshold of what's acceptable, I would expect a "faster" base image to be already in place.

> I (along with Linux Thorvalds himself) use Fedora on my personal desktop and have found it to be generally very solid while also very up-to-date; I know that Ubuntu 21.04 is also very solid. It doesn't really matter, but it would be nice to use new versions of (especially) gcc as that can affect the speed of Nim which uses it as a backend (Fedora default is gcc 11.2.1) and the latest version of LLVM is 12.0.1 which might make GHC Haskell faster and would also likely avoid having to compile LLVM from scratch.

I think we are indeed discussing a topic that is somewhat in the realm of personal preferences. As long as they don't come with undue side effects, that's fine.

GordonBGood:
@rbergen:

> For us, whatever we put in Docker Hub is only there to support this project ... However, we (the maintainers) have decided not to spend time on decorating the core content of our images. Maybe we'll get around to it at a later time, maybe we never will.

Fair enough.

> > My comment was more related to that there must be some kind of a trigger point where the compile time to create the compiler becomes unacceptably slow.
> 
> There is, but it is in no way an exact science. I'd say that I start looking more closely at the compile time of an image when it takes longer than 2 minutes, and I start having opinions when it takes more than 3. By the way, my frame of reference for this is not the i7 I use to generate benchmark reports, but a laptop with an 8th gen i5.

Okay, two or three minutes is the trigger point.  That's good to know.

> > Do you find the time to compile LLVM as used in the Haskell Dockerfile(s) acceptable (I can't remember how long that takes)?
> 
> If the compile time had crossed the threshold of what's acceptable, I would expect a "faster" base image to be already in place.

Total time to built from scratch for the current PrimeHaskell Dockerfile is about two minutes, so it's just less than your threshold above.  However, the GHC Haskell Docker Hub base images don't support AMD; it would seem that their primary intent is to be somewhat useful in being able to actually do GHC Haskell development on a machine without installing the compiler as long as one has Docker installed.  The images have been "stripped down" by GHC so that they are "only" about 3.X Gigabytes.

> I think we are indeed discussing a topic (Linux distro used for base Docker images) that is somewhat in the realm of personal preferences. As long as they don't come with undue side effects, that's fine.

For Nim, I like Fedora because one can easily get the latest GCC version, which Nim uses as a compilation backend, which versions tend to gain in performance with later developments.  I have changed my PrimeNim Dockerfile so that it builds Nim from source using the Fedora 34 Docker image as base and it only takes two or three minutes on x86_64/amd64, so quite close to your informal threshold.  I'll submit this for you consideration on whether a Docker Hub image is required or not, but I'll hold it up for a day or two to combine the PR with some performance tweaks.

As for Haskell, I just recently found out through GHC Haskell issues that (at least as of a year ago) the version of Debian that the Raspberry Pi computers come with is 32-bit so I guess they can't run arm64 docker images unless the OS is changed?  Apparently it is sometimes done to upgrade the OS to a 64-bit version, in which case the restrictions don't apply.  I see that marghidanu's  Raspberry Pi 4 has been upgraded to a 64 bit OS but I don't know if yours has, and also we should really support people that have not upgraded.  It seems as though the build tool chain that comes installed on all Raspberry's actually emits arm6 code so as to be compatible with their original model, but the new models like the "4"'s OS will run arm7-a code from a docker file, so we can use a Docker image built for that.  There are two possible approaches, as follows:
1.  I kind of like the approach as for Nim of building the GHC Haskell compiler from source so the same Dockerfile then applies to all of amd64, arm64, and now arm32; however, it takes about an hour on my machine to compile (per supported target), which is rather onerous.
2. There are pre-compiled binary distributions available from GHC for each of amd64, arm64, and arm32 for Debian 10 and they can be installed on the Debian 10 Docker image in only a few minutes each, so that is the approach I have taken.  The Dockerfile either needs to have some conditional processing based on the target supported or you can likely handle that with your multi-host support things.  Again, I will combine the PR with some tweaks to the performance of the code.

rbergen:
@GordonBGood 
 
> For Nim [...] I'll submit this for you consideration on whether a Docker Hub image is required or not, but I'll hold it up for a day or two to combine the PR with some performance tweaks.

Ok, I look forward to seeing it.

> As for Haskell, I just recently found out through GHC Haskell issues that (at least as of a year ago) the version of Debian that the Raspberry Pi computers come with is 32-bit so I guess they can't run arm64 docker images unless the OS is changed?

That's correct. 

> Apparently it is sometimes done to upgrade the OS to a 64-bit version, in which case the restrictions don't apply. I see that marghidanu's Raspberry Pi 4 has been upgraded to a 64 bit OS but I don't know if yours has

They both have. The system information block in marghidanu's reports show that he is running Ubuntu, and so am I. 

> and also we should really support people that have not upgraded. 

I don't think arm32 support in the context of this project is on the cards anytime soon. Broadening support for arm64 is already a significant challenge, as you are now experiencing. We're also not aiming for Raspberry Pi support as such, it's just interesting (and in my opinion, amusing) that the whole thing works on a device that costs no more than 40 euros. 
Ubuntu 64 bit has been available for Raspberry Pi 3 and 4 for quite some time now, and is very stable. Raspberry Pi OS 64 is underway, as well.

All that said, there is absolutely nothing stopping you from putting Docker images together that work on arm32, as well.

> Again, I will combine the PR with some tweaks to the performance of the code.

Looking forward to that one as well.

marghidanu:
@GordonBGood It's Linus Torvalds.

marghidanu (approved):

GordonBGood:
@marghidanu: RE: Linus Torvalds, I know, typing too fast; didn't check.  Thanks for the approval.

@rbergen:
> Ok, I look forward to seeing it (the new Nim version).

You'll have been flagged as to my discussion with @mike-barber on our work to make this benchmark even faster in Rust and Nim (which is, of course, applicable to any language that compiles to efficient native code), and that is likely the approach the new Nim implementation will take.

> > and also we should really support people that have not upgraded.
> 
> I don't think arm32 support in the context of this project is on the cards anytime soon. Broadening support for arm64 is already a significant challenge, as you are now experiencing. We're also not aiming for Raspberry Pi support as such, it's just interesting (and in my opinion, amusing) that the whole thing works on a device that costs no more than 40 euros.
> Ubuntu 64 bit has been available for Raspberry Pi 3 and 4 for quite some time now, and is very stable. Raspberry Pi OS 64 is underway, as well.

That's good to see that most Raspberry's will likely be upgraded to 64-bit.

> All that said, there is absolutely nothing stopping you from putting Docker images together that work on arm32, as well.

After thinking about this, I don't think arm32 support is all that important, as the world is at 64-bits and likely never going to look back.  Even almost all of the low-end smart phones sold these days use 64 bit SoC's even if using the current version of Android-Go is only 32 bit (similar to Raspberry up to now).  Running the faster versions of this benchmark on such processors is like putting Top Fuel into the family station wagon and entering a drag race - amusing but not really that meaningful other than to show the probable ratios of performance between the different levels of CPU.

> > Again, I will combine the PR with some tweaks to the performance of the code (in Haskell).
> 
> Looking forward to that one as well.

As with Nim, this will likely be based on the new techniques developed with Mike.

GordonBGood:
@rbergen:

I note that this Chapel implementation is up on the PrimeView test streams for Tudor and Dave's machines but not from yours even though your latest test seems to have a more recent time stamp.  Is there some sort of flag you need to set to include an extra language on the list of tested ones?

> 
> > Ok, I look forward to seeing it (the new Nim version).
> 
> You'll have been flagged as to my discussion with @mike-barber on our work to make this benchmark even faster in Rust and Nim (which is, of course, applicable to any language that compiles to efficient native code), and that is likely the approach the new Nim implementation will take.

We think our ideas for extending the "hybrid reset" technique toward an "ultimate" implementation will pass `faithful to base`, but I'll mention them.  We are aware that some "word" implementations were (correctly) rejected as not being `faithful to base` because they used a mask word that set/cleared multiple composite values with one instruction.  The "hybrid reset" technique presumably passes because although the order is changed, it still uses one instruction per culling/marking operation.  This new "ultimate" technique extends that to 64-bit words that are read once, have multiple marking operations done one bit at a time, and then are written out once, which we think passes.  The pseudo code would look something like the following if we were doing this for a step size of one bit per step, which of course is not a valid base prime:

```
v = composites[wordIndex] or 0x0000_0000_0000_0001
v = v or 0x0000_0000_0000_0002
v = v or 0x0000_0000_0000_0004
v = v or 0x0000_0000_0000_0008
v = v or 0x0000_0000_0000_0010
...
v = v or 0x4000_0000_0000_0000
composites[wordIndex] = v or 0x8000_0000_0000_0000
```
with one `or` operation per bit.  This is fast because it has better cache associativity in only doing one read and one write separated by 64 `or` operations, and also because the read and write to memory take about one CPU clock cycle other than for "cache thrashing" but the `or` register operations only take a third to a half (depending on CPU) so this takes only a little over an average of a third or a half of a CPU clock cycle per operation. on average.  We then use the generated patterns of doing this for the odd values up to 19, above which there is no benefit, but that covers about 50 percent of the total culling operations; we don't save as much as fifty percent each as the base prime values approach 19 but, along with the memory access improvements, the overall improvement can be thirty percent or more, especially for the more primitive CPU's.

We are aware that you prefer that no prior knowledge of the values of primes other than two are used, so the above technique would include code to handle the cases of nine and fifteen even though they are not primes.

For values above 19, my "Extreme Loop Unrolling" technique as used in Nim seems to be the best.

I think you will have been flagged, but I also got the idea of a challenge to any authors using languages that are capable of implementing this same "ultimate" technique (which also includes the "Extreme Loop Unrolling" for primes above 19) in their language for a better comparison of language's capabilities.  I propose using a new "ultimate" version this code in Chapel as a model for others to follow and may forgo the use of the "template" since many other languages have neither templates nor macros; code in Chapel has the advantage that it is easy to read and understand for most programmers as it is OOP, has a familiar syntax, and its few differing features aren't too hard to understand at a glance.

**What do you think?  Could we add an invitation to this challenge to the front page?**  It would then less penalize those languages whose authors don't have the expertise to come up with the "ultimate" technique themselves.  It may or may not apply to JIT'ed languages, although with the new low-level pointer capabilities in F# and C#, one may be surprised; also, there is WebAssembly/AssemblyScript.

rbergen:
@GordonBGood 

While doing some "lower-priority" administration on closed PRs, I noticed I missed this comment up to now. Please note that closed PRs get a lot less attention from me than open ones do, and I pay no attention to notification emails from GitHub; there are just too many.

> I note that this Chapel implementation is up on the PrimeView test streams for Tudor and Dave's machines but not from yours 

I believe this sorted itself out since you wrote this?

> We are aware that you prefer that no prior knowledge of the values of primes other than two are used, so the above technique would include code to handle the cases of nine and fifteen even though they are not primes.

This "forced unawareness" concerning what are primes and what are non-primes is not a preference on my part, but one of the "hard rules" for this project. An important part of the underlying reason is that Dave has been very clear that this is a non-negotiable for him.

> **What do you think? Could we add an invitation to this challenge to the front page?** It would then less penalize those languages whose authors don't have the expertise to come up with the "ultimate" technique themselves.

We could but if it is up to me, we won't. By design, the project issues an open invitation to (potential) contributors to be creative in their approaches, and be inspired by solutions developed by others. The implementations that perform best at any moment are easy to identify with the tooling we provide, and you can take personal credit for writing detailed documentation on the algorithms used therein. With all that in place, I think that people who are "meaningfully committed" to writing the fastest implementations in "their" language are properly facilitated in absorbing current insights, already.

GordonBGood:
@rbergen:

You may or may not see this, but just for closure:

> > We are aware that you prefer that no prior knowledge of the values of primes other than two are used, so the above technique would include code to handle the cases of nine and fifteen even though they are not primes.
> 
> This "forced unawareness" concerning what are primes and what are non-primes is not a preference on my part, but one of the "hard rules" for this project. An important part of the underlying reason is that Dave has been very clear that this is a non-negotiable for him.

Understood and that's fine.

> > **What do you think? Could we add an invitation to this challenge to the front page?** It would then less penalize those languages whose authors don't have the expertise to come up with the "ultimate" technique themselves.
> 
> We could but if it is up to me, we won't. By design, the project issues an open invitation to (potential) contributors to be creative in their approaches, and be inspired by solutions developed by others. The implementations that perform best at any moment are easy to identify with the tooling we provide, and you can take personal credit for writing detailed documentation on the algorithms used therein. With all that in place, I think that people who are "meaningfully committed" to writing the fastest implementations in "their" language are properly facilitated in absorbing current insights, already.

You are correct, this happens automatically as already I see some attempting to emulate the techniques mike_barbar and I use in our extreme unrolling and dense resetting techniques.  This is also why I have been increasing the size of my README.md files to more fully describe the techniques use and how to view the macro generated code, both for the benefit of you reviewer and others that are interested in how the techniques work with a view to doing the same.

rbergen:
@GordonBGood Just writing this to let you know I read your comment; I kept an eye on this one after yesterday. And I will stop doing so again, now. :)

PR #656, opened 2021-08-17
Rust - hybrid reset

mike-barber:
## Description
I've added a new algorithm (`hybrid`) that is a variation on the `striped-blocks`. Essentially, it's a specific reset algorithm used for small skip factors (under 8) that tackles each word and applies all masks for that word before proceeding to the next.

It can get a bit complicated, so I've added pretty extensive comments in the code, and added a bit of a writeup about the `striped` algorithms to the README, since they're being used in several other solutions now. 

I'm tagging @GordonBGood and @ItalyToast on this change. Hope you don't mind me mentioning you in the README, but your work definitely sparked this latest addition. I don't think it's quite the same as either of your solutions, but there are definitely some common themes. 

Apologies again @rbergen for another complex PR. I think this one is kind of fun though. I think it should be acceptable within the rules, but I'm happy to tweak a few things if you're concerned about any parts of it.

## Contributing requirements

### DIFF
--- a/PrimeRust/solution_1/README.md
+++ b/PrimeRust/solution_1/README.md
@@ -20,6 +20,37 @@ This is a somewhat Rust-idiomatic version that has the storage of prime flags ab
 
 Note that C++'s `vector<bool>` is a controversial specialisation of `vector` that `*may*` use more efficient storage than using a whole byte for 1 or 0. It's up to the compiler vendor, and is not standardised. Typically it's using individual bits for storage, within bytes or words. So the `bit storage` I have implemented closely resembles this. Rust has no such thing built in, although there are several crates available. For clarity, it's implemented manually in this solution.
 
+## Striped storage
+
+My `striped` variations have proved quite popular, and I'm very pleased to see that the approach has been successfully adopted in a number of other solutions. Given this popularity, it's worth writing a few notes about how they work.
+
+The original `striped` storage algorithm resulted from a hunch I had about the relative cost of creating a mask compared with the cost of applying it. I suspected that the cost of creating the mask, along with the cost of figuring out which address to apply it to, was quite high. So the `striped` storage solution took a slightly different approach to the storage. In the normal, linear, case, we store bits like the following. Pretend we have just two bytes of storage.
+
+```
+word0: [0   1   2   3   4   5   6   7], 
+word1: [8   9  10  11  12  13  14  15], 
+```
+
+In the `striped` storage case, we essentially _transpose_ this layout like this:
+
+```
+word0: [0   2   4   6   8  10  12  14]
+word1: [1   3   5   7   9  11  13  15]
+```
+
+This allows us to set a single mask for the first bit, then apply it with very regular `skip` factors to _all_ the words. Then proceed to calculate the mask for the second bit, and apply it to all the words again. This works really well, but it's not perfect: we're looping over all the words multiple times, and this has poor _locality_: we land up thrashing the cache, particularly on smaller processors.
+
+This brings us to the later algorithm, `striped-blocks`, that addresses this locality problem by splitting the problem into smaller chunks (blocks) of 16kB or 4kB. Instead of looping over all the words in the sieve and thrashing the cache, we apply the masks for each bit in turn to the first block. Once we're done with the first block, we proceed to the next. The compiler may be able to optimise it slightly better with known block sizes too, but I'm not sure if this is the case.
+
+Most recently, I've added an optional feature called `HYBRID` to `striped-blocks` that has a specific algorithm for resetting bits when we have smaller skip factors. The motivation for this is that we're typically touching every word in the block with small skip factors, and typically having to reset multiple bits in each word. We're not allowed to reset multiple bits in one operation, but we're definitely able to apply multiple (separate, single-bit) masks to a single word before proceeding to the next word. 
+
+This variation calculates the masks that will need to be applied: 8 of them, one for each bit, for each index. We have `SKIP` indices we need to keep track of, as the pattern repeats. So we can calculate all these masks before we loop through the words, then apply them in a very mechanical way with very little calculation required. Because the method is generic over the `SKIP` factor, the compiler can be quite aggressive with the loop optimisation. This seems to result in a 10-20% improvement in performance. And I think it's quite a fun application of const generics in Rust. 
+
+Part of the inspiration for this hybrid algorithm comes from the work of two people, whom I gratefully acknowledge here: 
+
+- @ItalyToast for the novel approach to resetting bits in a dense way in the traditional linear sieve in `PrimeCSharp/solution_4`. It got me thinking about how I could potentially apply a similar technique to the `striped-blocks`, although it required a bit of thinking.
+- @GordonBGood for his really interesting `PrimeNim/solution_3` implementation. I definitely had a good think about how I could apply a similar code-gen approach in Rust, and there's some interesting stuff I could potentially do with procedural macros. Before trying that, I decided to see what I could do with const generics and calculating masks, and that path led me to the hybrid algorithm.
+
 ## Run instructions
 
 - using Docker is the easiest way to get started without installing Rust:
--- a/PrimeRust/solution_1/src/main.rs
+++ b/PrimeRust/solution_1/src/main.rs
@@ -319,8 +319,19 @@ pub mod primes {
     /// This is a variation of [`FlagStorageBitVectorStriped`] that has better locality.
     /// The striped storage is divided up into smaller blocks, and we do multiple
     /// passes over the smaller block rather than the entire sieve.
-    pub struct FlagStorageBitVectorStripedBlocks<const N: usize, const HYBRID: bool> {
-        blocks: Vec<[u8; N]>,
+    ///
+    /// The implementation is generic over two parameters, making use of Rust's new
+    /// const generics.
+    /// - `BLOCK_SIZE` is the size of the blocks, in words (bytes seem to work best)
+    /// - `HYBRID` is a boolean specifying whether to enable a slightly different
+    ///   algorithm for resetting smaller factors.
+    ///   - `false` disables the algorithm, falling back on only the original striped block resets
+    ///   - `true` enables the new algorithm for smaller skip factors (under 8)
+    ///
+    /// Since there are a lot of bits to reset for smaller factors, there is a moderate
+    /// performance gain from using the `HYBRID` approach.
+    pub struct FlagStorageBitVectorStripedBlocks<const BLOCK_SIZE: usize, const HYBRID: bool> {
+        blocks: Vec<[u8; BLOCK_SIZE]>,
         length_bits: usize,
     }
 
@@ -338,6 +349,8 @@ pub mod primes {
     {
         const BLOCK_SIZE_BITS: usize = BLOCK_SIZE * U8_BITS;
 
+        /// Returns `1` if the `index` for a given `start`, and `skip`
+        /// should be reset; `0` otherwise.
         fn should_reset(index: usize, start: usize, skip: usize) -> u8 {
             let rel = index as isize - start as isize;
             if rel % skip as isize == 0 {
@@ -347,6 +360,23 @@ pub mod primes {
             }
         }
 
+        /// This is the new algorithm for resetting words in a different
+        /// order from the original `reset_flags_general` algorithm below.
+        ///
+        /// In the original algorithm, we reset all the first bits in the block,
+        /// then all the second bits, and so on. This works really well in the
+        /// general case, when we have a large skip factor, as we don't touch
+        /// most words.
+        ///
+        /// We use this algorithm when the skip factors are small, say less
+        /// than 8. In this case, we're typically touching every word in the block,
+        /// and we can expect each word to have multiple bits that need to be reset.
+        /// So we proceed in a different order, one word at a time. And for each
+        /// word, we reset the bits one by one.
+        ///
+        /// Note that the algorithm is generic over the `SKIP` factor, which
+        /// allows the compiler to do some extra optimisation. Each skip factor
+        /// we specify will result in specific code.
         #[inline(always)]
         fn reset_flags_dense<const SKIP: usize>(&mut self) {
             // earliest start to avoid resetting the factor itself
@@ -366,9 +396,10 @@ pub mod primes {
                     0
                 };
 
-                // Calculate the masks we're going to apply first.
-                // Note that we _could_ calculate a single mask word and apply it,
-                // but I believe that would be against the rules of the competition as
+                // Calculate the masks we're going to apply first. Note that each mask
+                // will reset only a single bit, which is why we have 8 separate masks.
+                // Note that we _could_ calculate a single mask word and apply it in a
+                // single operation, but I believe that would be against the rules as
                 // we would be resetting multiple bits in one operation if we did that.
                 let mut mask_set = [[0u8; U8_BITS]; SKIP];
                 for word_idx in 0..SKIP {
@@ -388,8 +419,8 @@ pub mod primes {
                     *word = masks.iter().fold(*word, |w, mask| w & mask);
                 }
 
-                // run through all exact size chunks; compiler loves to optimise
-                // known sizes
+                // run through all exact `SKIP` size chunks - the compiler is able to
+                // optimise known sizes quite well.
                 block.chunks_exact_mut(SKIP).for_each(|words| {
                     words
                         .iter_mut()
@@ -399,7 +430,7 @@ pub mod primes {
                         });
                 });
 
-                // run through the last stub of fewer than SKIP items
+                // run through the remaining stub of fewer than SKIP items
                 block
                     .chunks_exact_mut(SKIP)
                     .into_remainder()
@@ -417,6 +448,10 @@ pub mod primes {
             }
         }
 
+        /// This is the original striped-blocks algorithm, and proceeds to
+        /// set the first bit in every applicable word, then the second bit
+        /// and so on. This works really well for larger skip sizes, as the
+        /// words we need to reset are generally quite far apart.
         #[inline(always)]
         fn reset_flags_general(&mut self, start: usize, skip: usize) {
             // determine first block, start bit, and first word
@@ -485,11 +520,13 @@ pub mod primes {
             }
         }
 
+        /// Reset flags specified by the sieve. We use the optional
+        /// hybrid/dense reset methods for small factors if the
+        /// `HYBRID` type parameter is true, with the general
+        /// algorithm for higher skip factors. If `HYBRID` is false,
+        /// we rely only on the general approach for all skip factors.
         #[inline(always)]
         fn reset_flags(&mut self, start: usize, skip: usize) {
-            // Reset flags using the dense reset approach if HYBRID
-            // is enabled. Otherwise use the general approach for
-            // all skip factors.
             if HYBRID {
                 match skip {
                     // We only really gain an advantage from dense
--- a/PrimeRust/solution_1/src/main.rs
+++ b/PrimeRust/solution_1/src/main.rs
@@ -241,8 +241,8 @@ pub mod primes {
     /// There is a computation / memory bandwidth tradeoff here. This works well
     /// only for sieves that fit inside the processor cache. For processors with
     /// smaller caches or larger sieves, this algorithm will result in a lot of
-    /// cache thrashing due to multiple passes. It really doesn't work well on something 
-    /// like a raspberry pi. 
+    /// cache thrashing due to multiple passes. It really doesn't work well on something
+    /// like a raspberry pi.
     ///
     /// [`FlagStorageBitVectorStripedBlocks`] takes a more cache-friendly approach.
     pub struct FlagStorageBitVectorStriped {
@@ -319,7 +319,7 @@ pub mod primes {
     /// This is a variation of [`FlagStorageBitVectorStriped`] that has better locality.
     /// The striped storage is divided up into smaller blocks, and we do multiple
     /// passes over the smaller block rather than the entire sieve.
-    pub struct FlagStorageBitVectorStripedBlocks<const N: usize> {
+    pub struct FlagStorageBitVectorStripedBlocks<const N: usize, const HYBRID: bool> {
         blocks: Vec<[u8; N]>,
         length_bits: usize,
     }
@@ -333,27 +333,97 @@ pub mod primes {
     /// in parallel.
     pub const BLOCK_SIZE_SMALL: usize = 4 * 1024;
 
-    impl<const N: usize> FlagStorageBitVectorStripedBlocks<N> {
-        const BLOCK_SIZE: usize = N;
-        const BLOCK_SIZE_BITS: usize = Self::BLOCK_SIZE * U8_BITS;
-    }
+    impl<const BLOCK_SIZE: usize, const HYBRID: bool>
+        FlagStorageBitVectorStripedBlocks<BLOCK_SIZE, HYBRID>
+    {
+        const BLOCK_SIZE_BITS: usize = BLOCK_SIZE * U8_BITS;
 
-    impl<const N: usize> FlagStorage for FlagStorageBitVectorStripedBlocks<N> {
-        fn create_true(size: usize) -> Self {
-            let num_blocks = size / Self::BLOCK_SIZE_BITS + (size % Self::BLOCK_SIZE_BITS).min(1);
-            Self {
-                length_bits: size,
-                blocks: vec![[u8::MAX; N]; num_blocks],
+        fn should_reset(index: usize, start: usize, skip: usize) -> u8 {
+            let rel = index as isize - start as isize;
+            if rel % skip as isize == 0 {
+                1
+            } else {
+                0
             }
         }
 
         #[inline(always)]
-        fn reset_flags(&mut self, start: usize, skip: usize) {
+        fn reset_flags_dense<const SKIP: usize>(&mut self) {
+            // earliest start to avoid resetting the factor itself
+            let start = SKIP / 2 + SKIP;
+            debug_assert!(
+                start < BLOCK_SIZE,
+                "algorithm only correct for small skip factors"
+            );
+            for (block_idx, block) in self.blocks.iter_mut().enumerate() {
+                // Preserve the first bit of one word we know we're going to overwrite
+                // with the masks. Its cheaper to put it back afterwards than break the loop
+                // into two sections with different rules. Only applicable on the first block:
+                // this is the factor itself, and we don't want to reset that flag.
+                let preserved_word_mask = if block_idx == 0 {
+                    block[SKIP / 2] & 1
+                } else {
+                    0
+                };
+
+                // Calculate the masks we're going to apply first.
+                // Note that we _could_ calculate a single mask word and apply it,
+                // but I believe that would be against the rules of the competition as
+                // we would be resetting multiple bits in one operation if we did that.
+                let mut mask_set = [[0u8; U8_BITS]; SKIP];
+                for word_idx in 0..SKIP {
+                    for bit in 0..8 {
+                        let block_index_offset = block_idx * BLOCK_SIZE * U8_BITS;
+                        let bit_index_offset = bit * BLOCK_SIZE;
+                        let index = block_index_offset + bit_index_offset + word_idx;
+                        mask_set[word_idx][bit] = !(Self::should_reset(index, start, SKIP) << bit);
+                    }
+                }
+                // rebind as immutable
+                let mask_set = mask_set;
+
+                /// apply all 8 masks - one for each bit - using a fold, mostly
+                /// because folds are fun
+                fn apply_masks(word: &mut u8, masks: &[u8; U8_BITS]) {
+                    *word = masks.iter().fold(*word, |w, mask| w & mask);
+                }
+
+                // run through all exact size chunks; compiler loves to optimise
+                // known sizes
+                block.chunks_exact_mut(SKIP).for_each(|words| {
+                    words
+                        .iter_mut()
+                        .zip(mask_set.iter().copied())
+                        .for_each(|(word, masks)| {
+                            apply_masks(word, &masks);
+                        });
+                });
+
+                // run through the last stub of fewer than SKIP items
+                block
+                    .chunks_exact_mut(SKIP)
+                    .into_remainder()
+                    .iter_mut()
+                    .zip(mask_set.iter().copied())
+                    .for_each(|(word, masks)| {
+                        apply_masks(word, &masks);
+                    });
+
+                // restore the first bit on the preserved word in the first block,
+                // as noted above
+                if block_idx == 0 {
+                    block[SKIP / 2] |= preserved_word_mask;
+                }
+            }
+        }
+
+        #[inline(always)]
+        fn reset_flags_general(&mut self, start: usize, skip: usize) {
             // determine first block, start bit, and first word
             let block_idx_start = start / Self::BLOCK_SIZE_BITS;
             let offset_idx = start % Self::BLOCK_SIZE_BITS;
-            let mut bit_idx = offset_idx / Self::BLOCK_SIZE;
-            let mut word_idx = offset_idx % Self::BLOCK_SIZE;
+            let mut bit_idx = offset_idx / BLOCK_SIZE;
+            let mut word_idx = offset_idx % BLOCK_SIZE;
 
             for block_idx in block_idx_start..self.blocks.len() {
                 // Safety: we have ensured the block_idx < length
@@ -361,9 +431,8 @@ pub mod primes {
                 while bit_idx < U8_BITS {
                     // calculate effective end position: we might have a shorter stripe on the last iteration
                     let stripe_start_position =
-                        block_idx * Self::BLOCK_SIZE_BITS + bit_idx * Self::BLOCK_SIZE;
-                    let effective_len =
-                        Self::BLOCK_SIZE.min(self.length_bits - stripe_start_position);
+                        block_idx * Self::BLOCK_SIZE_BITS + bit_idx * BLOCK_SIZE;
+                    let effective_len = BLOCK_SIZE.min(self.length_bits - stripe_start_position);
 
                     // get mask for this bit position
                     let mask = !(1 << bit_idx);
@@ -390,17 +459,52 @@ pub mod primes {
                     }
 
                     // early termination: this is the last stripe
-                    if effective_len != Self::BLOCK_SIZE {
+                    if effective_len != BLOCK_SIZE {
                         return;
                     }
 
                     // bit/stripe complete; advance to next bit
                     bit_idx += 1;
-                    word_idx -= Self::BLOCK_SIZE;
+                    word_idx -= BLOCK_SIZE;
                 }
 
                 // block complete; reset bit index and proceed with the next block
                 bit_idx = 0;
+            }
+        }
+    }
+
+    impl<const BLOCK_SIZE: usize, const HYBRID: bool> FlagStorage
+        for FlagStorageBitVectorStripedBlocks<BLOCK_SIZE, HYBRID>
+    {
+        fn create_true(size: usize) -> Self {
+            let num_blocks = size / Self::BLOCK_SIZE_BITS + (size % Self::BLOCK_SIZE_BITS).min(1);
+            Self {
+                length_bits: size,
+                blocks: vec![[u8::MAX; BLOCK_SIZE]; num_blocks],
+            }
+        }
+
+        #[inline(always)]
+        fn reset_flags(&mut self, start: usize, skip: usize) {
+            // Reset flags using the dense reset approach if HYBRID
+            // is enabled. Otherwise use the general approach for
+            // all skip factors.
+            if HYBRID {
+                match skip {
+                    // We only really gain an advantage from dense
+                    // resetting up to skip factors under 8, as after
+                    // that, we're expecting the resets to be sparse.
+                    // We only get called for skip factors, so there's
+                    // no point adding cases for even numbers.
+                    1 => self.reset_flags_dense::<1>(),
+                    3 => self.reset_flags_dense::<3>(),
+                    5 => self.reset_flags_dense::<5>(),
+                    7 => self.reset_flags_dense::<7>(),
+                    _ => self.reset_flags_general(start, skip),
+                }
+            } else {
+                self.reset_flags_general(start, skip);
             }
         }
 
@@ -411,8 +515,8 @@ pub mod primes {
             }
             let block = index / Self::BLOCK_SIZE_BITS;
             let offset = index % Self::BLOCK_SIZE_BITS;
-            let bit_index = offset / Self::BLOCK_SIZE;
-            let word_index = offset % Self::BLOCK_SIZE;
+            let bit_index = offset / BLOCK_SIZE;
+            let word_index = offset % BLOCK_SIZE;
             let word = self.blocks.get(block).unwrap().get(word_index).unwrap();
             *word & (1 << bit_index) != 0
         }
@@ -581,6 +685,11 @@ struct CommandLineOptions {
     #[structopt(long)]
     bits_striped_blocks: bool,
 
+    /// Run variant that uses bit-level storage, using striped storage in blocks with
+    /// hybrid dense resets for smaller skip factors
+    #[structopt(long)]
+    bits_striped_hybrid: bool,
+
     /// Run variant that uses byte-level storage
     #[structopt(long)]
     bytes: bool,
@@ -606,6 +715,7 @@ fn main() {
         opt.bits_rotate,
         opt.bits_striped,
         opt.bits_striped_blocks,
+        opt.bits_striped_hybrid,
         opt.bytes,
     ]
     .iter()
@@ -676,7 +786,7 @@ fn main() {
             thread::sleep(Duration::from_secs(1));
             print_header(threads, limit, run_duration);
             for _ in 0..repetitions {
-                run_implementation::<FlagStorageBitVectorStripedBlocks<BLOCK_SIZE_DEFAULT>>(
+                run_implementation::<FlagStorageBitVectorStripedBlocks<BLOCK_SIZE_DEFAULT, false>>(
                     "bit-storage-striped-blocks",
                     1,
                     run_duration,
@@ -687,8 +797,34 @@ fn main() {
             }
 
             for _ in 0..repetitions {
-                run_implementation::<FlagStorageBitVectorStripedBlocks<BLOCK_SIZE_SMALL>>(
+                run_implementation::<FlagStorageBitVectorStripedBlocks<BLOCK_SIZE_SMALL, false>>(
                     "bit-storage-striped-blocks-small",
+                    1,
+                    run_duration,
+                    threads,
+                    limit,
+                    opt.print,
+                );
+            }
+        }
+
+        if opt.bits_striped_hybrid || run_all {
+            thread::sleep(Duration::from_secs(1));
+            print_header(threads, limit, run_duration);
+            for _ in 0..repetitions {
+                run_implementation::<FlagStorageBitVectorStripedBlocks<BLOCK_SIZE_DEFAULT, true>>(
+                    "bit-storage-striped-hybrid",
+                    1,
+                    run_duration,
+                    threads,
+                    limit,
+                    opt.print,
+                );
+            }
+
+            for _ in 0..repetitions {
+                run_implementation::<FlagStorageBitVectorStripedBlocks<BLOCK_SIZE_SMALL, true>>(
+                    "bit-storage-striped-hybrid-small",
                     1,
                     run_duration,
                     threads,
@@ -758,7 +894,7 @@ fn run_implementation<T: 'static + FlagStorage + Send>(
         // print results to stderr for convenience
         print_results_stderr(
             label,
-            &sieve,
+            sieve,
             print_primes,
             duration,
             total_passes,
@@ -791,10 +927,17 @@ mod tests {
     }
 
     #[test]
-    fn sieve_known_correct_bits_striped_blocks() {
+    fn sieve_known_correct_bits_striped_blocks_general() {
         // check both sizes
-        sieve_known_correct::<FlagStorageBitVectorStripedBlocks<BLOCK_SIZE_DEFAULT>>();
-        sieve_known_correct::<FlagStorageBitVectorStripedBlocks<BLOCK_SIZE_SMALL>>();
+        sieve_known_correct::<FlagStorageBitVectorStripedBlocks<BLOCK_SIZE_DEFAULT, false>>();
+        sieve_known_correct::<FlagStorageBitVectorStripedBlocks<BLOCK_SIZE_SMALL, false>>();
+    }
+
+    #[test]
+    fn sieve_known_correct_bits_striped_blocks_hybrid() {
+        // check both sizes
+        sieve_known_correct::<FlagStorageBitVectorStripedBlocks<BLOCK_SIZE_DEFAULT, true>>();
+        sieve_known_correct::<FlagStorageBitVectorStripedBlocks<BLOCK_SIZE_SMALL, true>>();
     }
 
     #[test]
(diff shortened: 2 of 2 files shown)


rbergen:
> Apologies again @rbergen for another complex PR. 

Well, if you apologise for that then I'll have to reciprocate by apologising for the time it can take me to wrap my head around some of the more complex contributions. The depth and breadth of skills and knowledge that are being applied to this project are, at times, both mind-blowing and humbling.

> I think it should be acceptable within the rules, but I'm happy to tweak a few things if you're concerned about any parts of it.

I'll try to take my time to make up my mind about this. I appreciate the time you and others have spent and may have to spend on getting me there. :)

GordonBGood:
@mike-barber:

> I'm tagging @GordonBGood on this change. Hope you don't mind me mentioning you in the README, but your work definitely sparked this latest addition. I don't think it's quite the same as your solution, but there are definitely some common themes.

Hi, thanks for the tag.

I don't think this "hybrid" approach much resembles my "Extreme Loop Unrolling" from "PrimeNim/solution_3".  I'll comment further at the end of this post.

From your "README.md" file in this submission:
> @GordonBGood for his really interesting `PrimeNim/solution_3` implementation. I definitely had a good think about how I could apply a similar **code-gen approach** in Rust, and there's some interesting stuff I could potentially do with **procedural macros**. Before trying that, I decided to see what I could do with const generics and calculating masks, and that path led me to the hybrid algorithm.

In fact, it doesn't take macros or templates to use my technique:  one could do the whole thing in languages without either by using a separate code generation step for the thousand or two lines of code for the whole Monty, but in this trivial benchmark there are only four cases out of the 64 different patterns actually used (odd base primes modulo eight all have the same marking start bit index modulo eight when starting from their first appearance), so it could even be done by hand.

I have done this in Rust (the full Monty) using `macro_rules` templates, but have been waiting to see what you would do with it before submitting my solution and if your version was as fast as mine then I wouldn't bother, as your version is likely to be more Rust idiomatic since Rust isn't one of my primary languages.  I couldn't see that Rust guarantees generating a computed goto, so implemented the functionality of a computed goto manually using an Look Up Table array of functions with the above mentioned `macro_rules` "template" generating the functions for the entries in the table.  I think one could use a full procedural macro to do the whole thing, but read that these need to be compiled separately (something like Template Haskell) so couldn't be bothered with the extra effort as the resulting code would likely be the same.

If you want to see an example of how I did this in Rust, look at my recently submitted Chapel PR, which essentially does the above using a Chapel parameterized generic class as a template.

mike-barber:
> I'll try to take my time to make up my mind about this. I appreciate the time you and others have spent and may have to spend on getting me there. :)

Thanks @rbergen -- I think everyone really appreciates all the effort it takes to moderate the myriad different languages and approaches. I hope you're still enjoying it. I definitely am, and definitely learning plenty of cool techniques from all the solutions.

mike-barber:
> If you want to see an example of how I did this in Rust, look at my recently submitted Chapel PR, which essentially does the above using a Chapel parameterized generic class as a template.

Thanks @GordonBGood -- your Chapel implementation makes a lot of sense, and I understand how the algorithm works properly now. I really like it. I think I like Chapel too :) 

> I don't think this "hybrid" approach much resembles my "Extreme Loop Unrolling" from "PrimeNim/solution_3". I'll comment further at the end of this post. 

I agree. It's fairly different, as a result of me having transposed the storage order for the bits. But I do have to give you credit for making me think about different ways to approach the problem by giving the compiler more constants to work with. 

> I have done this in Rust (the full Monty) using macro_rules templates, but have been waiting to see what you would do with it before submitting my solution and if your version was as fast as mine then I wouldn't bother, as your version is likely to be more Rust idiomatic since Rust isn't one of my primary languages.

Personally, I'd be really happy to see you submit your Rust solution either way. Nobody else has really gone to town with macros just yet :)

rbergen:
> Thanks @rbergen -- I think everyone really appreciates all the effort it takes to moderate the myriad different languages and approaches. 

I don't think everyone does, and that's okay. "Everyone" is too high a goal to aim for; it's inevitable that I make some decisions that certain people don't like. I do really appreciate your positive feedback, and similar feedback I get of others. I can honestly say that I'm doing the best I can, and it is nice that that is appreciated. 

> I hope you're still enjoying it. I definitely am, and definitely learning plenty of cool techniques from all the solutions.

Generally speaking, I do. I love the fact we now have 70+ languages, including some I couldn't imagine implementing a prime sieve with, and/or I thought died years (decades?) back. I love the group of contributors who keep throwing all they have at squeezing just a little more out of their language(s) of choice (you're one of them). I love what we've been able to do regarding CI and benchmarking. And I love I've developed something for fun in my favorite language (C#), that more than just a handful of people use (that something being PrimeView).
The slightly less enjoyable things, and they are the exceptions, get more than compensated by the trust and patience I'm given by so many skilled, dedicated and overall nice people.

rbergen (approved):
LGTM, @mike-barber with thanks for the helpful code comments. Can I merge this or are you considering making changes?

mike-barber:
Thanks @rbergen 

> I love what we've been able to do regarding CI and benchmarking. And I love I've developed something for fun in my favorite language (C#), that more than just a handful of people use (that something being PrimeView).

I'm really glad you're enjoying it still. It's a pretty cool setup you've got now for collecting results and reporting. I didn't realise your favourite language was C# :) It's my primary language these days, but I get perhaps too much of it at work, so I play around with Rust and stuff in my free time. 

I'm ready for you to merge it. I'm always thinking of ways to tweak it further, but I've run out of bright ideas for now. :) Thanks for your prompt review!

GordonBGood:
@mike-barber:

> I think I like Chapel too :)

Chapel has it's points if you don't mind "curly bracket languages **plus** semicolons" and OOP although I think its implementation of generics needs some more work as it may not be "type complete" from a functional sense; however generics work well enough for their purpose.

The main wart on Chapel is fairly long compile times, I think even longer than those of Rust:  it can take ten seconds just to compile a "hello World" in Chapel in `--fast` optimizing mode, and about seven seconds in "ordinary" mode.

> > I have done this in Rust (the full Monty) using macro_rules templates, but have been waiting to see what you would do with it before submitting my solution and if your version was as fast as mine then I wouldn't bother, as your version is likely to be more Rust idiomatic since Rust isn't one of my primary languages.
> 
> Personally, I'd be really happy to see you submit your Rust solution either way.

I'm on the fence on this, as the "race" rules now have us try to avoid duplication of ideas and code and I don't want to keep all the fun to myself ;-)  I am coaching a Julia contributor to implement this and if you want to take a stab at it in Rust, I will do the same for you.  I have enough on my plate as I think I will show how to do this in GHC Haskell, for which no one else seems to have my expertise.

After I've finished with Haskell and tuning Nim, I will likely implement this in Rust and Julia if no one else has done it yet.  It isn't too hard in Rust or Julia because there are templates/macros.

> Nobody else has really gone to town with macros just yet :)

Yes, nobody else has used macros much yet, if at all.  In my Nim solution_3, I use a AST macro as a code generator to fairly good effect.  Of course, not all languages have macros; C/C++ preprocessor "macros" are really just text substitution templates, which can in fact be used for this requirement as I used the better templating capabilities of Chapel.  As I said, Rust's `macro_rules` are really templates and not macros but procedural macros are the real thing.  Julia has real AST macros that can generate the required expressions and can be used for code generation.  Many/most of the other languages don't have macros, and many don't even have templates, meaning that to implement "Extreme Loop Unrolling" would need a code generation step (or a lot of patience writing all that code manually) or cutting the cases down to only those actually used (as I said, only four).

The problem with this "unrolling" technique is that it requires some reasonably good branch prediction for the multiple branches that are fairly random, so less processors such as Dave Plummer's Bay Trail Celeron J1800 con't really cut it, perhaps due to insufficient Branch Look-aside Buffer's (BLB's).  Thus, it runs pretty well on more modern CPU's or even newer ARM but not-so-well on old clunkers, where just what I call "loop unpeeling" (which is similar to what you call "striping" but doesn't change the bit order) can run as fast.  I am going to look into this a bit to see what can be done.

mike-barber:
> I'm on the fence on this, as the "race" rules now have us try to avoid duplication of ideas and code and I don't want to keep all the fun to myself ;-) I am coaching a Julia contributor to implement this and if you want to take a stab at it in Rust, I will do the same for you. I have enough on my plate as I think I will show how to do this in GHC Haskell, for which no one else seems to have my expertise.

Thanks @GordonBGood -- I would like to take a stab at this over the weekend. I think it'll be a fun thing to implement. I'll dust off the Rust macros :) And it'll be fantastic having your input on it. 

> The problem with this "unrolling" technique is that it requires some reasonably good branch prediction for the multiple branches that are fairly random, so less processors such as Dave Plummer's Bay Trail Celeron J1800 con't really cut it, perhaps due to insufficient Branch Look-aside Buffer's (BLB's). Thus, it runs pretty well on more modern CPU's or even newer ARM but not-so-well on old clunkers, where just what I call "loop unpeeling" (which is similar to what you call "striping" but doesn't change the bit order) can run as fast. I am going to look into this a bit to see what can be done.

I didn't realise the J1800 is really an Atom part until I had a closer look at it this morning. They're pretty slow on a number of fronts by the looks of it. I had a look through [Agner Fog's notes](https://www.agner.org/optimize/microarchitecture.pdf) on the Silvermont microarchitecture. Interesting, definitely, but I'm not sure what modifications one could make to code to improve performance on it: that's way beyond the scope of my knowledge. It's also pretty difficult not actually having one available to hack on. 

It does sound like the two integer ports are contended for what we're doing: masking and address calculation. And there might be a bit of a load/store bottleneck too. I'm sure you're right about the branch prediction issue as well: I'm sure they've kept it as simple as possible in order to hit their power targets.

GordonBGood:
@mike-barber:

> I didn't realize that the J1800 is really an Atom part until I had a closer look at it this morning. They're pretty slow on a number of fronts by the looks of it. I had a look through Agner Fog's notes on the Silvermont microarchitecture.

You'll note that the Intel J1800 is actually Atom Bay Trail which is the oldest of Intel's attempts to get into the mobile/phone/tablet market as well as being slightly better processors for Netbooks (remember those, which is perhaps what Dave Plummer actually has).  Due to the lack of sophistication of the architecture, it has roughly half of the instructions per clock cycle of more modern architectures such as even Sandy Bridge, let alone Sky Lake and newer.  But you are reading the same sources as I am, so you likely know (Agner Fog is a great resource isn't he?).

>  Interesting, definitely, but I'm not sure what modifications one could make to code to improve performance on it: that's way beyond the scope of my knowledge.
>
> It does sound like the two integer ports are contended for what we're doing: masking and address calculation. And there might be a bit of a load/store bottleneck too. I'm sure you're right about the branch prediction issue as well: I'm sure they've kept it as simple as possible in order to hit their power targets.

As well as the bottlenecks I (and Agner Fog) mentioned, I think that the cache implementations are much more primitive than a more modern CPU and, since we aren't allowed to use full page segmentation, we are suffering from "cache thrashing".  We have already improved things by culling by "blocks" but our culling target is currently a window on the full deliverable sieve buffer.  I think something can be done by allocating a separate "block"/cache sized buffer and doing all of the culling/marking to that with a fast copy memory operation (a low level operation that calls the CPU instruction directly on powerful low level languages such as C/C++/Rust/Nim, etc.) may improve things on such a primitive CPU, as `cpymem` instructions bypass the cache handling for speed.

I am going to try this and if it works, all implementations that use "blocks" can do the same.  It still passes the `base` requirements as the algorithm still stays the same as culling per base prime value by blocks, just we have moved the culling out to a separate smaller buffer. 
 We still aren't doing full page segmentation where we would find the base prime values as a separate loop and just apply all of them per page which is not allowed.  I have already implemented a full page segmented version which was not allowed which was faster, but more importantly it showed that the cost of the memory copy is only about 10 percent of even very fast versions such as my Nim ( or your fast Rust) contribution, where as the cost of "cache thrashing" is about 25 percent or more even on newer CPU's and likely to be much more on primitive CPU's such as this.

> It's also pretty difficult not actually having one available to hack on.

I actually have a Windows tablet computer with a slightly newer Airmont CPU but I've lent/given it to my daughter for online schooling.  I'll see if I can borrow it for an hour or two between when she is doing homework assignments, etc.  It seems to have about the same micro architecture as the Bay Trail CPU's, just tweaked for lower power consumption, four cores, and no Hyper Threading, none of which is important for purposes of this benchmark.

GordonBGood:
@mike-barber:

Just spewing:

> I think something can be done by allocating a separate "block"/cache sized buffer and doing all of the culling/marking to that with a fast copy memory operation

The above doesn't work without full page segmentation since the results of culling for each of the individual base prime values otherwise need to be combined, which would destroy any savings made.

The more I think about it, the more I think that the bottleneck my "Extreme Loop Unrolling" technique runs into on primitive CPU's is not the branch prediction so much as less efficient "cache thrashing", which we are all running into.  The reason that some implementations of just "striping"/"unpeeling" are so fast and approach (or pass on primitive CPU's) my fancy technique is likely that the LLVM back-ends as for Rust and Julia are turning the simple code loops into something that reduces the number of machine cycles and by using less op codes and/or simpler op codes, I haven't looked at the assembly produced yet to see if this is true, but it must be.  I tried using clang, which is the LLVM C/C++ compiler as my Nim backend, and for "striping"/"unpeeling" I get about seven percent faster results in line with those of the same technique used in Julia, where for my macro based "Extreme Loop Unrolling" the results are actually worse.  I think I know why it's worse, as LLVM has a known miss optimization issue in misusing Loop Strength Reduction to actually de-optimize for the complex loops in the latter.

So we are running into the limits of the use of our tools.

I have had a look at why this "hybrid" technique is so fast and I believe it has nothing to do with code presenting better branch prediction possibilities or being more cache friendly but because LLVM is able to reduce the number of complex machine instructions used for the loops into simple machine instructions (two or three times faster) for the small base prime value cases, and since for this trivial range about half of the culling/marking operations qualify for the "hybrid" approach, the time saving is quite appreciable.

I have already run my loop unrolling code on my daughters Windows tablet which has close enough to the same CPU as Dave Plummer is using in the testing, and when I reduce the sieving range so it fits into the CPU L1 cache, it runs at about 1.25 CPU clock cycles per cull, but when I increase the range (not using page segmentation) to sieve to a million, the clock cycles per cull increases to about three, where the simple "unpeeling" loops don't increase as much because they are more register based.

In order to use something like your "hybrid" technique, I would likely use my macro capabilities to just cover the special cases within the generated code, and it will then likely run about as fast as this or maybe a bit faster.  My point is that, due to not being able to use page segmentation, we are doing things just get a high score on the race which is starting to get further and further away from a "real" Sieve of Eratosthenes application to sieve to non-toy ranges like billions or trillion.  In such an application (which of course would use page segmentation), most of the small base prime values which make the "hybrid" technique work wouldn't be marked manually anyway, as they would be initialized by a `cpymem` from a pre-culled pattern array, which is even faster than your "hybrid" technique, but can't be used here.

My rant is about the definition of the benchmark to some extent, as it doesn't represent reality; to represent reality, it would be an odds-only (bit-packed where possible) sieve over a range that fits into typical CPU L1 cache like 16 Kilobytes (all of the test CPU's have 32 Kilobytes, although it looks like the ARM one doesn't use it's cache so efficiently) or the benchmark specification would be built around using page segmentation and even, perhaps, pre-filling with all the small base primes up to about 17 as would normally be used in a real application.

Anyway, I guess I'll look into expanding my macro to include the "hybrid" optimizations for small base prime values in order to present competitive results, and may have to have a "striped"/"unpreeled" block implementation for primitive CPU's that can't handle the macro algorithm with "cache thrashing".  My problem is that the extra work to do this isn't something that I can use in my sieving applications for the above reasons.

mike-barber:
Hi @GordonBGood - apologies for the slow reply. I've had a very busy filled with dev-ops "fun" and herding cats with tons of pull requests. Fridays are usually a bit quieter. 

> You'll note that the Intel J1800 is actually Atom Bay Trail which is the oldest of Intel's attempts to get into the mobile/phone/tablet market as well as being slightly better processors for Netbooks (remember those, which is perhaps what Dave Plummer actually has). Due to the lack of sophistication of the architecture, it has roughly half of the instructions per clock cycle of more modern architectures such as even Sandy Bridge, let alone Sky Lake and newer. But you are reading the same sources as I am, so you likely know (Agner Fog is a great resource isn't he?).

Absolutely! I really enjoy reading his material. I'm really not an expert on microarchitecture, but I do have an appreciation for the engineering that goes into it. My knowledge gets particularly hazy around caches: I know about the different latencies and so on, but it gets a bit fuzzy when looking at ways and sets and so forth. 

> The more I think about it, the more I think that the bottleneck my "Extreme Loop Unrolling" technique runs into on primitive CPU's is not the branch prediction so much as less efficient "cache thrashing", which we are all running into. The reason that some implementations of just "striping"/"unpeeling" are so fast and approach (or pass on primitive CPU's) my fancy technique is likely that the LLVM back-ends as for Rust and Julia are turning the simple code loops into something that reduces the number of machine cycles and by using less op codes and/or simpler op codes, I haven't looked at the assembly produced yet to see if this is true, but it must be.

I think this is the case too, although I am quite surprised they're still fast on the Celeron: it doesn't have a huge vector unit to do smart things. It could go some way to explaining why they're so fast on my machine, though. I also haven't had a good look at the assembly lately: the project got quite big, so I need to to a bit of fiddling to get something readable out of it for perusal.

> I have already run my loop unrolling code on my daughters Windows tablet which has close enough to the same CPU as Dave Plummer is using in the testing, and when I reduce the sieving range so it fits into the CPU L1 cache, it runs at about 1.25 CPU clock cycles per cull, but when I increase the range (not using page segmentation) to sieve to a million, the clock cycles per cull increases to about three, where the simple "unpeeling" loops don't increase as much because they are more register based.

Looking at the processors, it seems that most common ones have 32kB of L1d, but the Celeron is only about 24kB. So rough thumb-suck, it's doing 3 full caches every loop through the sieve, whereas other processors only have to do about 2. There might be more to it though: maybe it's really slow getting data from L2 into L1 compared to more normal processors. Or, as you point out, a simpler caching strategy could be at fault. Or some other factor, like it's really bad at doing read-ahead or something. 

> My rant is about the definition of the benchmark to some extent, as it doesn't represent reality; to represent reality, it would be an odds-only (bit-packed where possible) sieve over a range that fits into typical CPU L1 cache like 16 Kilobytes (all of the test CPU's have 32 Kilobytes, although it looks like the ARM one doesn't use it's cache so efficiently) or the benchmark specification would be built around using page segmentation and even, perhaps, pre-filling with all the small base primes up to about 17 as would normally be used in a real application.

I agree - it is quite an "artificial" benchmark in many ways. I guess what we're trying to achieve within the boundaries of the problem is going to be dominated by cache effects. 

I think the "hybrid" approach could work quite well for your unrolled solutions, and would be an interesting thing to try. The other thing I'd like to do at some stage is see if I can expand the striped one to be a bit smarter in its use of constants: I know I've cut some corners there. Didn't really have time to fully explore it. Although at this point, the idea of implementing your approach in Rust is sounding most interesting to me :)

GordonBGood:
@mike-barber:

Hi Mike, this post is mostly about caches and "cache thrashing":

> I know about the different latencies and so on, but it gets a bit fuzzy when looking at ways and sets and so forth.

I'm no expert either, but a comparison like this benchmark will probably help put some things together ;-)

> > The more I think about it, the more I think that the bottleneck my "Extreme Loop Unrolling" technique runs into on primitive CPU's is not the branch prediction so much as less efficient "cache thrashing", which we are all running into. The reason that some implementations of just "striping"/"unpeeling" are so fast and approach (or pass on primitive CPU's) my fancy technique is likely that the LLVM back-ends as for Rust and Julia are turning the simple code loops into something that reduces the number of machine cycles and by using less op codes and/or simpler op codes, I haven't looked at the assembly produced yet to see if this is true, but it must be.
> 
> I think this is the case too, although I am quite surprised they're still fast on the Celeron: it doesn't have a huge vector unit to do smart things. It could go some way to explaining why they're so fast on my machine, though. I also haven't had a good look at the assembly lately: the project got quite big, so I need to to a bit of fiddling to get something readable out of it for perusal.

It seems that the cache "ways" on the ARM Cortex 72 is somewhat worse than even the Celeron as evidenced by that your small "block" versions run about five percent faster on ARM than the larger "block" versions even though the ARM L1 data cache is 32 Kilobytes and thus bigger than either.  I have some ideas as to why these "lesser" CPU's cache performance is so poor for our benchmarks, which I'll get into at the end of this post.

As for "a good look at the assembly code produced", the following is the crucial unrolled "unpeeling" (which is the ordered version of your "striped" technique) loop from Julia, which uses LLVM to produce its code like Rust:

```x86_64
L992:
        orb     %dl, -1(%r12,%rsi)
        orb     %dl, (%r13,%rsi)
        orb     %dl, (%r8,%rsi)
        orb     %dl, (%rax,%rsi)
        addq    %rcx, %rsi   ; add prime to base
        cmpq    %rdi, %rsi   ; compare base to limit
        jle     L992
```
As you can see, it is pretty well stripped down to minimum!  This loop will be run eight times per block, plus another loop by ones'ys just to complete the "block".

> > I have already run my loop unrolling code on my daughters Windows tablet which has close enough to the same CPU as Dave Plummer is using in the testing, and when I reduce the sieving range so it fits into the CPU L1 cache, it runs at about 1.25 CPU clock cycles per cull, but when I increase the range (not using page segmentation) to sieve to a million, the clock cycles per cull increases to about three, where the simple "unpeeling" loops don't increase as much because they are more register based.
> 
> Looking at the processors, it seems that most common ones have 32kB of L1d, but the Celeron is only about 24kB. So rough thumb-suck, it's doing 3 full caches every loop through the sieve, whereas other processors only have to do about 2. There might be more to it though: maybe it's really slow getting data from L2 into L1 compared to more normal processors. Or, as you point out, a simpler caching strategy could be at fault. Or some other factor, like it's really bad at doing read-ahead or something.

As for Celeron, like my daughter's Cherry Trail CPU, it has only 24 Kilobytes, but even your biggest "block" is only 16 Kilobytes so that's still about four pages going through the about 62.5 Kilobyte deliverable bit sieve buffer, so it's the same in that regard.  I have some ideas on why the cache performance is so slow to be discussed at the end of this post.

> > My rant is about the definition of the benchmark to some extent, as it doesn't represent reality; to represent reality, it would be an odds-only (bit-packed where possible) sieve over a range that fits into typical CPU L1 cache like 16/32 Kilobytes (all of the test CPU's have 32 Kilobytes, although it looks like the ARM one doesn't use it's cache so efficiently) or the benchmark specification would be built around using page segmentation and even, perhaps, pre-filling with all the small base primes up to about 17 as would normally be used in a real application.
>
> I agree - it is quite an "artificial" benchmark in many ways. I guess what we're trying to achieve within the boundaries of the problem is going to be dominated by cache effects.

Amen to that - the problem with this benchmark is that it is so influenced by cache effects, particularly for the "lesser" CPU's.

> I think the "hybrid" approach could work quite well for your unrolled solutions, and would be an interesting thing to try.

I am in process or trying it and have come to the realization that using macros as for my unrolled approach is just another way of implementing it and can optionally cover all the cases that we don't use for this benchmark:  because this benchmark must cull the whole sieve buffer for each prime value, it is just another way of doing what you do with your generic code based on the step value, which is then specialized only for the step values actually used.

> The other thing I'd like to do at some stage is see if I can expand the striped one to be a bit smarter in its use of constants: I know I've cut some corners there. Didn't really have time to fully explore it. Although at this point, the idea of implementing your approach in Rust is sounding most interesting to me :)

I encourage you to do so, but warn you to be disappointed with the result unless you go further than my first version of "unrolling" in that it will be slower than even your "striped-blocks" version on the primitive CPU's for the discussed cache reasons.  In order to be competitive with "hybrid", we need to go a step further in handling the smaller base prime values in a similar way to "hybrid".

My approach uses templates/macros to generate the constants, but for this benchmark that doesn't really matter all that much because they can be generated by any means and remain the same for the entire culling pass; with page segmentation that would be different and the generation of the variables for the registers to use would take an appreciable time.  But I think you should implement my approach, too, and the following will make it clear why:

First, a data point:  in Nim I have the option of using clang/LLVM as the back end, and when doing so it makes "unpeeling" faster as it generates the critical loop in the same form as the assembly code listed above, but when run on my "unrolled" technique it is slower.  Fortunately, I previously ran into this problem in Haskell (which also has an optional LLVM back end, which is generally the fastest option), so know that there is a bug in Loop Strength Reduction that actually reverses my "unrolled" optimization and found the means to turn it off.  The command line options to run to turn Loop Strength Reduction off on clang are "-mllvm --disable-lsr", which in Nim I can pass to the compiler stage, which made the code run at the same speed as the gcc generated code (although the gcc generated code isn't quite as good as the above assembly for "unpeeled" and I don't know how to fix that).  So if you implement my approach in Rust, you will likely need to find a way to pass the disabling of the LSR optimization to the LLVM back end, if you can.

**This is the promised reveal on why your "hybrid" technique here is faster than my "unrolled" approach and why my unrolled approach appears to get less performance comparatively with the lesser CPU's:**

The difference between "unpeeled" and "unrolled" as far as "cache thrashing" goes is that "unpeeled" (or "striped") jumps through byte indexes by prime values for eight separate loops whereas "unrolled" culls/marks all the composite bits in a single pass with a single loop.  The observation is that the difference isn't so much and in "unrolled"'s favour for modern CPU's with better "cache thrashing"/more execution units but "unrolled" loses out to "unpeeled"/"striped" for lesser CPU's.  Why is that?  I think it has to do with the span of the jumps:  "unpeeled"/"striped" jumps by spans eight times as large and for the smaller primes will quickly get to the point that they are covering more than one cache line per span, after which pass the whole cache will have been loaded into the L1 data cache meaning the the seven subsequent loops will have normal memory access speed of about one CPU clock cycle on average.  This is true for base prime values up to the cache line size which includes base prime values up to 31 even for a cache line size of 32 bytes so covers about 66 percent of the actual culling work.  For the current version of "unrolled", memory access is eight times as dense meaning the the execution units will be waiting on the same cache line for all eight loop memory accesses for all of the base prime values up to the same limit, so the wait times will be much longer.  Added to this for primitive CPU's like the Celeron is that there are less integer execution units to do the waiting and the result is that "cache thrashing" time is about three times as long on a percentage base for the current "unrolling" on the Celeron than on a more modern CPU on a percentage basis.

So why does the "hybrid reset" technique fix this, for "unpeeled"/"striped" and will fix it for my new "unrolled" technique?  It is a special case for the small base prime values (since we aren't allowed to eliminate "cache thrashing" by page-segmentation or eliminate the small base prime values through pre-initialization) so that they are converted into a single load followed by multiple `or` operations (from two to twenty-one, averaging about ten) taking only about a half or a third of a clock cycle each followed by a single store meaning that we will get an average of about seven clock cycles per ten markings/cullings and even more importantly, even if there are only two integer execution units, a maximum throughput of twice this and all of the benefits of sparser cache line accesses separated by more CPU clock cycles.

I suspect that with the extra freedom I gain from using a macro to generate these small prime value special cases, I will likely get back to even faster than this "striped hybrid" version posted here, and for all CPU's.  To make it simpler, since @rbergen says its acceptable to tailor the solution to this benchmark and not cover the cases that are never used, I may reduce the cases to keep the macro code generation phase a little quicker, else it needs to cover 64 start indices times 11 base prime values is 704 cases instead of just 11 (const start indexes for all of the base prime values); thus this specialization becomes very similar to the code you use here, but just generated differently (which may be clearer to read and maintain if one can understand the macro code generation).

Of course, this technique is primarily of use in this benchmark, as a real application would use page-segmentation so that "cache thrashing" isn't much of a problem and even if it were, typically all the base prime values up to and including 19 are pre-culled by array initialization from a pattern array, leaving only 23, 29, and 31 and meaning that the complexity of this technique is probably not worth it.

mike-barber:
Thanks @GordonBGood -- this is a really good analysis, and I think you're pretty spot on. 

On Rust, 

> I encourage you to do so, but warn you to be disappointed with the result unless you go further than my first version of "unrolling" in that it will be slower than even your "striped-blocks" version on the primitive CPU's for the discussed cache reasons. In order to be competitive with "hybrid", we need to go a step further in handling the smaller base prime values in a similar way to "hybrid".

I think so too -- I'm not expecting the first attempt to be particularly fast, but hopefully it'll provide a good basis for some experimentation. I usually need to have some prototype code to tinker with before I can come up with more ideas :)

> Fortunately, I previously ran into this problem in Haskell (which also has an optional LLVM back end, which is generally the fastest option), so know that there is a bug in Loop Strength Reduction that actually reverses my "unrolled" optimization and found the means to turn it off.

This is an interesting observation, and it's something I think I've fighting with too in some of the other approaches in my solution. I know there is an option in Rust somewhere to pass flags directly to LLVM (and I have seen loop strength reduction mentioned before). Being a global switch, though, it might mess up some of the existing solutions. But it's definitely worth a try. It might also be possible to bypass it by formulating the iterations slightly differently -- hack around the detection logic somehow. I'll look for the option and see what effect it has at some point.

> Why is that? I think it has to do with the span of the jumps: "unpeeled"/"striped" jumps by spans eight times as large and for the smaller primes will quickly get to the point that they are covering more than one cache line per span, after which pass the whole cache will have been loaded into the L1 data cache meaning the the seven subsequent loops will have normal memory access speed of about one CPU clock cycle on average.

I think this is spot on. Another possible contributing factor with the `striped` implementations is this: the `skip` (from an address perspective) is always the same for a whole stripe. In my thinking (which might be wildly off), this should enable a simplistic cache controller to predict which lines to read next more effectively. The jitter resulting from the the modulo in the un-striped/linear storage might be just enough to confuse it into pre-fetching the wrong cache lines sometimes. 

Presumably it'll get them from L2, but there might be other weird differences between processors: like whether L1 gets evicted directly to main memory, or just to L2. And whether L2 is inclusive of L1, etc. 

> So why does the "hybrid reset" technique fix this, for "unpeeled"/"striped" and will fix it for my new "unrolled" technique? It is a special case for the small base prime values (since we aren't allowed to eliminate "cache thrashing" by page-segmentation or eliminate the small base prime values through pre-initialization) so that they are converted into a single load followed by multiple or operations (from two to twenty-one, averaging about ten) taking only about a half or a third of a clock cycle each followed by a single store meaning that we will get an average of about seven clock cycles per ten markings/cullings and even more importantly, even if there are only two integer execution units, a maximum throughput of twice this and all of the benefits of sparser cache line accesses separated by more CPU clock cycles.

This is roughly the motivation for doing it. Also part of it is that LLVM can do further optimisations too, because in the `striped-blocks` case, the block size is known. The initial prototyping I did separately indicated that it would be able to optimise operations on 32b or 64b words, and do multiple of these on separate registers to break the data dependence. I'm not sure if this has made it into the actual version, but it's quite possible we're tackling multiple adjacent bytes on each clock cycle.

> I suspect that with the extra freedom I gain from using a macro to generate these small prime value special cases, I will likely get back to even faster than this "striped hybrid" version posted here, and for all CPU's. To make it simpler, since @rbergen says its acceptable to tailor the solution to this benchmark and not cover the cases that are never used, I may reduce the cases to keep the macro code generation phase a little quicker, else it needs to cover 64 start indices times 11 base prime values is 704 cases instead of just 11 (const start indexes for all of the base prime values); thus this specialization becomes very similar to the code you use here, but just generated differently (which may be clearer to read and maintain if one can understand the macro code generation).

I think so -- it'll be very interesting to see how it turns out. 

The approach I took with the permutations that we need is to specialise is relatively simple. We know what the `start` is from the `skip`. And I think it's fair enough to just generate code for the all the odd skip factors below a certain limit, even if there are a few unused paths there. We could do it specifically for all the ones we know are prime, but it's maybe getting a little close to encoding knowledge of primes in the system. 

As for me, I'll try to put together a minimal Rust implementation on a branch on my repo, and let you know where it is when it's ready for feedback. I'm feeling like procedural macros will be a fun thing to try -- I haven't used them in a while, so it'll be a good refresher. I've never used them for anything useful before. I'll have a look at your Chapel solution again for guidance - it seems a little easier to read than Nim for now.

GordonBGood:
@mike-barber:

As to LLVM optimization settings being global, Docker files can include multiple "cargo" builds and result in multiple entry points to be run in sequence, so it seems to me that you could apply the optimization overrides for a cargo such as "Extreme Loop Unrolling" if it was necessary and not for a separate cargo including "striping" where it may require that optimization.

> The initial prototyping I did separately indicated that it would be able to optimise operations on 32b or 64b words, and do multiple of these on separate registers to break the data dependence. I'm not sure if this has made it into the actual version, but it's quite possible we're tackling multiple adjacent bytes on each clock cycle.

I don't know that LLVM will optimize to that level; I think your gains with "hybrid reset" are mostly due to the better cache associativity and the reduced time for many of the marking operations as they are register to register (or immediate to register in my case).  For modern high-end CPU's the internal execution of the microops may actually be doing some of what you say in simultaneous  execution streaming to the various available execution units, but we need to trust the compiler to transform the code into the best form to suit that (which is why we don't do manual assembly coding anymore, even for tight loops).

I am using manually generated 64-bit words and get a remarkable improvement in performance - about 15 percent (or much more on less CPU's) just for doing the dense thing for base prime of three and I estimate another 15 percent or more may be added to performance using the other small base primes up to 19 or so.  It seems to me that it becomes pretty much a wash after about the prime of 19.

> The approach I took with the permutations that we need is to specialise is relatively simple. We know what the `start` is from the `skip`. And I think it's fair enough to just generate code for the all the odd skip factors below a certain limit, even if there are a few unused paths there. We could do it specifically for all the ones we know are prime, but it's maybe getting a little close to encoding knowledge of primes in the system.

There is that - that the "race committee" may consider any prior knowledge of the small base primes not to be allowed.  It doesn't really matter to me as once I generate the code with a macro, it doesn't really matter that there are branches that are never used.

> As for me, I'll try to put together a minimal Rust implementation on a branch on my repo, and let you know where it is when it's ready for feedback. I'm feeling like procedural macros will be a fun thing to try -- I haven't used them in a while, so it'll be a good refresher. I've never used them for anything useful before. I'll have a look at your Chapel solution again for guidance - it seems a little easier to read than Nim for now.

Yes, I think you may need the procedural macros to do some code generation if you want to get into that.  Up to now, just templates/`macro_rules` have been adequate if that is all that's easily available, but they won't generate variable length code based on conditions as will be needed here (or at least I don't think so).  The Chapel solution is easy to read, but it doesn't use a macro, just a template.

In my Nim solution I use a full AST macro to generate the code, probably something like a procedural macro in Rust, so of course it is a little harder to understand.  There are some lines at the end that can be uncommented so as to view the generated code in whole of in part, as one can see [on this Wandbox link](https://wandbox.org/permlink/P8qDMd3mIaSwthXB); this link contains what I am working on right now and by inserting a space as per `# [` at line 130 you can enable "dense hybrid" for the only manually written base prime value of three and see the remarkable improvement when re-running.  The Wandbox server processor is a Sandy Bridge Xeon running single threaded at about 2.9 GHz so it quite a bit better than Dave Plummer's J1800, yet it has a huge "cache thrashing" slowdown of almost a factor of two compared to using page-segmentation (page-segmented implementation not shown) which is helped immensely by even this just handling three.  Sandy Bridge is also about 25 percent slower in culls per CPU clock for this implementation because it doesn't optimize branches as well, always taking two cycles to do the compare-and-branch even correctly predicted.  Newer CPU's such as Sky Lake and newer elide this two clock cycles to almost zero when correctly predicted.

In order to do this dense thing in Chapel without using the macros it doesn't have, one would either need to use a code generation pass external to actual code and include the results of the code generation or just hand code the nine cases for odd numbers from three to nineteen (with the cases for nine and fifteen unused, but present to keep the "race commitee" happy).  It would be quite easy for me to output the results of the my Nim macro once I finish it and just copy and past it into the Chapel code, then do the minor refactoring for syntax differences.  One could do the same thing for C/C++ and any language that doesn't have true macros (or even templates for that matter, as there are only four odd step modulo eight values in the "unrolling" for larger base primes).

Since this "Open Software Drag Race" isn't like a real car drag race where there may be secrets in the implementation of the individual cars, in this benchmark all of the details are available between the competitors.  Thus, to me it would seem to make sense that a new challenge is opened using code such as a new Chapel implementation that includes this "dense hybrid" technique" since, being OOP and of very conventional syntax for the most part, most programmers will find it as easy to read and understand as yourself;  The challenge would be to implement the same technique in their language as efficiently, effectively, and elegantly as they know how.  Then, when Dave Plummer eventually does the YouTube video comparing the implementations, he can look at the comparative speed on the various test machines, but also at the code to show the effectiveness and elegance (with the last two based on how concise and easy to read the code is as well as if it has templates and/or macros).

I think that might be fun, and will post to @rbergen about it.  Then the benchmark actually would show something in implementing the same "ultimate" algorithm for all languages capable of doing that.

mike-barber:
> I don't know that LLVM will optimize to that level; I think your gains with "hybrid reset" are mostly due to the better cache associativity and the reduced time for many of the marking operations as they are register to register (or immediate to register in my case). For modern high-end CPU's the internal execution of the microops may actually be doing some of what you say in simultaneous execution streaming to the various available execution units, but we need to trust the compiler to transform the code into the best form to suit that (which is why we don't do manual assembly coding anymore, even for tight loops).

I still haven't looked into what the actual assembly looks like, but generally that is the motivation for the approach: do several mask operations on a given register, then write it back. I'll do some digging at some point -- it'll be interesting to see what's going on at assembly level. There's bound to be some unrolling: we've got quite a lot of registers, and they're quite wide.

> I am using manually generated 64-bit words and get a remarkable improvement in performance - about 15 percent (or much more on less CPU's) just for doing the dense thing for base prime of three and I estimate another 15 percent or more may be added to performance using the other small base primes up to 19 or so. It seems to me that it becomes pretty much a wash after about the prime of 19.

Fantastic! This is similar to the approach I'd like to take with Rust. I'll experiment with both bytes and larger words. 

I have made some progress, but pretty light on the code at this point. It took a little time to figure out what I was doing with procedural macros again, and realised it's quite tricky to get literal parameters into them. It's always the little things, but that should be resolved now.

Another thing I realised is that I didn't actually know _why_ you had 64 mask sets (or functions) that you could recycle. I understood the repetition in the masks for a given skip/start factor, but I hadn't clicked why the _sets_ of masks repeated. I'm probably just being slow, but it took a lot of head-scratching and visualising things on paper and in Excel. But I do get it, finally. 

The approach I'm going to attempt in Rust is: 
- create functions to generate the mask sets
- use these in the procedural macro to generate the `dense` reset functions, odd numbers 3..N; will experiment a bit to see where the efficiency drops off. 
- for factors above N, call into the `general` reset function where the word indices are calculated on the fly, but still using the calculated mask sets
- perhaps replace the `general` reset function with generated specific functions, one for each mask set, so it's using literals instead; but it might not be necessary.

GordonBGood:
@mike-barber:

> > I am using manually generated 64-bit words and get a remarkable improvement in performance - about 15 percent (or much more on less CPU's) just for doing the dense thing for base prime of three and I estimate another 15 percent or more may be added to performance using the other small base primes up to 19 or so. It seems to me that it becomes pretty much a wash after about the prime of 19.
> 
> Fantastic! This is similar to the approach I'd like to take with Rust. I'll experiment with both bytes and larger words.
> 
> I have made some progress, but pretty light on the code at this point. It took a little time to figure out what I was doing with procedural macros again, and realised it's quite tricky to get literal parameters into them. It's always the little things, but that should be resolved now.
> 
> Another thing I realised is that I didn't actually know _why_ you had 64 mask sets (or functions) that you could recycle. I understood the repetition in the masks for a given skip/start factor, but I hadn't clicked why the _sets_ of masks repeated. I'm probably just being slow, but it took a lot of head-scratching and visualising things on paper and in Excel. But I do get it, finally.

Actually, we don't need 64 mask sets for this application as there are only four odd primes values modulo eight and each of those base prime modulo eight values has a fixed starting index since we aren't allowed to do page-segmentation and always cull from the start of the sieve buffer array to the end; thus only four entries are actually used and necessary.  I provide the full Monty of mask sets as this technique can be used generally and to keep the "race committee" happy that we are not using any prior knowledge of primes other than the only even prime is two; also, it doesn't cost me anything as the mask sets are generated by just invoking the "template" 64 times.  However, rbergen has since told me that reducing the cases here to just the used four would be allowed, as it still fits into the general processing without requiring knowledge of what the actual base prime values are, which would make it much easier to implement this in languages that don't even have templates.

> The approach I'm going to attempt in Rust is:
> 
> * create functions to generate the mask sets
> * use these in the procedural macro to generate the `dense` reset functions, odd numbers 3..N; will experiment a bit to see where the efficiency drops off.
> * for factors above N, call into the `general` reset function where the word indices are calculated on the fly, but still using the calculated mask sets
> * perhaps replace the `general` reset function with generated specific functions, one for each mask set, so it's using literals instead; but it might not be necessary.

I guess I'm away ahead of you here:

I have chosen to implement this "dense base prime hybrid" technique as an upgrade to my Chapel code as PR #671, first because other programmers such as yourself likely find it easier to read and understand Chapel code, and also because I would like to issue this code as a challenge to others such as yourself to implement in their chosen language(s).

The performance increase of this above just using straight "extreme loop unrolling" is phenomenal!  It is about half again faster on modern CPU's and likely much more than than on more primitive CPU's due to the improvements in the use of the cache.

Since Chapel does not have code generating macros (only something that can be used as templates), I had to use external code generation for the dense culling code patterns so the code is a little long (but repetitive); However, that may make it easier for others to understand what is done.  In languages such as Julia, Rust, Nim, Haskell, etc. that have code generating macros, the code cases can be generated by invoking a single macro just as I do for "extreme loop unrolling" in Nim in only a few tens of lines of code.

I think this should pass the `faithful to base` algorithm considerations as I have been careful to use at least one operation per composite number bit representation culled (no combining into a single masking operation which wouldn't be allowed) for the added dense base primes technique, and have used no prior knowledge of what any of the odd base prime values are.

This may help your thinking of how to proceed.  Enjoy.

mike-barber:
> I have chosen to implement this "dense base prime hybrid" technique as an upgrade to my Chapel code as PR #671, first because other programmers such as yourself likely find it easier to read and understand Chapel code, and also because I would like to issue this code as a challenge to others such as yourself to implement in their chosen language(s).

Nice! I will check it out a little later. I'm half way through hacking a minimal version together in Rust. When that's working, I'll compare and contrast :) 

> The performance increase of this above just using straight "extreme loop unrolling" is phenomenal! It is about half again faster on modern CPU's and likely much more than than on more primitive CPU's due to the improvements in the use of the cache.

That's promising -- that's more than I expected. And it sounds like it's going to be the fastest algorithm by some margin!

> This may help your thinking of how to proceed. Enjoy.

Thanks! This is really quite fun!

mike-barber:
@GordonBGood your Chapel implementation works really well! I've done the first cut on the Rust implementation, but there are a few issues I still need to resolve. It's looking reasonably positive so far: https://github.com/mike-barber/daves-garage-primes/pull/9

GordonBGood:
@mike-barber:

> > I have chosen to implement this "dense base prime hybrid" technique as an upgrade to my Chapel code as PR #671, first because other programmers such as yourself likely find it easier to read and understand Chapel code, and also because I would like to issue this code as a challenge to others such as yourself to implement in their chosen language(s).
> 
> Nice! I will check it out a little later. I'm half way through hacking a minimal version together in Rust. When that's working, I'll compare and contrast :)
> 
> > The performance increase of this above just using straight "extreme loop unrolling" is phenomenal! It is about half again faster on modern CPU's and likely much more than than on more primitive CPU's due to the improvements in the use of the cache.
> 
> That's promising -- that's more than I expected. And it sounds like it's going to be the fastest algorithm by some margin!

Maybe the fastest on some CPU's; it does more or less the expected on both new CPU's and on ARM64, and it will gain on the Celeron, but there is some question about whether it is better to complete the non-dense culling by using extreme loop unrolling or "striping"/"unpeeling" by blocks - for this type of primitive CPU only.  The PR has been merged so we'll see in a day or two.  Also, my daughter is with me today doing on-line schooling, so I'll run the Chapel version on Cherry Trail to see how it compares.

This Chapel version may be the fastest for now, but I suspect that the equivalent Nim version will be even faster, both because the back end compiler(s) optimize just a little bit better and because by using macros I won't need to limit the dense algorithm to only base primes up to 19 and will likely go to 31.  The Nim version should be out in a day or two.

> > This may help your thinking of how to proceed. Enjoy.
> 
> Thanks! This is really quite fun!

Isn't it just?

Kulasko:
I followed the thread for a bit, you both really go up and beyond with your optimization efforts :)
Just asking, since you're looking at generated assembly and stuff: Are you using the stable or nightly compiler? Because at least on my machine, the nightly compiler provides a bit better runtime performance, so it might be a good idea to use that as target to optimize.

For this reason, I would generally like to use the nightly compiler for the competition, but afaik there is no docker image containing it, so I'm still on the fence (and didn't do it on my solution yet). Reliability wise, it really is stable enough.

Also a note on the Atom CPUs, Silvermont luckily isn't the oldest Atom, because that was Saltwell and boy was it slow, it could barely run Windows 7 back in the day. From Silvermont to Airmont (I think) it's basically the same core with light optimizations, Tremont is the first core with interesting performance improvements, its successor is said (by Intel) to already approach Skylake.

Also, I have on of those "lesser" CPUs (an AMD FX8350 and currently an AMD Athlon II X4 860k), the hybrid version went from ~8000 passes to ~8200 passes on the FX, to provide a data point. I'm interested to see how far you guys will be pushing it until the competition is over. :D

GordonBGood:
@Kulasko:

> Just asking, since you're looking at generated assembly and stuff: Are you using the stable or nightly compiler? Because at least on my machine, the nightly compiler provides a bit better runtime performance, so it might be a good idea to use that as target to optimize.

I, at least, am not using nightly/devel versions of any language to which I have contributed.  As you say, it won't be easy to find Docker images for the latest non-stable versions from many languages such as Rust, and also, being "non-stable", the performance could change one way or the other when the backing Docker image was updated, even if one were available as they are for some languages.

> Also a note on the Atom CPUs, Silvermont luckily isn't the oldest Atom, because that was Saltwell and boy was it slow, it could barely run Windows 7 back in the day. From Silvermont to Airmont (I think) it's basically the same core with light optimizations, Tremont is the first core with interesting performance improvements, its successor is said (by Intel) to already approach Skylake.

The discussion related to Atom processors was really just about how to confirm the lack of performance we are getting with Dave Plummer's Celeron J1800 and how we could confirm it with CPU's we have at hand.  There is less and less of a reason to use Atom CPU's at all, so unless that market segment recovers and there gets to be the newer "Atom" CPU's commonly available in laptops/tablets, I won't be upgrading my Silvermond/Airmont machine to whatever Intel is replacing that range with.

> Also, I have on of those "lesser" CPUs (an AMD FX8350 and currently an AMD Athlon II X4 860k), the hybrid version went from ~8000 passes to ~8200 passes on the FX, to provide a data point.

I've also have an even older original "Bulldozer" in the AMD FX8120 which still runs occasionally and also have one of the original AMD Athlon II X4's which I think would still run if I bothered.  The latter was a very good CPU for its day and for many jobs probably was at least the equal of the "Bulldozer" even though its clock rate is lower (2.4 or 2.5 GHz IIRC); the "Bulldozer" was a dog and I bought it knowing this but wanting to experiment to understand why this was do - it has a different cache problem than even the lowest of the Atom processors in that it has a bottleneck of only a few Kilobytes in its Cache Coalescing Buffer so that for sieving such as this that need to use cache in blocks larger than that CCB, the performance is throttled in passing through this buffer.  This design almost killed AMD, but fortunately Zen has saved the day for them.

>  I'm interested to see how far you guys will be pushing it until the competition is over. :D

There will be incremental increases, especially using some other languages, but we are starting to run into "physics" limits with the latest dense resetting ideas.  Theoretical maximum culling rate is about two to three culls per clock cycle, but that would only be possible if all culling was "dense" and there were no cache delays at all; on a modern CPU, **EDITED** my Chapel version is at about ~~1.25~~ 0.75 clock cycles per cull but almost a third of that is cache delays and about a third of the culling operations are not dense, so there isn't very far to go.  For lesser CPU's, a higher percentage of the time is lost due to cache delays, and as well, the minimum number of clock cycles per instruction is larger due to the simpler microops processing used, so there will tend to eventually be a fixed ratio between these and the top CPU's performance.

I have a new Nim version I'm about to PR which runs at an average of about half a CPU clock cycle per operation including time lost for "cache thrashing" and I think that's about the limit; that's about 42K passes per five seconds on my 3.6 GHz processor.  I don't know how it will do on the Celeron J1800 or the Raspberry Pi 4, but I'm guessing about 13 thousand and eight thousand passes per five seconds, respectively.

Mostly what we are experimenting with now is ways to reduce cache bottle necks, especially with the Celeron J1800, but with this there may be no better technique than just Mike's original "striped by blocks" one, which is pretty much the same as "unpeeled by blocks" or "strided by blocks" as just alternate names for about the same culling order.  So we are just about done...

PR #659, opened 2021-08-18
Refactor the 4th C# solution to reduce duplicated code

teo-tsirpanis:
## Description

This PR optimizes the fourth C# solution to recycle the same bit buffer and convert the sieves from classes to structs.

As a result, allocations in the main loop were eliminated, with the drawback that __the solution is no longer faithful__.

Additionally, the runner was refactored to reuse code, time is measured using .NET's `Stopwatch` class, and the `Dockerfile` references the .NET images for Alpine.

Tagging @ItalyToast to provide feedback/objections.

## Contributing requirements

### DIFF
--- a/PrimeCSharp/solution_4/PrimeCS.cs
+++ b/PrimeCSharp/solution_4/PrimeCS.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
@@ -12,7 +12,12 @@ namespace PrimeSieveCS
         IEnumerable<uint> EnumeratePrimes();
     }
 
-    class PrimeCS
+    interface ISieveRunner
+    {
+        ISieve RunSieve(uint sieveSize);
+    }
+
+    static class PrimeCS
     {
         static Dictionary<uint, int> validationDict = new Dictionary<uint, int>
         {
@@ -28,70 +33,38 @@ namespace PrimeSieveCS
 
         static bool ContainsValidResults(ISieve sieve)
         {
-            return validationDict.ContainsKey(sieve.SieveSize) && validationDict[sieve.SieveSize] == sieve.EnumeratePrimes().Count();
+            return validationDict.ContainsKey(sieve.SieveSize) &&
+                   validationDict[sieve.SieveSize] == sieve.EnumeratePrimes().Count();
         }
 
-        static void Main(string[] args)
+        static void Main()
         {
             CultureInfo.CurrentCulture = new CultureInfo("en-US", false);
 
-            RunSieveDefault();
-            RunSieveStriped();
+            const int sieveSize = 1_000_000;
+
+            RunSieve(default(SieveDenseAndSparseRunner), sieveSize);
+            RunSieve(default(SieveStripedRunner), sieveSize);
         }
 
-        static void RunSieveDefault()
+        static void RunSieve<TRunner>(TRunner runner, uint sieveSize) where TRunner : ISieveRunner
         {
-            //setup
-            const int sieveSize = 1_000_000;
-            
             //warmup - 5 seconds permitted
-            var wStart = DateTime.UtcNow;
-            while ((DateTime.UtcNow - wStart).TotalSeconds < 5)
-                new SieveDenseAndSparse(sieveSize).RunSieve();
+            while (sw.Elapsed.TotalSeconds < 5)
+                runner.RunSieve(sieveSize);
 
             //Forcing a GC to give us as much space on the heap as possible (dont do this in real code).
+            //It's not part of the measurement either way.
             GC.Collect(0, GCCollectionMode.Forced, true, true);
 
             //running the dragrace
             var tStart = DateTime.UtcNow;
             var passes = 0;
-            SieveDenseAndSparse sieve = null;
+            ISieve sieve = default;
 
             while ((DateTime.UtcNow - tStart).TotalSeconds < 5)
             {
-                sieve = new SieveDenseAndSparse(sieveSize);
-                sieve.RunSieve();
-                passes++;
-            }
-
-            var tD = DateTime.UtcNow - tStart;
-
-            if (sieve != null)
-                PrintResults(sieve, false, tD.TotalSeconds, passes);
-        }
-
-        static void RunSieveStriped()
-        {
-            //setup
-            const int sieveSize = 1_000_000;
-
-            //warmup - 5 seconds permitted
-            var wStart = DateTime.UtcNow;
-            while ((DateTime.UtcNow - wStart).TotalSeconds < 5)
-                new SieveStriped(sieveSize).RunSieve();
-
-            //Forcing a GC to give us as much space on the heap as possible (dont do this in real code).
-            GC.Collect(0, GCCollectionMode.Forced, true, true);
-
-            //running the dragrace
-            var tStart = DateTime.UtcNow;
-            var passes = 0;
-            SieveStriped sieve = null;
-
-            while ((DateTime.UtcNow - tStart).TotalSeconds < 5)
-            {
-                sieve = new SieveStriped(sieveSize);
-                sieve.RunSieve();
+                sieve = runner.RunSieve(sieveSize);
                 passes++;
             }
 
--- a/PrimeCSharp/solution_4/PrimeCS.cs
+++ b/PrimeCSharp/solution_4/PrimeCS.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
 
@@ -49,6 +50,8 @@ namespace PrimeSieveCS
 
         static void RunSieve<TRunner>(TRunner runner, uint sieveSize) where TRunner : ISieveRunner
         {
+            var sw = Stopwatch.StartNew();
+
             //warmup - 5 seconds permitted
             while (sw.Elapsed.TotalSeconds < 5)
                 runner.RunSieve(sieveSize);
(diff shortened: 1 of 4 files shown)


rbergen:
This PR changes the key characteristics of a faithful implementation in a way that it no longer is. That's not the type of "improvement" we merge as a rule. 

@ItalyToast: this being a PR on a solution you submitted, I am wondering what's your opinion?

teo-tsirpanis:
Hmm, I understand your point @rbergen, but I wouldn't say that "key characteristics" of the solution changed. Both the algorithm itself and its implementation remain identical. This PR indeed deprives the solution of its faithfulness but changes secondary technical aspects of it. For example I wouldn't see value in forking the PR's changes into a fifth C# solution.

rbergen:
Faithfulness is one of the key characteristics of an implementation in the context of this repo. The others are using one thread vs using multiple, using the base algorithm vs not doing so, and the number of bits per prime candidate. Changing a faithful implementation to an unfaithful one disqualifies it from inclusion on the leaderboard, and most authors of leaderboard-qualifying implementations made a concerted effort to make them that.

> For example I wouldn't see value in forking the PR's changes into a fifth C# solution.

I agree. With 4 C# solutions in place, and one of them exploring a wide range of approaches, we may have reached a point where there need to be clear reasons to add another one.

ItalyToast:
First of all thanks for showing my solution some love BUT breaking faithfulness is a big nono for me.

As I state in the readme for the solution the goal is to remain faithful, use the base algorithm and not recycle memory buffers. As long as those criteria are met I'm all ears. Why not allow recycling of buffers? In my opinion its kind of cheating, the task of generating primes is only useful once so it doesnt make sense to reuse a buffer used to calculate primes. Also the GC is very fast on Gen1 so it shouldnt hinder us too much anyway. The allocations should just be bumping the heap pointer and GC should be very fast aswell since we dont have any promotions to Gen2.

As for some of the other changes I'm much more interested, I was thinking of doing something like a sieverunner when I created the striped version but didnt bother since I was unsure of the performance characteristics and I only have 2 solutions so the duplicated code is still very readable and easily modifiable. I dont remember where I heard it but there is a rule of thumb that you should refactor when you have duplicated code 3 times because then you will find the edge cases. 

The preferred way of doing a sieve runner would be if we could just have a generic method that takes an ISieve
`public static void RunSieve<TSieve>() where TSieve : ISieve, new(int sievesize){ }`
unfortunately `new(int sievesize)` is not currently possible and we dont have static methods in interfaces yet. (Coming soon🥳) That means we need to keep the SieveRunners if we dont want to pass any function pointers to create the new sieves but with the requirement of not recycling buffers they should be very simple
```
interface ISieveRunner<TSieve>
{
    TSieve RunSieve(int sieveSize);
}

readonly struct SieveDenseAndSparseRunner : ISieveRunner<SieveDenseAndSparse>
{
    public SieveDenseAndSparse RunSieve(uint size)
    {
        var sieve = new SieveDenseAndSparse(size);
        sieve.RunSieve();
        return sieve;
    }
}
```

I think the structs is nice. The size is small and is reducing unnessecary GC pressure.

I would definitely prefer if we could keep stylistic changes to a minimum. By which I mean it doesnt change the behaviour in any meaningful way. For example:

- removing args in main: `static void Main(string[] args)`
- making PrintResults a generic method

The change from focal to alpine is welcome. That should help reduce image size for the repo. I copied my dockerfile from solution_2 so this problem also applies there @davepl.

ItalyToast:
There have been some discussions about using `ArrayPool` to recycle buffers. The argument is that it's a part of the language and should still be considered faithful. I dont agree with that sentiment.

teo-tsirpanis:
Hello @ItalyToast and thanks for your quick feedback. I understand the value you give to the faithfulness of your solution, and since the garbage collector's impact is tiny (only **0.21%** of the execution time was spent in the GC), I will revert the buffer reusing logic from my PR.

Making the sieves structs is still a violation of faithfulness. Quoting from the contributor's guide:

> [The implementation is faithful if it] uses a class to encapsulate the sieve, or a (closest) equivalent feature in your language if a class construct is not available.

Since classes are available in C# and we use a struct which is not a class, we still violate faithfulness, which means that I will revert it as well.

And now the PR is reduced to the runner's refactoring. Despite your subsequent removal of the objection about the `Stopwatch` class, I will still revert it as it did not actually make a difference, and to make as little changes as necessary.

I also considered `stackalloc`ing the buffer but it would require non-trivial restructuring of the solution, and the sieves would have to be represented as a `ref struct` which as I said earlier is not a class and breaks faithfulness.

---

I changed the sieve runner interface into this:

``` csharp
interface ISieveRunner
{
    ISieve RunSieve();
}
```

It does not need to return a generic sieve, opting for the `ISieve` interface instead, and the `RunSieve` method does not accept any arguments; I preferred it so to abstract away any parameterization from the runner's consumer.

Let me know if everything is OK.

rbergen:
> First of all thanks for showing my solution some love BUT breaking faithfulness is a big nono for me.

@ItalyToast I'll interpret this as a statement from you that you do not want me to merge this PR in its current form.

> I think the structs is nice. The size is small and is reducing unnessecary GC pressure.

The rules have stipulated for a while now that in languages that have a class construct, that is the one that needs to be used for an implementation to be considered faithful. So in light of your earlier statement, structs are also a nono in the C# context.

> The change from focal to alpine is welcome. 

I'd like to note here that multiple contributors have found that other languages perform worse in Alpine than Ubuntu (or other distros). I don't know if that's the case for C#, but it is something that may warrant some testing.

>  I don't agree with that sentiment.

Duly noted, but we're not going to change the decision on this.

teo-tsirpanis:
> I'd like to note here that multiple contributors have found that other languages perform worse in Alpine than Ubuntu (or other distros).

I happened to notice that the contributor's guide prefers Alpine images if they exist, which is the reason I changed it. I think that these performance differences among the Ubuntu and Alpine images should be discussed in a separate issue about changing the policy.

ItalyToast:
@rbergen Correct, faithfulness is of top priority.

> I'd like to note here that multiple contributors have found that other languages perform worse in Alpine than Ubuntu (or other distros). I don't know if that's the case for C#, but it is something that may warrant some testing.

Ohh that is interesting, @teo-tsirpanis did you notice any performance difference between alpine, ubuntu and focal? If there is no major performance degradation I see no reason not to change to either alpine or ubuntu. 

> Duly noted, but we're not going to change the decision on this.

Understandable, it was a constraint I wanted for my solution. I know languages like C++ got much slower heap allocations unless they use an arena allocator of some kind, so it wouldnt really be fair either way.

@teo-tsirpanis 
> Despite your subsequent removal of the objection about the Stopwatch class, I will still revert it as it did not actually make a difference, and to make as little changes as necessary.

First I thought that it wasnt a worthwhile change since we are basically doing the same thing(measuring time). I changed my mind after a quick google search of datetime precision. Datetime.Now precision is in the range of 10ms-50ms which is quite a lot. If you want to add the stopwatch back in I'm fine with that (didnt expect you to see the comment so fast 😅)

Could you remove the `CreateSieve()` in the SieveRunner implementations?

`static void RunSieve<TRunner>(TRunner runner) where TRunner : struct, ISieveRunner` The struct requirement shouldn't be needed I think.

> the RunSieve method does not accept any arguments; I preferred it so to abstract away any parameterization from the runner's consumer.

I prefer the way of passing the parameter. It simplifies the code of the runner. The interface "contract" is much more explicit and if we need new parameters we'll fix that when we get to it.

teo-tsirpanis:
OK @ItalyToast.

ItalyToast:
@teo-tsirpanis Thanks for your contribution, if you want some recognition in the readme, feel free to add a line. I noticed you added one before but its not present now.

PrimeCSharp/solution_4/PrimeCS.cs
```diff
@@ -28,77 +34,45 @@ class PrimeCS
...
 
             //running the dragrace
-            var tStart = DateTime.UtcNow;
             var passes = 0;
-            SieveStriped sieve = null;
-
-            while ((DateTime.UtcNow - tStart).TotalSeconds < 5)
+            ISieve sieve = default;
+            sw.Restart();
+            while (sw.Elapsed.TotalSeconds < 5)
```
rbergen, line 67: The creation of the sieve object needs to take place at every iteration within the timed loop, for the solution to be faithful, as per the following rule in the contributing guidelines: _It uses a class to encapsulate the sieve, or a (closest) equivalent feature in your language if a class construct is not available. This class must contain the full state of the sieve. **Each iteration should re-create a new instance of this class from scratch.**_

teo-tsirpanis: It does actually. This is the implemenation of `SieveDenseAndSparseRunner.RunSieve`:

https://github.com/PlummersSoftwareLLC/Primes/blob/3f56243fe9eca20d99e5de8abdc0a08e17dd1cd9/PrimeCSharp/solution_4/SieveDenseAndSparse.cs#L11-L16

teo-tsirpanis:
Done @ItalyToast, I added a line in the README.

ItalyToast:
LGTM

PrimeCSharp/solution_4/PrimeCS.cs
rbergen: I see what's happening here, now. Check.

rbergen:
To finish this thread: personally, I would have chosen the factory pattern. That is to say, create an `ISieveFactory` interface instead of `ISieveRunner`, with one `CreateSieve(int sieveSize)` method. I'd then create implementations (`SieveDenseAndSparseFactory` and `SieveStripedFactory`) that only invoke the constructor of their sieve counterparts. It would save duplicating the run logic, which is what the runners now do. 

But, with @ItalyToast's approval already given, I will merge it as it is.

rbergen:
Final final comment in response to this:

> I happened to notice that the contributor's guide prefers Alpine images if they exist, which is the reason I changed it. I think that these performance differences among the Ubuntu and Alpine images should be discussed in a separate issue about changing the policy.

The policy is applied in the sense that Alpine is preferred, except if there is a reason to deviate. This being a drag race, we do accept increasing performance as a valid reason for such deviation.

teo-tsirpanis:
I dropped the commit that switches to Alpine.

PR #667, opened 2021-08-20
Run "everything" in docker

AstolfoKawaii:
Adds the ability to run on systems without gcc or nodejs, the only requirements are docker and git.


rbergen:
@AstolfoKawaii In an earlier version of the benchmark tooling, we actually executed the report generator in a Docker container. 
The reason we don't do that anymore is that we collect some system information that is relevant to the benchmark reports, which only works if we run the benchmark tool on the host.

PR #670, opened 2021-08-22
Clipper Implementation

BradleyChatha:
## Description

Draft since, for the moment, we're just trying to get an environment set up before a solution.

Following on from our discussion, I use the following commands within the `PrimeClipper/solution_1` directory:

```
docker build -t clipper .
docker run --rm -t clipper
```

## Contributing requirements

### DIFF
--- a/PrimeClipper/solution_1/Dockerfile
+++ b/PrimeClipper/solution_1/Dockerfile
@@ -10,7 +10,7 @@ RUN apt-get update && \
     apt-get install -y "./${HARBOUR_PKG}" && \
     rm "${HARBOUR_PKG}" 
 
-COPY . /opt
+COPY . /opt/
 
 RUN hbmk2 ./test.prg
 
--- a/PrimeClipper/solution_1/Dockerfile
+++ b/PrimeClipper/solution_1/Dockerfile
@@ -1,0 +1,17 @@
+FROM ubuntu:20.04 AS build
+
+ENV HARBOUR_PKG=r2017-12-15-18_53_harbour_3.2.0-1_amd64.deb
+
+WORKDIR /opt
+
+RUN apt-get update && \
+    apt-get install -y wget build-essential && \
+    wget "https://sourceforge.net/projects/harbour-project/files/binaries-linux-ubuntu/nightly/${HARBOUR_PKG}/download" -O ${HARBOUR_PKG} && \
+    apt-get install -y "./${HARBOUR_PKG}" && \
+    rm "${HARBOUR_PKG}" 
+
+COPY . /opt
+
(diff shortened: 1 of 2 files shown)


rbergen:
@BradleyChatha Try putting a slash (/) at the end of the COPY directive. It's effectively a wildcard copy, and then the target must end with a slash.

BradleyChatha:
The make file is still failing for me.

It's weird since if I call docker normally it'll work, it's just that the make file fails for it.

rbergen:
@BradleyChatha Yes, I was ruling out things. I know Docker is inconsistent across platforms with the COPY-and-slash requirement. I'll have a look at what happens when I try this.

marghidanu:
Ok, the Dockerfile seems to be building just fine in CI

rbergen:
Yes, it builds and runs on my Windows box as well, and the benchmark run also works (`make DIRECTORY=PrimeClipper`). However, the ANSI codes that ncurses generates do cause problems with the report output.

BradleyChatha:
The makefile fails for any project I try, so considering what you two have said, it seems to be an issue with my setup, apologies for the needless hassle.

@rbergen I can't seem to find any way to change things into a "plain" output to stop the ANSI codes. How should I proceed from here?

Also I knew it was a good idea to try and get this sorted *before* letting him get to work on a solution, what a nightmare.

rbergen:
@BradleyChatha passing -gtcgi to hbmk2 seems to fix it for me.

BradleyChatha:
Awesome, I think that's the final hurdle solved now? I keep apologising, but sorry for using up your time for something like this.

rbergen:
I like adding new languages to the project and I have a soft spot for the classics. To some extent, it comes with the territory. 
Also, I've done a bit of coding against ncurses myself, albeit in C++. I've gotten familiar with some of the challenges that come with its use.

rbergen:
@BradleyChatha Using -gtstd instead of -gtcgi also works. The benefit of -gtstd is that it provides a few more output capabilities than -gtcgi, newlines being one of them.

BradleyChatha:
@rbergen Thanks for the note.

I won't be seeing Andy (the other person) until probably the end of the week because of COVID, so don't be alarmed if the PR seems stale, I'll probably do a push just to get the README and stuff sorted out in the mean time though

rbergen:
@BradleyChatha You're welcome.

Now that we have this figured out, you could also decide to close the PR while you work on it, and reopen it when it's ready for review. History on the PR will be preserved but it won't be in our (my) review list, which will save you inquisitive prods from maintainers while you (plural you) work on putting the actual implementation together.

PR #671, opened 2021-08-22
Add dense base primes to Chapel solution_1...

GordonBGood:
## Description

This adds a dense base prime value technique which handles base prime values up to and including nineteen; base prime values above that point use the same extreme loop unrolling technique as before.

It speeds up the extreme loop unrolling in two ways:  it reduces "cache thrashing" due to accessing memory less frequently and by larger spans closer to the cache line size, and it reduces a large proportion of the dense culling/marking operations to register/immediate value operations not involving memory access for a large reduction in the average  number of CPU clock cycles per operation.

I believe that it is `faithful to base` in that, unlike some "word" techniques it does not combine the culling/marking of composite bits within the same word with a single operation using a multi value mask but has a marking operation for each and every bit; it also assumes no prior knowledge of the base prime values other than as allowed that the only even prime is two, and covers the cases of culling by all odd step values such as nine and fifteen, which will never by used as they are not prime.

The code is long because Chapel does not have code generating macros (only something that can be used as a template) so the required dense culling patterns had to be externally code generated.  However, as Chapel is an Object Oriented Programming language, the code should be easy to read by most programmers and easily translated to other languages.

The code is issued as a program that authors of other languages can emulate the techniques used and thereby see how well optimized the code as generated by their language is; it is more interesting than the general "drag race" because authors would implement this very same technique and thus compare the compiler's ability.  Also, authors should try to write as efficiently and elegantly as possible in their chosen language, as the code will be reviewed at end culmination of this race.

Other languages, especially those that have code generating macros, will be able to greatly shorten the code and may be able to extend the threshold of dense base primes up to the value of above 31 for an extra small increase in speed.

## Contributing requirements

### DIFF
--- a/PrimeChapel/solution_1/README.md
+++ b/PrimeChapel/solution_1/README.md
@@ -1,4 +1,4 @@
-# Chapel implementation by GordonBGood
+# Chapel implementations by GordonBGood
 
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
@@ -29,10 +29,11 @@ The Dockerfile uses an official Debian image and then builds and installs Chapel
 
 ## Output
 ```
-GordonBGood_1of2;7798;5.00044;1;algorithm=base,faithful=yes,bits=1
-GordonBGood_unpeeled;8969;5.00027;1;algorithm=base,faithful=yes,bits=1
-GordonBGood_unpeeled_block;10487;5.00019;1;algorithm=base,faithful=yes,bits=1
-GordonBGood_unrolled;16356;5.00022;1;algorithm=base,faithful=yes,bits=1
+GordonBGood_bittwiddle;7776;5.00044;1;algorithm=base,faithful=yes,bits=1
+GordonBGood_unpeeled;9838;5.00047;1;algorithm=base,faithful=yes,bits=1
+GordonBGood_unpeeled_block;10319;5.00005;1;algorithm=base,faithful=yes,bits=1
+GordonBGood_unrolled;15865;5.0002;1;algorithm=base,faithful=yes,bits=1
+GordonBGood_unrolled_hybrid;28818;5.00012;1;algorithm=base,faithful=yes,bits=1
 ```
 
 ## Benchmarks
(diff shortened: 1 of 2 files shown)


rbergen:
@GordonBGood Quoting from your README: _but this implementation **clears them bit by bit with individual operations** in a manner similar to the technique used in the Rust "striped hybrid reset" technique but using larger words instead of bytes_ 

Could you please explain to me in a little more detail how the part printed in bold is achieved? I'm particularly interested in how the effective twiddling of multiple bits in one operation is prevented when the masks are 64 bit words.

GordonBGood:
@rbergen:

> @GordonBGood Quoting from your README: _but this implementation **clears them bit by bit with individual operations** in a manner similar to the technique used in the Rust "striped hybrid reset" technique but using larger words instead of bytes_
> 
> Could you please explain to me in a little more detail how the part printed in bold is achieved? I'm particularly interested in how the effective twiddling of multiple bits in one operation is prevented when the masks are 64 bit words.

Sure, I think I can explain,  In fact, there is no twiddling of bits for the dense base primes masking, as the masks are literal/immediate 64-bit values with just one bit set in each (inverted algorithm with the composites array starting as zeros and bits set to one when marked).  So, like the Rust "striped hybrid" version, for the dense base primes (except it does it by byte), I read a 64-bit word and then do multiple operations `orq`ing the one-bit-set masks for the pattern, then write out the 64-bit word to its original location.

You can see this clearly in the code, for instance for the pattern for the base prime value of three the first of the three words is read at line 123, then all of the or'ing is done until line 143, and then this is repeated for the next two 64-bit words until line 187 when the base pointer is incremented by three 64-bit word sizes and the whole process is repeated to the limit near the end of the buffer, leaving just the few remaining marking operations that are less than the word loop span.

The fact that the masks are 64-bits wide is no different than the Rust "striped hybrid reset" technique of using masks that are only a byte wide; Mike Barber thinks that the speed gain is due to the compiler and/or CPU doing such a thing as combining the masks, but it's not - the speed gains are due to using a high ration of higher throughput register instructions and better cache associativity due to less frequent memory accesses, even when accessing memory by byte sizes as in the Rust implementation.

Now, the CPU itself could be optimizing all of these multiple `or` operations into one mask but it's not since we would know that from the time used as it would be a huge multiple faster than even this, so it is only doing the usual CPU optimizations of fusing microops to achieve the stated throughput of immediate to register `or`'s of a half or a third of a CPU clock cycle (depending on the CPU) as it always does, and even if the CPU or even the compiler could recognize the pseudo random pattern and combine it into a single mask (which no CPU or compiler that I know about is that smart), it would be out of our control as we have still issued individual commands.

That's why I state that it is clearing them bit-by-bit with one operation per bit; as you can see the code is doing exactly that but using the modulo pattern over the words so as to form a loop which spans exactly the step value number of 64-bit words for each base prime value within the dense processing limit, here up to and including nineteen.  As stated, there is likely some small benefit of setting a threshold a little higher, like 31, but the code generated code would get about twice as long for just a little benefit.

I hope you see that this issuing of individual single-bit masking operations means that the bits are being cleared individually.

rbergen:
@GordonBGood Thank you. Based on what I'm seeing, I would summarize things as follows: the algorithm sets the bits in the 64 bit `rp(n)` words individually, albeit in a static fashion, by the use of manually unrolled bit indexing for factors <= 19. While doing this, it uses the local variable `v` to set all the bits for a 64-bit word in sequence. When `v` has been completed for a 64-bit range, then it is committed to the reverse prime memory space.

Would that be an accurate description?

GordonBGood:
@rbergen:

> @GordonBGood Thank you. Based on what I'm seeing, I would summarize things as follows: the algorithm sets the bits in the 64 bit `rp(n)` words individually, albeit in a static fashion, by the use of manually unrolled bit indexing for factors <= 19. While doing this, it uses the local variable `v` to set all the bits for a 64-bit word in sequence. When `v` has been completed for a 64-bit range, then it is committed to the reverse prime memory space.
> 
> Would that be an accurate description?

Yes, that sums it up.  Rather than doing the marking in place as most algorithms do, for the dense base prime values it marks them into the intermediately held value `v`.

rbergen:
@GordonBGood 

> Yes, that sums it up. Rather than doing the marking in place as most algorithms do, for the dense base prime values it marks them into the intermediately held value `v`.

Thank you. Then I agree that this is a faithful base algorithm approach.

rbergen (approved):
LGTM

marghidanu (approved):

PR #674, opened 2021-08-23
TeX/solution_2: arm-compatible Dockerfile

jfbu:
## Description

I modify the Dockerfile to be based on Ubuntu:20.04 and its texlive-base as this is the simplest solution I have found for compatibility with both arm and amd. Inspired by the Dockerfile of TeX/solution_1.

As a bonus the Docker runs are a bit faster, don't know why (previous one was Alpine based texlive-minimal install).

Also some tidbits on usage with Windows/MikTeX or TeXLive

## Contributing requirements

### DIFF
--- a/PrimeTeX/solution_2/Dockerfile
+++ b/PrimeTeX/solution_2/Dockerfile
@@ -1,6 +1,10 @@
-FROM phipsgabler/texlive-minimal@sha256:f1e37453ef11783e582033adc07ba908839c00889634d17364a3a7869bc7f56d
+FROM ubuntu:20.04
 
 WORKDIR /opt/app
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get install -y texlive-base
 
 COPY shared_batteries.tex *_sieve.tex *_benchmark.tex runpdftex.sh texmf.cnf ./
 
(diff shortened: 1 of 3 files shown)


jfbu:
@rbergen I noticed that  [your latest Report at 24/08/2021 13:43:27](https://plummerssoftwarellc.github.io/PrimeView/report?id=rbergen-1629805407.json&hi=False&hf=False&hp=False&fi=&fp=&fa=&ff=&fb=&tp=False&sc=pp&sd=True) which probably incorporated this merged PR does not show anymore this TeX/solution_2... should I worry about this? I did test locally the new Dockerfile... (and from previous report on your hardware I did not expect even the wheel48of210 to exceed 294 passes which would exhaust pdfTeX memory)

rbergen:
I had a look at the logging of the latest automated run, and a build error was reported on PrimeTeX/solution_2. I've just manually built and run the Dockerfile on the benchmark machine successfully, and I could tell the earlier failure occurred during the apt-get install. My guess would be that an intermittent failure occurred while downloading one of the packages.
As the build did complete now, the Docker image is cached locally and I expect the solution to show up again in tomorrow's benchmark report.

jfbu:
...maybe I am not at fault, what a relief! thanks!

PR #675, opened 2021-08-25
Lean4 solution by badly-drawn-wizards

badly-drawn-wizards:
## Description
An implementation in the Lean4 proof assistant.

## Contributing requirements

### DIFF
--- a/PrimeLean4/solution_1/Dockerfile
+++ b/PrimeLean4/solution_1/Dockerfile
@@ -1,0 +1,22 @@
+FROM ubuntu:21.04
+
+RUN apt-get update
+
+# Install elan and lean4
+RUN apt-get install curl -y
+RUN curl https://raw.githubusercontent.com/leanprover/elan/master/elan-init.sh -sSf | sh -s -- --default-toolchain none  -y
+
+# Set shell to login shell so that ~/.profile is loaded
+SHELL ["/bin/sh", "-lc"]
+
+# Install lean toolchain early so that changes in source code don't
+# trigger a redownload of lean
+RUN elan toolchain install leanprover/lean4:4.0.0-m2
+RUN apt-get install build-essential libgmp-dev -y
+
+WORKDIR /app
+COPY . .
+
+RUN leanpkg configure
+RUN leanpkg build bin
+ENTRYPOINT [ ./build/bin/PrimeLean4 ]
--- a/PrimeLean4/solution_1/Dockerfile
+++ b/PrimeLean4/solution_1/Dockerfile
@@ -1,22 +1,13 @@
 FROM ubuntu:21.04
 
-RUN apt-get update
-
-# Install elan and lean4
-RUN apt-get install curl -y
-RUN curl https://raw.githubusercontent.com/leanprover/elan/master/elan-init.sh -sSf | sh -s -- --default-toolchain none  -y
-
-# Set shell to login shell so that ~/.profile is loaded
-SHELL ["/bin/sh", "-lc"]
-
-# Install lean toolchain early so that changes in source code don't
-# trigger a redownload of lean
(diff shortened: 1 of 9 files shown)


PrimeLean4/solution_1/PrimeLean4.lean
```diff
@@ -0,0 +1,115 @@
...
+  let start_ms := ← IO.monoMsNow
+  let stop_ms := start_ms + dur_ms
+  let mut iterations : Nat := 0
+  let mut last_ms := start_ms
+  for _ in Loop.mk do
+    last_ms := (← IO.monoMsNow)
+    if stop_ms ≤ last_ms then break
+    don't_optimize $ impl.run size
+    iterations := iterations + 1
+  pure ⟨impl.toImplMetadata, iterations, last_ms - start_ms⟩
```
rbergen, line 101: For an implementation to be marked as faithful, it needs to recreate the sieve class (or equivalent) instance from scratch for every run. This implementation seems to operate on a single `Impl` instance that is reused within the timed loop. Can you please take me through how this implementation handles the rule I just referred to?

PrimeLean4/solution_1/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# Install lean toolchain early so that changes in source code don't
+# trigger a redownload of lean
+RUN elan toolchain install leanprover/lean4:4.0.0-m2
+RUN apt-get install build-essential libgmp-dev -y
+
+WORKDIR /app
+COPY . .
+
+RUN leanpkg configure
+RUN leanpkg build bin
```
rbergen: Please consolidate the various series of `RUN` directives to one directive per set of commands.

PrimeLean4/solution_1/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+ENTRYPOINT [ ./build/bin/PrimeLean4 ]
```
rbergen: When I `docker run` the image, nothing happens:
```
$ docker run -ti --rm lean4
$
```

PrimeLean4/solution_1/PrimeLean4.lean
badly-drawn-wizards: Hi. This is a confusion of naming between FP and OOP communities. Classes here are typeclasses, which are distinct from the notion of classes in OOP. The equivalent of the sieve class being recreated here is the recreation of the array of primes each run. If this is not deemed equivalent, then I don't mind unmarking it as unfaithful.

badly-drawn-wizards: You can think of the classes here as closer to Rust's traits.

PrimeLean4/solution_1/Dockerfile
badly-drawn-wizards: Ah, forgot quotes. Is fixed now.

PrimeLean4/solution_1/PrimeLean4.lean
rbergen: Ok, thanks for clarifying that.

PrimeLean4/solution_1/PrimeLean4.lean
```diff
@@ -0,0 +1,115 @@
...
+  run : Nat → array
+
+structure Output :=
+  impl : ImplMetadata
+  iterations : Nat
+  total_ms : Nat
+
+instance : ToString Output where
+  toString o := 
+    let tags := ";".intercalate (o.impl.tags.map toString)
```
rbergen: The separator that has to be used between tags is the comma (,), not the semicolon (;). The semicolon _is_ the correct separator for the fields in the output line (line 89).

rbergen (approved):
LGTM

marghidanu (approved):

PR #678, opened 2021-08-25
new V solution_2

Ekopalypse:
## Contributing requirements

### DIFF
--- a/PrimeV/solution_2/README.md
+++ b/PrimeV/solution_2/README.md
@@ -1,0 +1,37 @@
+# V solution by Penguindark,Ekopalypse ...
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-unknown-yellowgreen)
+
+## Run
+
+### Run locally
+
+You can find an installation guide on the official website at https://vlang.io/.
+Basically call v -cc gcc -prod run primes.v 1_000_000
+
+### Docker
+
+The official images have no support for ARM64 therefor; this Dockerfile also installs the latest version of V.
+
+```
+docker build -t primes .
+docker run --rm primes
+```
+
+## Benchmarks
+
+```
+Passes: 4978, Time: 5, Avg: 0, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: true
+Passes: 4967, Time: 5, Avg: 0, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: true
+Passes: 4969, Time: 5, Avg: 0, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: true
+Passes: 4967, Time: 5, Avg: 0, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: true
+Passes: 4914, Time: 5, Avg: 0, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: true
+Passes: 4909, Time: 5, Avg: 0, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: true
+Passes: 4962, Time: 5, Avg: 0, Limit: 1000000, Count1: 78498, Count2: 78498, Valid: true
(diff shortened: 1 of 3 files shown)


rbergen:
Please explain the differences between the existing solution and this one, and in particular why those changes warrant a new solution to be added.

Ekopalypse:
Sorry for ignoring this requirement. Two key features have changed: According to the new (?) rules, it is a faithful implementation, since both size and memory buffer are set at runtime, and a different memory solution is used, namely a 64-bit variable instead of a bit array.

PrimeV/solution_2/README.md
```diff
@@ -0,0 +1,37 @@
+# V solution by Penguindark,Ekopalypse ...
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-unknown-yellowgreen)
```
rbergen, line 6: I don't understand why this solution is marked as having an unknown bit count. You've indicated in the PR comments that this implementation uses a 64-bit variable for prime candidate storage. Therefore, I think there are two possibilities:

1. The solution uses bit twiddling to set individual bits in the sieve memory area, in which case the bit count is 1
2. The solution uses a 64-bit variable for each prime candidate, in which the bit count is 64.

Looking at code like the following, I think the first is what's happening. In that case, the bit count (in README.md badges and in the program output) should be indicated as being 1:
```v
sieve.buf[num >> 6] |= (u64(1) << (num & 0x3F))
```

PrimeV/solution_2/primes.v
```diff
@@ -0,0 +1,114 @@
...
+	if show_results {
+		println('')
+	}
+
+	avg := f64(duration / passes)
+	count_primes := sieve.count_primes()
+	valid := (count_primes == dictionary[sieve.size])
+	eprintln('Passes: $passes, Time: $duration, Avg: $avg, Limit: $sieve.size, Count1: $count, Count2: $count_primes, Valid: $valid')
+
+	println('Penguindark;$passes;$duration;1;algorithm=base,faithful=yes')
```
rbergen, line 77: See earlier comment.

PrimeV/solution_2/primes.v
```diff
@@ -0,0 +1,114 @@
...
+
+fn main() {
+	sw := time.new_stopwatch(time.StopWatchOptions{})
+
+	size := os.args[1].u64()
+	mut passes := 0
+
+	mut sieve := Sieve{
+		size: size
+		buf: unsafe { &u64(vcalloc(8 * int(size / 64 + 1))) }
```
rbergen, line 98: For an implementation to be marked as faithful, the sieve instance needs to be recreated from scratch for every iteration of the timed loop, as per the following rule in the contributing guidelines: _"It uses a class to encapsulate the sieve, or a (closest) equivalent feature in your language if a class construct is not available. This class must contain the full state of the sieve. **Each iteration should re-create a new instance of this class from scratch.**"_

Ekopalypse:
I am closing this in favor to the solution suggested by @GordonBGood. #688

GordonBGood:
@Ekopalypse:

> I am closing this in favor to the solution suggested by @GordonBGood. #688

I'm sorry I didn't notice that you already had this PR open or I would have communicated before opening my own, but I just started to get interested in V about when you were submitting this.

Now that I do know about it, I have had a look at your code and see that your solution is to add a one bit solution but accessing the array by 64-bit words.  As you can see in my PR, I have moved well beyond this for my most advanced technique in accessing the array by 8-bit words but marking one bit at a time for base prime values above 19 and accessing by 64-bit words one bit a time for the dense marking for base prime values below this where both use pre-computed (by code generation) constants to do the marking according to the modulo pattern based on the base prime value factor.

Thanks for closing this and letting my PR proceed.

Ekopalypse:
@GordonBGood - no problem and thanks for your V contribution. It's nice to see what someone can accomplish who knows what he is doing. I am impressed that your exterme-hybrid solution runs 5x faster on my machine than the solution I was going to add. As a novice programmer, this is a wonderful chance to learn from your code. Thanks again.

PR #681, opened 2021-08-26
Befunge: use 32 bits per cell for the sieve, not just 8

tjol:
## Description

I've spent some time optimising the Befunge solution, enough to squeeze out an additional iteration (up to 6!) on my machine. The main improvement is a switch to using 32 bits per cell rather than 8 for storing the sieve state. I've also added an preliminary sieve reset before starting the actual benchmark to make sure the interpreter has enough memory allocated before the 1st real iteration, and added an error message to make the program fails gracefully if you try it on an interpreter that doesn't support the required fingerprints.

## Contributing requirements


rbergen (approved):
LGTM

marghidanu (approved):

PR #685, opened 2021-08-26
drag-race typescript - results array has half the size

TrimVis:
## Description
The original TypeScript solution1 used 8 bits / flag. (by Tudor Marghidanu)
The original TypeScript solution2 then used 1 bit / flag. (by Michael van der Bokke)
solution2 also added the support to run different data structures for the flag storage.
My solution is based on solution2, but using an results array of half the size.

## Contributing requirements

### DIFF
--- a/PrimeTypeScript/solution_3/.gitignore
+++ b/PrimeTypeScript/solution_3/.gitignore
@@ -1,0 +1,2 @@
+node_modules/
+build/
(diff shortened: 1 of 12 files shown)


rbergen:
May I ask why you're not submitting this as an improvement on solution_2?

TrimVis:
I didn't think of that.
Now that you say it, that's probably a good idea.
I will change solution 2 instead and will reopen this PR, when I did that.
Thanks! :+1:

PR #686, opened 2021-08-26
[D solution 2] Solution that can't even compile.

BradleyChatha:
## Description
I really don't expect you to allow this, but I wanted to try anyway.

This is a solution that *very technically works*, except you need at least 136GB of RAM (Windows gave up growing my swap file after that, so you need even more) to compile it for the full one million sieve size.

The class is never instantiated with the full 1_000_000 (in fact, never at all) so it's effectively disabled, meaning this PR changes practically nothing except adds a technically working, fully unrolled sieve.

It's a shame since 100_000 would compile (with around 20GB RAM) but another 10x of that is far too much for the poor D compilers.

Just to clarify: This new class is *not compiled at all*, so it's purely just code for anyone interested in looking at it, the builds won't suddenly start needing 300GB to run.

This was inspired by me trying to make the other solutions faster, getting frustrated, giving up and then going "frick it, let's inline the entire thing".

## Contributing requirements

### DIFF
--- a/PrimeD/solution_2/README.md
+++ b/PrimeD/solution_2/README.md
@@ -56,6 +56,104 @@ There are 3 test types currently: single, dynamicThreads, and staticThreads:
 
 In general "RT" means "Runtime" and "CT" means "Compile time". Compile time does not mean it runs completely at compile time, but only parts of it might.
 
+## Implementations
+
+When an algorithm says it uses the 'base-unrolled' algorithm, it means it uses the base algorithm with a slightly unrolled loop.
+
+### SieveCT
+
+**Algorithm:** base-ex
+
+**Bits:** 1
+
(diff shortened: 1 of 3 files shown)


BradleyChatha:
Speed wise, it's roughly 2.5x faster compared to the other solutions when PRIME_COUNT is set to 100_000, which may or may not reflect what it'd be like at 1_000_000.

rbergen:
We have more code in the project that is not actually benchmark-executed. Which is fine, if it doesn't add an undue amount of reviewing workload. In this case, I don't think it does, which means I'm okay with merging this.

That said, I would like to suggest you consider spending some text in the README.md on this specific implementation. People who browse the source code of this solution may wonder what purpose the class serves when it clearly is not compiled or run.

PR #688, opened 2021-08-29
Add fast V lang solution_2...

GordonBGood:
## Description

This adds single bit techniques that have successively better performance than solution_1.

The following techniques are provided:
1. Bit twiddling, shifting and masking a byte buffer by single bits.
2. Stride8, which breaks the above loops into more simpler loops marking by fixed mask values across the modulo patterns by bytes.
3. Stride8 by blocks does the above but performs the eight individual marking loops per base prime value across more-or-less cache sized blocks, 16 Kilobytes used here
4. Extreme loop unrolling, where the above eight loops are recombined into a single loop with literal/immediate values pre calculated for each of the modulo bit patterns.  In this case, there are only four patterns as there only eight base prime values modulo eight and as each prime value starts from the base of the sieve buffer to the top, there is only one start index for each of the modulo prime values.
5. The final technique combines the above for larger base prime values above 19 with a "dense" technique for the smaller primes as described in the README.md file, but essentially moving the marking of composite bits into the CPU by 64-bit words, which when completed marking the word, the word is committed to the original location in the sieve buffer.

## Contributing requirements


rbergen:
We already have #678 open aiming to add a second V solution, meaning we can't add both without causing a merge conflict. 

@GordonBGood I'd like to suggest that you and @Ekopalypse discuss how to move forward with these two PRs. Full disclosure: personally, I'd prefer that discussion ending up with one PR that adds one additional V solution, containing whatever the discussion between you concludes should be included.

GordonBGood:
@rbergen:

> @GordonBGood > We already have #678 open aiming to add a second V solution, meaning we can't add both without causing a merge conflict.
> 
> @GordonBGood I'd like to suggest that you and @Ekopalypse discuss how to move forward with these two PRs. Full disclosure: personally, I'd prefer that discussion ending up with one PR that adds one additional V solution, containing whatever the discussion between you concludes should be included.

@Ekopalypse kindly closed his PR in favour of this one; I used the delay to tweak this solution a little further, and it is now ready for your full review.  It should be easy to determine that I use my usual "extreme" and "hybrid" techniques as the V language does not have templates or macros, so the code is hanging right out there...

PrimeV/solution_2/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM alpine:3.13 AS build
+
+RUN apk add --no-cache build-base bash git
+
+WORKDIR /opt/vlang
+RUN git clone https://github.com/vlang/v /opt/vlang && make
```
rbergen: Could you please pin the V version down to a specific one? You can do this by either checking out a tag (release) or a specific commit between the `git clone` and the `make`.

GordonBGood: @rbergen:

> Could you please pin the V version down to a specific one? You can do this by either checking out a tag (release) or a specific commit between the `git clone` and the `make`.

I had just copied the Dockerfile from marghidanu's solution_1 and hadn't looked at it too much.  That Dockerfile is cloning from master.  When I did as you suggested and checked out the latest released version 0.2.2, it won't build on the Alpine nor Ubuntu stable images, perhaps due to missing installations.  So I checked out the actual tag value for the next release, 0.2.4, and that does compile and run.  Since 0.2.4 is near release, there shouldn't be any performance changes up to its release and, of course, it will be frozen once released.

I can't see how to do anything else without doing a lot of research on why 0.2.2 doesn't compile, which seems pointless given it is soon to be replaced.  I have already spent far too much time in this V language project already.

But someone else is welcome to try...

I've submitted a new Dockerfile checking out tagged version 0.2.4.

rbergen: Yes, after writing my comment I realized I should have checked the Dockerfile for solution_1, I think we should make a similar (if not identical) change there, but I'll discuss this with @marghidanu; he authored that solution,

Using 0.2.4 is perfectly fine with me. It became a release 13 hours ago, according to GitHub.

rbergen (approved):
LGTM. Thanks to both @Ekopalypse and @GordonBGood for the smooth alignment on #678 and this PR.

marghidanu (approved):

PR #689, opened 2021-08-29
unrolled hardcoded looping based on rbergen-x64ff-bitshift

joonicks:
a simple loop unroll operating on bytes rather than words
only two skipvalues exist for byte operations, each getting its own hardcoded loop

performs about 36% faster on my machine


rbergen:
Please structure your PR as a modification to PrimeAssembly/solution_1, being one that adds your unrolled approach as an additional implementation to the the implementations that that solution already contains. It will save us an additional docker build and docker run during benchmarks.

joonicks:
my version has now evolved further into a more hardwired loop (+48% from x86ff_bitshift)

I was thinking of saving the new version as yet another program, making solution_3 contain two variants

or do you still want my versions merged back in solution_1?

rbergen:
It's hard to know for sure unless I see the second implementation you'd add, but if the second implementation is a more or less linear progression in the use of hardwired loops, then I would suggest only adding the most "aggressive" version. And yes, I would add that as an additional implementation to solution_1. At a source code level, the link between the original implementations and yours is still clearly visible.

Of course, when it comes to credits, you can mark your implementation as yours in the program output, and also add yourself to the list of authors in the solution_1 README.md.

joonicks:
pushed changes back to solution_1, + added hardwired variant.

Im still git ignorant, did I get it right?

PrimeAssembly/solution_1/primes_ff_hardwired.asm
```diff
@@ -0,0 +1,353 @@
...
+; * r13d = initBlockCount
+
+    mov         r12, rax                    ; sievePtr = &sieve
+    mov         r13d, [r12+sieve.sieveSize] ; sieveBytes = sieve.sieveSize
+    shr         r13d, 7                     ; sieveBytes /= 128
+    inc         r13d                        ; sieveBytes++
+    shl         r13d, 4                     ; sieveBytes *= 16
+
+    mov         edi, r13d                   ; ask sieveBytes
+	add			edi, 10000                  ; lazy -- add 10kb to allow overextending writes
```
rbergen, line 214: I don't like this. For starters, one of the rules for a faithful implementation is that _"The sieve size and corresponding prime candidate memory buffer (or language equivalent) are set/allocated dynamically at runtime. **The size of the memory buffer must correspond to the size of the sieve.**"_ Adding 10K to what is actually needed dilutes conformity to that rule, does indeed feel lazy, and makes me feel that the algorithm is not finished.

Lines 207 to 211 make a concerted effort to limit the overallocation of memory to a maximum of 15 bytes. I think we should stay within those or very similar boundaries.

PrimeAssembly/solution_1/primes_ff_unroll4.asm
```diff
@@ -0,0 +1,372 @@
...
+; * r13d = initBlockCount
+
+    mov         r12, rax                    ; sievePtr = &sieve
+    mov         r13d, [r12+sieve.sieveSize] ; sieveBytes = sieve.sieveSize
+    shr         r13d, 7                     ; sieveBytes /= 128
+    inc         r13d                        ; sieveBytes++
+    shl         r13d, 4                     ; sieveBytes *= 16
+
+    mov         edi, r13d                   ; ask sieveBytes
+	add			edi, 10000
```
rbergen, line 214: Same comment as before.

rbergen:
> Im still git ignorant, did I get it right?

You did, thank you.

PrimeAssembly/solution_1/primes_ff_hardwired.asm
joonicks: overallocating is an extremely cheap way of avoiding bounds checking in an unrolled loop.

my question would be; what is the point of this restriction?

allocating 108% of the memory needed does "correspond" to the size of the sieve. I could even calculate a narrower margin if necessary. but now we're getting into semantics and personal preference. personal interpretation should never be the basis for rules.

I'd wager that if you did an in depth study of what the top entries in the 'base,faithful' algorithm entries did, you would find far more egregious violations of the base rules.

joonicks: be mindful, my assembly version spends 80% or so of its cpu cycles on instructions flipping single bits, any variant claiming 50%+ higher speed or so cant be doing it flipping single bits unless they have a magic flawless algoritm to determine if an earlier factor flipped that bit already. so im pretty confident in saying your current top 5-10 is not base algorithm single bit flippers.

rbergen: I'd wager against that, because I did review all of the top entries for rule compliance, and so have others. I'm also not going to defend the rules, they are what they are. 

It's simple: I won't approve this PR unless you remove the overallocation.

joonicks: hypocrisy. your own "faithful" version allocates extra memory just because it uses xmm0 writes to blank the buffer.

rules being rules, why are you personally allowed to bend the rules when others are not?

blanking exactly 125000 bytes (1000000 bits) isnt exactly difficult.

rbergen: I'd argue that over-allocating a maximum of 15 bytes is not quite the same as over-allocating 10K, _on top of the max 15 additional bytes you still also claim_. And I'll happily grant you and anybody else those 15 bytes, as I've already stated, so I don't see the hypocrisy. You labeled the +10K as lazy yourself, and that is exactly what it is.

It's going to be a case of "take it or leave it". I'm fine either way.

PR #690, opened 2021-08-30
Go/solution_2 reworked

ssovest:
## Description
In order to continue developing solution_2, i had to do something with all the code duplication, so i've reworked it completely. Storage types are now generated from templates with options defined in a json config. `block` variations are created with decorators, so any storage type can now be `block`. Also, some simple tooling for creating, testing, benchmarking and profiling implementations is now an integral part of the solution.

It's a rather big PR, so, please, let me know if i can do something to simplify the review process.

## Contributing requirements


rbergen:
Does this PR add anything in terms of performance improvements, approaches, implementations, etc.?

ssovest:
Yes, it does.
 - No new storage types are added, but existing ones are optimized.
 - `rblock` variations are added. They're similar to `block`, but perform better in most cases
 - Also, Go version changed to 1.17 (It has [some changes](https://golang.org/doc/go1.17#compiler) that affect performance)

rbergen (approved):
LGTM

marghidanu (approved):

PR #692, opened 2021-09-01
Java add unroll and strided block implementation

chrvanorle:
## Description
Added an unrolled and an unrolled strided block version. The strided variant uses the idea of the repeating patterns used in I64PatternCalc, while not using the multibit masks that made that an `other` algorithm. Using blocks prevents the cache trashing that would otherwise occur.

From the striped/strided names i've seen, i think strided makes the most sense for this one (final name being chrvanorleStrided32Blocks16k). I removed the W(armup) from the name as it doesn't really add anything since all runs use it.

There is a significant drop in performance on (ry)zen due to the changes in PrimeSieveBase::runSieve. This is due to an issue i don't fully understand (#679), but only seems to affect the implementations using the simple loops, and not the unrolled variants. As this only affects the slow implementations for a single architecture, i decided to just go forwards with this PR.

## Contributing requirements

### DIFF
--- a/PrimeJava/solution_4/README.md
+++ b/PrimeJava/solution_4/README.md
@@ -17,7 +17,8 @@ This is a collection of prime sieve algorithms implemented in Java. All solution
 * I64: Using 64 bit integers to store primes, 1 bit per prime
 * I64C: Using 64 bit integers to store primes, 1 bit per prime, precache masks
 * I64PatternCalc: Using 64 bit integers to store primes, 1 bit per prime, using masks to mark multiple values per array write, making it an "other" algorithm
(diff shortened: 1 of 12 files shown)


ItalyToast:
@chrvanorle striped/stride are 2 distinct algorithms. the Striped approach transposes the bit array:
Striped
```
1 5 9
2 6 10
3 7 11
4 8 12
```
Stride
```
1  2  3
4  5  6
7  8  9
10 11 12
```

It looks like you are using the strided approach

rbergen (approved):
LGTM

marghidanu (approved):

PR #695, opened 2021-09-03
Add Crystal solution_2...

GordonBGood:
## Description

Add better as in faster to much faster techniques than for solution_one, which just uses the `bitarray` class available from the standard library.  The following techniques are used, all odds-only bit-packed and `faithful to base` using one bit to represent each number:
1.  A "bit-twiddling" technique which has slight improvements over use of the "bitarray" available from the standard library.
2. A "stride8" technique which breaks the composite marking operations up to do them out of order for eight different constant bit indices as mask value for eight separate loops.
3. The same as the above but advancing by cache-sized blocks of sixteen Kilobytes to help reduce "cache-thrashing".
4. The technique of "Extreme Loop Unrolling" where the above eight loops are re-assembled into a single loop advancing across the entire deliverable sieve buffer (use of blocks is not possible for this case due to the rules not allowing page-segmentation); the eight marking operations per loop are by constant values according to the modulo pattern, and in this case are generated by a macro.
5. The "extreme hybrid" technique which uses the above for base prime values above 64 but below that threshold uses a "dense marking" technique whereby 64-bit words are read in, completely marked by individual single marking operations, then committed back to the original location, for each 64-bit word in the modulo marking pattern according the the base prime value used to step across the composite numbers to be marked.

While  use of these techniques makes the results from this solution fairly competitive, Crystal is limited in ultimate speed by the use of the LLVM back end, just as other languages that use this back end are as in Rust, Julia, etc.  While LLVM can optimize extremely well for simple code loops, it actually tries too hard and the multiple optimizations for more complex code often turn out to be de-optimizations.   If one knows what optimization(s) is(are) going wrong, one can pass the input to LLVM back through the compilation processes with those passes turned off, but I didn't do this custom step for any of these languages, so didn't do it for this Crystal submission.  This is generally not necessary for languages that use gcc as their back end.

A new Dockerfile is used based on Debian 11 which installs the Crystal compiler and its dependencies.  @rbergen, this installation comes close to your approximate limit of two to three minutes to do the first build, mostly due to the time to install the dependencies, so you may want to create a  Crystal Docker Hub image with the usual pair of tags for the two CPU architectures on which we want to be able to benchmark; accordingly, I have formatted the Dockerfile in a way similar to the one you did for me for the Nim language to hopefully make this easy for you should you decide this is necessary.

## Contributing requirements


rbergen (approved):
LGTM

@marghidanu personally, I'm fine with the docker build duration for this solution, so in itself that doesn't warrant a pre-baked Crystal base image to be created as far as I'm concerned. However, doing so it would allow us to create a base image that also supports arm64, which the current solution_1 Dockerfile doesn't. As solution_1 was authored by you, I'll leave it up to you if we're going to use the Dockerfile in this PR as the basis for that solution, also. If you do think it's a good idea to do so, I'm obviously happy to help out with putting together a primeimages/crystal:1.1 base image.

GordonBGood:
@rbergen:

> LGTM

Forgive me for making changes after the PR has already been reviewed by you, but I wasn't happy with how many steps were required to view the macro expansions so made some slight changes to make it very easy.

In making these changes and reviewing the generated code again, I realized that there was a slight error that didn't affect the validation of the results but could slightly degrade the performance, so corrected that while I was at it.  The `diff`'s on the two files will make it obvious what I did as follows:
1. Added a conditional compilation on a `expand_macros` define for the macro expansion so no source code changes are required to switch between viewing expansions mode and regular benchmark mode.
2. Fixed the minor problem in the `dense_setbits` macro as it was doing some redundant operations; the problem was that for the larger dense base prime values (approaching the limit of 63), the lowest index value for a given 64-bit word was calculated incorrectly so was generating an operation that didn't have any results.  The LLVM back end might see that it was a redundant operation and elide it but might not.
3. Updated the "README.md" file to reflect these changes; in the `diff` of the macro expansion section for the value of 61, you will easily see that the redundant operation has been eliminated.

If you would rather, I can back out these changes and submit a new PR after this PR is merged.

marghidanu:
LGTM

@rbergen The only reason I prefer the build packages is that they take care of the dependencies, I feel we're trading one thing for another if we compile it from source.  In reality, aarch64 is not actually fully supported by Crystal. Here are the supported platforms from their official website: https://crystal-lang.org/reference/platform_support.html#tier-2

marghidanu (approved):

rbergen:
@marghidanu You're right, it is a trade-off. The question is which option's pros weigh more heavily against the other option's cons. Your considerations on it are fair enough.

marghidanu:
Since `aarch64` is not fully supported by the compiler we should expect more issues until that's finalized. I suggest we don't concern with the issue until is officially resolved.

PR #697, opened 2021-09-03
Rust solution_1: small fix

mike-barber:
## Description
This is a small fix/optimisation to the unrolled implementation submitted last week. I missed an important optimisation: essentially, it's quite expensive setting up the sparse resetters because we need to calculate the relative indices of the words we're going to be resetting. This only makes sense when we have lots of iterations to do. When we've only got a few to do (with large prime factors), it's a lot faster to just smash the bits in the simple way (like the basic implementations in this solution do). 

@GordonBGood had this implemented in his solutions, and it makes a pretty big difference on the humble Pi.  Apologies for missing this in the previous submission @rbergen. At least this is a small PR this time :)

## Contributing requirements


mike-barber:
Apologies for the noise - I'll reopen this later. Gordon raised an interesting point, and I suspect I'm not addressing an underlying problem with my sparse resets in this PR. I'm going to close this for now, and open another when I've double checked everything.

PR #698, opened 2021-09-04
Rust solution_1 - start location fix

mike-barber:
## Description
This is a small fix to the unrolled implementation submitted last week. I opened a PR yesterday evening, but it turns out I was treating the symptom rather than the cause. Thanks @GordonBGood for taking a look. 

There was indeed something wrong with the sparse resetter: it was starting from the beginning of the array in all cases, even for large skip factors. While this is functionally still correct, it is a bit of a performance killer from both a cycles and L1 cache point of view.

This PR is a simple fix: start resetting bits only at the square of the factor, rather than the start of the array. 

@rbergen, my apologies for all the noise, as well as missing this the first time. Luckily this PR is really straightforward in comparison to previous ones. 

## Contributing requirements

### DIFF
--- a/PrimeRust/solution_1/prime-sieve-rust/src/unrolled.rs
+++ b/PrimeRust/solution_1/prime-sieve-rust/src/unrolled.rs
@@ -132,7 +132,6 @@ impl FlagStorage for FlagStorageUnrolledHybrid {
     /// ```
     #[inline(always)]
     fn reset_flags(&mut self, skip: usize) {
-        const U64_BITS: usize = u64::BITS as usize;
         const MIN_ITERATIONS: usize = 1200;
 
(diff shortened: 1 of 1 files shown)


rbergen:
> my apologies for all the noise, as well as missing this the first time. Luckily this PR is really straightforward in comparison to previous ones.

@mike-barber please don't worry about it. I've been developing software for 35 years by now and this is how it works, sometimes. :)

mike-barber:
> please don't worry about it. I've been developing software for 35 years by now and this is how it works, sometimes. :)

Thanks! It was one of _those_ bugs: subtle enough to hide for a week or two, despite a lot of searching. And then blindingly obvious in retrospect ;)

PR #706, opened 2021-09-07
Updates Zig to use calloc, and unroll functions

ityonemo:
## Description

Further updates Zig solution #2 to use c stdlib calloc memory allocator, which is faster, and also unroll the hot
loops manually.  The zig complier generates (N) functions as a function lookup table, where N is the bit-width
of the desired integer type, and these tick down the sieve with patterns that match the first N odd numbers.  Also 
generates (N) functions which capture the patterns for *all other* odd integers I, indexed by I mod (bit-width).  See src/unrolled.zig for more information.

also implements contribution by @SpexGuy which uses SIMD vector operations, but it's not as fast as unrolling
the loops.

## Contributing requirements


marghidanu (approved):

rbergen:
@ityonemo Could you please walk me through the algorithm used in the VecSieve.runSmallFactor function? 

As it stands, it looks to me like a bitmask vector of 64 bytes is composed at the beginning of the function, which is then projected throughout the prime sieve towards the end of the function. Is that interpretation correct?

ityonemo:
I believe that's correct, I'm only 99% sure as this was the work of @SpexGuy

SpexGuy:
That's close but not quite correct.  The pattern buffer is `64 + the_prime - 1` bytes long.  For any given prime, there is a repeated sequence of that many bytes.  So with 3, the repeating sequence is 3 bytes long, with 7 the sequence is 7 bytes long, etc.  The piece of the pattern buffer that needs to be stamped into a given location must start with one of these bytes, and the bytes which follow are determined by that starting byte.  So the pattern buffer is prepared to be long enough that we can always pick a valid start byte and load 63 bytes after it to get a full vector.

rbergen:
@ityonemo and @SpexGuy Thank you for your clarification. It does mean that the algorithm in question must then be marked as "other" instead of "base". The "base" algorithm definition includes that all non-primes must be cleared (or set, depending on which bit value represents a prime vs a non-prime) individually, as in: one at a time.

ityonemo:
Ok cool will update it later tonight.

PR #711, opened 2021-09-08
Dockerlint part 1

Macleykun:
The following programming languages and their solutions have been checked and fixed based on Dockerlint:
PrimeAda
PrimeAmd64
PrimeAssembly
PrimeAssemblyScript
PrimeAWK

~~Note that somehow, i was unable to build PrimAmd64 for example. The dockerfiles look right in my eyes and even with some testing i can see the files. I assume it's just me and i only added versioning on them.

```
#9 [5/6] RUN ls -al
#9 sha256:3daebede1a2c80e94db8b4f17304c512b69bc8e5e1e381de464b26d0d6d380d7
#9 0.325 total 104
#9 0.325 drwxr-xr-x    1 root     root          4096 Sep  8 19:17 .
#9 0.325 drwxr-xr-x    1 root     root          4096 Sep  8 19:17 ..
#9 0.325 -rwxr-xr-x    1 root     root           705 Sep  7 20:11 build.sh
#9 0.325 -rwxr-xr-x    1 root     root         13407 Sep  7 20:11 primes_ff_bitbtr.asm
#9 0.325 -rwxr-xr-x    1 root     root         14217 Sep  7 20:11 primes_ff_bitshift.asm
#9 0.325 -rwxr-xr-x    1 root     root         12315 Sep  7 20:11 primes_ff_byte.asm
#9 0.325 -rwxr-xr-x    1 root     root         11023 Sep  7 20:11 primes_uff_bitbtr.asm
#9 0.325 -rwxr-xr-x    1 root     root         10989 Sep  7 20:11 primes_uff_bitshift.asm
#9 0.325 -rwxr-xr-x    1 root     root          9189 Sep  7 20:11 primes_uff_byte.asm
#9 0.325 -rwxr-xr-x    1 root     root           160 Sep  7 20:11 run.sh
#9 DONE 0.3s

#10 [6/6] RUN ./build.sh
#10 sha256:24f3007ef2678ffd943c784cb2ca58d2f1a23b4e988c223c540aa612a9ce1576
#10 0.506 /bin/sh: ./build.sh: not found
#10 ERROR: executor failed running [/bin/sh -c ./build.sh]: exit code: 127
```~~

Feel free to review and give feedback, next i'll do another 5 after primeawk :)

I did test in docker playground and it just works from my fork. So i assume it's just a me thing :P

### CODE REVIEW

File: PrimeAssembly/solution_2/Dockerfile [hunk 1]
```diff
@@ -1,15 +1,11 @@
...
 WORKDIR /opt/app
 
-COPY *.sh ./
-COPY *.s ./
-
-RUN chmod a+x ./*.sh && \
-    ./build.sh
-
-ENTRYPOINT [ "./run.sh" ]
+COPY *.s* ./
```
rbergen: This is not the same as the COPY commands it seeks to replace. `COPY *.s*` would also copy any other files with an extension starting with .s, where the original commands only copy files that either have a .s or .sh extension. There are no such files in the solution right now, but the new copy is less exact.

PrimeAssembly/solution_1/Dockerfile
```diff
@@ -1,15 +1,11 @@
...
-RUN apk add --no-cache build-base nasm
+RUN apk add --no-cache build-base=0.5-r2 nasm=2.15.05-r0
 
 WORKDIR /opt/app
 
-COPY *.sh ./
-COPY *.asm ./
-
-RUN chmod a+x ./*.sh && \
-    ./build.sh
```
rbergen, line 11: The chmod a+x is necessary because at the moment, the execute bit is _not_ set on the .sh files in this solution. That's also why the CI run on this PR currently fails.

PrimeAssembly/solution_2/Dockerfile
```diff
@@ -1,15 +1,11 @@
...
-RUN apk add --no-cache build-base
+RUN apk add --no-cache build-base=0.5-r2
 
 WORKDIR /opt/app
 
-COPY *.sh ./
-COPY *.s ./
-
-RUN chmod a+x ./*.sh && \
-    ./build.sh
```
rbergen, line 11: See previous comment.

rbergen:
I'm getting the impression that you're working on WSL2 on Windows, with your local copy of the GitHub repo residing on NTFS. If I'm wrong about that then you can ignore the rest of this comment. If I'm right then you're obviously free to do so, but my recommendation would be to work on an ext4 file system under WSL2, instead.

Using NTFS will lead to all kinds of issues with that filesystem setting executable bits being where they don't belong, and git then removing them when committing, including where they are indeed supposed to be set.

PrimeAssembly/solution_2/Dockerfile
Macleykun: True! I believed it wouldn't be an issue as it minimized COPY and the less exact doesn't seem an issue to me.
If you wish to prefer exacter COPY instead of less COPY's then i'll do so! It's up to you! I don't have a preference in this :).

Macleykun:
> I'm getting the impression that you're working on WSL2 on Windows, with your local copy of the GitHub repo residing on NTFS. If I'm wrong about that then you can ignore the rest of this comment. If I'm right then you're obviously free to do so, but my recommendation would be to work on an ext4 file system under WSL2, instead.
> 
> Using NTFS will lead to all kinds of issues with that filesystem setting executable bits being where they don't belong, and git then removing them when committing, including where they are indeed supposed to be set.

Well that's something i haven't heard before! I will use a vm instead to do docker testing then! Will fix those .sh files and then those comments are resolved :)

PrimeAssembly/solution_2/Dockerfile
rbergen: I deliberately made the COPY as specific as possible when I authored the Dockerfile, to limit the chances of stuff being accidentally pulled into the image at a later stage. So yes, I would prefer the "exact" COPYs over the other one.

rbergen:
> Well that's something i haven't heard before! I will use a vm instead to do docker testing then! Will fix those .sh files and then those comments are resolved :)

I had never heard of it either, until I ran into the exact same thing a few months ago, when getting started on this project. :)

PrimeAssembly/solution_2/Dockerfile
Macleykun: Fixed :D

Macleykun:
haha i see :) I fixed the one thing ya asked me to do, how does it look?
The second pull will be... quite a bit larger...

rbergen (approved):
LGTM

rbergen:
@Macleykun I'm sorry for not seeing this earlier, but it just occurred to me that you're likely to have used a different hadolint configuration while putting this PR together, than the one we use during our CI. That latter configuration is set in the [config/hadolint.yml](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/config/hadolint.yml) file in the repo.
One issue we ignore in that is 3008, being the one flagging the pinning down of apt-get installs to specific package versions.

I think we'll have to limit the changes in the existing Dockerfiles to those that are reported by hadolint when this project's hadolint configuration is used.

(The reason this now occurs to me is that I'm putting a PR together to add a section concerning hadolint to CONTRIBUTING.md, and I had to figure out how to use custom configuration when running hadolint inside a Docker container.)

Macleykun:
Aaah I see, yeah I used default config and fixed everything. It’ll take a bit to re-do all of those but will get to it tonight :).
Would it be possible to merge this maybe soonish after I fixed the issues?

rbergen:
For PRs like this, we (the maintainers) consistently apply a double review approach. Because of this, the speed with which PRs can be merged partly and sometimes largely depends on when two of us get round to performing the reviews. I should be able to review an update to the PR pretty quickly but I can't speak on behalf of the others, nor do I want to. It's a matter of nobody getting paid for this, lives to live, day jobs, etc.

Macleykun:
true true! tottaly understand! Would it maybe be easier for you two if i segment the pulls even more? I don't mind adjusting to the preference of what's best for the maintainers!

Macleykun:
I'll go ahead and go trough all the others i've made, gonna see if can make multiple pulls that are seperate of eachother :P. Then i can atleast keep better up with any changes that are done.

rbergen (approved):
LGTM, again

rbergen:
More PRs aren't necessarily better for us. :) I mean that to the point that if you have other fixes ready of the same types as you apply here, it may actually be best to roll them up in this very PR. I also say that because I really do think we have now identified the right set of changes to make (primarily clean-up after install and consolidating RUNs).

Macleykun:
> More PRs aren't necessarily better for us. :) I mean that to the point that if you have other fixes ready of the same types as you apply here, it may actually be best to roll them up in this very PR. I also say that because I really do think we have now identified the right set of changes to make (primarily clean-up after install and consolidating RUNs).

Oh, then i will immediatly start to put all the changes i've made into this PR then! I really though it was a easier, but on second though, it's indeed better to just put it all into one PR. Will have to do another re-review. But hopefully it'll be the last for this PR.

rbergen:
Build fails on PrimeCOBOL/solution_1/Dockerfile, due to a change you made to address DL3047. That message is ignored in our config, so that change is not necessary.

Macleykun:
> Build fails on PrimeCOBOL/solution_1/Dockerfile, due to a change you made to address DL3047. That message is ignored in our config, so that change is not necessary.

Thanks for pointing that one out! I'm currently going trough all the dockerfiles and testing them, i believed i specfically tested COBOLT but didn't. The issue was an oudated wget which fixed the issue by installing it. I've kinda already addressed DL3047 in the solutions so i'll keep those fixes in.

Every so often i'll push my changes, so technically the CI doesn't have to be re-ran each time. Ofcours it's up to you :).
I'm now at Forth and have about 50 more solutions to go trough :P.

rbergen:
I see you've addressed the build failure by bumping wget to a newer version, In line with my earlier comment that changes should in principle be limited to what's needed to address issues flagged by this project's hadolint configuration, I do think the appropriate thing to do is revert the use of the --progress option when running wget, instead.

Macleykun:
> I see you've addressed the build failure by bumping wget to a newer version, In line with my earlier comment that changes should in principle be limited to what's needed to address issues flagged by this project's hadolint configuration, I do think the appropriate thing to do is revert the use of the --progress option when running wget, instead.

Addressed in current and upcoming changes!

rbergen:
Thanks! I'll hold on my next review and CI run until you notify me that all changes you want to pull into this PR, are indeed in.

Macleykun:
Ok this is now ready for CI testing!
Also, i found some left over files i think from vscode?
./PrimeCPP/solution_3/.vscode
./PrimeCPP/solution_2/.vscode

Is it ok if i also delete those in this pull? Or rather not?

PrimeHaskell/solution_1/Dockerfile
```diff
@@ -2,19 +2,17 @@ FROM haskell:8.10
...
-RUN ./llvm.sh 9
-
-COPY stack.yaml package.yaml /opt/sieve/
+COPY stack.yaml package.yaml ./
 
 RUN stack build --only-dependencies
 
-COPY src/ /opt/sieve/src/
-COPY app/ /opt/sieve/app/
+COPY src/ app/ ./
```
rbergen: This change seems to break the build.

PrimeKos/solution_1/Dockerfile
```diff
@@ -1,25 +1,17 @@
 FROM alpine:3.13 as builder
 
 RUN apk update && \
-    apk add curl make g++
+    apk add --no-cache curl make0 g++
```
rbergen: This change seems to break the build.

PrimeWren/solution_1/Dockerfile
```diff
@@ -1,14 +1,14 @@
 FROM ubuntu:20.04
 
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN apt-get update \
-    && apt-get install -y curl busybox 
-
-RUN curl -sL https://github.com/wren-lang/wren-cli/releases/download/0.3.0/wren_cli-linux-0.3.0.zip | busybox unzip - -d / \
+    && apt-get install --no-install-recommends -y curl busybox3 \
```
rbergen: This change seems to break the build.

PrimeProlog/solution_1/Dockerfile
```diff
@@ -1,13 +1,11 @@
...
-RUN apt-get update && apt-get -y install gcc
-
 WORKDIR /opt/app
-
 COPY primes-*.pl bitvector.c run.sh ./
 
-RUN swipl-ld -cc-options,-w,-O3 -shared -o bitvector bitvector.c
-
-RUN chmod a+x run.sh
+RUN apt-get update && apt-get --no-install-recommends -y install gcc \
```
rbergen: This change seems to break the build.

PrimeZig/solution_3/Dockerfile
```diff
@@ -1,9 +1,8 @@
 FROM debian:buster-slim as builder
 
 RUN apt-get update && \
-    apt-get install -y curl xz-utils
+    apt-get install --no-install-recommends -y curl xz-utils
```
rbergen: This change seems to break the build.

PrimeV/solution_1/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM alpine:3.13 AS build
 
 ENV V_VER="0.2.4"
 
-RUN apk add --no-cache build-base bash git
+RUN apk update && apk add --no-cache build-base bash git
```
rbergen: This change seems to break the build (although I have to admit I don't know how.)

rbergen (commented):
There are 6 build failures in CI. Did you test these changes locally before committing?

Also, I see some Hadolint findings still remain. Did you mean to leave those?

rbergen:
> Is it ok if i also delete those in this pull? Or rather not?

Please don't.

PrimeHaskell/solution_1/Dockerfile
Macleykun: Correct, COPY doesn't copy the folder itself, rather it's content. Reverted back to two COPY commands.

PrimeKos/solution_1/Dockerfile
Macleykun: Didn't correctly removed the version tag. Installing make instead of the wrong make0.

PrimeWren/solution_1/Dockerfile
Macleykun: Same issue as before, didn't removed the version correctly from busybox.
Also it seems that ca-certificates was also needed.

PrimeProlog/solution_1/Dockerfile
Macleykun: libc-dev was also needed.

PrimeZig/solution_3/Dockerfile
Macleykun: ca-certificates had to also be installed

PrimeV/solution_1/Dockerfile
Macleykun: Yep, this is something that's sadly unrelated to my pull but.
When i try to compile it with make and gcc on ubuntu. I get the same error.
When i just use the latest, it works.

I'm guessing it has to do with a newer build-base version/gcc/make which doesn't work with version 0.2.4.

You wish to:
use this version: weekly.2021.38 ? It's the only one that works and has a fixed version.
want me (or ask someone else) to find out if an older version of the build tools at 31 august (or earlier) works?

Macleykun:
Ok all of the build issues are resolved! The PrimeV one is the only one a decision has to be made on.
I did check every not and then the build, but i figured testing in CI in this pull is quicker/easier. Hence the commited build errors.

PrimeV/solution_1/Dockerfile
marghidanu: Adding versions to alpine will break more stuff in the future as previous versions are discarded. Also there’s no need to use `apk update` if using `apk add —no-cache`.  Why do we need these improvements for the linter, we can simply ignore more rules.

marghidanu: Also `—no-install-recommends` can skip essential packages, not sure about these rules.

Macleykun: The apk update should be removed. That was unintentional!
--no-install-recommmends avoids installing extra packages which aren't needed. Thus minimizing image size! I've already done this for all the current dockerfiles that i adressed in this pull. I do have to catch up with some newer solutions.

But most issues i've come across is people forgetting to install libc-dev or ca-certificates. It's already done but if you guys don't want to have --no-install-recommends for the future, i'll skip those :-).

marghidanu: I can tell you from experience that `—no-install-recommends` causes problems. I have setup Hadolint for this project :)

Macleykun: Aah i see, well if you want, i can remove all the occurances of --no-install-recommends if you want, all the builds do work at the moment (exepct the vlang one)

rbergen: I think removing --no-install-recommends is the way to go, particularly if @marghidanu says so. :) Also, this is one of the issues our hadolint config ignores, so it is again a recommendation that should not be implemented based on the premise that we should only fix issues our hadolint config flags. For one, builds working is one thing, but the solutions also need to work when they are run. I'm guessing you haven't tested that to exhaustion (if you did, I obviously immediately retract that last statement, but the principle of only fixing what's broken remains.) 

Concerning V: if moving up to weekly.2021.38 is the only way to get things working at the current time, then I do think that is the best thing to do. I can't help observing that V is the language that takes the most effort to just keep working.

Macleykun: I though we tested if it works using CI :P (sorry correction!)
will make that change for vlang in a second!
and also remove all the --no-install-recommends

rbergen: CI tests builds, not runs.

Macleykun: Removed --no-install-recommends everywhere and also fixed vlang (tested build, it works)

Macleykun:
The pull is now ready to be reviewed again :-)

rbergen:
I think that in some of yesterday's commits. you explicitly added ca-certificates and libc-dev packages to the apt-get installs of some Dockerfiles, to fix build errors when --no-install-recommends was still in. To keep things clean (as in: only change what's needed) these additions should to be reverted.

Macleykun:
> I think that in some of yesterday's commits. you explicitly added ca-certificates and libc-dev packages to the apt-get installs of some Dockerfiles, to fix build errors when --no-install-recommends was still in. To keep things clean (as in: only change what's needed) these additions should to be reverted.

Checked all my changes and removed what packages I added, which already get installed with recommended (did this on iPad hence the individual commits).

rbergen:
The build of PrimeV/solution_1 fails, again.

Macleykun:
> The build of PrimeV/solution_1 fails, again.

I'm not sure how or why this is happening yet, but setting up my vm, the build worked, i pruned everything and rebuilding also fails on my side.

This error is different, i'll try the older version again and will let you know once i have another update.

Macleykun:
I've managed to fix it again, by using the latest tag of vlang.
To show i've really, really tested this throughout.
Logs of it working after the fix: https://pastebin.com/w5Ny6Xr2
Logs doing a system prune -a and rebuilding completely again: https://pastebin.com/C51BTgHW

I sadly lack the knowledge/experience with vlang/compilers to really determine why this happens, the only thing that i can imagine is that the dependencies from build-base changed and might cause this behavior.

Anyhow, i'm sorry that it has taken this long to get the builds all working, i prefer finishing this pr and then picking up in a new pr to do the last files so that everything is dockerlint compliant (as configured/desired).

rbergen:
I'm comfortable with concluding that you are not at fault here. It seems the relationship between V and its dependencies is extremely fragile, and the build reliability completely horrendous as a consequence. I'm close to concluding we should disable the automated builds of V solutions until the language('s tooling) stabilizes to a point that builds keep working for more than a week.

Macleykun:
> I'm comfortable with concluding that you are not at fault here. It seems the relationship between V and its dependencies is extremely fragile, and the build reliability completely horrendous as a consequence. I'm close to concluding we should disable the automated builds of V solutions until the language('s tooling) stabilizes to a point that builds keep working for more than a week.

One proposal i can do is that the dependencies it relies on, can be a fixed version. I know it's not desirable but that might keep the Vlang more stable. Ofcours it's up to you if you prefer to disable the automation for Vlang.

For the future, the dependencies we installed alongside have the versions:
build-base-0.5-r2
git-2.30.2-r0
bash-5.1.0-r0
The full list is: https://pastebin.com/Q4HNbqxz

I'll hear if any changes have to be made :-)

rbergen:
Maybe pinning down the versions of dependencies is the way to go in this particular case. If you can apply that to the V Dockerfile then we'll see if we can keep it building until we merge this PR, at least.

Macleykun:
> Maybe pinning down the versions of dependencies is the way to go in this particular case. If you can apply that to the V Dockerfile then we'll see if we can keep it building until we merge this PR, at least.

done :-)

rbergen (approved):
LGTM

marghidanu (approved):

PR #715, opened 2021-09-13
Rust solution_1: Small tweaks

mike-barber:
## Description

These are a few small tweaks to improve performance _very_ slightly; just correcting a few small inefficiencies. 
- remove unnecessary restore in the sparse resets
- starting dense resets slightly later when possible
- check and apply sparse resets first -- the nested case statement was causing issues with inlining

I've also corrected the Dockerfile, so it should get rid of one lint complaint too. More importantly, I've also added an explicit `cargo search` command to the build layer to force cargo to download the index from crates.io. If we do the same in the other Rust solutions, this might improve build performance a bit as Docker will locally cache this layer with the already-retrieved index. Doing this before copying stuff into the image makes this stage independent of the specific code that goes into each solution. I think the other Dockerfiles are all based on this one anyway, at least initially.

## Contributing requirements

### DIFF
--- a/PrimeRust/solution_1/Dockerfile
+++ b/PrimeRust/solution_1/Dockerfile
@@ -5,13 +5,17 @@
 
 # Note: Debian based, not Alpine; for some reason, the 
 # Alpine malloc call is slower than it should be.
-#
+# 
+# Also, immediately run `cargo search` to force
+# cargo to load the package index from crates.io.
+# This allows it to be cached as part of the build
+# layer, saving a lot of time on repeated builds.
 FROM rust:1.54 AS build
+RUN cargo search num_cpus -q --limit 1
 
 WORKDIR /app
 COPY . .
-RUN cargo test
-RUN cargo build --release
+RUN cargo test && cargo build --release
 
 # 2. runtime stage
 #   - start from clean container
--- a/PrimeRust/solution_1/prime-sieve-rust/src/main.rs
+++ b/PrimeRust/solution_1/prime-sieve-rust/src/main.rs
@@ -80,13 +80,13 @@ pub mod primes {
     }
 
     /// Recommended start for resetting bits -- at the square of the factor
-    pub fn square_start(skip_factor: usize) -> usize {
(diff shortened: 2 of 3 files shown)


mike-barber:
@Macleykun just a heads up that I'm changing my Dockerfile in PrimeRust/solution_1 -- and it might be a useful trick to apply to the other Rust solutions. It should result in a common, cached, build layer. If some of them are still using the prime images Rust image, this should be a better option. That one was based on Alpine, and I had some performance issues with it when allocating memory. Having said that, I can't authorise changing other developer's Docker files, so we'd need to OK it with them if you think it's a good idea.

Macleykun:
> @Macleykun just a heads up that I'm changing my Dockerfile in PrimeRust/solution_1 -- and it might be a useful trick to apply to the other Rust solutions. It should result in a common, cached, build layer. If some of them are still using the prime images Rust image, this should be a better option. That one was based on Alpine, and I had some performance issues with it when allocating memory. Having said that, I can't authorise changing other developer's Docker files, so we'd need to OK it with them if you think it's a good idea.

Sure thing :) if yours happen to be accepted earlier then mines, i'll take inspiration from yours and fix my pull.
If the other is true, i'll again make a pull to improve what is left to improve with dockerlint.

Interesting to hear that the alpine image is better then the official Rust image.

mike-barber:
> Interesting to hear that the alpine image is better then the official Rust image.

@Macleykun it's not -- the Alpine image doesn't work as well as the normal Rust image. That's why I'm sticking to the official Debian-based one.

Macleykun:
> > Interesting to hear that the alpine image is better then the official Rust image.
> 
> @Macleykun it's not -- the Alpine image doesn't work as well as the normal Rust image. That's why I'm sticking to the official Debian-based one.

sorry your right! Didn't read correctly :P

mike-barber:
> sorry your right! Didn't read correctly :P

@Macleykun no worries! I wrote it quite badly :) 

Thanks for review & merge @rbergen -- much appreciated!

PR #718, opened 2021-09-14
Fix zig docker images

ityonemo:
## Description
This minor patch switches from alpine, which has an imperformant malloc/calloc system, and also removes compilations for arm architecture, which is probably under more memory pressure and is likely to be OOMing during compilation step.

## Contributing requirements

### DIFF
--- a/PrimeZig/solution_3/Dockerfile
+++ b/PrimeZig/solution_3/Dockerfile
@@ -1,9 +1,7 @@
-FROM alpine:3.13 as builder
+FROM debian:buster-slim as builder
 
-RUN apk update && \
-    apk add \
-        curl \
-        xz
+RUN apt-get update && \
+    apt-get install -y curl xz-utils
 
 RUN mkdir -p /deps
 WORKDIR /deps
@@ -11,7 +9,7 @@ RUN curl https://ziglang.org/download/0.8.0/zig-linux-"$(uname -m)"-0.8.0.tar.xz
     tar xf zig-linux-"$(uname -m)"-0.8.0.tar.xz && \
     mv zig-linux-"$(uname -m)"-0.8.0 local/
 
-FROM alpine:3.13
+FROM debian:buster-slim
 COPY --from=builder /deps/local/ /deps/local/
 RUN ln -s /deps/local/zig /usr/bin/zig
 
--- a/PrimeZig/solution_3/src/main.zig
+++ b/PrimeZig/solution_3/src/main.zig
@@ -25,6 +25,8 @@ var scratchpad: [SIZE]u8 align(std.mem.page_size) = undefined;
 
 const ARCH_64 = std.builtin.target.cpu.arch.ptrBitWidth() == 64;
 const ARCH_32 = std.builtin.target.cpu.arch.ptrBitWidth() == 32;
+// arm systems tend to have constrained memory so compile less stuff
+const NOT_ARM = (!std.builtin.target.cpu.arch.isARM());
 
 pub fn main() anyerror!void {
     const run_for = 5; // Seconds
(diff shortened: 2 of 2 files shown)


ityonemo:
also, is there some documentation on how the build process works?  I'm getting 

`ManDeJan&ityonemo&SpexGuy-zig-parallel-gustafson-bitSieve-unrolled-run-u64-find-u32;131433;5.00018;8;faithful=yes,algorithm=base,bits=1` on my
4-core `Intel(R) Core(TM) i5-10210U CPU @ 1.60GHz` and the benchmark 6-core `Core™ i7-8750H @ 2.2` is performing worse at 68728; I'm concerned that the CPU feature detection might not be working at build time if the docker build platform is different than the run platform.

rbergen:
> also, is there some documentation on how the build process works? 

I'm assuming you're referring to the Docker builds that are used in the context of the benchmark runs, which is what the benchmark machines execute daily. The best documentation on this is the source code for the benchmark tool, most specifically [/tools/src/commands/benchmark.ts](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/tools/src/commands/benchmark.ts). The long and short of it is that the tool executes a `docker build` and `docker run` for each solution that includes a Dockerfile. That means that all benchmark images are built on the same machine on which they are run, and any image caching across benchmark runs is done locally. 

> I'm concerned that the CPU feature detection might not be working at build time if the docker build platform is different than the run platform.

Considering the previous, I don't think that's possible. Other differences do and may exist: your 4-core i5 is a 10th gen Intel whereas my 6-core i7 is an 8th-gen, ventilation on your system may be better than on mine (the latter is a laptop, for starters), etc.

ityonemo:
ah ok, thanks!

ityonemo:
we just ran it on i7-8700 one of the teammembers has.  Relative to the i7-8750, the i7-8700 should have lower performance, and we got the following results:

```
ManDeJan&ityonemo&SpexGuy-zig-single-bitSieve-unrolled-run-u64-find-u32;48449;5.00002;1;faithful=yes,algorithm=base,bits=1
ManDeJan&ityonemo&SpexGuy-zig-single-bitSieve-unrolled-run-u64v8h-find-u32;42855;5.00002;1;faithful=yes,algorithm=base,bits=1
ManDeJan&ityonemo&SpexGuy-zig-single-bitSieve-unrolled-run-u64-find-u32-5760of15015;51938;5.00009;1;faithful=yes,algorithm=wheel,bits=1
ManDeJan&ityonemo&SpexGuy-zig-single-inverted-bitSieve-unrolled-run-u64-find-u8-advanced-5760of15015;51481;5.00001;1;faithful=yes,algorithm=wheel,bits=1
ManDeJan&ityonemo&SpexGuy-zig-single-vecSieve;42737;5.00003;1;faithful=yes,algorithm=other,bits=1
ManDeJan&ityonemo&SpexGuy-zig-parallel-gustafson-bitSieve-unrolled-run-u64-find-u32;280275;5.00016;12;faithful=yes,algorithm=base,bits=1
ManDeJan&ityonemo&SpexGuy-zig-parallel-gustafson-inverted-bitSieve-unrolled-run-u64-find-u8-advanced-5760of15015;301319;5.00016;12;faithful=yes,algorithm=wheel,bits=1
ManDeJan&ityonemo&SpexGuy-zig-parallel-gustafson-bitSieve-unrolled-run-u64-find-u32-8of15;272509;5.00015;12;faithful=yes,algorithm=wheel,bits=1
ManDeJan&ityonemo&SpexGuy-zig-parallel-gustafson-bitSieve-unrolled-run-u64-find-u32-48of105;246864;5.00013;12;faithful=yes,algorithm=wheel,bits=1
ManDeJan&ityonemo&SpexGuy-zig-parallel-gustafson-bitSieve-unrolled-run-u64-find-u32-480of1155;246868;5.00009;12;faithful=yes,algorithm=wheel,bits=1
ManDeJan&ityonemo&SpexGuy-zig-parallel-gustafson-bitSieve-unrolled-run-u64-find-u32-5760of15015;232491;5.00019;12;faithful=yes,algorithm=wheel,bits=1
```

ityonemo:
@rbergen is the suite running under docker on linux or docker on windows?  We're able to reproduce this performance regression and it seems that it is happening only when we run docker on windows.

rbergen:
It's running on Linux on bare metal, with stuff that's not needed (like the graphical desktop) disabled.

ityonemo:
thanks!  I remembered had substituted podman for docker on my own dev machine.  It seems that there is a 2x performance regression between docker and podman for reasons I can't fathom, so now I am reproducing this regression on "linux on linux", but I at least have a lead on figuring out how to fix this.
Screenshot at 2021-09-14 17-18-43

PR #719, opened 2021-09-14
COMAL solution that runs within Docker

rbergen:
## Description
This is an adaptation of the COMAL solution submitted by @mkott that runs in Docker.

## Contributing requirements

### DIFF
--- a/PrimeCOMAL/solution_2/Dockerfile
+++ b/PrimeCOMAL/solution_2/Dockerfile
@@ -1,0 +1,12 @@
+FROM primeimages/opencomal:0.3.0
+
+RUN apt-get update && apt-get install -y socat
+
+WORKDIR /opt/app/
+
+COPY opencomal.txt run.sh build.sh .
+
+RUN sh build.sh
+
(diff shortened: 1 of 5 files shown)


marghidanu:
Wow! LGTM

marghidanu (approved):

rbergen:
Thanks, @marghidanu :)

kottm:
This looks good, thanks for doing this! As you say, the timing precision is hindered by SYS(now) only being in whole seconds, but at least COMAL is now in the race, albeit way back near the end.

rbergen:
> This looks good, thanks for doing this! 

Thank you, and you're welcome!

> As you say, the timing precision is hindered by SYS(now) only being in whole seconds, but at least COMAL is now in the race, albeit way back near the end.

To be frank, with the duration of the sieve run being what it is, I don't think the digits to the right of the decimal point matter too much for the solution's ranking. :)
 
From a conceptual standpoint, I just love the fact that we have this "1970s/1980s language" running on technology that is current in 2021, regardless of its place on the leaderboard.

PR #722, opened 2021-09-16
Add APL Language

arcfide:
## Description

This adds a benchmark for the Dyalog APL programming language. Please note that the use of the binary dws is standard practice in APL, and would be the typical "source file". The source code in the .apln file is the canonical source but a copy of that source is kept in the dws format because this is how the interpreter expects the code. 

## Contributing requirements


rbergen:
@arcfide Some time ago, #549 was opened to also add an APL solution based on Dyalog. As the discussion in that PR explains, we can't include a Dockerfile in the repository that pulls in Dyalog for legal reasons. We also really do not merge binary files.

Where that leaves us is that we can't merge this PR in its current form, but we could merge a "showcase" version of your solution. That is, a version that does not include a Dyalog-based Dockerfile, and does not include the dws file. Obviously, this would mean that the solution can't be included in the benchmark runs, or hence the reports that document their results.

arcfide:
I read that, but I'm pretty sure that it's misinterpreting the licensing for Dyalog (I work quite closely with them). However, what if I took the Dockerfile and "disabled" it by renaming the file or the like, with a note in the README.md about how to enable it and the requirements to accept the non-commercial licensing for Dyalog or already having a commercial or academic license? This would prevent people from accidently running the solution and would require that people accept the licensing terms, but it would also ensure that people could easily run the solution without having to invent their own Dockerfile to make things work. I think this would be *strongly* encouraged by Dyalog, as I am actually working with Dyalog as a part of this submission, and this contribution comes directly from work with Dyalog members (including myself, Roger Hui, and Morten Kromberg). 

I can try to find a non-binary solution, but it's significantly less elegant, IMO.

arcfide:
Also, just curious, but what do you do when you have a language that *only* stores its source code in binary format?

arcfide:
Just to clarify, what sort of licensing would allow you to incorporate the solution and include the benchmark numbers?

arcfide:
I have removed the binary dependency, so I believe that all that remains is clarification regarding the Dyalog licensing terms, yes?

rbergen:
> However, what if I took the Dockerfile and "disabled" it by renaming the file or the like, with a note in the README.md about how to enable it and the requirements to accept the non-commercial licensing for Dyalog or already having a commercial or academic license?
> This would prevent people from accidently running the solution and would require that people accept the licensing terms, but it would also ensure that people could easily run the solution without having to invent their own Dockerfile to make things work.

That could work, as far as I'm concerned. The point is that we cannot bring ourselves or others in a position where they either "blindly" accept license terms, or inadvertently and unknowingly break them, by running a benchmark. For disabling the Dockerfile, you can use the respective flag file, [as described in the contributing requirements](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md#disabling-build-and-benchmark).

> I think this would be _strongly_ encouraged by Dyalog, as I am actually working with Dyalog as a part of this submission, and this contribution comes directly from work with Dyalog members (including myself, Roger Hui, and Morten Kromberg).

I appreciate that, but Dyalog are as of yet not a party in this conversation, or any conversation with this project's maintainers, as far as I'm aware.

> I can try to find a non-binary solution, but it's significantly less elegant, IMO.

Duly noted.

> Also, just curious, but what do you do when you have a language that only stores its source code in binary format?

Then it can't be submitted for inclusion in the automated benchmark, and instructions would have to explain how to enter the textual representation of the source code into the language's environment in question. This came up in #50 as well; that solution ended up being added as a "plain" Lua solution, in the end.  

> Just to clarify, what sort of licensing would allow you to incorporate the solution and include the benchmark numbers?

As a premise, compatible with BSD-3 or more permissive towards the licensee. 

Just for sake of completeness: I take no issue with Dyalog's license as such, and I understand that commercial licenses are effectively inevitable if one wants to run a business. We just can't accept them on other people's behalf. I also don't think that the grantors of such licenses would want us to do so.

arcfide:
> Dyalog are as of yet not a party in this conversation, or any conversation with this project's maintainers, as far as I'm aware.

Sorry, I probably wasn't clear. I essentially am the "Dyalog rep" on this right now, though I can have Morten (CTO/CXO) weigh in directly on this with his Github account if you need him to do so. If you need more people to weigh in from the Dyalog side we can do that, but no one else has weighed in because they're essentially letting me do that work at the moment. 

> > Just to clarify, what sort of licensing would allow you to incorporate the solution and include the benchmark numbers?
> 
> As a premise, compatible with BSD-3 or more permissive towards the licensee.

It's my understanding that we *are* compatible with BSD-3, in that you are perfectly capable of running our code within a BSD-3 codebase without violating BSD-3. Moreover, running benchmarks using Dyalog would not violate the Dyalog licensing. 

> I also don't think that the grantors of such licenses would want us to do so.

In this case, they explicitly do want to encourage people to experiment with Dyalog, and this is a perfect example. 

If a "more senior" Dyalog representative were to provide some assertion on this thread, would that be sufficient? What if they extended the licensing to say something to this effect?

    Non-commercial use explicitly includes distribution of Docker Containers created to demonstrate 
    Dyalog APL, or to deliver solutions to programming competitions.

rbergen:
If Dyalog is of the opinion that we can explore a licensing solution that addresses the concerns I've described, then I think that a) that's great news, and b) it's best to have a conversation outside of the context of this PR. That would indeed have to involve formal/legal representation from Dyalog, and others besides myself at the side of this project. The conclusion of that conversation could then be used to bring this PR forward. Alternatively, we can merge this PR "now" with the Dockerfile disabled, and change this if and when that is opportune.

arcfide:
I think we can merge now with the dockerfile disabled for the moment and then hopefully get this resolved further OOB if you think that's best.

rbergen:
As I've said before, I'd love to have APL included in the project in some form, so if you ask me, I would like to get there as soon as possible. 

That said, I really will happily engage in an OOB conversation aiming to include the language in the benchmark runs, as well.

arcfide:
Alright, I've disabled the build using the `build-no` flag file as described in the document you linked. I'd really love to get APL in the benchmarks as soon as possible, so what would you need on your end to get the ball rolling on that? I believe that the PR is safe to merge as is now based on what you've described above. If you would like, I can open a separate PR specifically as a point of reference for enabling the APL build after this PR is merged if that works for you.

rbergen:
Thank you. I'll then proceed with the "regular" review of this solution.

Concerning what's needed on the other topic, I'd suggest that someone who can legally represent Dyalog contacts me via email on rbergen (at) xs4all dot nl, so we can get that conversation going.

PrimeAPL/solution_1/PrimeSieveAPL.apln
```diff
@@ -0,0 +1,38 @@
...
+	warmup←{expected≠+⌿sieve sieveSize:'BAD SIEVE'⎕SIGNAL 99 ⋄ 0}
+
+	runOne←{
+		passes←0 ⋄ dur←0 ⋄ start←⎕AI[2]
+		msk←{z←sieve sieveSize ⋄ passes+←1 ⋄ z}⍣{5000<dur∘←⎕AI[2]-start}⍬
+		passes dur}
+
+	print←{
+		passes dur threads←⍵ ⋄ tag←'arcfide','Single' 'Multi'⊃⍨threads>1
+		flags←'algorithm=wheel,faithful=yes,bits=1'
```
rbergen: The `faithful` qualification comes with a few specific requirements. Could you please talk me through how the following one is implemented in this solution? 
> _It uses a class to encapsulate the sieve, or a (closest) equivalent feature in your language if a class construct is not available. This class must contain the full state of the sieve. Each iteration should re-create a new instance of this class from scratch._ 

I'm afraid that my unfamiliarity with APL('s syntax) makes that I can't deduce this from the source code, myself.

arcfide: Sure, I understand. :-) The sieve buffer `b` is created on line 13 and initialized with a wheel factorization style algorithm before proceeding to the main recursion in line 14, which zeros out the buffer on each iteration according to the factors.  Each call to `sieve` allocates a new `b` buffer. Idiomatic APL code tends to be done in a more functional style, so the buffer would typically be allocated afresh inside of the sieve function rather than outside of it and referenced as a free variable. The core sieve algorithm and code is really just lines 13 and 14. Line 9 is an early termination check for small inputs (we could probably remove that for this benchmark), while lines 10 through 12 setup the basic limits like the terminating limit `r` and the set of initial primes for wheel factorization in `p`. All the other lines are just configuration code to run the benchmark, collect timings, and print the results in accordance with the format of the benchmark. We check that sieve is giving us the correct results during `warmup` with the values provided by `expected` against the warmup `sieve` result.

rbergen: Thank you. Based on your explanation, I'm still not sure how the sieve encapsulation in a class-like construct is implemented. You also say that the buffer is "zeroed out" on each iteration, which does not really sound like the sieve structure is recreated from scratch, as the rules require. Could you expand a little on these specific points?

As you've used the word "idiomatic" in your explanation, I'd like to note pre-emptively and potentially unnecessarily that idiomaticity is not a reason to deviate from the requirements for a faithful qualification. This application of the rules has been done consistently throughout the project (admittedly to the frustration of a number of earlier contributors).

arcfide: Sorry, confusing use of iteration there. The iterations that are counted in the benchmark each individually create a fresh buffer structure from scratch and each is separately initialized, including all the sieve data, the buffer, wheel factors, and other data. What I meant above was that the buffer was mutated *within* an individual sieve computation per each prime factor. 

I think I'm possibly confused by your confusion regarding sieve encapsulation now. The `sieve` function creates and initializes fresh state that is encapsulated within that invocation, and line 14 does the sieve calculation. The state is fully encapsulated and initialized afresh each time you call `sieve`. Am I missing something?

rbergen: What you're describing is that the sieve state and its execution are encapsulated within one function. That's not the same as creating (and destructing) an instance of a class-like construct. I took another look at the code in #549, and that solution does seem to create a sieve class in the PrimeSieve.aplc file. That approach would certainly be a faithful one.

I'll be frank: as your implementation is a `wheel` algorithm and as such already not a contender for the "main" leaderboard, you may just choose not to care about faithfulness, and mark it as `faithful=no` in the output. That change would make the entire aspect irrelevant.

arcfide: I'll take that under consideration. I'll need to think about it. I'm a bit confused though still, what constitutes a class-like object in languages that don't inherently have classes? Do you need to construct an entire class system on top of your language to be faithful? The C implementations marked faithful seem to do the same thing that I'm doing here. At any rate, if I wanted to submit a solution in the pure dfns dialect of APL (which doesn't have classes), what would I do?

rbergen: It depends on the language. In C, we've settled on the use of `struct` as being sufficiently "class-like". Concerning quite a few other languages, the conclusion has indeed been that writing faithful implementations is practically impossible. Solutions in those languages are therefore indeed all marked as "not faithful". 

If I'm honest, I think part of the problem may be related to "real-world semantics", where (un)faithfulness mostly constitutes a value statement. That's not the case here: it really is just a record of whether an implementation employs certain constructs/paradigms, or not. 

Concerning your question: I would indeed submit your solution in its purest form and not care about the fact that `faithful=no` is the correct faithfulness flag for that solution.

arcfide: I've committed an update that removes the faithfulness=yes flag, but I still want to submit a faithful implementation in pure dfns that doesn't use wheel factorization. Given that the above implementation already does in fact create and destruct the sieve state on each instance of `sieve`, what would I need to do to count as a class-like construct given that dfns doesn't have classes? What makes it a class-like construct?

PrimeAPL/solution_1/README.md
```diff
@@ -1,7 +1,7 @@
 # APL solution by arcfide
 
 ![Algorithm](https://img.shields.io/badge/Algorithm-wheel-yellowgreen)
-![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-no-green)
```
rbergen: Please change the colour of the Faithfulness badge from `-green` to `-yellowgreen`.

PrimeAPL/solution_1/PrimeSieveAPL.apln
rbergen: Thank you. Concerning dfns, the situation may simply be that it's not possible to write a faithful implementation if no class-like constructs are available. As I mentioned in an earlier comment, it would most certainly not be the first time that this is in fact the end conclusion. 

If you really do want to pursue a faithful implementation, then your implementation should include a proposal for the class-like construct you'd like to use. As indicated in the definition of "faithfulness" within the contributing guidelines, key criteria are that the construct's instances must contain the full state of the sieve, and each sieve run iteration should re-create a new instance of from scratch. (In most cases, the instances contain at least the sieve items in an array, bit set or something similar, as well as a variable that holds the sieve size.)

PrimeAPL/solution_1/README.md
arcfide: Ah, thanks. Done.

PrimeAPL/solution_1/PrimeSieveAPL.apln
arcfide: I think the problem I'm having understanding the criteria is what you mean by "construct". I had thought that I had a "construct" that was created from scratch on each iteration in this solution, but clearly you don't think so, so I think if I understood what makes something a construct versus doing everything that a class does mechanically (in terms of memory allocation, destruction, &c.) but still not being a construct, I can make a proposal. At any rate, I think maybe it would be worth it to just merge this solution and we can maybe start a new PR around trying to get a faithful version up and running?

rbergen: Merging this solution is exactly what I was aiming to move forward with. The tiny thing that was blocking that (the badge colour) is now fixed (thank you for that), so I think we're good to continue on this one.

arcfide: Great, I'll submit a new PR once this one is merged to deal with alternative approaches.

rbergen:
Concerning the definition of "class-like construct": I think the most relevant aspect is that the "construct" is an entity that itself uses memory, and encapsulates the actual variables and data structures that contain the information required for the sieve to run. This means that the encapsulating entity itself needs to be created and torn down (in memory) with every iteration.

rbergen (approved):
LGTM

marghidanu (approved):

PR #727, opened 2021-09-19
New Verliog implementation using Verilator to run as simulation on PC

alwayslinux2:
## Description
Verilog implementation for the Prime Number Drag Race

## Contributing requirements


rbergen:
Thanks for submitting this, I think Verilog would certainly be an interesting language to add.

As the moment, the solution is labeled as "faithful", which means a number of specific requirements apply. One of them is that _"It uses a class to encapsulate the sieve, or a (closest) equivalent feature in your language if a class construct is not available. This class must contain the full state of the sieve. Each iteration should re-create a new instance of this class from scratch."_

In this case I think the main module in `prime_number_core.v` could serve as the class-like construct (the `new Vprime_number_core` at line 6 of `prime_number_core_tb.cpp` basically illustrates that it is), but that is currently not recreated for each iteration. 
I therefore think that either the creation of `Vprime_number_core` needs to be pulled into the timed loop, or the solution needs to be labeled as not faithful.

alwayslinux2:
@rbergen Fair enough, I have pushed another update to do as you requested. The entire Vprime_number_core instance gets deleted and re-created each time. I updated the output in the Readme as well, since this does affect the runtime.

As an aside, some of this is follow on to an instructible I did a couple years ago https://www.instructables.com/Arduino-As-a-Cheap-Breadboard-FPGA/    using verilator to go from Verilog to arduino.

rbergen:
@alwayslinux2 Thank you. Docker linting found some issues with your Dockerfile. Please refer to the "Files changed" tab for details.

alwayslinux2:
@rbergen Updated Dockerfile to make hadolint happy.
Suggest contibuting.md be updated to point out that this hadolint is going to be run, and how to run this locally before commit. These aren't warnings that docker itself triggers.

rbergen:
@alwayslinux2 Thanks for addressing this so quickly.
That's good feedback/a good suggestion regarding CONTRIBUTING.md, but it may be a while until one of the other maintainers or I get around to it. We could definitely consider a PR that adds this before we do. The pointer I can easily give, if only for any future submissions by yourself, is that the hadolint configuration we use is in [config/hadolint.yml](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/config/hadolint.yml).

PrimeVerilog/solution_1/README.md
```diff
@@ -0,0 +1,26 @@
+# Verilog solution by alwayslinux2
+
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-1-green)
```
rbergen, line 4: Just for completion, I think it would be good to also add algorithm and faithfulness badges:
```markdown
```

rbergen (commented):
Only one tiny documentation-related thing left, as far as I'm concerned.

rbergen (approved):
LGTM

rbergen:
Regarding some text on hadolint in the contributing guidelines, I've opened #732 to add this.

marghidanu (approved):

PR #728, opened 2021-09-21
CONTRIBUTING.md: make link with C++ sample point to a simpler version of PrimeCPP.cpp

mayerrobert:
## Description

CONTRIBUTING.md contains a link to a C++ solution that can be used as a reference. After some optimizations to PrimeCPP.cpp is faster but less readable than it originally was. This PR make the link point to an older version that IMO works better as an easily readable reference implementation.

Please decide for yourself if you want to merge this PR as is, or reject this PR leaving things as is, or do something else such as using an even better reference (maybe into the original branch?), I'm ok either way.

## Contributing requirements

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -142,7 +142,7 @@ The base algorithm is defined as follows:
 
 * Starting the clearing loop at factor * factor (instead of 3 * factor, as done in the original implementation) is permissible.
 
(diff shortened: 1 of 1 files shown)


rbergen:
I think this is a good change, thank you for proposing it. It's basically a realignment to what the link claims to point to, that being the original implementation in C++, which the current version no longer is.

PR #729, opened 2021-09-22
Clipper Implementation running on XHarbour

BradleyChatha:
## Description
Not much to say here, I'm handling this PR on Andy's behalf.

## Contributing requirements


PrimeClipper/solution_1/README.md
```diff
@@ -0,0 +1,47 @@
+# Clipper Solution by Andy Radford
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
```
rbergen: You're labeling this solution as "faithful", which means it must comply with a number of specific rules. Quoting from the contributing guidelines, one of them is that _"It uses a class to encapsulate the sieve, or a (closest) equivalent feature in your language if a class construct is not available. This class must contain the full state of the sieve. Each iteration should re-create a new instance of this class from scratch."_

This solution does not do this, and can therefore not be marked as faithful. I'd love to be proven wrong about this, but I actually think it's most likely that Clipper simply doesn't include the language features to make a faithful implementation possible.

From the perspective of the "official leaderboard" it doesn't matter either, due to the bit count per prime being >1. Which...

PrimeClipper/solution_1/README.md
```diff
@@ -0,0 +1,47 @@
...
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-112-yellowgreen)
```
rbergen, line 6: ...brings me to my second point, which is that 112 bits per prime is a very unusual number indeed. Primarily out of curiosity, I was wondering if you could maybe explain how you get to this specific number?

PrimeClipper/solution_1/README.md
BradleyChatha: No problem with that. I thought the rules were a bit more lax regarding languages that don't have classes/structs. We tried to emulate this by recreating the array each time.

PrimeClipper/solution_1/README.md
BradleyChatha: I can't quite remember where to get the reference for it, but Clipper uses 14 bytes per record in an array, hence 112 bits.

Very weird, I know. You'll love when you see the second implementation he's working on which is something like 39 bytes per record.

BradleyChatha: https://www.itlnet.net/programming/program/Reference/ierror/ng1de8.html

`Because every array element in Clipper requires 14 bytes`

PrimeClipper/solution_1/README.md
rbergen: Thank you. And no, based on earlier conversations on the topic (and I've had quite a few) we've settled on the "no exceptions" approach concerning the rules. It keeps things clear, which makes it much easier for the maintainers to keep things consistent.

PrimeClipper/solution_1/README.md
rbergen: That's bizarre in a way that I absolutely love. 112 bits it is, then!

rbergen (commented):
Dockerfile linting found one issue that I think would be best taken care of before we merge.

Furthermore, when running the solution locally, the build fails with the following error:
```
 > [5/5] RUN hbmk2 -gtstd -optim -cflag=-O3 -cflag=-mtune=native ./sieve.prg:
#10 0.602 Harbour 3.2.0dev (r1712151853)
#10 0.602 Copyright (c) 1999-2016, http://harbour-project.org/
#10 0.602 Compiling './sieve.prg'...
Lines 259, Functions/Procedures 7
#10 0.603 Generating C source output to '/tmp/hbmk_S3UIwN.dir/sieve.c'... Done.
#10 0.680 cc1: error: bad value ('tigerlake') for '-mtune=' switch
#10 0.680 cc1: note: valid arguments to '-mtune=' switch are: nocona core2 nehalem corei7 westmere sandybridge corei7-avx ivybridge core-avx-i haswell core-avx2 broadwell skylake skylake-avx512 cannonlake icelake-client icelake-server cascadelake bonnell atom silvermont slm goldmont goldmont-plus tremont knl knm intel x86-64 eden-x2 nano nano-1000 nano-2000 nano-3000 nano-x2 eden-x4 nano-x4 k8 k8-sse3 opteron opteron-sse3 athlon64 athlon64-sse3 athlon-fx amdfam10 barcelona bdver1 bdver2 bdver3 bdver4 znver1 znver2 btver1 btver2 generic native
#10 0.683 cc1: error: bad value ('tigerlake') for '-mtune=' switch
```

Long story short, if you're planning to use `-mtune=native`, then I think it would be wise to use a more recent version of GCC so that newer CPU architectures are also recognized. 
You can either explicitly install and invoke gcc-10 on Ubuntu 20.04, or use a more recent Ubuntu version that uses gcc-10 by default (both 20.10 and 21.04 do).

BradleyChatha:
Could also just get rid of the flag since it's slow as hell either way :o)

rbergen:
You know, that's a really good point. Maybe that's why I said "if you're planning to use..." :o)

rbergen (approved):
LGTM

marghidanu (approved):

PR #730, opened 2021-09-22
Add Faithful APL implementations

arcfide:
## Description

This PR updates the APL solution with two new faithful implementations using two different dialects of the language. The two implementations both create a separate explicitly reified state construct that contains all sieve state, and then updates that state to compute the sieve before returning the sieve mask and destroying the construct on each iteration. 

## Contributing requirements


arcfide:
I'm not sure I understand why the commit history here is so long, given that I merged the latest drag-race branch before I made the PR. I'm not sure if that matters or how to fix it.

rbergen:
> I'm not sure I understand why the commit history here is so long, given that I merged the latest drag-race branch before I made the PR. I'm not sure if that matters or how to fix it.

It's because a) you continued development on the fork/branch you created when you started work on the original APL implementation and b) we squash commits on merge. The PR diff will only show the actual file-level changes between your branch and what is currently in the repo, but the commit history will show everything that happened since the branch-off took place.
The easiest way to fix it is to create a separate (feature) branch from this repo for every PR you intend to open. The current list of 18 commits is still manageable, so I wouldn't worry too much about it in this particular instance.

rbergen (approved):
LGTM

marghidanu (approved):

PR #731, opened 2021-09-23
Nim solution_2 code adjustment/improvement...

GordonBGood:
## Description

This code completes all five Techniques of using the methods of bit twiddling, stride8, stride8-block16K, extreme loop unrolling, and extreme hybrid (which combines extreme loop unrolling for larger base prime values and dense marking for smaller values, with the threshold used here as 129).  It also adds a sleep/wait of one second before each test to allow the CPU to cool down in the case of thermal throttling.

## Contributing requirements


develooper1994:
I get small performance boost with cpp and orc flags

nim cpp -d:release --os:linux -f --passC:"-flto" --implicitStatic:on --opt:speed --app:console --threads:on --benchmarkVM:on --tlsEmulation:on --seqsv2:on --deepcopy:off --hints:on --objChecks:off --fieldChecks:off --rangeChecks:off --boundChecks:off --overflowChecks:off --floatChecks:off --nanChecks:off --infChecks:off --refChecks:off --gc:orc -d:danger -t:-march=native -d:lto Primes.nim

GordonBGood:
There are a few comments that need to be made on your choice of Nim
compiler command line flags as follows:

1. `-d:danger` implies release and all checks turned off, so they are
redundant.
2. Threadvar's and threading would only apply to multi-threading, which
this implementation does not do.
3. You are turning on Link Time Optimization (LTO) in two different ways,
and as this code doesn't call other modules in time critical code, it is
not likely to be improved much if at all by LTO.
4. Since you are still using GCC to compile the code and since no features
such as exceptions are used that could be improved by being compiled as C++
rather than through C, it is fairly unlikely that explains any improvement
in performance.
5. Remaining is ARC versus ORC, which could explain a small difference.
ARC is fully deterministic which means that memory is immediately
deallocated when the reference count drops to zero, whereas ORC is not
fully deterministic unless one calls a `GC_FullCollect` when the references
go out of scope.  That means that all memory allocated may not yet be
deallocated when the time measurement is made, and such deallocation has a
small but measurable cost in time, especially when about ten thousand
allocations are made per second.  It's a bit like the non-deterministic
performance of regular GC unless one can call a full collect that
guarantees the return of memory allocation to the original state within the
timing span.
6. Thus, I think it likely that your small performance gain is likely due
to this indeterministic behavior and not "real".

OTOH, using Clang/LLVM with exactly the right set of optimization options
can give a real about two percent performance gain, but as it is an onerous
pain to apply "just the right optimization options", I didn't bother, as my
Nim solution is already the leader in the Race.

Regards, Gordon

PR #734, opened 2021-09-26
Improve Haskell solution_2...

GordonBGood:
## Description

This new version adds the same "extreme loop unrolling" and "extreme hybrid" using dense base base primes composite marking techniques as used in the current Nim sollution_3.

With Nim and Rust, the compiler emitted code is further optimized by the compiler back-ends, whether GCC as used currently in the Nim solution or LLVM as used by Rust, to auto vectorize the "extreme hybrid" solutions to use SSE/AVX/AVX2 128 and 256 bit registers when available on the host CPU; the back-ends also automatically convert the single bit-by-bit markings into multiple marking word masks.  These compiler generated optimizations are part of the reason that these techniques are so fast with Nim and Rust, gaining up to about 30 percent in speed due to these optimizations.

Even though the code algorithm is exactly the same as used in the Nim algorithm, the GHC Haskell compiler is not emitting LLVM IR code that the LLVM optimizer can recognize so as to do the automatic high bit width vector operations; however, GHC Haskell does have low-level primitive operations allowing the code to specify the length of vector operations used as intermediate storage within this algorithm.  This version switches out the usual low level primitive operations that are directed to the normal 64-bit registers for the equivalent using the higher bit width vector operations when they are available for the host CPU, and provides a means to specify the register bit width to use according to the detected host CPU's capability.

I believe that this is still `faithful to base` as it still marks the composite number representation bits on a bit-by-bit basis (as can be verified by expanding the macros that generate the code), just that the results are loaded and stored to memory using the wider available bus width SIMD operations to make the reading and writing to memory faster and thereby gain a considerable percentage in performance.

The generation of the Docker image file has also been changed to be based on Debian:11 and to use the distribution's latest version of the GHC compiler (version 8.8.4) and LLVM (version 11) so that the code can be run on both the ARM64 and AMD64 targets for the leader board.  Especially the "bit twiddle" technique but also the "stride8" techniques don't run quite as fast as they do on later versions of these, but the closer to "bare metal" techniques of "extreme loop unrolling" and "extreme hybrid" run well as they aren't as dependent on GHC Haskell compiler optimizations or even LLVM optimizations.  The distribution versions of these are used to avoid building these two applications from source which can take hours on the Raspberry Pi platform.  If the Race Committee would care to generate Docker Hub images of these, then that build time wouldn't be a concern as it would be done once and made available to Pull's.  I am not too much concerned if current build times are found acceptable, as the gains are in the slower of the techniques and not the "Top Fuel" ones.

## Contributing requirements


rbergen:
We will have to create/use a Haskell base image for this. Haskell and LLVM pull in a lot of dependencies, and that combined with the long compile times of the actual programs makes the build take too long.

GordonBGood:
@rbergen:

> We will have to create/use a Haskell base image for this. Haskell and LLVM pull in a lot of dependencies, and that combined with the long compile times of the actual programs makes the build take too long.

Yes, I expected as much:  Previous Haskell solutions pulled in a Haskell Docker image to avoid having to build GHC Haskell from source, which takes an hour or more, but still built LLVM from source which still takes several minutes on a fast machine; however, the official Haskell Docker images only support amd64 and I would like to try to run this on arm64 as well.

The PR'ed Dockerfile takes about three to four minutes to install GHC on Debian:11 including all the dependencies, plus another over a minute to compile the application when starting from bootstrap - installing just as required by the distro repos takes a few minutes on a fast machine; I thought you might almost find that acceptable, but probably not.

Mike Barber ran this Dockerfile for me on a fast AMD desktop to install GHC and LLVM in just a couple of minutes, but he found it took about twenty minutes to install them and compile the application on his Raspberry Pi 4.  Yes, it is expected that we need a custom Docker Hub image in order to support both amd64 and arm64 and to reduce these installation times for the two platforms.

So how do we proceed?  Can we do as before and have you generate the primeimages/haskell:8.4.4 Docker images that are required, or do you want me to try to do it?  Mike and I have tried to construct this Dockerfile to make it easy to convert.

If it isn't too much trouble, as we have proceeded on other languages such as Nim and Chapel was easy (at least for me ;-) ) as I just had to change my Dockerfile to pull in your custom Docker image for GHC Haskell and remove the commands to install the dependencies before building the application.  Do you have the time?  If not, try to give me instructions on how to do it myself.

Regards, Gordon

rbergen:
@GordonBGood I've just pushed primeimages/haskell:8.8.4, backed by [PlummersSoftwareLLC/Haskell-docker](https://github.com/PlummersSoftwareLLC/Haskell-docker).

GordonBGood:
@rbergen:

> @GordonBGood I've just pushed primeimages/haskell:8.8.4, backed by [PlummersSoftwareLLC/Haskell-docker](https://github.com/PlummersSoftwareLLC/Haskell-docker).

Thanks for that, I changed my Dockerfile to use the above Docker image, but I thought that you would have installed LLVM into the Docker image as well?  The majority of time is saved in not having to build/install GHC Haskell, but there are still a few tens of seconds used to install LLVM as required by the application, and indeed by any application whose goal is to run at the fastest speed..  The current Dockerfile I am testing with is as follows:

```
FROM primeimages/haskell:8.8.4 AS build

# install llvm
RUN apt-get -y update && apt-get install -y llvm 

# build the app
WORKDIR /opt/sieve
COPY Primes*.hs ./
COPY build.sh .
Run sh build.sh

# runtime container
FROM debian:11 AS runtime
WORKDIR /opt/sieve
COPY --from=build /opt/sieve/Primes Primes
ENTRYPOINT [ "/opt/sieve/Primes" ]
```
If LLVM were installed in the "primeimages/haskell:8.8.4" Docker image, then lines three, four, and five would not be necessary and everything would be loaded on the first pull other than the application source files.

Should I commit the Dockerfile as is or do you think that the "primeimages/haskell:8.8.4" Docker image should include LLVM?

Regards, Gordon

rbergen:
@GordonBGood LLVM is (or definitely should be) installed in the image already, as shown in [the Haskell image Dockerfile](https://github.com/PlummersSoftwareLLC/Haskell-docker/blob/master/Dockerfile) and documented in [the image's README](https://github.com/PlummersSoftwareLLC/Haskell-docker/blob/master/README.md).

GordonBGood:
@rbergen:

> @GordonBGood LLVM is (or definitely should be) installed in the image already, as shown in [the Haskell image Dockerfile](https://github.com/PlummersSoftwareLLC/Haskell-docker/blob/master/Dockerfile) and documented in [the image's README](https://github.com/PlummersSoftwareLLC/Haskell-docker/blob/master/README.md).

Yes, according to the above "README.md" file, LLVM is said to be installed; however, when I run the above Dockerfile with line four commented out that installs LLVM, I get the following:

```
[Gordon@localhost solution_2]$ docker build -t primesh .
Sending build context to Docker daemon   3.89MB
Step 1/9 : FROM primeimages/haskell:8.8.4 AS build
 ---> 953bee3fdb4c
Step 2/9 : WORKDIR /opt/sieve
 ---> Using cache
 ---> eab02a999a54
Step 3/9 : COPY Primes*.hs ./
 ---> Using cache
 ---> 59575b5b21e7
Step 4/9 : COPY build.sh .
 ---> Using cache
 ---> 51474fb81cfa
Step 5/9 : Run sh build.sh
 ---> Running in 57e86d8ced0e

<no location info>: error:
    Warning: Couldn't figure out LLVM version!
             Make sure you have installed LLVM 11

<no location info>: error:
    Warning: Couldn't figure out LLVM version!
             Make sure you have installed LLVM 11
[1 of 3] Compiling PrimesTH         ( PrimesTH.hs, PrimesTH.o )

<no location info>: error:
    Warning: Couldn't figure out LLVM version!
             Make sure you have installed LLVM 11
ghc: could not execute: opt
The command '/bin/sh -c sh build.sh' returned a non-zero code: 1
[Gordon@localhost solution_2]$ 
```

With that line that installing LLVM included, everything works as it should.

My conclusion:  LLVM somehow did not get fully installed in the "primeimages/haskell:8.8.4" Docker image...

The output when installing LLVM to a fresh "primeimages/haskell:8.8.4" image is as follows:

```
[Gordon@localhost solution_2]$ docker run --rm -it primeimages/haskell:8.8.4 bash
root@f5068b2d26eb:/# apt-get install llvm
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following additional packages will be installed:
  llvm-runtime
The following NEW packages will be installed:
  llvm llvm-runtime
0 upgraded, 2 newly installed, 0 to remove and 1 not upgraded.
Need to get 19.2 kB of archives.
After this operation, 106 kB of additional disk space will be used.
Do you want to continue? [Y/n] n
```
which shows that `llvm` and its dependency `llvm-runtime` are not installed.

**EDIT_ADD:**  I think I see the problem:  In your Dockerfile that builds the image you install `llvm-11` but it seems that the debian distro mistakenly doesn't include `llvm` as a dependency, which would then also install `llvm-runtime`.  It seems that one needs to install both `llvm` and `llvm-11` in order to get something can can be run.  The only reason I can think of that debian has done it this way is that if their default stable LLVM version is eventually upgraded to a new `llvm` version such as version 12 but one has software that isn't compatible with the new version, one can downgrade back to version 11 by installing `llvm-11` on top.

As GHC Haskell is moving to a new LLVM support system where instead of specifying an exact required version, they require a specific version or newer, and most applications written in GHC Haskell will possibly run faster with a new version, there is no reason not to just install as `llvm` other than you might be concerned with consistency of run times.  GHC Haskell version 8.8.4 as installed in the Docker image officially supports LLVM up to version 11, but runs without problems using LLVM version 12, and some techniques used in this benchmark run faster under 12 than 11.

Since you need to correct the build Dockerfile anyway, you might consider using the following lines to install LLVM version 12 instead (new stable 13 is also available, but I haven't tested it):

```
# install LLVM 12
RUN apt-get -y update && apt-get install -y wget lsb-release software-properties-common gnupg && \
    wget https://apt.llvm.org/llvm.sh && chmod +x llvm.sh && ./llvm.sh 12
```
There will be some warnings during compilation of the benchmark application that it officially supports only up to veersion 11, but it proceeds anyway and everything in this benchmark runs fine and faster.

rbergen:
@GordonBGood Thanks for looking into this and figuring out what's going on. Considering that the Haskell image repo is a public one, may I suggest you submit your suggested changes as a PR on it?

GordonBGood:
@rbergen:

> @GordonBGood Thanks for looking into this and figuring out what's going on. Considering that the Haskell image repo is a public one, may I suggest you submit your suggested changes as a PR on it?

I made a PR on the Haskell image repo, also changing to fully installing stable LLVM 12 which makes my benchmark run faster; if you don't want to instal LLVM 12 but rather stick to LLVM 11; there are very minor changes to make it do that.

rbergen:
@GordonBGood After merging the PR on Haskell-docker, I found that the arm64 docker buildx fails. The tail of the log looks as follows:
```
#8 441.0 + apt-get install -y clang-12 lldb-12 lld-12 clangd-12
#8 441.3 Reading package lists...
#8 445.7 Building dependency tree...
#8 446.7 Reading state information...
#8 446.8 Package clang-12 is not available, but is referred to by another package.
#8 446.8 This may mean that the package is missing, has been obsoleted, or
#8 446.8 is only available from another source
#8 446.8
#8 447.2 E: Package 'clang-12' has no installation candidate
#8 447.2 E: Unable to locate package lldb-12
#8 447.2 E: Unable to locate package lld-12
#8 447.2 E: Unable to locate package clangd-12
------
Dockerfile:4
--------------------
   3 |     # Install GHC Haskell 8.8.4 and LLVM 12...
   4 | >>> RUN apt-get -y update && apt-get install -y ghc=8.8.4-2 wget \
   5 | >>>                            lsb-release software-properties-common gnupg && \
   6 | >>>     wget https://apt.llvm.org/llvm.sh && chmod +x llvm.sh && ./llvm.sh 12
   7 |     ENV PATH=/usr/lib/llvm-12/bin:$PATH
--------------------
error: failed to solve: executor failed running [/bin/sh -c apt-get -y update && apt-get install -y ghc=8.8.4-2 wget                            lsb-release software-properties-common gnupg &&     wget https://apt.llvm.org/llvm.sh && chmod +x llvm.sh && ./llvm.sh 12]: exit code: 100
```

I think this means we have two options: supporting arm64 and as a consequence, sticking to LLVM 11 for now, or choosing LLVM 12 and dropping arm64 support for now. From a Primes project perspective we strongly prefer arm64 support over a higher-version build toolchain, but I am happy to take your input on this. In either case, the primeimages/haskell images haven't yet been updated due to this build failure.

GordonBGood:
@rbergen:

I'm sorry I couldn't test my PR on ARM64 as I have no ARM64/Linux (Raspberry Pi 4):

> @GordonBGood After merging the PR on Haskell-docker, I found that the arm64 docker buildx fails. The tail of the log looks as follows:
> 
> ```
> #8 446.8 Package clang-12 is not available, but is referred to by another package.
> ...
> ------
> Dockerfile:4
> --------------------
>    3 |     # Install GHC Haskell 8.8.4 and LLVM 12...
>    4 | >>> RUN apt-get -y update && apt-get install -y ghc=8.8.4-2 wget \
>    5 | >>>                            lsb-release software-properties-common gnupg && \
>    6 | >>>     wget https://apt.llvm.org/llvm.sh && chmod +x llvm.sh && ./llvm.sh 12
>    7 |     ENV PATH=/usr/lib/llvm-12/bin:$PATH
> --------------------
> error: failed to solve: executor failed running [/bin/sh -c apt-get -y update && apt-get install -y ghc=8.8.4-2 wget                            lsb-release software-properties-common gnupg &&     wget https://apt.llvm.org/llvm.sh && chmod +x llvm.sh && ./llvm.sh 12]: exit code: 100
> ```

It looks like the failure is related to that there is no Clang 12 available for Debian, which ARM requires for its build tool chain as GCC isn't supported.  If that is the case, just changing the "12" to "11" in both the install run line as above and tne `ENV` path setting line should work.

> I think this means we have two options: supporting arm64 and as a consequence, sticking to LLVM 11 for now, or choosing LLVM 12 and dropping arm64 support for now. From a Primes project perspective we strongly prefer arm64 support over a higher-version build toolchain, but I am happy to take your input on this. In either case, the primeimages/haskell images haven't yet been updated due to this build failure.

I agree with the general Primes project perspective of wanting to support ARM64 if possible, which is why I have gone through these hoops rather than just use the GHC Haskell Docker Hub image as Haskell/solution_1 does, which then also can use the latest versions of GHC Haskell but which then only runs on `x86_64`.  However, there may be some other options, as follows:

1. I suppose that rather than use Debian as a base, we could use latest stable Fedora:34 which does support both ARM64 and AMD64 and whose distro's repos have both this same GHC version 8.8.4 and LLVM 12 available for immediate installation without having to go through the LLVM build process.  I know you don't care for Fedora as much as Debian or Ubuntu and I seem to recall you said that it was because of a somewhat longer pull time because is is more "complete" (also doesn't have a "slim" version) and it is true that it is about 178 Kilobytes rather than 125 KiloBytes for Debian; as well, the "update" process takes a little longer as it includes more and has more to check, but that is only a few seconds when the Docker Hub image(s) are built; finally, the actually build sizes don't really matter so much as GHC Haskell is huge and takes a couple of GigaBytes by itself including its dependencies (in fact it plus LLVM are smaller on Fedora than on Debian due to the more direct build process), so the size difference between them doesn't mean much.
2. Alternatively, could there be an option of using LLVM 12 on AMD64 and LLVM 11 on ARM64 in the "primeimages/haskell" images?  It would seem to be a simple matter to detect which architecture is the host using something like `uname` and just substitute either "12" or "11" in the build script in two places as appropriate to the architecture.

I actually prefer option 1) above as the build process is even simpler, but I am open to discussion.  I've tested both on my `x86_64` machine and either works, and see no reason they won't work as well on ARM64, but might be surprised.

If  these options won't work or you don't feel they are appropriate, then it is a simple matter to just file a new PR changing the "12" to "11" in the mentioned two places and we will be using LLVM 11 for both, with a slight cost in performance for my solution_2 benchmark for the "lower" techniques of "bit twiddling", "stride8" and "stride8 block" but not for the fastest two techniques of "extreme" and "extreme hybrid" so it doesn't concern me too much.

OTOH, if the Fedora option 1) works, then we still get the older stable GHC Haskell version 8.8.4 (which doesn't concern me much for this benchmark) but would get LLVM 12 for both AMD64 and ARM64 which would make them more comparable, which is my preference.

Regards, Gordon

rbergen:
@GordonBGood My _preference_ for Debian-based distributions is indeed based in my better knowledge of those distros and the fact that they do tend to be smaller in many common scenarios. My _issue_ with Fedora-based images is that they often include a `dnf upgrade` somewhere in the command list. A full system upgrade is something that really shouldn't be done in a situation where determinism of Docker image builds matters; for this reason hadolint will consider it an error if it is included.
That said, if it is possible to put a Fedora-based image together that does not do this then I won't block Fedora out of principle, either. (I now understand from the dnf documentation that the regular "apt-get update" can be rolled up in any dnf command by adding the `--refresh` switch, which may be the way out, here.)

I'm not a fan of doing things like using different versions of key applications in an image, based on whatever architecture the image is used on. It's fragile when in place and will cause confusion in maintenance scenarios.

GordonBGood:
@rbergen:

> @GordonBGood My _preference_ for Debian-based distributions is indeed based in my better knowledge of those distros and the fact that they do tend to be smaller in many common scenarios.

While it is indeed true that the Fedora 34 standard distribution is about 53 MegaBytes larger than the standard distribution of Debian 11, that is fairly tiny compared to the installed size of especially GHC Haskell but also combined with LLVM, and the combined install size isn't that much different than the combined install size for Debian 11.

> My _issue_ with Fedora-based images is that they often include a `dnf upgrade` somewhere in the command list. A full system upgrade is something that really shouldn't be done in a situation where determinism of Docker image builds matters; for this reason hadolint will consider it an error if it is included. That said, if it is possible to put a Fedora-based image together that does not do this then I won't block Fedora out of principle, either. (I now understand from the dnf documentation that the regular "apt-get update" can be rolled up in any dnf command by adding the `--refresh` switch, which may be the way out, here.)

Indeed, we can skip doing included package upgrades/updates before installation by just a command as per `dnf --refresh install...`; this will not upgrade any of the packages and will only install as specified on the install list.

> I'm not a fan of doing things like using different versions of key applications in an image, based on whatever architecture the image is used on. It's fragile when in place and will cause confusion in maintenance scenarios.

We won't need to do that if we use an image based on Fedora 34, and I will as well gain the extra speed due to the LLVM 12 back end.

Since it seems you are okay with this if the above conditions are fulfilled, I tried submitting a PR, but it seems you are also working on this; I'll wait for your response before going further...

rbergen:
Ok, status update from my end:
1. About an hour ago, I pushed an updated Debian 11 based image to Docker Hub that has a fully working version of LLVM 11 installed, for both amd64 and arm64 architectures. This should make it possible in itself to build the Haskell solution without installing LLVM in the solution Dockerfile.
2. I've written a Fedora 34 based Dockerfile that installs ghc-8.8.4 and llvm-12.0.0. The build on amd64 completed successfully, but the buildx for arm64 is horrendously slow. It will therefore take quite a while before I can confirm if this indeed yields a working arm64 image.

If the Fedora 34 based image does build for arm64 then I will push it to docker hub for both architectures, overwriting the current Debian 11 based images.

rbergen:
@GordonBGood Build and push to Docker Hub of the Fedora 34 images for amd64/arm64 just finished. Please let me know if this works as a basis for this solution's Dockerfile.

GordonBGood:
@rbergen:

> @GordonBGood Build and push to Docker Hub of the Fedora 34 images for amd64/arm64 just finished. Please let me know if this works as a basis for this solution's Dockerfile.

Firstly, there is a minor "whoops" in your Dockerfile in that it installs LLVM version 12.0.0 which is a release candidate version with version 12.0.1 the production release version.

Secondly, the new Fedora 34 base docker image works, at least on amd64/x86_64 as I don't have an arm64 machine to test it with.  It took me so long to check this out as I didn't realize that docker doesn't automatically update the local copy of the image when it sees it is stale so I had to do a manual pull to force it to update.

As you can see above, I have updated my Dockerfile which is a part of this PR to use the new "primeimages/haskell:8.8.4" Docker image, and while I was at it tuned the build process to save one copy operation by renaming the build script to "Primes.sh".

I'm sure you'll let me know if there are any further problems...

rbergen:
@GordonBGood Concerning the "whoops": I took my info on the LLVM version from https://pkgs.org/search/?q=llvm, which lists 12.0.0 as the primary version for LLVM on Fedora 34. When drilling down on that I now see that indeed there is an 12.0.1 version available as well. I'll update the image to use that, and push an updated version to Docker Hub if and when it builds.

rbergen:
Images using LLVM 12.0.1 have been pushed to Docker Hub.

rbergen (approved):
LGTM

GordonBGood:
@rbergen:

> @GordonBGood Concerning the "whoops": I took my info on the LLVM version from https://pkgs.org/search/?q=llvm, which lists 12.0.0 as the primary version for LLVM on Fedora 34. When drilling down on that I now see that indeed there is an 12.0.1 version available as well. I'll update the image to use that, and push an updated version to Docker Hub if and when it builds.

> Images using LLVM 12.0.1 have been pushed to Docker Hub.

> LGTM

Now that this Haskell solution and my Julia "full" solution have been approved by you and are just awaiting Tudor's review and approval, I'm pretty much finished with what I wanted to accomplish in the "Drag Race" unless problems come up with my various solutions, so there remains just to again give you a heartfelt "Thank you" for all your efforts on the project and, in particular, for your assistance in submitting my solutions.

There are some languages that would likely benefit from some "technique love" in applying some of my fastest techniques to their solutions, namely I suspect the languages Zig, Odin, Swift, C, and C++ (at least) could be made considerably faster using these techniques, to the extent that they could be close to or even the speed leaders in the case of C and C++.  However, I am not likely going to do it, although I would be happy to advise others who want to make the attempt.  I have several reasons for this:  I find some languages too "opinionated" on what they support or do not support such as not looking for better ways of doing memory management, some such a Swift have a somewhat limited potential user base as they don't really support use on Windows (and only due to IBM does Swift even support running on Linux), and some I just find downright ugly such as C and C++, and since Nim can generate these (C is generated in my Nim solution, but C++ can also optionally be emitted) I don't have anything I want to prove there.  Of this list, if I were to write a solution for any of them it might be Swift, as other than not supporting the Windows platform it isn't too bad a language.  I also have thought about what it would take to write C/C++ solutions using my "sparse"/"dense" marking techniques and think it might be possible to make them somewhat easier to write by using `#define` text templates/macros to generate the immediate/constant single bit mask values and thus perhaps not need to use an external code generator as I resorted to for some languages.

That said, it would be interesting to see how much the above languages could move up the rankings if someone who knew those languages' idioms applied my techniques...

rbergen:
@GordonBGood Thank you for the kind words. From my end, after we needed a few interactions to tune into each other's frequencies, I have experienced our collaboration as constructive and your contributions a clear shift in the ambition level of the project. The latter is clearly illustrated by the fact that others have indeed already picked up your approaches in some other languages. I would like to extend my sincere thanks to you for that.

If you feel you've done what you wanted to do in the context of the project, then it is perfectly fine to leave it at that (nobody gets paid for this, so to speak). That said, if you change your mind about this conclusion then you're very welcome to contribute again. And I say that knowing that others have said they had reached the end of their contributions and did come back, afterwards. :)

I share your interest in where this project will, and to some extent can, go. I can honestly say that it's already developed to a point I never imagined possible.

marghidanu (approved):

PR #735, opened 2021-09-26
Switch from lut to unrolled loop

ityonemo:
## Description
This update fixes the strange docker 50% regression, though there is still a very significant penalty to using docker.

## Contributing requirements


rbergen (approved):
LGTM

marghidanu (approved):

PR #737, opened 2021-09-29
Clipper variants to Solution 1

BradleyChatha:
## Description

This adds two variants to the first clipper solution.

1. An XHarbour-specific version that doesn't have to abide by the 4096 array limit.
2. A genuine Clipper 5.2e solution using Clipper's Database functionality.

## Contributing requirements


PrimeClipper/solution_1/README.md
```diff
@@ -24,6 +24,24 @@ Special thanks to the maintainers of this repository for spending some of their
 
 The original Clipper has some quite *unique* flaws, so there are some implementation notes at the top of the file detailing our workarounds.
 
+<<<<<<< HEAD
```
marghidanu: Are these from a conflict?

BradleyChatha: huh, weird. They must've shown up when I pulled from upstream/drag-race, but I didn't get any notification of it. Will fix it.

PrimeClipper/solution_1/sieve_xharbour.prg
```diff
@@ -0,0 +1,223 @@
...
+      ? "WARNING: result is incorrect!"
+
+   ENDIF
+
+   OutputString = "AndyRadford,BradleyChatha,XHarbour"
+   OutputString = OutputString + ";" + ALLTRIM(STR(PassCount))
+   OutputString = OutputString + ";" + ALLTRIM(STR(Duration))
+   OutputString = OutputString + ";" + ALLTRIM(STR(1))
+   OutputString = OutputString + ";" + ALLTRIM("algorithm=base")
+   OutputString = OutputString + "," + ALLTRIM("faithful=yes")
```
rbergen: This implementation can't be marked faithful for the same reason that the implementation in #729 can't.

PrimeClipper/solution_1/sieve_xharbour.prg
```diff
@@ -0,0 +1,223 @@
...
+*±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±
+*    FUNCTION InitArray()
+*±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±±
+*
+FUNCTION InitArray(ElementCount,initvalue)
+
+    LOCAL array := ARRAY(ElementCount)
+
+    FOR I = 1 TO ElementCount
+        array[I] := .F.
```
rbergen: This is not an actual review comment, but an honest question: don't you intend to set the array item to `initvalue` instead of the constant `.F.`?

PrimeClipper/solution_1/sievedb.prg
```diff
@@ -0,0 +1,376 @@
...
+      ? "WARNING: result is incorrect!"
+
+   ENDIF
+
+   OutputString = "AndyRadford,Clipper_5.2e,FileVersion"
+   OutputString = OutputString + ";" + ALLTRIM(STR(PassCount))
+   OutputString = OutputString + ";" + ALLTRIM(STR(Duration))
+   OutputString = OutputString + ";" + ALLTRIM(STR(1))
+   OutputString = OutputString + ";" + ALLTRIM("algorithm=base")
+   OutputString = OutputString + "," + ALLTRIM("faithful=yes")
```
rbergen, line 150: Just for completeness: I'm willing to accept this implementation as faithful, because one could argue that the database is the "container entity" in the context of this implementation.

PrimeClipper/solution_1/sieve_xharbour.prg
BradleyChatha: Aha, the woes of copy pasting the original code! Will fix that.

PrimeClipper/solution_1/sieve_xharbour.prg
BradleyChatha: Also correct, just an oversight from a quick job being done.

PrimeClipper/solution_1/sievedb.prg
BradleyChatha: Even though marking it faithful was an accident, that's an accident I'm willing to accept.

rbergen:
When running the solution locally, I noticed that the output lines are now being broken up; I'm guessing Clipper has an old-fashioned (default) line length of 80 characters. This causes a problem with the parsing of your output by our benchmark tooling, so I think we need to try to keep the output on a single line for each of the implementations. 

My suggestions are:

1. Removing the "parallel=no" tag. It really shouldn't be added anyway, as parallelism is deduced from the number of threads reported. I actually missed this in my earlier review(s). Also,
2. Making the implementation labels shorter. You could consider shortening "Clipper_5.2e" to "Cl" and "XHarbour" to "XH"?

BradleyChatha:
If you need me to also make sure there's no new line between each output, let me know. Since that requires slightly more effort

BradleyChatha:
Also regarding line length, I think it dynamically chooses based off of the console size.

I was running it in VSCode's integrated console and it outputted just fine, but in Tilix with its default window size, it broke it up.

rbergen:
No, it's actually very important that the newlines are there.

I'm pretty sure that the 80-column cut-off is related to the use of -gtstd with hbmk2, combined with Docker. The documentation for -gtstd states that _it uses stdout output but tries to support full screen output but without collor support and cursor shape. **It format text to number of row and columns if is able to detect these values on given platform.**_

I'm guessing it can't figure out the screen size when running within Docker and then defaults to a conservative screen width of 80 columns.

In any case, the longest line is now 73 characters, which should mean we're good.

rbergen (approved):
LGTM

marghidanu (approved):

PR #744, opened 2021-10-08
Dockerlint Part 2 (final?)

Macleykun:
During #711 a few languages popped up and this pr looks at these and fixes also those docker lint issues.

All builds are verified and do build and work.

My only concern would be: PrimeMixed/solution_3/Dockerfile as i believe venv is meant to use pip to install what's in the setup.py with it's src dir. I did a bit of googeling but the use of . venv/bin/activate is still unknown to me.

Even having the same solution and removing that part still works.

### DIFF
--- a/PrimeMixed/solution_3/Dockerfile
+++ b/PrimeMixed/solution_3/Dockerfile
@@ -17,7 +17,7 @@ COPY --chown=michaelgrn setup.py .
 COPY --chown=michaelgrn src/ ./src/
 
 # The following is ignored, due to way venv is used. If you're a Python developer with knowledge of Docker. venv and user installs consider giving your insight here: https://github.com/PlummersSoftwareLLC/Primes/pull/744
-# hadolint ignore=SC1091
+# hadolint ignore=DL3059,SC1091
 RUN . venv/bin/activate
 # hadolint ignore=DL3059
 RUN python -m pip install -e . --user --no-cache-dir
--- a/PrimeMixed/solution_3/Dockerfile
+++ b/PrimeMixed/solution_3/Dockerfile
@@ -17,8 +17,9 @@ COPY --chown=michaelgrn setup.py .
 COPY --chown=michaelgrn src/ ./src/
 
 # The following is ignored, due to way venv is used. If you're a Python developer with knowledge of Docker. venv and user installs consider giving your insight here: https://github.com/PlummersSoftwareLLC/Primes/pull/744
-# hadolint ignore=DL3059,SC1091
+# hadolint ignore=SC1091
 RUN . venv/bin/activate
+# hadolint ignore=DL3059
 RUN python -m pip install -e . --user --no-cache-dir
 
 CMD ["python", "-m", "pysieve"]
--- a/PrimeMixed/solution_3/Dockerfile
+++ b/PrimeMixed/solution_3/Dockerfile
@@ -9,8 +9,8 @@ RUN python -m venv venv \
 
 USER michaelgrn
 
-ENV HOME=/home/michaelgrn
-ENV PATH=$HOME/.local/bin:$PATH
+ENV HOME=/home/michaelgrn \
+    PATH=$HOME/.local/bin:$PATH
 
 COPY --chown=michaelgrn setup.py .
 COPY --chown=michaelgrn src/ ./src/
--- a/PrimeMixed/solution_3/Dockerfile
+++ b/PrimeMixed/solution_3/Dockerfile
@@ -16,7 +16,7 @@ ENV PATH=$HOME/.local/bin:$PATH
 COPY --chown=michaelgrn setup.py .
 COPY --chown=michaelgrn src/ ./src/
 
-# The following is ignored, due to way venv is used. If you're a Python developer with knowledge of Docker and venv and user installs consider giving your insight here: https://github.com/PlummersSoftwareLLC/Primes/pull/744
+# The following is ignored, due to way venv is used. If you're a Python developer with knowledge of Docker. venv and user installs consider giving your insight here: https://github.com/PlummersSoftwareLLC/Primes/pull/744
 #hadolint ignore=SC1091,DL3059
 RUN . venv/bin/activate
 RUN python -m pip install -e . --user --no-cache-dir
--- a/PrimeMixed/solution_3/Dockerfile
+++ b/PrimeMixed/solution_3/Dockerfile
@@ -10,8 +10,8 @@ RUN python -m venv venv \
 
 USER michaelgrn
 
-ENV HOME=/home/michaelgrn \
-    PATH=$HOME/.local/bin:$PATH
+ENV HOME=/home/michaelgrn
+ENV PATH=$HOME/.local/bin:$PATH
 
 COPY --chown=michaelgrn setup.py .
 COPY --chown=michaelgrn src/ ./src/
--- a/PrimeMixed/solution_3/Dockerfile
+++ b/PrimeMixed/solution_3/Dockerfile
@@ -17,7 +17,7 @@ COPY --chown=michaelgrn setup.py .
 COPY --chown=michaelgrn src/ ./src/
 
 # The following is ignored, due to way venv is used. If you're a Python developer with knowledge of Docker. venv and user installs consider giving your insight here: https://github.com/PlummersSoftwareLLC/Primes/pull/744
-# hadolint ignore=SC1091,DL3059
+# hadolint ignore=DL3059,SC1091
 RUN . venv/bin/activate
 RUN python -m pip install -e . --user --no-cache-dir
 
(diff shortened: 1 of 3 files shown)


rbergen:
> My only concern would be: PrimeMixed/solution_3/Dockerfile as i believe venv is meant to use pip to install what's in the setup.py with it's src dir. I did a bit of googeling but the use of . venv/bin/activate is still unknown to me.

It's shell short-hand for `source venv/bin/activate`, which is a required step to activate the Python virtual environment that is created a few lines higher up in the Dockerfile (line 7 in your modified version, to be precise). 

> Even having the same solution and removing that part still works.

Yes, but the actual benchmark run will not take place in the Python virtual environment that the solution's contributor intends to use. This may impact the performance of the prime sieve, which I think the author would be justified to object to. I therefore think that the ". venv/bin/activate" command does need to be restored.

Macleykun:
> > My only concern would be: PrimeMixed/solution_3/Dockerfile as i believe venv is meant to use pip to install what's in the setup.py with it's src dir. I did a bit of googeling but the use of . venv/bin/activate is still unknown to me.
> 
> It's shell short-hand for `source venv/bin/activate`, which is a required step to activate the Python virtual environment that is created a few lines higher up in the Dockerfile (line 7 in your modified version, to be precise).
> 
> > Even having the same solution and removing that part still works.
> 
> Yes, but the actual benchmark run will not take place in the Python virtual environment that the solution's contributor intends to use. This may impact the performance of the prime sieve, which I think the author would be justified to object to. I therefore think that the ". venv/bin/activate" command does need to be restored.

So after a bit of digging and trying etc. I got it to work, i did remove the --user when the prime program is going to be installed. The reason is because of [here](https://github.com/microsoft/vscode-python/issues/14327#:~:text=Since%20the%20interpreter,9).
And i believe this solution never used venv to begin with. See [this](https://pythonspeed.com/articles/activate-virtualenv-dockerfile/#:~:text=Every%20RUN%20line%20in%20the%20Dockerfile%20is%20a%20different%20process.%20Running%20activate%20in%20a%20separate%20RUN%20has%20no%20effect%20on%20future%20RUN%20calls%3B%20for%20all%20practical%20purposes%20it%E2%80%99s%20a%20no-op.).

The only "draw-back" of my solution would that we aren't using the --user tag. But everything else does work as intended if not actually what the developer would want.

I'm curious to your opinion of that last commit/above.

rbergen:
I'm sorry, but I fail to see the point in the restructuring of the PrimeMixed/solution_3 Dockerfile you're proposing here. The Dockerfile works as it was composed by the author, and we discussed in #711 that the changes in these PRs should basically be limited to those required to address the findings that hadolint reports when this project's configuration is used. I don't see a reason to deviate from that premise, here.

Macleykun:
sadly the: RUN . venv/bin/activate triggers: [SC1091](https://github.com/koalaman/shellcheck/wiki/SC1091)
and i can't really find a reason a fix, other then to implement venv differently.

We can do: # hadolint ignore=SC1091 . Unless someone can help me with that specific issue.

rbergen:
I think a `#hadolint ignore=SC1091` is fine, here. We know what the command is about and that it's a necessary step to create the specific environment the Dockerfile author intended to. Changing the entire Dockerfile structure just to prevent this message feels like wagging the dog, to me.

Macleykun:
Sure thing!
The idea behind changing the entire Dockerfile structure is that i read here: https://pythonspeed.com/articles/activate-virtualenv-dockerfile/#:~:text=Every%20RUN%20line%20in%20the%20Dockerfile%20is%20a%20different%20process.%20Running%20activate%20in%20a%20separate%20RUN%20has%20no%20effect%20on%20future%20RUN%20calls%3B%20for%20all%20practical%20purposes%20it%E2%80%99s%20a%20no-op.
that doing this methode, wouldn't enable the vnev

gonna edit the pr so you can check it out later o/

rbergen:
I understand your motivation, but the thing is that these PRs (#711 and #744) should focus on addressing issues that "our" hadolint config reports, either by fixing the issue or justifiably ignoring the report. 
It is possible to open (a) PR(s) to improve Dockerfiles in other areas and on different grounds, but those PRs would have to follow the regular path of PRs changing other people's code. That means that the original author is then asked to review the change to the solution they contributed, and a decision on whether to move forward with the change then incorporates any feedback they give. 
For purposes of focus and clarity, we shouldn't mix the two types of changes in one PR.

Macleykun:
Ready for review :D 
After this pull, i saw that the other pulls aren't changing anything Dockerfile related, so i assume from now on the hadolint will be followed by the (new) contributors :-) 
And i see! Totally makes sense! For now we can best let it to be ignored, technically i lack the Python knowledge to say if my implementation is indeed better.

rbergen:
It seems we have a hadolint error on PrimeMixed/solution_3/Dockerfile.

Macleykun:
Sorry, it’s fixed now!

PrimeMixed/solution_3/Dockerfile
```diff
@@ -16,6 +16,7 @@ ENV PATH=$HOME/.local/bin:$PATH
 COPY --chown=michaelgrn setup.py .
 COPY --chown=michaelgrn src/ ./src/
 
+#hadolint ignore=SC1091
 RUN . venv/bin/activate
 RUN python -m pip install -e . --user --no-cache-dir
```
rbergen: Let's combine these RUNs to get rid of the DL3059.

Macleykun: Fixed! Weird how I overlooked this..

Macleykun:
I’m so sorry that the build fails again, however I do see why this fails.

because we activate the venv in the same layer when we want to install the prime app. The —user install won’t work anymore.

This is because we are now actually using the venv.

We can either:
Remove —user (preferred)
Not activate venv as we aren’t technically using it when we run the container. It technically wasn’t used before (hence the error we are getting)

I don’t mind if we use or not use venv. I’m not sure if it gives a performance boost or penalty.

rbergen:
I think neither of us really understands the intended setup, nor any side effects that may still be relevant to the solution's performance.

Because of this, I think the best thing to do is ignore the DL3059 in this particular instance. This can (and should) be undone if and when someone with enough Python-specific knowledge decides to take up revising this particular Dockerfile.

Macleykun:
> I think neither of us really understands the intended setup, nor any side effects that may still be relevant to the solution's performance.
> 
> Because of this, I think the best thing to do is ignore the DL3059 in this particular instance. This can (and should) be undone if and when someone with enough Python-specific knowledge decides to take up revising this particular Dockerfile.

Sure thing! This should do it!

rbergen (approved):
LGTM

PrimeMixed/solution_3/Dockerfile
```diff
@@ -16,6 +16,7 @@ ENV PATH=$HOME/.local/bin:$PATH
 COPY --chown=michaelgrn setup.py .
 COPY --chown=michaelgrn src/ ./src/
 
+#hadolint ignore=SC1091,DL3059
 RUN . venv/bin/activate
```
marghidanu: These are consecutive RUN statements, why are these ignored in this case?

Macleykun: It has to do with how venv is implemented. Me and rbergen agreed we don’t really understand the intended way of running this solution.

If we put these two RUNs into one RUN. Venv is used when installing the Prime app. This new environment is ran by root (?) and using —user results in an error.

The only other way to address this approach is to change it fundamentally. But as said before, we both don’t really understand the intended way.

rbergen: What we could do is add a comment with the `#hadolint ignore` explaining this reason for ignoring this specific DL3059. And [I do agree](https://github.com/PlummersSoftwareLLC/Primes/pull/744#issuecomment-939310161) that this should be cleaned up if and when someone with sufficient knowledge of Python (including virtual environments, user installs, the way those interact, etc.) takes this Dockerfile under their wing.

Macleykun: Added a comment to it, is that comment good enough?

mgdotdev (approved):

PrimeMixed/solution_3/Dockerfile
mgdotdev: I read through the documentation that @Macleykun sent me and it is clearly the case that I was implementing the venv incorrectly; the final implementation below appears to fix the issue
```python
>>> import sys
>>> sys.executable
'/opt/venv/bin/python'
```

Macleykun: Thanks mike for your insights!

I'll adres this in another pull request, that'll make later tonight.

Macleykun:
Reverted the Mixed solution 3. Will make a new pull that addresses all the concerns in one go.
Pull is here: https://github.com/PlummersSoftwareLLC/Primes/pull/751

rbergen (approved):
LGTM. @marghidanu can you approve this PR now that the controversial Dockerfile has been pulled out of it?

marghidanu (approved):

Issue #745, opened 2021-10-08
PrimesV/solution_2 does not build


VinWare:
When building PrimesV/solution_2, it gives an error. Manually building with docker shows this error:

```
make fresh_vc
make[1]: Entering directory '/opt/vlang'
rm -rf ./vc
git clone --depth 1 --quiet --single-branch https://github.com/vlang/vc ./vc
make[1]: Leaving directory '/opt/vlang'
cd ./vc && git clean -xf && git pull --quiet
make fresh_tcc
make[1]: Entering directory '/opt/vlang'
rm -rf ./thirdparty/tcc
git clone --depth 1 --quiet --single-branch --branch thirdparty-linux-amd64 https://github.com/vlang/tccbin ./thirdparty/tcc
make[1]: Leaving directory '/opt/vlang'
cd ./thirdparty/tcc && git clean -xf && git pull --quiet
cc  -std=gnu99 -w -o ./v ./vc/v.c -lm -lpthread 
./vc/v.c:1311:2: error: #error VERROR_MESSAGE Header file <atomic.h>, needed for module `sync` was not found. Please install the corresponding development headers.
 1311 | #error VERROR_MESSAGE Header file <atomic.h>, needed for module `sync` was not found. Please install the corresponding development headers.
      |  ^~~~~
./vc/v.c:5677:2: error: unknown type name 'atomic_uintptr_t'
 5677 |  atomic_uintptr_t write_adr;
      |  ^~~~~~~~~~~~~~~~
./vc/v.c:5678:2: error: unknown type name 'atomic_uintptr_t'
 5678 |  atomic_uintptr_t read_adr;
      |  ^~~~~~~~~~~~~~~~
./vc/v.c:5679:2: error: unknown type name 'atomic_uintptr_t'
 5679 |  atomic_uintptr_t adr_read;
      |  ^~~~~~~~~~~~~~~~
./vc/v.c:5680:2: error: unknown type name 'atomic_uintptr_t'
 5680 |  atomic_uintptr_t adr_written;
      |  ^~~~~~~~~~~~~~~~
make: *** [Makefile:91: all] Error 1
The command '/bin/sh -c git clone https://github.com/vlang/v /opt/vlang && git checkout 0.2.4 && make' returned a non-zero code: 2
```

rbergen:
Closing after merge of #746.

PR #746, opened 2021-10-08
Fixed Dockerfile of PrimeV/solution_2

VinWare:
## Description

PrimeV/solution_2 did not work, due to some issue with building V. Changed the V version in the Dockerfile to latest one (similar to solution_1). Now it works


VinWare:
Will close #745

rbergen:
Many thanks for addressing this.

PR #749, opened 2021-10-09
Rust solution_1 explicit "extreme" technique

mike-barber:
## Description

This PR adds an _explicit_ extreme reset technique. It's functionally-equivalent to the existing unrolled technique, but uses a procedural macro to write the functions. The existing unrolled technique achieves this implicitly through the use of const-generics.

I've been sitting on this a while as it needed a bit of testing, clean up and documentation before submitting it, and the performance impact relatively small. From what I've seen, it does vary a bit between processors though. I believe it's worth including as it is very similar (implementation-wise) to @GordonBGood's extreme implementations, and it's easier to see the relationship between our various solutions this way. The existing const-generic approach, while essentially similar, looks quite different. 

I've included a snippet of the code generated by the macro in a note in the README. It's probably easier to understand what the macro is doing by reviewing this alongside the macro itself; they can be fairly challenging to parse at times.

## Contributing requirements


rbergen (approved):
LGTM

mike-barber:
@rbergen before merging this, I'd like to confirm if it'll be acceptable to increase the current delay between successive runs. I've currently got a 1s cool-off delay in there. 

I've been discussing this with @GordonBGood and we do think there might be a bit of thermal load on the benchmark machines causing variance in run times.

rbergen:
I think #750 already made you aware that we condone a pause of a maximum of 5 second between runs within a solution. :)

With a runtime of 5 seconds, adding a 5 seconds pause already doubles the overall execution duration, which seems like a healthy "logical boundary". The 5+5 seconds also matches the 10 second run duration that we have previously allowed in some grandfathered pre-guideline solutions.

mike-barber:
Thanks -- I'll update the PR shortly. I think it will help a bit. Let me know if it'll be useful to ditch some of the variants in the benchmark run. I don't strictly need to run all of the earlier solutions, as they're not really competitive anyway.

mike-barber:
Thanks @rbergen I've updated it. Also note there's a small correction to memory allocation I noticed yesterday. It helps to actually delete the old sieve _before_ creating the new one, otherwise it's flip-flopping between different memory regions each time. The allocator tends to give you the same block you just released on the next allocation, so this helps a tiny bit.

rbergen:
> Let me know if it'll be useful to ditch some of the variants in the benchmark run. I don't strictly need to run all of the earlier solutions, as they're not really competitive anyway.

This is not a very exact-science answer, but if you feel you can drop some variants without losing the spirit of your solution, then please do.

mike-barber:
Sounds good, thanks. I've removed 3 of the older ones from the default run. It still has a reasonable story from the earlier solutions to the later ones.

mike-barber:
We may just need to re-run the CI job; it looks like a random failure of one of the other jobs.

rbergen:
> We may just need to re-run the CI job; it looks like a random failure of one of the other jobs.

That does happen, sometimes. Everything seems to be working again, now.

rbergen (approved):
Still looks good to me.

marghidanu (approved):

PR #750, opened 2021-10-09
More zig optimizations

ityonemo:
## Description
speedup of all parts by switching back to LUTs and eliminating one set of hot loop assembly instructions, and running in a shell script to prevent CPU throttling

## Contributing requirements


PrimeZig/solution_3/run.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+SLEEP=7
```
rbergen: In earlier discussions on preventing CPU throttling, the project has settled on a maximum pause duration of 5 seconds between actual sieve runs.

rbergen:
CI on this solution now fails with the following error:
```
------
 > [stage-1 6/6] RUN zig build -Drelease-fast -Darm-is-rpi4:
#14 7.884 ./src/main.zig:102:1: error: expected token '}', found '='
#14 7.884 =
#14 7.884 ^
#14 7.888 PrimeZig...The following command exited with error code 1:
#14 7.888 /deps/local/zig build-exe /opt/app/src/main.zig -lc --pkg-begin build_options /opt/app/zig-cache/PrimeZig_build_options.zig --pkg-end -OReleaseFast --cache-dir /opt/app/zig-cache --global-cache-dir /root/.cache/zig --name PrimeZig --enable-cache 
#14 7.889 error: the following build command failed with exit code 1:
#14 7.889 /opt/app/zig-cache/o/95d79a7316652e1d898193b64dba2d65/build /deps/local/zig /opt/app /opt/app/zig-cache /root/.cache/zig -Drelease-fast -Darm-is-rpi4
------
```

ityonemo:
thanks for the bug catch.  I probably didn't quite delete all of those merge conflict equals signs.

rbergen:
You're welcome, although it was really CI that did the catching.

It does seem your PR now aims to add 14 additional solutions, on top of modifying the 3rd?

ityonemo:
Yes!  That way I don't need a pause between different runs.

rbergen:
We really can't carry the resource footprint of 14 additional solutions just to prevent the pause. Please put all implementations in one solution as they were, and just reduce the pause duration from 7 to 5 seconds if you do want to add one.

mike-barber:
As a practical suggestion, if the code is the same, would it be easier to just have multiple Docker files? It might require a bit of input on the build script side from @rbergen and @marghidanu to pick them up, but it should be a lot easier to review the changes than duplicated code.

As a side question: what is the acceptable pause between different "runs" within a solution? I suspect we are seeing the effects of some thermal throttling -- there's a fair bit of variance. This happens on my desktop machine, but not on more thermally stable server hardware.

mike-barber:
Thanks @rbergen, you've just answered my question above. 

Practically speaking, the 5s pause might be better than separate Docker images anyway. The system gets pretty heavily loaded building the Docker image right before it runs it, so it's not exactly cooling off right before the run.

ityonemo:
I will comply, but this unfairly penalizes implementations which are modular over implementations in languages that cannot be modular or are sufficiently difficult to make modular that no one bothers (e.g. C, asm), in those languages to achieve the number of implementations we do you *would* have to make 14 different solutions.

Also I have benchmarked it on my personal machine, rebuilding docker image is better than a 5s pause.  I don't know why.  I suspect the per-process throttling rules are complicated and also governed by Linux, not just the hardware thermal detector

mike-barber:
Hi @ityonemo there are probably a few ways this could be tackled without putting you at a disadvantage. 

If I understand it correctly, you're doing multiple builds with different parameters, using the same source files (essentially what you have in the existing solution_3). I might have this wrong from a brief look at it, though.

If that is the case, perhaps something like this approach would work:
- keep your existing solution_3 only (which makes the PR more review-able)
- in the Dockerfile, build all the executables to separate files (I'm sure they're quite small anyway), so they're all in the same docker image.
- use a run.sh script as the docker entrypoint: this just calls each executable in turn, with a 5s shell delay between invocations

This _should_ behave about the same as if you had multiple Dockerfiles each with their own image in it, since it's a fresh process launch for each variation.

mike-barber:
I had a bit of time to play around with it, and this approach does seem to work. It also resolves issues you're having with the build, like the errant `}` in solution_3/main.zip -- there are lots of other small differences between the files that are quite hard to resolve when you have N copies of them. Also solution 7 and 8 are identical, both running 82 from what I can see.

If it helps, I just started with a list of variants to build called `variants-list.txt`:
```
45 46 48 50 82 84 86 87 88 89 90 91 92 93
```

Then created a `build-all.sh` file:
```bash
#!/bin/bash

LIST=$(<variants-list.txt)

for ONLY in $LIST
do
    echo "Building variant $ONLY ..."
    zig build -Drelease-fast -Darm-is-rpi4 -Donly=$ONLY -p "output${ONLY}"
done
```

and a `run-all.sh` file:
```bash
#!/bin/bash

LIST=$(<variants-list.txt)

for ONLY in $LIST
do
    ./output${ONLY}/bin/PrimeZig
    sleep 5
done
```

Then updated solution_3/Dockerfile thus:

```Dockerfile
FROM debian:buster-slim as builder

RUN apt-get update && \
    apt-get install -y curl xz-utils

WORKDIR /deps
RUN curl https://ziglang.org/download/0.8.0/zig-linux-"$(uname -m)"-0.8.0.tar.xz  -O && \
    tar xf zig-linux-"$(uname -m)"-0.8.0.tar.xz && \
    mv zig-linux-"$(uname -m)"-0.8.0 local/

FROM debian:buster-slim
COPY --from=builder /deps/local/ /deps/local/
RUN ln -s /deps/local/zig /usr/bin/zig

WORKDIR /opt/app
COPY . .

RUN ./build-all.sh

ENTRYPOINT [ "./run-all.sh" ]
```

Maybe give it a bash on your machine and see if it solves the issue you're having. Times running the individual solutions look very similar to running them in combined docker image on my hardware at least. Anyway, hope this helps a bit!

rbergen:
I'll wait for an update to this PR that reverts the solution split ([commit beea2ba](https://github.com/PlummersSoftwareLLC/Primes/pull/750/commits/beea2ba5b2a1eb8b5c00903e08960f5de2e6ba69) perhaps, but with run.sh containing `SLEEP=5` instead of `SLEEP=7` and without the superfluous = [that broke CI](https://github.com/PlummersSoftwareLLC/Primes/pull/750#issuecomment-939375912)?), before I continue reviewing.

ityonemo:
> I had a bit of time to play around with it, and this approach does seem to work. It also resolves issues you're having with the build, like the errant `}` in solution_3/main.zip -- there are lots of other small differences between the files that are quite hard to resolve when you have N copies of them. Also solution 7 and 8 are identical, both running 82 from what I can see.
> 
> If it helps, I just started with a list of variants to build called `variants-list.txt`:
> 
> ```
> 45 46 48 50 82 84 86 87 88 89 90 91 92 93
> ```
> 
> Then created a `build-all.sh` file:
> 
> ```shell
> #!/bin/bash
> 
> LIST=$(<variants-list.txt)
> 
> for ONLY in $LIST
> do
>     echo "Building variant $ONLY ..."
>     zig build -Drelease-fast -Darm-is-rpi4 -Donly=$ONLY -p "output${ONLY}"
> done
> ```
> 
> and a `run-all.sh` file:
> 
> ```shell
> #!/bin/bash
> 
> LIST=$(<variants-list.txt)
> 
> for ONLY in $LIST
> do
>     ./output${ONLY}/bin/PrimeZig
>     sleep 5
> done
> ```
> 
> Then updated solution_3/Dockerfile thus:
> 
> ```dockerfile
> FROM debian:buster-slim as builder
> 
> RUN apt-get update && \
>     apt-get install -y curl xz-utils
> 
> WORKDIR /deps
> RUN curl https://ziglang.org/download/0.8.0/zig-linux-"$(uname -m)"-0.8.0.tar.xz  -O && \
>     tar xf zig-linux-"$(uname -m)"-0.8.0.tar.xz && \
>     mv zig-linux-"$(uname -m)"-0.8.0 local/
> 
> FROM debian:buster-slim
> COPY --from=builder /deps/local/ /deps/local/
> RUN ln -s /deps/local/zig /usr/bin/zig
> 
> WORKDIR /opt/app
> COPY . .
> 
> RUN ./build-all.sh
> 
> ENTRYPOINT [ "./run-all.sh" ]
> ```
> 
> Maybe give it a bash on your machine and see if it solves the issue you're having. Times running the individual solutions look very similar to running them in combined docker image on my hardware at least. Anyway, hope this helps a bit!

we were already doing this and measured that 5s is not enough of a cooldown.

rbergen:
@ityonemo Can I conclude from your comment that you are still working on this?

ityonemo:
close in favor of https://github.com/PlummersSoftwareLLC/Primes/pull/753

ityonemo:
sorry couldn't figure out how to roll back and do the re-merge in a sane way

PR #751, opened 2021-10-11
PrimeMixed sl3: Use Python from the venv location and fix hadolint

Macleykun:
This pull focuses on PrimeMixed sl3. Not only are the hadolint issues fixed. But the preferred venv Python is also now used.
As a bonus, removed/changed some steps to minimize the amount of steps and complexity.
Added comments how venv python is used for clarity.


Macleykun:
@1mikegrn This pr focuses on your implementation. I tested it and it's working. Checking with whereis also shows that the right python is used at later stages.
~I kept the USER in it, but removed the workdir part as we install the app, so there isn't a need to have that in there.~
EDIT: hadolint ain't happy with that. So added workdir back in :)

I also assumed that --user isn't needed when using the venv enviroment of Python. If you prefer to use --install i can swap the venv part with the install. I believe that would also work, but not sure if that is how you like to have it intended.

rbergen:
I'll allow some time for @1mikegrn to review and comment on this Dockerfile before moving forward with this.

marghidanu:
Why is this so complicated? Creating users, using venv, upgrading pip, manually adding things to PATH, explicitly setting UID, The entire file should be 12 lines (including whitespace):

```
FROM python:3.10-alpine

RUN apk add --no-cache build-base

WORKDIR /opt/app

COPY setup.py .
COPY src/ ./src/

RUN pip3 install --no-cache-dir -e .

ENTRYPOINT [ "python3", "-m", "pysieve" ]

```

Fully compliant with the `hadolint` configuration we have so far.

Macleykun:
> Why is this so complicated? Creating users, using venv, upgrading pip, manually adding things to PATH, explicitly setting UID, The entire file should be 12 lines (including whitespace):
> 
> ```
> FROM python:3.10-alpine
> 
> RUN apk add --no-cache build-base
> 
> WORKDIR /opt/app
> 
> COPY setup.py .
> COPY src/ ./src/
> 
> RUN pip3 install --no-cache-dir -e .
> 
> ENTRYPOINT [ "python3", "-m", "pysieve" ]
> ```
> 
> Fully compliant with the `hadolint` configuration we have so far.

While i agree, i wanted to stay more so true to what @1mikegrn made. Let's wait what he wants, and i'll follow on those suggestions!

mgdotdev:
Principle of Least Privilege; any process vulnerabilities ultimately give an attacker root access, so it's better to not be running containers as root. If an application doesn't need root access, so it shouldn't have root access. This PR can (should IMO?) be accepted as-is.

marghidanu:
@1mikegrn Fair enough, but even with the security feature enabled (the principle of least privilege), the rest of the commands are not justified.

Macleykun:
> @1mikegrn Fair enough, but even with the security feature enabled (the principle of least privilege), the rest of the commands are not justified.

So the reason why i implemented it this way, is that found it easier to implement (ofc i can remove one env)

```
# Set the venv location
ENV VIRTUAL_ENV=/opt/venv
# Install venv in the specified location
RUN python3 -m venv $VIRTUAL_ENV
# Add the Python from the venv location to the PATH so we can use it
ENV PATH="$VIRTUAL_ENV/bin:$PATH"
```
I specify where i would like to install the venv where python residence and add that to the path so that it will always be used in the upcoming RUN and CMD parts.
See [this](https://pythonspeed.com/articles/activate-virtualenv-dockerfile/#:~:text=The%20result%20is%20the%20following%20Dockerfile%3A)

When you use the python in the venv enviroment. The logs will say that pip is outdated. So i also updated b/c if we install the prime app, we can also just update pip with it aswell.

marghidanu:
In a conclusion, if you remove the virtual env, you don't need to: manage the PATH variable; configure the `venv` for a specific path on disk; update pip.

Macleykun:
> In a conclusion, if you remove the virtual env, you don't need to: manage the PATH variable; configure the `venv` for a specific path on disk; update pip.

Yup, but i believe it's up to the original dev who made the dockerfile in the first place, to determine if he wants to use venv of not.

rbergen:
Not entirely. When a PR constitutes an edit to someone else's code, we do tend to ask for their review, wait some time for them to put that review together, and consider it if it comes. In the end, it's always up to the maintainers to decide what gets merged and what doesn't.
In this case, we've established that the original Dockerfile for PrimeMixed/solution_3 contains a number of conflicting/unnecessary constructs. That being the case, we should make sure that an edit to that file is in fact an improvement to its overall setup.

Macleykun:
> Not entirely. When a PR constitutes an edit to someone else's code, we do tend to ask for their review, wait some time for them to put that review together, and consider it if it comes. In the end, it's always up to the maintainers to decide what gets merged and what doesn't. In this case, we've established that the original Dockerfile for PrimeMixed/solution_3 contains a number of conflicting/unnecessary constructs. That being the case, we should make sure that an edit to that file is in fact an improvement to its overall setup.

understood!
Just hit me up here when the changes are needed to remove venv :) and i'll do it ASAP

rbergen:
I'm not sufficiently versed in Python to be able to make that call, but I would say @marghidanu is. I think his opinion on the matter is a matter of public record already. :)

mgdotdev:
How much time are we going to waste debating whether an app should use its own isolated virtual machine vs. using a users virtual machine vs. using a virtual machine assigned to root? Feels like we're missing the forest through the trees at this point. There was a technical issue with the implementation, and a fix has been submitted that satisfies the new system constraints.

rbergen:
The shortest path to a merge is removing the use of the virtual environment, as suggested by @marghidanu in his earlier comment.

Macleykun:
As both maintainers suggested, i removed the virtual environment part and got the following to say and show:
First off, for the python:3-alpine using pip3/python3 isn't needed.
But if we remove virtual environment we get this:

```
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behavior with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
WARNING: You are using pip version 21.2.4; however, version 21.3 is available.
You should consider upgrading via the '/usr/local/bin/python -m pip install --upgrade pip' command.
```

If we change to the michaelgrn user we get this warning instead:

```
    ERROR: Command errored out with exit status 1:
     command: /usr/local/bin/python -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/home/michaelgrn/setup.py'"'"'; __file__='"'"'/home/michaelgrn/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' develop --no-deps --user --prefix=
         cwd: /home/michaelgrn/
    Complete output (4 lines):
    running develop
    running egg_info
    creating src/pysieve.egg-info
    error: could not create 'src/pysieve.egg-info': Permission denied
    ----------------------------------------
ERROR: Command errored out with exit status 1: /usr/local/bin/python -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/home/michaelgrn/setup.py'"'"'; __file__='"'"'/home/michaelgrn/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' develop --no-deps --user --prefix= Check the logs for full command output.
WARNING: You are using pip version 21.2.4; however, version 21.3 is available.
You should consider upgrading via the '/usr/local/bin/python -m pip install --upgrade pip' command.
```
The following Dockerfile follows what the maintainers propose, but will give the running as root and outdated pip warning:

```
FROM python:3-alpine

RUN apk add build-base --no-cache && \
    /usr/sbin/adduser -u 1001 -D michaelgrn

WORKDIR /home/michaelgrn

COPY setup.py .
COPY src/ ./src/

RUN pip install -e . --no-cache-dir

USER michaelgrn

CMD ["python", "-m", "pysieve"]
```

I don't have a preference to use or not use a virtual environment. But i prefer the methode that fixes all/most warnings.
Let me know if the warnings should be kept or the virtual environment.

Macleykun:
I really think that @1mikegrn is right, and that the solution we both came up with is the best way possible (no error's or warnings in the least amount of steps).

marghidanu (approved):

rbergen:
I've opened #754 to fix the build failure that is unrelated to this PR. Once that has been reviewed and merged, we can proceed with merging this.

Macleykun:
> I've opened #754 to fix the build failure that is unrelated to this PR. Once that has been reviewed and merged, we can proceed with merging this.

Dankjewel! Think we can merge it his now then ;)

rbergen:
Please sync your PR branch with this one, to incorporate the new Dockerfile for Zig/solution_1. That should get us to a point where CI works, at which time I can merge.

Macleykun:
> Please sync your PR branch with this one, to incorporate the new Dockerfile for Zig/solution_1. That should get us to a point where CI works, at which time I can merge.

Gotcha :)

Macleykun:
Thank you everyone for helping me to fix all the hadolint/dockerlint warnings with all the languages :)
I hope davepl is doing allright (given he has been absent for a few weeks) and that he'll be able to work on those two PR's if he still wants to ofc! :)

It was a blast to work with so many different Dockerfiles and i learned alot with the way of thinking and making solutions/automation like those! (Mostly that setting up an IDE/editor well with lint's is key to keep quality high!)

Take care! Hope we'll cross path's once again!

PR #754, opened 2021-10-25
Fix zig/solution_1 Dockerfile

rbergen:
This replaces the Zig/solution_1 Dockefile with that of Zig/solution_3. The reason for this is that the Zig/solution_1 Dockerfile started failing due to a dependency clash regarding libffi; glib2 requires libffi.so-7, whereas Zig started requiring libffi.so-8. 
I picked the Dockerfile for Zig/solution_3 as the replacement because Zig/solution_3 has been most recently maintained.


marghidanu (approved):

marghidanu:
LGTM!

PR #758, opened 2021-11-13
Enable manual CI runs

rbergen:
This adds the workflow_dispatch event to the list of triggering events in CI.yml. 
This will allow maintainers to manually trigger CI in GitHub Actions. I would like to be able to do that to periodically check if builds still work, now that the influx of PRs has decreased and this doesn't regularly happen organically anymore.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -5,6 +5,8 @@ on:
     branches: ["**"]
   pull_request:
     branches: ["**"]
+  workflow_dispatch:
+    branches: ["**"]
 
 jobs:
   collect-solutions:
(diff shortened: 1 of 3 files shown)


rbergen:
Classic. CI fails due to V falling apart, again. 
I promise I didn't know this when I opened this PR, but it does kind of prove my point...

rbergen:
Bumped V to weekly.2021.45 in both solutions, which seems to have fixed the builds.

@marghidanu This is now really ready for your review.

marghidanu (approved):

marghidanu:
LGTM!

PR #760, opened 2021-11-19
Pascal - types changed to improve performance

laprado:
## Description
I did some improvements without changing the main algorithm. 
* Changed function RunSieve to procedure RunSieve because the main function does not use the return value.
* Changed data type Integer to NativeUInt to better run under a 64bit system.
* Implemented the use of Generics for array types to use a different initialization approach.

## Contributing requirements
The are no pull request for pascal/solution_1 at this time.
I didn't change nothing related to identation, idiomatic, etc.

* [x] A README.md already exists with the right badge(s).

### DIFF
--- a/PrimePascal/solution_1/prime.pas
+++ b/PrimePascal/solution_1/prime.pas
@@ -6,35 +6,43 @@ uses
     fgl, sysutils;
 
 type
-    CheckMap = specialize TFPGMap<Integer, Integer>;
+    CheckMap = specialize TFPGMap<NativeUInt, NativeUInt>;
     PackedBoolArray = packed array of Boolean;
 
-    PrimeSieve = class
+    TArrayFor10 = packed array [0..4] of Boolean;
+    TArrayFor100 = packed array [0..49] of Boolean;
+    TArrayFor1000 = packed array [0..499] of Boolean;
+    TArrayFor10000 = packed array [0..4999] of Boolean;
+    TArrayFor100000 = packed array [0..49999] of Boolean;
+    TArrayFor1000000 = packed array [0..499999] of Boolean;
+    TArrayFor10000000 = packed array [0..4999999] of Boolean;
+    TArrayFor100000000 = packed array [0..49999999] of Boolean;
+
+    generic PrimeSieve<_T> = class
     private
-        SieveSize: Integer;
-        NotPrimeArray: PackedBoolArray;
+        SieveSize: NativeUInt;
+        NotPrimeArray: _T;
 
     public
-        constructor Create(Size: Integer);
+        constructor Create(Size: NativeUInt);
         procedure RunSieve;
-        function CountPrimes(): Integer;
+        function CountPrimes(): NativeUInt;
         function ValidateResults(var ReferenceResults: CheckMap): Boolean;
 end;
 
-constructor PrimeSieve.Create(Size: Integer);
+constructor PrimeSieve.Create(Size: NativeUInt);
 begin
     SieveSize := Size;
-    SetLength(NotPrimeArray, (Size + 1) Div 2);
 end;
 
 procedure PrimeSieve.RunSieve;
 var
-    Factor: Integer = 3;
-    Number: Integer;
-    SieveSqrt: Integer;
-    Iterations: Integer;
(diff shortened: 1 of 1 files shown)


laprado:
running on my CPU: i7-6700T, 8GB, Ubuntu 20.04, running docker linux for PrimePascal/solution_1
before change: 2822 passes
first change: alter function RunSieve to procedure to avoid unused return value.
result: 2892 passes.
second change: use of generic types to initialize NotPrimeArray.
result: 5155 passes.
third change: change Integer type to NativeUInt for better use of 64bit
result: 6607 passes.

PrimePascal/solution_1/prime.pas
```diff
@@ -6,35 +6,43 @@
...
 
-    PrimeSieve = class
+    TArrayFor10 = packed array [0..4] of Boolean;
+    TArrayFor100 = packed array [0..49] of Boolean;
+    TArrayFor1000 = packed array [0..499] of Boolean;
+    TArrayFor10000 = packed array [0..4999] of Boolean;
+    TArrayFor100000 = packed array [0..49999] of Boolean;
+    TArrayFor1000000 = packed array [0..499999] of Boolean;
+    TArrayFor10000000 = packed array [0..4999999] of Boolean;
+    TArrayFor100000000 = packed array [0..49999999] of Boolean;
```
rbergen, line 19: With this, you're effectively pinning down the size of the memory space for the sieve; the fact you do this for a number of sieve sizes is irrelevant. This is not compatible with a faithful implementation. For that reason, this is a change I'm not willing to apply.

PrimePascal/solution_1/prime.pas
```diff
@@ -6,35 +6,43 @@
...
+    generic PrimeSieve<_T> = class
     private
-        SieveSize: Integer;
-        NotPrimeArray: PackedBoolArray;
+        SieveSize: NativeUInt;
+        NotPrimeArray: _T;
 
     public
-        constructor Create(Size: Integer);
-        function RunSieve(): PackedBoolArray;
```
rbergen, line 19: The base rules stipulate that the sieve function must return the sieve results. The fact that the main function doesn't use them doesn't change this. As such, I'm not willing to apply this change.

PrimePascal/solution_1/prime.pas
```diff
@@ -56,14 +64,12 @@ function PrimeSieve.RunSieve(): PackedBoolArray;
 
         Factor := Factor + 2;
     end; 
-
-    RunSieve := NotPrimeArray;
```
rbergen, line 60: See previous comment.

rbergen (commented):
I'm happy to consider the type change from Integer to NativeUInt, provided it objectively increases performance in itself. The other changes I won't apply, for reasons indicated in the code comments, below.

PrimePascal/solution_1/prime.pas
laprado: Ok. It´s clear.

PrimePascal/solution_1/prime.pas
laprado: I you look at the original solution, we have a dynamic array initialization at the constructor:
SetLength(NotPrimeArray, (Size + 1) Div 2);
for a sieve size of 1.000.000 the array size was 500.000.  
Using a generic type like TArrayFor1000000 (array size still is 500.000) and this will force the compiler to use a static array instead of a dynamic one, but with the same size.

rbergen: Quoting from the contributing guidelines: _"The sieve size and corresponding prime candidate memory buffer (or language equivalent) **are set/allocated dynamically at runtime**."_
The use of a static array is exactly the problem.

If you want to move forward with this PR, you can update it to exclude this change and the switch from function to procedure. I will then check the performance improvement that remains and see if merging the remaining change (from Integer to NativeUInt) makes sense.

laprado: I did a isolated test with type change from Integer to NativeUInt and there are no perfomance improvement. For some reason this change only takes effect using the static array.
Thanks for your time reviewing this PR.

PR #761, opened 2021-11-20
Optimizations for Ruby solution

Topseader:
## Description
- Remove unnecessary check.
- Added a more accurate time counting function.

## Contributing requirements


rbergen:
I'm closing this. One change (the remove of the check) breaks the algorithm, the other (changing clock) is unnecessary as the precision of the old clock is sufficient.

PR #764, opened 2021-11-20
Drag race kinematics

Kinematics:
## Description
Update code for .NET 6.

## Contributing requirements

### DIFF
--- a/PrimeCSharp/solution_1/Dockerfile
+++ b/PrimeCSharp/solution_1/Dockerfile
@@ -18,5 +18,5 @@ COPY --from=build /app .
 
 # default entry point and command argument
 ENTRYPOINT ["./PrimeCSharp"]
-#CMD ["--all"]
+CMD ["--all"]
 
--- a/PrimeCSharp/solution_1/Dockerfile
+++ b/PrimeCSharp/solution_1/Dockerfile
@@ -1,9 +1,9 @@
 #See https://aka.ms/containerfastmode to understand how Visual Studio uses this Dockerfile to build your images for faster debugging.
 
-FROM mcr.microsoft.com/dotnet/runtime:5.0 AS base
+FROM mcr.microsoft.com/dotnet/runtime:6.0 AS base
 WORKDIR /app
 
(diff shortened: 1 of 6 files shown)


PrimeCSharp/solution_1/Dockerfile
```diff
@@ -1,19 +1,22 @@
...
+FROM mcr.microsoft.com/dotnet/runtime:6.0-focal AS publish
 WORKDIR /app
-COPY --from=publish /app/publish .
-ENTRYPOINT ["dotnet", "PrimeCSharp.dll", "--all"]
\ No newline at end of file
+COPY --from=build /app .
+
+# default entry point and command argument
+ENTRYPOINT ["./PrimeCSharp"]
+#CMD ["--all"]
```
rbergen: I notice that you've removed the --all argument from the invocation of your solution. This means that only the "Standard" implementation will be executed during benchmark runs, and therefore only its results will be included in the daily reports. Is this a change you indeed intend to make?

Kinematics: Not quite. There's a problem with running docker containers from within Visual Studio:  namely, it won't pass in any commandline arguments, and will thus always run the "Standard" implementation.  In order to get the commandline arguments to work, you have to run docker from the terminal.

Somehow the prior version (which included "--all" in the parameter list) was working, I thought, but I found out it actually wasn't, in either VS2022 or VS2019. I think I may have tested it from the terminal when originally submitted, and never realized that that was never actually functional in Visual Studio.

The above CMD option was part of the attempt to figure out how to get things working, and something I wasn't sure about leaving as the default.  It worked when run from the terminal, but ultimately, all options work from the terminal, so there was no real purpose in leaving that as a new default.  However, looking at the prior Dockerfile, it included "--all" as well (though under the ENTRYPOINT section).  So people may have expected that as the default, even if the program itself didn't use that default.

I left it commented out as a reminder that I may want to reconsider how it gets used.  And for consistency with the prior configuration, it seems I should. I'll uncomment that.

PR #765, opened 2021-11-22
C128 assembly solution

rbergen:
## Description
This is a solution in 6502 assembly, targeting the Commodore 128.

A Dockerfile is not provided because:
1. I have not been able to find a C128 emulator that works with only the text console.
2. The copyright situation concerning the Commodore ROMs that are required for emulation is somewhat muddy. I'm not completely sure that we can pull those in without exposing our users to potential litigation. 

Instead of a Dockerfile, instructions for running the solution have been provided.

## Contributing requirements

### DIFF
--- a/Prime6502Assembly/solution_1/.gitignore
+++ b/Prime6502Assembly/solution_1/.gitignore
@@ -1,0 +1,2 @@
+primes.prg
+primes.d64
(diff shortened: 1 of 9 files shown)


marghidanu:
LGTM

marghidanu (approved):

PR #766, opened 2021-11-23
Drag race kinematics2

Kinematics:
## Description

Last summer I'd made an attempt to rework the code into something more flexible, but it didn't work out, and performed poorly.

After update the code to .NET 6, I decided to try again with a different approach. This time it worked.

The primary purpose was to increase the flexibility of running sieves, and making it easier to compare them. In the original version you only had two options: 1 a single sieve, or run all sieves.  With the rework, you can now run any number of specific sieves, or give a property for the program to use to select sieves, and run all matching sieves (eg: all sieves that use the 1 of 2 algorithm; all sieves that draw from the array pool; all sieves that use 32 bit data storage; etc).

This also involved expanding out the number of sieve implementations, to every combination of property I was using, with consistent naming practices, rather than arbitrary specific combinations with sometimes confusing names.

In addition, this allowed me to review the implementation code and look for things that caused poor performance.  On average, this is about a 10% improvement over the .NET 6 upgrade, which itself was about 7% better than the .NET 5 version.

I kept the original sieves, flagged behind a 'v1' commandline filter, in case referencing the originals is still relevant.

This is a substantially larger PR, so I expect more issues to be brought up.

## Contributing requirements


rbergen:
> This is a substantially larger PR, so I expect more issues to be brought up.

@Kinematics I'm not exactly sure how to read this. Are you looking for a specific type of review, are there specific areas you think warrant discussion, ...?

Kinematics:
@rbergen :
> @Kinematics I'm not exactly sure how to read this. Are you looking for a specific type of review, are there specific areas you think warrant discussion, ...?

Just saying that a large amount of code change means there's more likely to be something I missed, and thus I'm ready to see more flags from code review.  If nothing comes up, then I'm happy with that, too.

rbergen (commented):
I think a few changes need to be made before we can merge this:
- The 2 of 6 algorithms are currently marked as "base" in the solution output. This should be "wheel".
- I believe the benchmark currently executes all permutations, adding up to 32 runs. I think this should be brought back down to about half that (comparable to the current solution in main). In situations like this, my usual recommendation is to pick those implementations that provide the best representation of the full field of implementations and/or are the most relevant ones from a performance perspective.

On the point of naming conventions, I have one suggestion that you're free to completely ignore: some of the sieve implementations are called `IBool...`. This confused me a little bit initially, as I expected to be looking at interface definitions based on general-practice C# naming conventions. You could consider using `InvBool...` instead. But again, it's entirely up to you what you want to do with this.

Kinematics:
> The 2 of 6 algorithms are currently marked as "base" in the solution output. This should be "wheel".

OK. I've never been entirely clear on what constitutes the "wheel" algorithm.  At first I thought it was the bitmasking algorithm, as it sort of stamped results onto the output array like a wheel. Then I included the 8 of 30 algorithm since it used the external array to keep track of how far to move.

If "wheel" is any math other than +1 or +2, then I have to ask what the bitmasking algorithm should be marked as, since it's notably different than the others.

> I believe the benchmark currently executes all permutations, adding up to 32 runs.

Yeah, the default left it at "all", since that's what the previous one was, but it is getting a bit excessive. 

I'll probably create a --demo commandline option that handpicks a reasonable sampling of the different versions.  Will take a bit to sort out which those will be.


> On the point of naming conventions, I have one suggestion that you're free to completely ignore: some of the sieve implementations are called IBool.... This confused me a little bit initially, as I expected to be looking at interface definitions based on general-practice C# naming conventions. You could consider using InvBool... instead. But again, it's entirely up to you what you want to do with this.

I was actually considering renaming them to BoolI (putting the I at the end, instead of the start), but it made it too ambiguous next to the lowercase L.  InvBool is more reasonable. Maybe BoolInv? Still don't like it, really. Yeah, I'll go with InvBool.

rbergen:
> OK. I've never been entirely clear on what constitutes the "wheel" algorithm. At first I thought it was the bitmasking algorithm, as it sort of stamped results onto the output array like a wheel. Then I included the 8 of 30 algorithm since it used the external array to keep track of how far to move.

"wheel" is the correct algorithm label for algorithms that use "prior knowledge" of prime number patterns and project that onto the sieve. If you feel the 2 of 6 algorithm doesn't do that, then there is also an "other" algorithm denomination you can use. That is the correct one for algorithms that aren't "base" or "wheel".

> I'll probably create a --demo commandline option that handpicks a reasonable sampling of the different versions. Will take a bit to sort out which those will be.

Thank you, and that's perfectly fine.

> I was actually considering renaming them to BoolI (putting the I at the end, instead of the start), but it made it too ambiguous next to the lowercase L. InvBool is more reasonable. Maybe BoolInv? Still don't like it, really. Yeah, I'll go with InvBool.

Nice. 

I look forward to the update commit(s). Do let me know when you feel this PR is ready for another review.

Kinematics:
@rbergen — OK, updates have been pushed, and should be ready for another review.

rbergen:
Thank you @Kinematics for the updates. I think we're nearly there. 

I see you've labeled some implementations as "wheel+other", and that's not possible. If an implementation doesn't use the "base" algorithm, it's either a "wheel" or "other" implementation, but not both. In that, "wheel" overrides "other", because the fact that an algorithm applies prime number projection has the biggest impact on performance, generally speaking. The fact that bit masking is applied on top of projecting primes is something you can express in the name of the implementation (the first field in the drag-race output format), as I believe you already do using letter indicators.

The approach for determining the algorithm label for an implementation can be expressed in pseudocode in the following way:

**if** _implementation adheres to [base algorithm rules](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md#base-algorithm)_ **then**
&nbsp;&nbsp;&nbsp;&nbsp;_algorithm =_ base
**else if** _implementation applies prime number projection ([wheel factorization](https://en.wikipedia.org/wiki/Wheel_factorization))_ **then**
&nbsp;&nbsp;&nbsp;&nbsp;_algorithm =_ wheel
**else**
&nbsp;&nbsp;&nbsp;&nbsp;_algorithm =_ other

Kinematics:
The thing with the bitmasking, though, is that it doesn't do the "trial division" on the entire dataset. It only applies it to a small subset, and then copy/pastes it to the rest of the dataset.  As far as the biggest impact on performance, it's definitely the bitmasking.

I added a new sieve that uses the 1 of 2 (base) algorithm with bitmasking, to allow a comparison:

```
@Kinematics: Starting (Array pool, 64 bit, 1 of 2)...
Passes: 6795, Time: 5.00057 s, 0.735835 ms/loop, Sieve Size: 1000000, Thread Count: 1, Primes: 78498, Valid: True
kinematics_poolq2;6795;5.00057;1;algorithm=base,faithful=yes,bits=1

@Kinematics: Starting (Array pool, 64 bit, 8 of 30)...
Passes: 13510, Time: 5.00035 s, 0.370096 ms/loop, Sieve Size: 1000000, Thread Count: 1, Primes: 78498, Valid: True
kinematics_poolq30;13510;5.00035;1;algorithm=wheel,faithful=yes,bits=1

@Kinematics: Starting (Array pool, 64 bit, 1 of 2, bitmasking)...
Passes: 15492, Time: 5.00015 s, 0.322747 ms/loop, Sieve Size: 1000000, Thread Count: 1, Primes: 78498, Valid: True
kinematics_poolq2m;15492;5.00015;1;algorithm=wheel+other,faithful=no,bits=1

@Kinematics: Starting (Array pool, 64 bit, 8 of 30, bitmasking)...
Passes: 17400, Time: 5 s, 0.287356 ms/loop, Sieve Size: 1000000, Thread Count: 1, Primes: 78498, Valid: True
kinematics_poolq30m;17400;5;1;algorithm=wheel+other,faithful=no,bits=1
```

1 of 2 with bitmasking is a couple thousand points higher than 8 of 30 without bitmasking.  8 of 30 with bitmasking is a relatively small gain over 1 of 2 with bitmasking compared to the difference between without and with bitmasking (though a much larger gain when neither uses bitmasking).

I'd be willing to change it to just "other", but the wheel aspect is definitely the lesser part compared to the bitmasking (unless you get to larger wheels than I'm implementing).

Also, I see that the wiki mentions 2 of 6 explicitly as a wheel factorization pattern, so I'll change that.  Though technically 1 of 2 is also a wheel factorization, it's a special case for our purposes.


Though as long as we're doing clarifications, are the bitmasking sieves faithful? I marked them "no" because they depart algorithmically a fair bit from the original, but when reading the contribution guidelines, they don't seem to violate any of the requirements for being marked faithful.

rbergen:
@Kinematics Looking at your reasoning for selecting "other" over "wheel" for the implementations that use both, I'd have to say that I can follow it. As such, that's okay with me. The most important distinction at an algorithmic level is between "base" and "not base", anyway.
 
As you have already deduced. faithfulness is not a matter of the algorithm being used. It (only) concerns certain specific "technical" aspects of how the algorithm in question is implemented. If an implementation deviates strongly from the base algorithm but adheres to the requirements regarding faithfulness, then it can (and should) still be marked as faithful.

I've reviewed your code again, and the only comment I was going to make but didn't is about the relatively large chunks of commented-out code in the bit getters and clearers for some implementations (the ones at the bottom of the diff, to be specific), that I'm not a fan of. However, I understand that it's documentation for the concisely coded single-line logic in the methods in question. That makes it a matter of personal preference regarding style more than anything else, and I certainly don't have a monopoly on wisdom concerning that. Therefore, you can leave that as it is, if you want to.

All that said, you can make any changes you'd still like to make, e.g. to the faithfulness labeling of your implementations (as long as they are in line with the guidelines of course), and I think we can move towards merging this, then.

Kinematics:
@rbergen I don't think I have any other changes to make at the moment, other than updating the faithful label (just did that). So everything should be ready.

rbergen:
Ok, I think we're good to go. Thanks for your patience with this one.

PR #769, opened 2021-11-24
6502 assembly solution for Commodore PET

rbergen:
## Description
This is a port of 6502Assembly/solution_1 to the Commodore PET. My main reason for doing this is that I found out yesterday that someone I know actually owns a functioning 32K PET.

An added benefit is that, while working on this port, I discovered a few pretty nasty bugs in my original 6502 assembly solution. They didn't cause issues in the original solution more or less by accident, but did impact performance.

## Contributing requirements


marghidanu:
Do we need a Commodore runner? :)

rbergen:
I'd love to add one. 
But seriously, I couldn't find any Commodore multi-system emulator that is console-only and would work within Docker. Another issue would be stopping the emulator after the run, because Commodore computers have no means to switch themselves off "from the inside" (as in, by the software running on the computer).

rbergen:
> But seriously, I couldn't find any Commodore multi-system emulator that is console-only and would work within Docker

Hold on that one. I may come back to this pretty soon. :)

rbergen:
@marghidanu So, there you have it. :)

What it took in the end was a self-built version of VICE and some specific arrangement of emulator command-line options. The only thing I'm not entirely happy with is that the `docker run` requires the `-ti` options, because VICE refuses to run without a tty attached. I could maybe `socat` my way out of that but as we won't include this in our benchmark runs anyway, I've decided to leave it for now.

I have a separate branch lined up to also dockerize the C128 solution, but I'll turn that into a PR of its own once we agree this setup is ok to merge.

marghidanu (approved):

marghidanu:
LGTM!

rbergen:
Cool, thanks @marghidanu! One C128 PR coming up! 😄

PR #770, opened 2021-11-25
SympleCode Implementation

Treidexy:
This does not have Dockerfile yet, but you can still compile & run by following README instructions

### DIFF
--- a/PrimeSympleCode/solution_1/.gitignore
+++ b/PrimeSympleCode/solution_1/.gitignore
@@ -1,0 +1,1 @@
+sandbox/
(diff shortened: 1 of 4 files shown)


rbergen:
@TeddyTelanoff Thanks for your contribution. May I ask why you did not include a Dockerfile?

rbergen:
I've taken a second look at this, and considering the fact that the language itself is in such an early state that it is specifically released for building this solution, the instructions will have to include a description of how to build SympleCode itself from source.
As an extension of that, the Dockerfile would have to do the same.

rbergen:
@TeddyTelanoff Do you have an expectation concerning when you will be able to work on developing this PR to a point where it could be merged?

rbergen:
Closing this due to lack of contributor response. The PR can be reopened if and when requirements for merging it have been met.

PR #775, opened 2021-12-03
Perl pdl

wlmb:
## Description
Perl solution using the Perl Data Language

## Contributing requirements


wlmb:
I just wonder about 'faithfullness'. My solution does use external packages: PDL and PDL::NiceSlice. However, they are general purpose extensions to the Perl language in order to do number crunching efficiently, and are not packages oriented to any particular problem, such as the prime sieve.

rbergen:
Thanks for contributing this. After looking at it and PDL in general, I think that in the context of this project we should probably consider PDL a language in itself. That means the solution should be put in a directory called PrimePDL/solution_1. For one, it will solve the issue of faithfulness in relation to the inclusion of PDL and PDL::NiceSlice, as those are more or less an integrated part of any PDL program.

wlmb:
Done!
Regards.

PrimePDL/solution_1/primes.pl
```diff
@@ -0,0 +1,100 @@
...
+    for(my $factor=3; $factor<=$q; $factor+=2) {
+	$bits($factor*$factor:-1:2*$factor).=1 unless $bits(($factor));
+    }
+    $self->{ran}=1;
+}
+
+sub print_results {
+    my ( $self, $show_primes, $show_stats, $duration, $passes ) = @_;
+    say $self->get_primes if $show_primes;
+    printf "Luis Mochán (wlmb) Perl/PDL;%d;%f;%d;algorithm=base,faithful=yes\n",
```
rbergen: Please make sure your label (the first field in your output) doesn't contain any spaces. I think replacing them with underscores (_) is an easy fix in this case, but do feel free to use a different approach if you're so inclined.

wlmb: Done!

rbergen (approved):
LGTM

marghidanu (approved):

PR #776, opened 2021-12-04
Perl pdl

wlmb:
## Description
Solution with Perl/PDL using bytes for bits instead of default doubles

## Contributing requirements


wlmb:
Using byte PDL's instead of the default the speed increases about 50%.

wlmb:
I added a solution_2 employing the PDL::PP pp_def mechanism of PDL that allows writing C-like threadable functions. Its speed is about 30Xperl's solution_1.

PrimePDL/solution_2/primes.pl
```diff
@@ -0,0 +1,122 @@
...
+    my $bits   = $self->{bits};
+    my $factors=zeroes(long,$q+1-3)->xvals*2+3;
+    say $factors if $DEBUG;
+    $bits->run_sieve_aux($factors);
+}
+
+sub print_results {
+    my ( $self, $show_primes, $show_stats, $duration, $passes ) = @_;
+    say $self->get_primes if $show_primes;
+    printf "Luis_Mochán_(wlmb)_Perl/PDL;%d;%f;%d;algorithm=base,faithful=yes,bits=8\n",
```
rbergen: Please use a different label for this solution than the one used for solution_1. I'd offer `Luis_Mochán_(wlmb)_Perl/PDL-PP`  as a suggestion, but as before you're free to deviate.

wlmb: Sure!

PR #780, opened 2021-12-08
Fix output of extra tags in FSharp Solution 4

sonicrules1234:
Fixed output of extra tags so that it uses commas instead of semicolons for separators.

## Contributing requirements


rbergen:
Thanks for fixing this!

PR #783, opened 2021-12-10
Fix label output in PrimeD/solution_2 so they are unique.

sonicrules1234:
## Description

Previously, the program output two outputs with the same label (BradleyChatha-Single-SieveRTBX) but different numbers of bits.  This gets fixed so that all the single threaded outputs have "-Xbit" appended where X is the number of bits for the output.  Note that this means that primeview might need to be changed if it keeps historical data.
## Contributing requirements


rbergen:
Thanks @sonicrules1234 for identifying the duplicate labels and proposing a fix. 

Tagging @BradleyChatha to give them the opportunity to review and comment on this proposed change to their solution.

BradleyChatha:
Ah, good catch!

Also LGTM.

sonicrules1234:
We might also want to add the bits part to the multi-threaded ones too.

BradleyChatha:
Yeah, go ahead with that as well if you want. Will keep things consistant.

sonicrules1234:
Ok, done.

PR #787, opened 2021-12-14
Fixing go solution 3 so that it doesn't have an s at the end of the duration.

sonicrules1234:
## Description

The s at the end of the duration keeps this solution from showing up as a match.  This fixes it.
## Contributing requirements


rbergen:
Good catch, thanks for fixing this!

PR #788, opened 2021-12-14
Fixing an apparent typo in pascal solution 3's Dockerfile

sonicrules1234:
## Description

Now it actually runs the benchmark.
## Contributing requirements

### DIFF
--- a/Prime6502Assembly/solution_1/README.md
+++ b/Prime6502Assembly/solution_1/README.md
@@ -59,7 +59,9 @@ docker run -ti --rm 6502assembly1
   .\build.ps1
   .\run.ps1
   ```
-  Note that `build.ps1` does not need to be executed in subsequent runs; executing `run.ps1` will then suffice.
+  Note that:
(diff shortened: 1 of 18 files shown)


rbergen:
I think you are correct in identifying an issue with this solution's Dockerfile, but it's of a slightly different nature. The target directory should not be `/` but `/opt/app/`. Your change copies /opt/app/PrimePas from the build image to /opt/app/run.sh; this makes that only one of the two intended implementations are executed. If both PrimePas and run.sh are copied to /opt/app/ then both the single-threaded and multi-threaded implementations are executed.

rbergen:
(Also, on a sidenote, I think that after this PR is merged, you may want to rebase your branch to the current state of this one. As it stands, the commit list grows with every PR you open.)

sonicrules1234:
Ok, I fixed pascal solution 3's Dockerfile.  I also fixed csharp solutions 2 and 3's dockerfiles by adding a ./ in front of their entry points because it was giving an error that they weren't in PATH.

PR #790, opened 2021-12-14
Adding a build-no file to PrimeCOMAL solution 1.  It needs to be run …

sonicrules1234:
## Description

…manually with some external stuff, and solution 2 is a port of solution 1 for OpenCOMAL.

## Contributing requirements


rbergen:
There is no need for a build-no file due to the absence of a Dockerfile. Solutions without a Dockerfile will not be built by CI or the benchmark, by definition.

PR #791, opened 2021-12-14
Optimize Ruby Solution

Souravgoswami:
## Description
Optimizing code by changing variables and stuff, but keeping the algorithm intact.

## Contributing requirements
It just modifies the ruby code: prime.rb.

### DIFF
--- a/PrimeRuby/solution_1/prime.rb
+++ b/PrimeRuby/solution_1/prime.rb
@@ -13,20 +13,26 @@ class PrimeSieve
     def run_sieve
         factor = 3
         sieve_sqrt = Integer.sqrt(@sieve_size)
+        sqrt_by_2_plus_one = sieve_sqrt / 2 + 1
+        sieve_size_by_2 = @sieve_size / 2
 
-        while factor <= sieve_sqrt
+        while factor < sieve_sqrt
             number = factor
+            div_2 = number / 2
 
-            until @prime_array[number / 2] || number > sieve_sqrt do number += 2 end
+            div_2 += 1 until @prime_array[div_2] || div_2 > sqrt_by_2_plus_one
 
-            break unless number <= sieve_sqrt
+            # number = div_2 << 1 | 1
+            number = div_2 * 2 + 1
+            break if number > sieve_sqrt
 
             factor = number
-            number = factor * factor
+            factor_times_2 = factor
+            number = number * number / 2
 
-            while number < @sieve_size
-                @prime_array[number / 2] = false
-                number += factor * 2
+            while number < sieve_size_by_2
+                @prime_array[number] = false
+		number += factor_times_2
             end
 
             factor += 2
@@ -52,11 +58,11 @@ pass_count = 0
 start_time = Time.now.to_f
 sieve = nil
 
-begin 
+until (Time.now.to_f - start_time) > 1.0
     sieve = PrimeSieve.new(sieve_size)
     sieve.run_sieve
     pass_count += 1
-end until (Time.now.to_f - start_time) > 5.0
+end
 
 duration = (Time.now.to_f - start_time).round(3)
 
--- a/PrimeRuby/solution_1/prime.rb
+++ b/PrimeRuby/solution_1/prime.rb
@@ -58,7 +58,7 @@ pass_count = 0
 start_time = Time.now.to_f
 sieve = nil
 
-until (Time.now.to_f - start_time) > 1.0
+until (Time.now.to_f - start_time) > 5.0
     sieve = PrimeSieve.new(sieve_size)
     sieve.run_sieve
     pass_count += 1
--- a/PrimeRuby/solution_1/prime.rb
+++ b/PrimeRuby/solution_1/prime.rb
@@ -23,7 +23,7 @@ class PrimeSieve
             div_2 += 1 until @prime_array[div_2] || div_2 > sqrt_by_2_plus_one
 
             number = div_2 * 2 + 1
-            break if number > sieve_sqrt
+            break unless number <= sieve_sqrt
 
             factor = number
             number = factor * factor / 2
--- a/PrimeRuby/solution_1/prime.rb
+++ b/PrimeRuby/solution_1/prime.rb
@@ -27,12 +27,11 @@ class PrimeSieve
             break if number > sieve_sqrt
 
             factor = number
-            factor_times_2 = factor
             number = number * number / 2
 
             while number < sieve_size_by_2
                 @prime_array[number] = false
-		number += factor_times_2
+                number += factor
             end
 
             factor += 2
--- a/PrimeRuby/solution_1/prime.rb
+++ b/PrimeRuby/solution_1/prime.rb
@@ -16,18 +16,17 @@ class PrimeSieve
         sqrt_by_2_plus_one = sieve_sqrt / 2 + 1
         sieve_size_by_2 = @sieve_size / 2
 
-        while factor < sieve_sqrt
+        while factor <= sieve_sqrt
             number = factor
             div_2 = number / 2
 
             div_2 += 1 until @prime_array[div_2] || div_2 > sqrt_by_2_plus_one
 
-            # number = div_2 << 1 | 1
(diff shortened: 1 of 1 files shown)


rbergen:
Please limit you changes to only those that objectively improve the solution's performance. Also, please indicate what the performance improvement of those remaining changes is.

Souravgoswami:
Hi, my solution only changes the code that makes it more perform. Every single line modified is carefully benchmarked.

Here's the change, benchmarking on my Intel NUC i5 11th gen processor:

image

Using Makefile:
```
$ make DIRECTORY=PrimeRuby/solution_1
info: Unconfined mode: false
info: [PrimeRuby][solution_1] Building...
info: [PrimeRuby][solution_1] Running...
                                                      Single-threaded                                                      
┌───────┬────────────────┬──────────┬─────────┬────────┬──────────┬─────────┬───────────┬──────────┬──────┬───────────────┐
│ Index │ Implementation │ Solution │ Label   │ Passes │ Duration │ Threads │ Algorithm │ Faithful │ Bits │ Passes/Second │
├───────┼────────────────┼──────────┼─────────┼────────┼──────────┼─────────┼───────────┼──────────┼──────┼───────────────┤
│   1   │ ruby           │ 1        │ rbergen │  271   │ 5.00500  │    1    │   base    │   yes    │      │   54.14585    │
└───────┴────────────────┴──────────┴─────────┴────────┴──────────┴─────────┴───────────┴──────────┴──────┴───────────────┘

```

Souravgoswami:
The changes made removes repetitive calculation in a loop by caching them to the memory first before entering the loop. Every changes are carefully checked. It's more or less 30% improvement over the original one.

Edit, it's just authentic as the previous one:

image

PrimeRuby/solution_1/prime.rb
```diff
@@ -13,20 +13,25 @@ def count_primes
     def run_sieve
         factor = 3
         sieve_sqrt = Integer.sqrt(@sieve_size)
+        sqrt_by_2_plus_one = sieve_sqrt / 2 + 1
+        sieve_size_by_2 = @sieve_size / 2
 
-        while factor <= sieve_sqrt
+        while factor < sieve_sqrt
```
rbergen: I fail to see how this change improves performance. Also and more importantly, it's an algorithmic error to not include the actual square root of the sieve size in the range of factors.

PrimeRuby/solution_1/prime.rb
```diff
@@ -13,20 +13,25 @@ def count_primes
...
             number = factor
+            div_2 = number / 2
 
-            until @prime_array[number / 2] || number > sieve_sqrt do number += 2 end
+            div_2 += 1 until @prime_array[div_2] || div_2 > sqrt_by_2_plus_one
```
rbergen, line 23: I fail to see how the restructure of the `until` ... `do` ... `end` improves performance.

PrimeRuby/solution_1/prime.rb
```diff
@@ -13,20 +13,25 @@ def count_primes
...
 
-            break unless number <= sieve_sqrt
+            # number = div_2 << 1 | 1
```
rbergen: This commented-out line does not serve any purpose.

PrimeRuby/solution_1/prime.rb
```diff
@@ -13,20 +13,25 @@ def count_primes
...
+            number = div_2 * 2 + 1
+            break if number > sieve_sqrt
```
rbergen: I don't see how the conversion of a `break unless` to a `break if` improves performance.

PrimeRuby/solution_1/prime.rb
```diff
@@ -13,20 +13,25 @@ def count_primes
...
 
             factor = number
-            number = factor * factor
+            number = number * number / 2
```
rbergen: I don't see how replacing `number = factor * factor` with `number = number * number` improves performance.

PrimeRuby/solution_1/prime.rb
```diff
@@ -52,11 +57,11 @@ def run_sieve
 start_time = Time.now.to_f
 sieve = nil
 
-begin 
+until (Time.now.to_f - start_time) > 5.0
     sieve = PrimeSieve.new(sieve_size)
     sieve.run_sieve
     pass_count += 1
-end until (Time.now.to_f - start_time) > 5.0
+end
```
rbergen: I don't see how the restructure of a `begin` ... `end until` to an `until` ... `end` improves performance.

PrimeRuby/solution_1/prime.rb
Souravgoswami: Oops, that's a typo I made, so you can see the count is one more at 1000. But in case of range 1, it says 1 primes in the original program, too.

PrimeRuby/solution_1/prime.rb
Souravgoswami: Yes, need to remove it.

PrimeRuby/solution_1/prime.rb
Souravgoswami: Hi, it's not suggested by Ruby creator himself, and this might get remove in the future. But this doesn't improve the performance here because it's not run more times. In general it also makes the code more unreadable. Well, I can revert it.

PrimeRuby/solution_1/prime.rb
Souravgoswami: It's not number * number, it's number * number / 2. This reduces the iteration below. It can be factor * factor / 2.

PrimeRuby/solution_1/prime.rb
Souravgoswami: It's more readable than writing

```
until @prime_array[div_2] || div_2 > sqrt_by_2_plus_one do div_2 += 1 end
```

It's what Rubyists use, and it's a rather beautiful feature of Ruby.

PrimeRuby/solution_1/prime.rb
Souravgoswami: Unless uses one more instruction (actually, much more than one). It's not an alternative to the simple and more readable if condition. So it's better to avoid unless where possible. It's like doing this:

```
break if !(number <= sieve_sqrt)
```
 
This is what Ruby will do under the hood, at C level then translate it to ruby level. More wastage. So it's better to use just the suggested code `break if number > sieve_sqrt`.

rbergen:
The point really is that I am very happy to merge those changes that do make a meaningful performance improvement. The changes that decrease the number of divisions and multiplications clearly fall in that category, and therefore clearly qualify for inclusion.
 
The other changes are of a semantical nature, and I don't want to include those. Part of the point of this particular solution is to demonstrate some of the less usual constructs that Ruby offers in comparison to other languages. The fact that some of the constructs I used may no longer be supported in the future is irrelevant; they are part of the language's semantics, now.

Souravgoswami:
Hi, before merging, I will make sure to change a little bit and follow suggested changes.

Souravgoswami:
Hi, I've reverted some meaningless changes back to the original one...

PrimeRuby/solution_1/prime.rb
rbergen: ok, I'll accept that

PrimeRuby/solution_1/prime.rb
rbergen: I see your point, but I think this will only have a marginal performance impact. I really do want to include the use of `unless` in this solution, because it's an uncommon bit of language syntax that I just like a lot. If you can find another/better place to introduce it then please suggest where that would be. Otherwise I do want to keep it in. In the latter case, I am okay with explicitly mentioning in the README, or even a source code comment, that the use of `unless` in this particular situation is atypical.

rbergen:
As far as I'm concerned, we only need to solve the "unless situation" before we can move towards merging this.

Souravgoswami:
> As far as I'm concerned, we only need to solve the "unless situation" before we can move towards merging this.

Done!

rbergen:
Thank you!

PR #793, opened 2021-12-15
Add a third Clojure solution

PEZ:
## Description

A faithful [Clojure](https://clojure.org/) implementation of the [Sieve of Eratosthenes](https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes) algorithm.

The solution uses Clojure [transients](https://clojure.org/reference/transients) to make it run faster than it would do with the regular immutable datastructure, gaining about 2X in speed this way.

It also uses [futures](https://clojure.org/about/concurrent_programming) to try gain some extra speed, (hence the **Parallelism = yes** tag). This actually does not gain us all that much, but it's notable enough so I have kept it. Also, I think that when you start to sieve much more than the first one million primes, you will start to see gains from the parallelism. (But I'm not an expert, so don't take my word for it.)

### Why a third solution?

This solution is quite different from the others and I really didn't see a time-efficient way to I could ”merge” the ideas into the existing solutions. 

## Contributing requirements

### DIFF
--- a/PrimeClojure/solution_3/Dockerfile
+++ b/PrimeClojure/solution_3/Dockerfile
@@ -1,4 +1,4 @@
-FROM clojure
+FROM clojure:openjdk-18-tools-deps-1.10.3.1040
 WORKDIR /primes
 COPY deps.edn sieve.clj run.sh ./
 ENTRYPOINT ["./run.sh"]
--- a/PrimeClojure/solution_3/Dockerfile
+++ b/PrimeClojure/solution_3/Dockerfile
@@ -1,0 +1,4 @@
+FROM clojure
+WORKDIR /primes
+COPY deps.edn sieve.clj run.sh ./
+ENTRYPOINT ["./run.sh"]
--- a/PrimeClojure/solution_3/sieve.clj
+++ b/PrimeClojure/solution_3/sieve.clj
@@ -123,7 +123,7 @@
          "Count: " (count primes) ", "
          "Valid: " (if valid? "True" "False")
          "\n"
-         "pez-clj;" passes ";" timef ";1;algorithm=base,faithful=yes,bits=8")))
+         "pez-clj;" passes ";" timef ";8;algorithm=base,faithful=yes,bits=8")))
 
 
 (defn run [{:keys [warm-up?]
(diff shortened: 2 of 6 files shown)


PrimeClojure/solution_3/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM clojure
```
rbergen: Please pin the Docker base image to a specific version.

PrimeClojure/solution_3/sieve.clj
```diff
@@ -0,0 +1,134 @@
...
+  (let [nanos (.toString (.toNanos time))
+        timef (str (subs nanos 0 1) "." (subs nanos 1))]
+    (str "Passes: " passes ", "
+         "Time: " timef ", "
+         "Avg: " (float (/ (/ (.toNanos time) 1000000000) passes)) ", "
+         "Limit: " limit ", "
+         "Count: " (count primes) ", "
+         "Valid: " (if valid? "True" "False")
+         "\n"
+         "pez-clj;" passes ";" timef ";1;algorithm=base,faithful=yes,bits=8")))
```
rbergen: The thread count (the 1 after `timef` in this output) can't be 1 if it is a parallelized solution.

rbergen (commented):
Thanks for submitting this! After a first look at it, there's a couple of things that stand out. Please refer to the comments below.

PrimeClojure/solution_3/sieve.clj
PEZ: Thanks! I am actually not completely sure how many threads are used, but I slice things up in 8 slices up there, and create 8 futures, so I _think_ that might translate directly to 8 threads. (Though I am still a bit puzzled about the rather tiny speed gains, but maybe the coordination is expensive.)

PEZ:
Thanks for the super fast feedback! 🙏  I've updated thing now.

PrimeClojure/solution_3/sieve.clj
rbergen: I am definitely far less qualified to make any assumptions on how these Clojure features translate at the JVM level than you are, so I'll refrain from acting otherwise. :) From the perspective of this project, the main point is the distinction between single-threaded and parallelized. A thread count of 8 is fine from that point of view.

rbergen (approved):
LGTM

PrimeClojure/solution_3/sieve.clj
PEZ: Haha, yes, I didn't mean to consult you about the number of threads. More being transparent about it being a bit of a tricky question when using a high level abstraction. But, yes, > 1 threads at least, and my best guess is 8. 😄

marghidanu (approved):

PR #795, opened 2021-12-18
Adding some rationale to Clojure solution 3

PEZ:
## Description

* Added a doc string explaining a bit why the second step of the sieving is needed
* Correcting the bits tag 8 -> 1

## Contributing requirements

### DIFF
--- a/PrimeClojure/solution_3/sieve.clj
+++ b/PrimeClojure/solution_3/sieve.clj
@@ -128,7 +128,7 @@
          "Count: " (count primes) ", "
          "Valid: " (if valid? "True" "False")
          "\n"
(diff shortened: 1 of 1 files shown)


rbergen:
As I understand it:
- a Clojure `boolean-array` represents an array of Java booleans
- Java booleans _typically_ consume a byte of memory each

Therefore, _typically_ the correct number of bits per prime is 8, not 1.

I've used the word _typically_ because I believe that in the JVM specification, the memory size of a boolean is undefined (as in, JVM-dependent). One could therefore also argue that the correct number of bits per prime is "unknown".

PEZ:
Ah, but then the best guess is 8 bits anyway. I'll revert that.

PR #797, opened 2021-12-19
Add two non-parallel Clojure Solution 3 variants

PEZ:
## Description

* Add a boolean-array solution that instead of post-processing the sieve to remove the even numbers, it does this step first, in a non-parallel fashion
* Add a BitSet solution which is just like that ^ boolean-array one, only storing in a BitSet
* Change the return value for the `limit < 2` case of the current variant, to return a vector (this doesn't much matter, but it's nice to get the same type out for all inputs.

## Contributing requirements

### DIFF
--- a/PrimeClojure/solution_3/README.md
+++ b/PrimeClojure/solution_3/README.md
@@ -79,7 +79,7 @@ Then find one of the Rich Comment blocks in the file, one looks like so:
   )
 ```
 
-Place the cursor in one of the forms (say `(sieve 1)`) and issue the command **Calva: Evaluate top level form** (default key binding `alt+enter`). Try `alt+enter` in some of the other forms too.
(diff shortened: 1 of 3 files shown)


PrimeClojure/solution_3/README.md
```diff
@@ -3,40 +3,45 @@
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
 ![Parallelism](https://img.shields.io/badge/Parallel-yes-green)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
 ![Bit count](https://img.shields.io/badge/Bits-8-yellowgreen)
+![Bit count](https://img.shields.io/badge/Bits-1-yellowgreen)
```
rbergen: The badge colour for `Bits-1` is `-green` instead of `-yellowgreen`

PEZ: Thanks! Now fixed.

PR #798, opened 2021-12-20
Add a parallelized pre-even-filter

PEZ:
## Description

* Add a parallelized pre-even-filter for the boolean array sieve
* Also add two basecases , one for the BitSet variant and one for the boolean array, where the sieve visits all bits/indexes.

Rationale here is to demonstrate/uncover some possible tradeoffs for the curious Clojure programmer.

I'm also adding a variant where the primes array returned is half the size and the all indexes except 0 represent the odd numbers. Allowing me to skip the filtering away of even numbers that the other variants do in some different ways.

## Contributing requirements

### DIFF
--- a/PrimeClojure/solution_3/README.md
+++ b/PrimeClojure/solution_3/README.md
@@ -63,7 +63,7 @@ If not, I suggest using [Calva](https://calva.io):
 
 1. Open the project root in in VS Code.
 1. Open `sieve.clj`
-1. Issue the command **Calva: Starta Clojure REPL in your Project and Connect (aka Jack-in)**
+1. Issue the command **Calva: Start a Clojure REPL in your Project and Connect (aka Jack-in)**
 1. When the REPL has started, issue **Calva: Load current file and its dependencies**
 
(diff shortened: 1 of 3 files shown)


PEZ:
I might come back with an update where I will try to parallelize more of the work. This is fun! 😄

rbergen:
@PEZ As this is the third PR on the same solution in a timeframe of two days, I'm going to hold on its review (and hence the merge) until you indicate the solution is stable. 
(Background is that I am genuinely happy you enjoy this, but I've reviewed almost 800 PRs on the project, so my frame of reference is slightly different from yours :) )

PEZ:
Totally fair @rbergen! I'll have a think about wether the idea I just had can even work. Brb!

PEZ:
OK, I was seeing a mirage there..., I am out of ideas for now, @rbergen. (So this solution should stay calm for a while now. 😄 )

PEZ:
In case you haven't started reviewing yet. I do have one more thing I'd like to try...

PEZ:
There. It's ready for review now.

PR #802, opened 2022-01-04
Nim: Change sleep time between tests to five seconds...

GordonBGood:
## Description

Change sleep time between test runs from one second to five seconds so as to be consistent with Rust, as we speculate that mobile CPU's may be throttling their clock rates due thermal limits for the top solutions that push them to the maximum...

## Contributing requirements

### DIFF
--- a/PrimeNim/solution_3/Primes.nim
+++ b/PrimeNim/solution_3/Primes.nim
@@ -309,7 +309,7 @@ proc bench(tec: Techniques) =
   for p in ps1(): count.inc
   isValid = isValid and count == EXPECTED # two checks
 
-  sleep 5000 # sleep 1000 milliseconds = 5000 milliseconds to let CPU cool down!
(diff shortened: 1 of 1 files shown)


PrimeNim/solution_3/Primes.nim
```diff
@@ -309,7 +309,7 @@ proc bench(tec: Techniques) =
   for p in ps1(): count.inc
   isValid = isValid and count == EXPECTED # two checks
 
-  sleep 1000 # sleep 1000 milliseconds = one second to let CPU cool down!
+  sleep 5000 # sleep 1000 milliseconds = five seconds to let CPU cool down!
```
rbergen: I accept that this a very pedantic comment to make, but I think the comment should read "# sleep **5000** milliseconds = ..."

GordonBGood: @rbergen, I've made the minor comment change as requested; indeed, your suggestion is more correct (and shorter)...

PR #804, opened 2022-01-06
Add Deno runtime

hktr92:
## Description
Deno is a new, secure JavaScript and TypeScript runtime written in Rust. It was created by Ryan Dahl, the same person that made NodeJS. He mainly did Deno because he wasn't pleased about the design that NodeJS reached, so he did Deno.

Deno provides a huge benefit over NodeJS from a security standpoint. For example, during this benchmark, I specifically allow Deno to read the `IS_COMPILED` environment variable so I can "detect" whether the solution is run as a script or compiled binary (which is, tbh, the whole JS code embedded into the binary anyway).

There are some apparent benefits of compiling a deno project as binary:
- if you have TS code, it gets compiled to JS
- the JS code is then bundled (e.g. every individual .js file is unified in a single block of code)
- the bundled JS code is appended at the end(-ish) of the `deno` binary (like copy/pasting the code and setting a var to interpret that code)

So I got curious and wanted to perform three comparisons: 
- deno js vs node js: how they do package management? how easy is to write and run code? how easy is to lint, format, test and bench the code?
- deno script vs deno compiled: since `deno compiled` gets rid of opening (and resolving) modules, it could be faster. but how much faster?
- deno js vs deno ts: how much faster is pure javascript compared to javascript produced by typescript compiler?

When making this one, I analyzed what was already implemented in Node. What works there, could also work here (with some differences). I wanted to diverge every JS or TS based solution into this `PrimeDeno` directory, but I took one step at the time: the `PrimeNode` solution 1.

Another benefits of using Deno over Node (and by "benefit" i mean "do X with ease without any headache"):
- Deno is as much faithful as possible to the source material, i.e. what runs in your browser have to work in the same way here;
- Every Deno specific API is thrown under the `Deno` namespace (e.g. you want to read a file? `Deno.readTextFile()` is there to help);
- Deno has (limited) support for multi-threading via the official web workers api. and by "limited" i mean it's available only for `{ type: 'module' }`, compared to other types that were defined in the spec. This would allow one to write Primes in multi-threading with JS/TS.

## Contributing requirements


rbergen:
Thanks for submitting this. This is not the first PR aimed at adding a Deno solution, the other being #358 that was opened half a year ago almost to the day.

As explained in that PR's comments, we can't add Deno as a language because it is a runtime environment. That being the case, there are two ways to add (a) Deno-driven solution(s):
- Open a PR to modify (an) existing solution(s) to use Deno instead of the runtime environment they currently use. This will generally only be a viable approach if Deno is faster than the solution(s)' current runtime environment.
- Open a PR to add a new, Deno-driven, solution in JavaScript or TypeScript (or one for each). This will require the new solution's approach to a) be meaningfully different from the existing solutions in the respective language while b) conforming to the contributing guidelines.

Mooshua:
What we could do would be to create a new directory, "PrimeJavascript", and then move all Node.JS/Deno/QuickJS benchmarks there. That would reduce ambiguity, and is currently what the "PrimeLua" solutions are doing with Lua 5.3/LuaJIT.

The deno benchmark could then be placed there without objection. After all, it isn't Deno or Node which provide the performance, it's the V8 engine underneath.

rbergen:
That is a good point, in the sense that the NodeJS solution directory would have to be renamed to JavaScript if a JavaScript solution using another runtime was added.

GerbenRampaart:
(disclaimer: I made the original PR 6 months ago)

@Mooshua That's the peculiar thing about some of the decision making in the project; you're not actually testing syntax, you're testing runtimes and (to a lesser degree) compilers. Syntax is just text in a file being immediately compiled down to a more machine-like form. This includes javascript.

Runtimes really affect the execution time of the primes, compilers influence the efficiency of the machine code (this defines the difference between, say, the VB.NET compiler and the C# compiler, while both dotnet, and both compile to intermediate language).

In case of Deno, which supports (somewhat ironically) both javascript AND typescript, the actual syntax difference makes even less sense. What you're prime testing then is the transpile of typescript to javascript using swc (not needed in node.js) and the rust bindings to v8 (C++ in nodejs), also the event loop is using tokio where node is using libuv.

All this make node and deno drastically different and pre-selecting them by syntax makes about as much sense as calling England and America the same country because of English.

Mind you: whilst this might sound critical I adore the Primes project. I love it and thanks for all the efforts that people make to make this happen and expand. Especially @rbergen

rbergen:
@GerbenRampaart First off, thank you for your kind words. Looking at the statistics there must be quite a few people enjoying this project in some way, but it's really nice to see it stated so clearly.

I'm not sure if you are challenging my earlier statement(s) on this, and it's fine either way. And in either case, I want to clarify that I do have to conclude that the classification of a solution should really be based on the language the solution is written in, not the runtime that is used to execute it. The selection of the runtime(/compiler/interpreter/assembler/...) is basically an implementation decision concerning a solution in a specific language. I'll add to that straight away that it is an _extremely important_ implementation decision from a performance perspective, but an implementation decision is what it is. That means that it's up to the solution's author to decide what runtime works best for them; often the decision will be based on considerations of performance, but not always.

Node.js is a runtime for the JavaScript language (the Node.js website literally says so), which means that the language directory should really be called PrimeJavaScript. The fact it isn't simply is proof of an earlier oversight on my part. We're not going to change it now because Node.js is the only runtime used for the JavaScript language, and "if it ain't broken, don't try to fix it" is a principle that has proven its value to me time and time again.

If a JavaScript solution is added at a later time that uses a different runtime then a rename of the root-level project directory would definitely be connected to it. To complete the message here: adding a copy of the existing JavaScript solution that uses Deno instead of Node.js is not a viable way forward: at a language/implementation level there needs to be a meaningfully different approach for another solution to be considered. The only other way to get Deno in would be replacing Node.js as the runtime for the current JavaScript solution. That decision would basically have to be made between the author of the JavaScript solution that is currently in and the contributor proposing such a change.

GerbenRampaart:
I'd argue the direct opposite. The language is the implementation decision, not the runtime. From an Enterprise perspective it might feel different, you usually have teams of development people who are being hired as 'Java developers', 'C# developers' and so on. From a technical perspective it works the other way around. Java, Scala, Groovy and Kotlin all produce slightly different byte code, but yet byte code it is, and the JVM executes it. Dotnet same thing, FSharp, C#, VB.NET all generate intermediate language and the CLR executes it.

I'm certainly not arguing changing the whole approach by not focussing on languages anymore, because from a human perspective, languages make sense, people like languages, favour them, defend them and specialise in them.

I am however arguing that in each language group, because that's the way Primes is structured, 'solutions' in those language folders should also encompass different runtimes, which affects the Primes result in a big way and gives very interesting insights.

An additional example comes to mind; does Java does better in the Primes challenge on OpenJDK or OracleJDK? I have no idea but yes I would like to know.

rbergen:
The Primes project's basic premise is and always has been making a comparison between languages, _not_ between runtimes. This is also reflected in the original YouTube video that triggered the whole thing, and the fact that the whole repo is structured with languages as the top-level distinction. Languages don't run without some form of runtime so their impact is unavoidable (and acknowledged), but not a point of distinction in themselves, let alone the primary one.
 
That said, on second thought I could maybe entertain the idea of having a Deno-based and a Node.js-based JavaScript solution that are largely the same, otherwise. I guess I'd have to see the solutions (and their performance) side-by-side to come to a final conclusion.

GerbenRampaart:
Yes I watched that video as well. The premise originated with C++, C# and python, running on windows. In other words; language == runtime. It was the same thing. Other implementations of the dotnet or python runtime MIGHT exist, but it's so default that others are not reasonable to consider.

The fact that language == runtime in the original launch video skewed the premise of the project somewhat. It evolved into the idea that the language was enough to define the outcome.

Having said all this; I just want to add the above to your consideration. I'm fine either way, because I find it more important that you can oversee the project than the proposed change. Like I mentioned; I value you spending your time here and I appreciate it.

rbergen:
I think part of it is actually semantics. The primary divider between solutions is the language BUT I really do recognize the importance of the runtime. Similarly, Windows was indeed the OS on which everything was executed, BUT I would not disqualify other runtimes as quickly as you now do AND for C++ the compiler used is also a "runtime-shaping" factor.

Concerning this PR, I have already indicated that I will not merge it in its current form. Either it is updated to address the key points of the feedback I've given, or I'm going close it. Considering the fact that it was opened 3 weeks ago tomorrow, I think it's not unlikely that that is what I'll end up doing about a week from now.

rbergen:
Closing for lack of follow-up by contributor.

PR #807, opened 2022-01-12
added Java

Dfmaaa:
## Description
<!--
Java version


rbergen:
We already have 4 Java solutions in place, and this PR seems to have taken no notice whatsoever of the contributing guidelines. 
Please review those guidelines and consider the existing solutions in line with the guidelines. After that, feel free to open a conforming PR if you can explain why that is suitable.

PR #812, opened 2022-01-18
further dart improvements

mmcdon20:
## Description

Apologies for opening up multiple dart PR's in a short period of time, but this should (hopefully) be the last for the time being.

I did a bit of refactoring so that the solution now compiles to a single exe file. This change reduces the size of the docker image to about 10MB.

There are also some minor improvements to the the parallel implementations. It now uses, `Isolate.exit`, and also there is no need to `await` the `Isolate.spawn` calls so I have `unawaited` them instead.

## Contributing requirements

### DIFF
--- a/PrimeDart/solution_1/Dockerfile
+++ b/PrimeDart/solution_1/Dockerfile
@@ -4,15 +4,11 @@ COPY pubspec.* ./
 RUN pub get
 COPY . .
 RUN pub get --offline \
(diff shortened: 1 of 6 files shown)


rbergen:
@mmcdon20 There is no need to worry about this kind of update cadence. If only because one of the recent ones was forced by circumstances.

PR #817, opened 2022-01-23
Java

PEZ:
## Description

The current Java solutions are either slow or has a lot of ceremony. This solution is straight-forward and still fast. Faster than the one Dave ran against a C# solution.

## Contributing requirements


rbergen:
Considering that this solution has the same primary characteristics as PrimeJava/solution_1 (boolean storage, base algorithm, faithful) I think this PR should be structured to change/improve that solution instead of adding a new one. 

Also, please make the .gitignore a solution-specific file if you want to include it. We don't support the inclusion of language-level files.

PEZ:
> Considering that this solution has the same primary characteristics as PrimeJava/solution_1 (boolean storage, base algorithm, faithful) I think this PR should be structured to change/improve that solution instead of adding a new one.

I did consider that. Will consider it some more. 😄

PEZ:
I've updated the PR now to instead update solution_1.

rbergen:
For the record: on the machine I'm writing this on, the changes in this PR consistently increase the number of passes from just over 8,000 for the original solution, to just under 10,000 for the modified one.

rbergen:
Tagging @MansenC and allowing some time for them to review the proposed changes to their original solution.

PEZ:
Note to @MansenC: The reason this PR is a bit all over the place is because I first proposed my changes be a solution of its own. But it makes more sense to update solution_1, and I didn't find a time efficient enough way to figure out how to introduce the changes more ”surgically”.

rbergen:
Merging this based on a) the changes providing a consistent and meaningful performance improvement and b) lack of feedback/objections from the original author.

PR #819, opened 2022-01-26
Clojure Solution 3 Perf Tuning

PEZ:
## Description

For some reason I can't trust the performance measurements I do on my Mac. The 8-bit and 1-bit sieves perform very different (much worse) on the machines running the drag-racing. I've now gotten some help from the creator of [Clojure Goes Fast](http://clojure-goes-fast.com/) and I have reason to believe that this update should 2-3X the performance of the 8-bit sieve. (A bit less for the 1-bit, since there are other bottlenecks there.)

Check this tweet-thread for some more on the issue. https://twitter.com/pappapez/status/1486339516520673283 (Some little more. I plan to blog about it later.)

I've also made the two other ”featured” storage implementations a bit more readable.

## Contributing requirements

### DIFF
--- a/PrimeClojure/solution_3/sieve.clj
+++ b/PrimeClojure/solution_3/sieve.clj
@@ -16,18 +16,16 @@
     (let [sqrt-n (long (Math/ceil (Math/sqrt n)))
           half-n (bit-shift-right n 1)]
       (loop [p 3
-             primes (into [] (repeat (bit-shift-right n 1) true))]
-        (if-not (< p sqrt-n)
-          primes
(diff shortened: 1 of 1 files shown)


PEZ:
Those results are so weird! Super happy my tweaks delivered exactly as expected for the 1-bit variant. Super strange that the 8-bit one still runs at only 20% of the speed it does on my Macs (an x64 and an M1 with otherwise similar specs).

I've now ordered a machine that I hope is similar enough to the ones running this race. https://twitter.com/pappapez/status/1486976204943073282 I'm going to the bottom of this! Wish me luck!

PR #823, opened 2022-02-01
Clojure solution 3 optimize 8-bit solution some

PEZ:
## Description

* Flipping the true/false semantics of the boolean array variant to avoid filling it when it is created. This speeds it up some 10%.
* Also cleaned the boolean array variant up some to make for more readable Clojure code, without losing the help to the compiler. (clj-java-decompiler to the rescue).
* I've also added a version of the BitSet solution that runs faster on the machines I test it on. However, when run with Docker it runs at 0.5X it's max speed about 60% of the times. Super strange. **NB: This version of the function is not run in the drag-racing, just including it to be make it convenient for Clojure experts to clone and examine the mystery.

## Contributing requirements

### DIFF
--- a/PrimeClojure/solution_3/.gitignore
+++ b/PrimeClojure/solution_3/.gitignore
@@ -1,3 +1,8 @@
 .cpcache/
 .nrepl-port
 .calva/output-window/
+*.class
+classes/
+.lsp/.cache/
+.clj-condo/.cache/
+decompiled*.java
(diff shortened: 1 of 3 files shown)


rbergen:
@PEZ Do you expect to make more changes to Clojure/solution_3 in the near future?

PEZ:
@rbergen I'm sorry for creating work for you. I'm spending a lot of time and effort investigating a rather big mystery with unpredictable JVM execution performance, and it has made me find ways to optimize the solution.

As for your question. I will probably not find much more small/medium performance improvements now and even if I do I can hold them. for a while. (I will definitely send a PR if I find a way to solve the mystery, but that could take everything from a day up to never happening.)

rbergen:
I don't think apologies are necessary, but if we're looking at updates occurring once or twice a day then I do think it's better to bundle some. You can do that by holding back changes for a while as you mention, but I can also just not process this PR yet, until you tell me it's stable.
Let me know what you prefer.

PEZ:
By that definition, I'd say this is stable. 😀

No updates coming today, that's for sure.

PR #828, opened 2022-02-16
Attiliosbrana patch 3

attiliosbrana:
## Contributing requirements


attiliosbrana:
Adding another Python solution based on Numba JIT.

rbergen:
Could you please motivate why this Python solution should be added, next to the three others that are already in the repository?

attiliosbrana:
This solution uses Just in Time Compilation in the Python language, using the Numba library, together with Numpy. I was motivated to include it because (i) it's faster than the other Python solutions, (ii) it's basically using the pipeline of operations that most people should be using when doing scientific/math work in Python, and (iii) to bring awareness to the solution. There is a reason why researchers still use Python, and it is highly associated with the existence of libraries like Numba (https://numba.pydata.org/), which is "Built for Scientific Computing", as it states in the website. I believe people can only fully grasp the true computational effectiveness of Python when using Numba + Numpy. Numba, Numpy, Anaconda, and the whole framework of nd-arrays are products of Travis Oliphant, who is the person who was truly behind the popularization of Python, and these tools are kind of made to work together.

rbergen:
As well documented in the history of this repo, the use of JIT is considered a runtime aspect. It's not an aspect of the implemented algorithm, and therefore not a distinction that warrants the inclusion of a separate solution. As we also already have a solution based on Numpy in place (solution_3), I'd suggest modifying the PR so that it adds the use of Numba JIT to solution _3. We can then request feedback from the author of that solution on your proposed improvement.

rbergen:
@attiliosbrana I was wondering if you're of the intention to update your PR in line with my previous comment?

rbergen:
Closing for lack of contributor response.

PR #831, opened 2022-04-12
Fixes #741: APL Legal Questions and Licensing

### LINKED ISSUE #741: Handling Legal Questions re: Dyalog APL
> Concerning what's needed on the other topic, I'd suggest that someone who can legally represent Dyalog contacts me via email on rbergen (at) xs4all dot nl, so we can get that conversation going.

_Originally posted by @rbergen in https://github.com/PlummersSoftwareLLC/Primes/issues/722#issuecomment-922211574_

Just wanted to create a separate issue for tracking this. We got the ball rolling on the Dyalog side and @rbergen should have an email regarding this to move forward on.

arcfide:
## Description

In accordance with side-channel communications, this PR should fix any remaining issues associated with #741 and permit APL to be built and run in the standard benchmarking listings. 

Obvious licensing requirements in the contribution requirements list below is left unchecked, but the licensing should be appropriate now. 

Fixes #741 

## Contributing requirements


rbergen:
Hi @arcfide, thank you for submitting this. I have a couple of comments/requests:

- The code you are submitting should still be licensed in a way compatible with BSD-3, even if it relies on code external to the repository, tooling, etc. that is not.
- Please update the PR's description to include the literal text "`fix #741`" (you could do this by removing the words "any remaining issues associated with"), so GitHub will pick up that this PR will close the issue, when merged.

arcfide:
Okay, I've added the fix line and confirmed that the licensing of the submitted code is BSD-3 compatible. I had thought that the commit message would have been sufficient to close the issue and that it wasn't necessary to include that verbiage in the PR description, but it is in both now.

rbergen:
Thank you. You can see in the issue and PR lists that GitHub has now made the connection. I also don't know why things need to be phrased so specifically for that to work.

rbergen:
@arcfide This PR contains documentation on two applicable licenses:

- The license that applies to the solution's source code itself. This is either BSD-3 or another compatible, free license.
- The license terms that apply to the use of Dyalog in the context of the solution.

I think the current way this is documented could lead to confusion. I would recommend the following:
- Change the "License" header in README.md to "Terms and conditions for use of Dyalog"
- Rename the current `LICENSE` file to `LICENSE.Dyalog`
- Add another file named `LICENSE` to the solution that specifies the license that applies to the solution's actual source code.

Using the established side-channels, I have already verified that this is solidly within the boundaries of the terms and conditions that the use of Dyalog is subject to in a scenario like this.

arcfide:
Okay, I have pushed your recommended changes, so we should be good now.

PR #834, opened 2022-04-15
check for sleep process without kill

aunfl0w:
## Description
Check for sleep process pid without spawning kill.

As mentioned in https://www.youtube.com/watch?v=i4yhDUj2N7Q&t=673s 
as a possible optimization.  My test show no consistent reduction in runtime.
However, it is a way to check for the existence of a process without an extra 
kill process.

## Contributing requirements

### DIFF
--- a/PrimeBash/solution_1/PrimeBash.sh
+++ b/PrimeBash/solution_1/PrimeBash.sh
@@ -148,7 +148,7 @@ function main {
 	# we are working in nanoseconds (10^9)
 	tStart=$(date +%s%N)
 
-	while kill -0 "$sleepPid" 2>&3; do
+	while [[ -d /proc/$sleepPid ]]; do
 		emptyBitArray
 		runSieve
 		((++passes))


rbergen:
@aunfl0w Thanks for contributing.

Personally, I think the two approaches (current and proposed) are very similar. One could argue that using the [[ compound command makes that a little more of a Bashy approach, but I'm not convinced that warrants replacing the current test with this one. @Blzut3 I'm curious to your opinion, considering this PR would modify your contribution to the original solution?

Blzut3:
Relying on /proc makes the script less portable.  Specifically kill works on macOS which does not have /proc.  Granted the use of the date command isn't fully portable either so it's kind of moot.

In terms of which is more "bashy" both [[ and kill are builtin commands to bash so I can't say either way is better in that regards.  Additionally this means in either case there's no kill process as the OP suggests since as a builtin, kill is run by the current shell process just as [[ is.

My opinion on the matter is thus if there's no reduction in runtime then the change has no known advantage with at least one disadvantage.

rbergen:
Closing this based on lack of consensus that the proposed change is an improvement, rather than an alternative approach.

PR #836, opened 2022-04-18
Ruby 3.1 implements a much faster yjit compiler we can take advantage of

cFire:
## Description
Switch to Ruby 3.1 so we can make use of the new YJIT compiler

## Contributing requirements


rbergen:
Thank you for your contribution!

PR #838, opened 2022-04-22
Add CMake sieve implementation

madebr:
## Description

This PR adds a 100% CMake sieve implementation, 
The implementation should be faithful.

This is my first submission here, so some things might be off.

Result of the CI on my fork: https://github.com/madebr/Primes/runs/6133647005?check_suite_focus=true

## Contributing requirements


PrimeCMake/solution_1/README.md
```diff
@@ -0,0 +1,66 @@
+# Implementation in R
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
```
rbergen: I don't think this solution can be marked as faithful, by this repo's definition. For one, the solution does not use a class-like construct to contain the sieve.

PrimeCMake/solution_1/README.md
```diff
@@ -0,0 +1,66 @@
...
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-16-yellowgreen)
+
+This is an implementation of the [Sieve of Eratosthenes](https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes) algorithm in R. This implementation makes use of the reference class object oriented method that is available in R. This results in a solution that is closer to the rules defined in [CONTRIBUTING.md](../../CONTRIBUTING.md).
```
rbergen: I don't think this is accurate for this solution?

PrimeCMake/solution_1/primes.cmake
```diff
@@ -0,0 +1,134 @@
...
+        set(valid_str "false")
+    endif()
+
+    message("Passes: ${passes}, Time: ${duration}, Avg: ${average_duration} (sec/pass), Limit: ${size}, Count: ${count}, Valid: ${valid_str}")
+    message("")
+    message("madebr_cmake;${passes};${duration};1;algorithm=base,faithful=yes,bits=16")
+endfunction()
+
+if(NOT DEFINED SIEVE_SIZE)
+    set(SIEVE_SIZE 1000)
```
rbergen: For a fair comparison, the sieve size should be 1,000,000 for the automated benchmark runs, at least. You can choose one of the following:
- Change the benchmark run sieve size to 1,000,000. You can do this by changing the default size or explicitly setting the sieve size to 1,000,000 for benchmark runs. Note that this is only an option if the solution run time is within reasonable limits (about a minute, max).
- Removing the solution from the daily benchmark runs by adding a `build-no` flag file.

PrimeCMake/solution_1/README.md
madebr: `CONTRIBUTING.md` says this:
> * It uses a class to encapsulate the sieve, or a (closest) equivalent feature in your language if a class construct is not available. This class must contain the full state of the sieve. Each iteration should re-create a new instance of this class from scratch.

In this implementation, the state of the sieve is stored in the `sieve` list variable.

PrimeCMake/solution_1/README.md
madebr: Indeed, copy/paste error from the R readme, which I used as template.
```suggestion
This is an implementation of the [Sieve of Eratosthenes](https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes) algorithm in CMake. The complete state of the sieve is stored in a `sieve` cmake variable.
```

PrimeCMake/solution_1/README.md
rbergen: The point is that the list variable itself has to be encapsulated by the class construct, which must: 
- also include whatever else is needed to contain the full state of the sieve. The nearly 400 implementations in this repo have defined "the full state of the sieve" to include the sieve's size as a separate property, at least.
- itself be created from scratch for every iteration.

I'm happy to accept that the CMake "language" may well have no class-like construct available, but that does not change the situation concerning faithfulness. This has also been extensively documented in previous solutions/languages.

PrimeCMake/solution_1/primes.cmake
rbergen: I ran a test locally, and found that running the solution in Docker with a sieve size of 100,000 (so an order of magnitude smaller than the required sieve size for a benchmark run) takes about 25 minutes. That puts the execution time of a 1,000,000 limit sieve well beyond what we can support in benchmark runs.

rbergen:
@madebr I've created a [`make-pr838-conform`](https://github.com/rbergen/Primes/tree/make-pr838-conform) branch in my fork that includes the changes I think would be needed to make this solution conform to requirements. If you're okay with including these, I'm happy to open a PR on your `cmake-solution1` branch so you can pull them into this PR.

madebr:
> @madebr I've created a [`make-pr838-conform`](https://github.com/rbergen/Primes/tree/make-pr838-conform) branch in my fork that includes the changes I think would be needed to make this solution conform to requirements. If you're okay with including these, I'm happy to open a PR on your `cmake-solution1` branch so you can pull them into this PR.

Thanks, I will merge your changes.
Also, I will look into optimization strategies since the CMake solution is not deemed "faithful", I'm not bound by the original structure anymore.

rbergen:
> Also, I will look into optimization strategies since the CMake solution is not deemed "faithful", I'm not bound by the original structure anymore.

That's entirely fair, in my opinion. Do keep in mind that if you apply optimizations that modify the actual algorithm in a meaningful sense, it may no longer be possible to mark the algorithm as "base". Obviously, I will consider this aspect as part of future reviews.

madebr:
> > Also, I will look into optimization strategies since the CMake solution is not deemed "faithful", I'm not bound by the original structure anymore.
> 
> That's entirely fair, in my opinion. Do keep in mind that if you apply optimizations that modify the actual algorithm in a meaningful sense, it may no longer be possible to mark the algorithm as "base". Obviously, I will consider this aspect as part of future reviews.

I pushed a new revision, using cmake variables instead of a single cmake list.
Because this avoid a lot of copying/look-up, the time to calculate a 1,000,000 sieve is reduced to ~20 seconds on my *old* laptop.
So I took the liberty of removing the `build-no` file as this new time is more reasonable.

PrimeCMake/solution_1/primes.cmake
```diff
@@ -135,11 +134,11 @@ function(primes_drag_race size seconds)
 
     message("Passes: ${passes}, Time: ${duration}, Avg: ${average_duration} (sec/pass), Limit: ${size}, Count: ${count}, Valid: ${valid_str}")
     message("")
-    message("madebr_cmake;${passes};${duration};1;algorithm=base,faithful=no,bits=16")
+    message("madebr_cmake;${passes};${duration};1;algorithm=base,faithful=no,bits=unknown")
```
rbergen: Please drop the `bits` tag from the benchmark output altogether when the bit count is unknown. To avoid confusion: the README.md badge should be included with the "unknown" value. Also, as a reminder: please update the README.md example output accordingly.

rbergen (commented):
I'd say this is a significant improvement over the previous version, because we will indeed be able to include this result in the daily benchmark runs. 
There's one minor thing I came across when reviewing, please see below. Once that's addressed, I'm happy to put this solution forward for its second review.

PrimeCMake/solution_1/primes.cmake
```diff
@@ -0,0 +1,148 @@
...
+
+    if(reference_count)
+        set(valid_str "true")
+    else()
+        set(valid_str "false")
+    endif()
+
+    message("Passes: ${passes}, Time: ${duration}, Avg: ${average_duration} (sec/pass), Limit: ${size}, Count: ${count}, Valid: ${valid_str}")
+    message("")
+    message("madebr_cmake;${passes};${duration};1;algorithm=base,faithful=no,bits=unknown")
```
madebr:
```suggestion
    message("madebr_cmake;${passes};${duration};1;algorithm=base,faithful=no")
```

PrimeCMake/solution_1/README.md
```diff
@@ -0,0 +1,82 @@
...
+    ```
+
+## Output
+
+Below is an example of the output on my machine, running with Docker.
+
+```bash
+Passes: 1, Time: 20.169636, Avg: 20.169636 (sec/pass), Limit: 1000000, Count: 78498, Valid: true
+
+madebr_cmake;1;20.169636;1;algorithm=base,faithful=no,bits=unknown
```
madebr:
```suggestion
madebr_cmake;1;20.169636;1;algorithm=base,faithful=no
```

rbergen (approved):
LGTM

marghidanu (approved):
LGTM

PR #839, opened 2022-04-28
fix to Batch solution 2 to include ',' as a delimiter

Richhoef:
## Description
The previous version gave a `Missing operand.` error when it was run. The addition of `,` as a delimiter fixes this. 

## Contributing requirements


rbergen:
@Richhoef I presume you're running the solution under a non-English locale? Or more accurately, a locale with a number format that uses a comma as the decimal separator?

Richhoef:
Yes, that is true. Since the period is also a delimiter and it didn't seem to give me a problem, I would assume the comma won't give a problem with English locales (but I could be wrong off course)..

rbergen:
@Richhoef I didn't mean to imply that the inclusion of the comma as a delimiter could cause a problem for locales that use a decimal point. I was just trying to establish what problem the PR is solving. :)

Richhoef:
@rbergen Ah ok :)

PR #846, opened 2022-07-02
Optimisation to the existing PHP solution (solution_1)

sqonk:
## Description

Changing SPLFixedArray to a simple string yields a noticeable performance increase. In local tests - up to a factor of 2x.

I have also updated the Docker build to the use the main official PHP image running the latest version at time of writing (8.1.7). V8.1 has its own minor speed increases over the previous 8.0. Particularly, the Opcache JIT is now supported on ARM64. 

## Contributing requirements

### DIFF
--- a/PrimePHP/solution_1/Dockerfile
+++ b/PrimePHP/solution_1/Dockerfile
@@ -1,8 +1,11 @@
-FROM php:8.0-cli-alpine3.13
+FROM php:8.1.7-cli
+
+# Use the default production configuration
+RUN mv "$PHP_INI_DIR/php.ini-production" "$PHP_INI_DIR/php.ini"
+# Enable opcache & JIT
+RUN printf "zend_extension=opcache.so\nopcache.enable=1\nopcache.enable_cli=1\nopcache.jit_buffer_size=100M\nopcache.jit=tracing\nmemory_limit=-1\n" >> "$PHP_INI_DIR/php.ini"
 
 WORKDIR /opt/app
-
 COPY *.php .
-COPY opcache.ini /usr/local/etc/php/conf.d/
 
 ENTRYPOINT [ "php", "PrimePHP.php" ]
--- a/PrimePHP/solution_1/PrimePHP.php
+++ b/PrimePHP/solution_1/PrimePHP.php
@@ -14,7 +14,7 @@ declare(strict_types=1);
 
 class PrimeSieve
 {
-    private SplFixedArray $rawbits;
+    private string $rawbits;
 
     private int $sieveSize;
     private int $rawBitsSize;
@@ -42,11 +42,12 @@ class PrimeSieve
         $sieveSize = $this->sieveSize;
         $q = sqrt($sieveSize);
 		$rawBitsSize = $this->rawBitsSize;
-        $rb = new SplFixedArray($rawBitsSize);
+        $rb = str_repeat('0', $rawBitsSize);
 
         while ($factor < $q) {
             for ($i = $factor; $i <= $sieveSize; $i += 2) {
-                if ($rb[$i * 0.5] === null) {
+                $rbi = (int)($i * 0.5);
+                if ($rb[$rbi] == '0') {
                     $factor = $i;
                     break;
                 }
@@ -55,7 +56,7 @@ class PrimeSieve
             $ft2 = $factor;
             $start = (int)($factor * $factor * 0.5);
             for ($i = $start; $i < $rawBitsSize; $i += $ft2) {
-                $rb[$i] = 1;
+                $rb[$i] = '1';
             }
 
             $factor += 2;
@@ -66,7 +67,7 @@ class PrimeSieve
     public function printResults(): void
     {
         for ($i = 1; $i < $this->sieveSize; $i++) {
-            if ($i % 2 && $this->rawbits[$i * 0.5] === null) {
+            if ($i % 2 && $this->rawbits[(int)($i * 0.5)] === null) {
                 echo $i . ", ";
             }
         }
@@ -75,9 +76,9 @@ class PrimeSieve
     public function getRawbitCount(): int
     {
         $sum = 0;
-        $sz = $this->rawbits->getSize();
+        $sz = strlen($this->rawbits);
         for ($i = 0; $i < $sz; $i++){
-            if ($this->rawbits[$i] === null){
+            if ($this->rawbits[$i] == '0'){
                 $sum++;
             }
         }
@@ -120,7 +121,7 @@ printf(
 
 // Following 2 lines added by rbergen to conform to drag race output format
 echo "\n\n";
-printf("DennisdeBest;%d;%f;1;algorithm=base,faithful=yes\n", $passes, ((float)$tD) / 1000);
+printf("DennisdeBest;%d;%f;1;algorithm=base,faithful=yes,bits=8\n", $passes, ((float)$tD) / 1000);
 
 function getTimeDiffInMs(float $tStart): float
 {
--- a/PrimePHP/solution_1/PrimePHP.php
+++ b/PrimePHP/solution_1/PrimePHP.php
@@ -67,7 +67,7 @@ class PrimeSieve
     public function printResults(): void
     {
         for ($i = 1; $i < $this->sieveSize; $i++) {
-            if ($i % 2 && $this->rawbits[(int)($i * 0.5)] === null) {
+            if ($i % 2 && $this->rawbits[(int)($i * 0.5)] == '0') {
                 echo $i . ", ";
             }
         }
--- a/PrimePHP/solution_1/README.md
+++ b/PrimePHP/solution_1/README.md
@@ -6,7 +6,10 @@ Leaned the loop logic.
 Uses factor * factor.
 Enabled JIT, small improvement in docker, native it gives me about 30% boost.
 
+Optimisations by sqonk:
+Changing SPLFixedArray to a simple string yields a nearly 2x performance increase.
+
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
 ![Parallelism](https://img.shields.io/badge/Parallel-no-green)
-![Bit count](https://img.shields.io/badge/Bits-unknown-yellowgreen)
+![Bit count](https://img.shields.io/badge/Bits-8-yellowgreen)
--- a/PrimePHP/solution_1/opcache.ini
+++ b/PrimePHP/solution_1/opcache.ini
@@ -1,5 +1,0 @@
-zend_extension=opcache.so
-opcache.enable=1
-opcache.enable_cli=1
-;opcache.jit_buffer_size=32M
-;opcache.jit=1255


rbergen:
@sqonk Thanks for submitting this. For the future, it would help if you could spend a few lines of text in your PR description on the purpose of/motivation for the changes you're proposing.

@DennisdeBest On my machine, these changes bump the performance of your solution from approx. 220 iterations to around 410. Would you see any reason not to merge this?

sqonk:
@rbergen Just looking under the posted description now and am surprised to see it has nothing at all between the title and the requirements! My apologies for this, I had typed out about 3 small paragraphs under the bracketed section when I posted the PR, not sure how it got lost. Would you like me to re-post my description to the to a comment?

rbergen:
@sqonk I just checked the PR description markdown, and your lines are indeed there. They just weren't showing in the rendered description. Maybe it's because they followed an empty comment tag? Technically that shouldn't have this result, but it could have confused the GitHub markdown renderer. In any case, I removed the comment tag, and now your text does show. Which is exactly what I was looking for. So, thanks for including it in the beginning. :)

EDIT: I now see you edited your description two minutes before I did. Which means I don't know what was where when. It doesn't really matter, either.

sqonk:
@rbergen  No I think it was me all along. I'd gone and put the original text in the actual markup comment, hence why it never showed. I investigated and moved it down, which was the cause of your confusion. My apologies.

rbergen:
@sqonk That would indeed explain what we weren't seeing without the markdown renderer misbehaving. In response to what we saw yesterday, I have already changed the PR template to not include the comment tags in the Description section. That should avoid this type of confusion in the future.

I think we can now just wait for the response of @DennisdeBest. I'll give them a few more days to give one before proceeding on this.

sqonk:
@rbergen yes not a worry. I am actually quite new to the whole PR thing on Github. In hindsight I should have probably picked up that it was a comment but I had most of my focus on the contributing guidelines and making sure there was nothing I had missed.

Assuming there are no issues, I look forward to seeing the benchmarks in any case. I clocked around 930 passes on my own laptop (up from the mid 400s to memory), and hit about 1040 with the docker build.

I tried a bunch of different approaches to boost the score but admittedly the SPLFixedArray is actually reasonably good at a task like this. On the other hand, Strings in PHP are a simple byte array behind the scenes and are technically not immutable to the same degree as some other languages, as I understand it, so modifying the string with array syntax for individual chars, should essentially be changing the 0s and 1s directly.

At the moment that is about as close to the horses mouth as I think I can get it without resorting to external dependancies.

rbergen (approved):
Approving this in lieu of a response from the solution's original author.

DennisCodeBuds:
Hello, sorry I didn't reply earlier, thanks @sqonk it's very interesting to see these changes !
Thanks @rbergen for merging this.

DennisdeBest:
Hello, sorry I didn't reply earlier, thanks @sqonk it's very interesting to see these changes !
Thanks @rbergen for merging this.

ralfstrobel:
Hello everyone, I have been testing past variants of the algorithm on newer PHP versions and wanted to ask for your opinion on how to proceed...

As much as I like the clever string solution by @sqonk, it turns out that using a naive array approach has become the fastest solution again starting with PHP 8.2 and even more so in PHP 8.3:
```
$rb = array_fill(0, $rawBitsSize, true);
```
This is now about 20% faster than the string approach, while using SplFixedArray has seen no improvement. The performance boost also requires the JIT compiler to be enabled, so PHP seems to have implemented further optimization there for its native data structures.

Since this is basically the original approach by @DennisdeBest, I would suggest that I make a PR that updates the PHP version and converts this solution back to the array code (preserving other optimizations that have been added in the meantime). I would also like to add a few minor optimizations myself, such as using bitshift operations instead of multiplications by 0.5.

@rbergen This would no longer be a bits=8 solution, however, since native arrays in PHP internally store 16Byte zval structs (yes 16 bytes to store one bit). So I wonder if we should also preserve the current string approach, maybe as a new solution_2, properly credited to @sqonk. With a few more bit operations, we could even feasibly turn it into a 1-bit solution, that would be slower but essentially explore the other end of the spectrum from the fast but memory-wasting native approach.

rbergen:
@ralfstrobel I'm inclined to leave the current solution as it is, unless @DennisdeBest agrees to modifying his in accordance with what you suggest (with the consequence of it becoming a 16-bit per prime solution instead of an 8-bit per prime one), first. The difference between @sqonk's earlier change and this one, is exactly that his change pinned down the bits-per-prime to 8 bits per prime aspect where it was marked as "undefined" before, and you are now suggesting to double the bits-per-prime in comparison to the current implementation. That can be considered a "descent" of the solution, if that's the right way to put it.

Unless and until we get that approval from @DennisdeBest, I _would_ consider adding a solution_2 with both your ideas as implementations:

- One using a native (16-bit) array, using one element per prime candidate.
- A 1-bit implementation, which would yield a PHP implementation that is included on the leaderboard - although not at the top of it.

ralfstrobel:
I'm not suggesting to double the bits per prime, I'm suggesting to 16x them. :stuck_out_tongue_winking_eye:
A zval is a 16 byte struct.

But yeah we can wait for feedback from the others first.

rbergen:
That is an appropriate correction and I apologize for not getting this right the first time, but I think we can agree this kind of solidifies my main point. :)

I'll wait for either feedback of those tagged, or a new solution PR from yourself.

ralfstrobel:
Out of curiosity, how has this been handled for other dynamically typed languages?
To my knowledge, Python uses 28 bytes to store a boolean value object and JS (V8) uses 4-8 bytes for primitive types.

rbergen:
Throughout my tenure as maintainer I've tried to establish the actual bits used per prime candidate for every solution('s implementation), based on information available about the programming language and/or the approach and data type chosen. In all cases where I felt I couldn't unequivocally establish the actual answer to the question, the bit sizes of the solutions/implementations in question have been marked as "unknown".

Mistakes may have been made in the process; if there are concrete and well-motivated arguments for corrections I am open to them.

ralfstrobel:
To that end, there is both good [documentation](https://www.phpinternalsbook.com/php7/zvals/basic_structure.html#the-zval-struct) and [practical proof](https://github.com/mrsuh/php-var-sizeof) that a boolean is indeed stored as 128 bits in all current versions of PHP. So the result entry would not have to become undefined but simply bits=128.

Whether that constitutes a "descent" of the solution, as you put it, depends on the optimization criteria. If computation speed really is the only goal, then this is currently the best solution. But in the end I would also prefer to have multiple solutions with multiple bit-sizes and then anyone who objects to large memory footprints can use the respective filters in the leader board.

The reason why I would still prefer the array solution as "solution_1" is that it represents the most native way to solve the problem within PHP. And it's not a coincidence that it is now also the fastest, since the PHP developers continue to optimize for the way the language is intended to be used. The string solution is a bit of a clever hack. You're not supposed to use strings as byte arrays and you pay a performance penalty, e.g. because you have to use string comparisons rather than true byte value comparisons. If we go for the 1-bit solution, that overhead is going to get even larger, because you will have to involve string to integer casting and back. None of that is a reason not to do it, but it just doesn't have the feel of a "solution_1" to me.

sqonk:
@ralfstrobel I concur with your findings and I'd actually discovered the same as you back when 8.2  was released - native arrays became a lot faster (and yeah my string mod was a bit of a sneaky or lucky find).  I'd embarked upon rewriting solution 1 with a more modern approach of how PHP code is typically structured today with full type hinting. I don't have any hard numbers but it did appear to help speed up the program further, which would stand to reason as you're effectively allowing opcache to make more assumptions about what each variable can be.

I'd be happy to share my copy with you if you wanted to take bits from it that you deemed useful?

@rbergen I hadn't got round to submitting this rewrite as I was unsure whether a fully rewritten version of sol#1 would qualify as an appropriate addition?

rbergen:
@ralfstrobel I think it's important to keep in mind that there is no value statement in "solution_1" vs "solution_2", "solution_3", etc. It just means "solution_1" got submitted first. Because of the way things worked out at the beginning of this repo's existence, one of Dave's original solutions actually became solution_2 for the language in question.   

As per the purpose of the drag race, it's a solution's performance that decides its ranking, with the sidenote that the core ranking consists of those solutions that use the base algorithm and are faithful (both as defined in the contributing guidelines) and use 1 bit per prime candidate. Beyond that I think there is value in showing different approaches in various languages, provided they are indeed sufficiently different.

As I've stated before in this comment chain:

- solution_1 can be rewritten if @DennisdeBest agrees.
- That said, I think the approaches that are discussed now are sufficiently different to warrant the addition of a second solution, if you (@ralfstrobel and/or @sqonk) choose to create one. If one of the implementations in that second solution is faster than the one in solution_1, then that mainly demonstrates the progress PHP has made since the current iteration of solution_1 was submitted.

DennisdeBest:
Hello, 

Thanks for adding to this. 

When running this locally I get some weird results between the current version [ec584306ce7222b8f3ad426bc85d9c0cd9c75f7e](https://github.com/PlummersSoftwareLLC/Primes/commit/ec584306ce7222b8f3ad426bc85d9c0cd9c75f7e)  and the one just before [cd23cae8928c13770cf56ce764d7f62c40cfdcbf](https://github.com/PlummersSoftwareLLC/Primes/commit/cd23cae8928c13770cf56ce764d7f62c40cfdcbf)

The main difference here was moving from the SplFixedArray to string but reverting to SplFixedArray I do not see improvements in the newer php versions.

# Benchmark Results

### PHP 8.1.7

|  | Passes | Execution Time (seconds) |
|:---|:------:|:------------------------------:|
| **Current Code**           | 1137 | 5.012901 |
| **Current Code**           | 1164 | 5.011611 |
| **Current Code**           | 1164 | 5.010423 |
| **Current Code**           | 1175 | 5.011783 |
| **Reverting to SplFixedArray** |  411 | 5.012219 |
| **Reverting to SplFixedArray** |  406 | 5.013436 |
| **Reverting to SplFixedArray** |  418 | 5.016499 |
| **Reverting to SplFixedArray** |  418 | 5.016009 |

### PHP 8.2.25

|  | Passes | Execution Time (seconds) |
|:---|:------:|:------------------------------:|
| **Current Code**           | 1110 | 5.012697 |
| **Current Code**           | 1110 | 5.013165 |
| **Current Code**           | 1100 | 5.012483 |
| **Current Code**           | 1111 | 5.013720 |
| **Reverting to SplFixedArray** |  391 | 5.010501 |
| **Reverting to SplFixedArray** |  400 | 5.017447 |
| **Reverting to SplFixedArray** |  380 | 5.016591 |
| **Reverting to SplFixedArray** |  393 | 5.009321 |

### PHP 8.3.13

|  | Passes | Execution Time (seconds) |
|:---|:------:|:------------------------------:|
| **Current Code**           | 1145 | 5.010931 |
| **Current Code**           | 1158 | 5.011457 |
| **Current Code**           | 1136 | 5.010918 |
| **Current Code**           | 1157 | 5.013485 |
| **Reverting to SplFixedArray** |  412 | 5.015968 |
| **Reverting to SplFixedArray** |  401 | 5.005544 |
| **Reverting to SplFixedArray** |  408 | 5.013197 |
| **Reverting to SplFixedArray** |  414 | 5.009879 |


I can not see many other changes to take into account, what are your benchmarks @ralfstrobel ?

ralfstrobel:
@DennisdeBest as described in my [original comment above](#issuecomment-2442123547), the new jit optimizations do not seem to affect SplFixedArray but only native arrays.

PR #847, opened 2022-07-07
Optimization and updates to Lua #3 (~7043 -> ~8731)

Mooshua:
## Description
Improve performance and QOL of Lua solution 3, and added several more options/benchmarks.

* Improved cache performance by not allocating space for even numbers
* Added experimental marking method which aims to break up dependency chains (and likely will run like a sloth in prod :/)
* Added jitanalysis option to benchmark with certain optimizations disabled
* Improved test script
* Removed useless opt tuning (close enough to error margin that it's practically useless)
* Compiled scripts are now in the compiled directory for ease of browsing (three are included for visibility on github)
* Updated readme

## Contributing requirements


rbergen:
@Mooshua Thanks for submitting this.

I noticed that according to your example output, 3 implementations are _slower_ with the changes applied than before. Those implementations being mooshua_luajit_slow_ffi, mooshua_luajit_vm_ffi and mooshua_luajit_vm_hash. I'm willing to merge this PR regardless because you're augmenting your own solution, but I just wanted to double-check if this is indeed what you want to do.

Mooshua:
[`slow_ffi`'s degrade is caused by several dead guards (about 6) winding their way into JIT code as a result of using `bit.rshift`.](https://gist.github.com/Mooshua/2449fb99c22cd0604d626edfb4e04083) I've removed bit.rshift from this benchmark completely, despite a patch fixing all but 1 of the dead guards, simply because it was faster to do so.

`vm_ffi` and `vm_hash` were invoking functionality from the LuaJIT standard library, which included things like conversion which bloated up the loop. This has been fixed by simply not doing that.

Thanks for pointing this out, as quite frankly I hadn't noticed :) 
They've been patched up to about where they were before, and I think they're ready to go in now.

rbergen (approved):
You're welcome! 

LGTM

PR #848, opened 2022-07-08
PrimeBun - JavaScript and TypeScript

DiefBell:
## Description
JavaScript and TypeScript implementations running on the [Bun runtime](https://bun.sh). Taken from the existing JavaScript implementation for PrimeNodeJS.


marghidanu:
I was waiting for this 👍

marghidanu (approved):

rbergen:
I don't think we can merge this as a new language. Bun is a runtime. The language is either JavaScript or TypeScript, and in this case both. 

The situation is quite similar to the PRs that aimed to add a Deno solution (#358 and #804), which we also didn't merge as a language. As was the case then, it's possible to:

- submit JavaScript and/or TypeScript solutions that use the Bun runtime, provided that the implementations themselves are different, or
- propose changes to existing JavaScript/TypeScript solutions to use Bun. The proposed change can be to use Bun instead of the runtime(s) the solution(s) use, or to use Bun alongside their current runtime(s). In case you propose to replace, the criterion is that the new runtime is faster than the current one.

Personally, I'd choose to open a PR that runs (an) existing solution(s) within Bun in addition to their current runtime(s), to illustrate the performance benefits of the Bun runtime compared to others.

DiefBell:
Closing in favour of new PR: https://github.com/PlummersSoftwareLLC/Primes/pull/849

PR #851, opened 2022-07-13
Janet solution added

Musab-Nazir:
## Description
Added an implementation for the [janet language](https://janet-lang.org/)

## Contributing requirements


PrimeJanet/solution_1/README.md
```diff
@@ -0,0 +1,22 @@
+# Janet solution by Musab-Nazir
```
rbergen, line 1: Could you consider adding solution characteristics badges, as described in the relevant section of the contributing guidelines? I'm specifically saying this because the bit count per prime candidate is undefined (see one of my next comments), which is easier to identify when explicitly indicated by a badge.

PrimeJanet/solution_1/main.janet
```diff
@@ -0,0 +1,35 @@
...
+  [sieve-fn limit]
+  (var pass 1)
+  (let [single-pass-res     (sieve-fn limit)
+        start-time  (os/clock)
+        end-by      (+ start-time 5.0)
+        _           (while (<= (os/time) end-by)
+                      (let [primes   (sieve-fn limit)]
+                        (set pass (inc pass))))]
+    (print (string "Musab-Nazir;" pass ";" (- (os/clock) start-time) ";" 1 
+                   ";algorithm=base,faithful=yes,bits=8"))))
```
rbergen: Your benchmark output indicates that your solution uses 8 bits per prime. I've taken a look at the Janet source code and I believe every Janet variable is of the "Janet" internal type, the size of which depends on the platform Janet is built and run on and/or the build options (defines) set when compiling Janet. I therefore think that the number of bits per prime is actually undefined. That means the "bits" tag should be left out of the benchmark output.

rbergen (commented):
You've added an arch-arm64 flag file. This means that the Janet solution will only run on arm64 architecture machines. I think it would be best to remove the flag file, as Janet is clearly made to also run on the amd64 architecture.

PrimeJanet/solution_1/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM levischuck/janet-sdk
```
rbergen: Please select a specific version of your base Docker image (meaning: add a version tag to your base image specification).

PrimeJanet/solution_1/main.janet
Musab-Nazir: I've removed the tags from the print output and added the badges in the readme instead as per your earlier comment

PrimeJanet/solution_1/main.janet
```diff
@@ -23,13 +23,11 @@
...
-  (let [single-pass-res     (sieve-fn limit)
-        start-time  (os/clock)
+  (let [start-time  (os/clock)
         end-by      (+ start-time 5.0)
         _           (while (<= (os/time) end-by)
                       (let [primes   (sieve-fn limit)]
                         (set pass (inc pass))))]
-    (print (string "Musab-Nazir;" pass ";" (- (os/clock) start-time) ";" 1 
-                   ";algorithm=base,faithful=yes,bits=8"))))
+    (print (string "Musab-Nazir;" pass ";" (- (os/clock) start-time) ";" 1))))
```
rbergen: I may have caused confusion with my earlier comments, in which case I apologize. I meant to indicate that the `bits` tag should be removed from the benchmark output because the number of bits per prime candidate is undefined, but the other tag output (effectively `;algorithm=base,faithful=yes`) should remain. Otherwise, the solution will be interpreted as using an "other" algorithm and not being faithful, which is not correct.

PrimeJanet/solution_1/main.janet
rbergen: Thank you. As indicated in a new comment, the `algorithm` and `faithful` tags were correct and should be retained, however.

PrimeJanet/solution_1/main.janet
Musab-Nazir: No worries, I've added them back in :)

rbergen (approved):
LGTM

marghidanu (approved):
LGTM

PR #852, opened 2022-07-13
Fix Lua/solution_3

Mooshua:
## Description
Oops. I broke the docker build with the last push. This fixes the docker build, adds a test (and removes several faulty ones) and introduces some generous optimizations to the party.

Major Changes:

* rshift is now lifted to the top of the loop, hopefully increasing index performance (it's just an add & modRM now, yay)
* removed dependency unrolling tests (they didn't work as planned)
* Added an (i think) array test, which should hopefully use LuaJIT arrays as a backing.

## Contributing requirements


rbergen (approved):
Yes, always test your code, kids. ;)

Will proceed with merging because this once again is an edit to your own solution,

PR #854, opened 2022-07-21
JavaScript/solution_1: Remove unnecessary pinned down versions

rbergen:
This removes the version specification for `zip`, `unzip` and `nodejs` from the JavaScript/solution_1 `Dockerfile`. The rationale is as follows:
- `zip` and `unzip` are stable generic tools, updates to which are unlikely to break their current usage
- `nodejs` is acquired from NodeSource, which regularly updates the versions it provides. The explicit pulling in of an 18.x version (which is already a type of version pinning) should be enough.

The pulling in of a specific version of `curl` is maintained; the version that is pulled in without a version specification complains about one of its options (`S`) not making sense in conjunction with the other ones.  

I've also:
- bumped the V solutions to V version 0.3, and Alpine 3.16 to fix another periodic build breakage. Let's hope version 0.3 will prove less brittle than the versions we've seen so far.  
- made a small change to V solution_1, to address an error the new V version started issuing when comparing int with u64.

### DIFF
--- a/PrimeJavaScript/solution_1/Dockerfile
+++ b/PrimeJavaScript/solution_1/Dockerfile
@@ -7,15 +7,12 @@ WORKDIR /opt/app
 SHELL [ "/bin/bash", "-o", "pipefail", "-c" ]
 
 RUN apt-get update &&\
-    apt-get install --no-install-recommends -y \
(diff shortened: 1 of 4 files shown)


marghidanu (approved):
LGTM

PR #861, opened 2022-09-29
PrimeV: bump V version to weekly.2022.39

rbergen:
V did its regular thing of breaking on an updated base image. This bumps its version to weekly.2022.39, which seems to work on Alpine 3.16, for now.


marghidanu (approved):
LGTM

PR #865, opened 2022-10-27
Cache-optimized variants in PrimeLua/solution_3

Mooshua:
## Description

Added several variants to the traditional linear sieve which iterate over the `is_prime` array in a way which reduces cache evictions. There are 8 new benchmarks, each with different parameters (which should cover most cpu-to-cpu divergence). So far I'm seeing at least a 2x improvement on most benchmarks!

From the `README`:
> Cache optimized versions (ending in `_c--k`) use a non-linear sieve algorithm, which is broken into two steps:
> 1. A naive, linear sieve goes over the first `sqrt(size)` numbers, putting the ones that are prime into a list,
> 2. A non-linear sieve goes over every "block" of memory, and within each block sieves out the primes that stage 1 collected.
> 
> This reduces cache evictions as each block of memory is processed once, instead of `sqrt(size)` times like the standard linear sieves.

I've used the verification tools included in the solution, so to the best of my knowledge this new sieve type is fully correct. And yes, I've tested the dockerfile this time ;)

**Results on an Intel i9-9900**:
```
--   Cache-optimized: 20206
mooshua_luajit_16_c48k;20206;5.001;1;algorithm=base,faithful=no,bits=8
--   Linear: 8721
mooshua_luajit_16;8721;5.001;1;algorithm=base,faithful=no,bits=8
```

## Contributing requirements


Mooshua:
I should note, I'm not sure that this solution still qualifies as unfaithful. 

The only disqualifying aspect (that I remember) is that it does not re-create a class structure for every iteration, and simply reuses the same configuration variables in a loop. ([see for yourself](https://github.com/Mooshua/Primes/blob/dc94df32c9a6610cea9a79267e6d5e35223429e1/PrimeLua/solution_3/compiled/mooshua_luajit.lua#L15))

Given this is the only concern, can I bump it up to faithful or do you folks think it's better to keep it as-is?

rbergen:
The full build-up (and tear-down) of the memory structure ("class") used for the sieve/sieving is one of the most fundamental requirements for a faithful implementation, and has been that from the very beginning of the Primes project. In this case, the specific implementation aspect that goes against this rule most still is the reuse of the main sieve buffer (`ARENA`) across sieve runs.

Phrased differently, since the last time we discussed this the requirements for a faithful implementation haven't changed, and the relevant aspects of this solution's implementations haven't either. Therefore, the conclusion about this solution's unfaithfulness has to remain the same.

Mooshua:
Yep, that was my reasoning when I first created the solution. Thanks for the clarification :)

PrimeLua/solution_3/l1_sieve.lua
```diff
@@ -0,0 +1,254 @@
...
+
+                local k2 = k*2
+                
+                for x = (k*k), PRIME_LEN, k2 do
+
+                    local index = %s
+
+                    ARENA[ index ] = 1
+
+                end
```
rbergen, line 87: I'm not sure I understand why this logic is (still) in. Isn't it effectively duplicated by the unrolled writes in (the compiled output of) the code below?

Mooshua: It's "duplicated" but in reality it reduces the number of primes recorded by the first stage of the loop, and thus the time spent in the second stage.

If this wasn't done, then the per-block iteration of the sieve would run 1000 factors (`sqrt(SIZE)`), while with this code it will only run 168 (all primes below `sqrt(SIZE)`) for a size of 1 million.

rbergen: Yes, I see what you're saying; the output of the code limits the number of cycles of the loop this is in. But in that case, couldn't the second stage be optimized by setting the `last_value` for a factor to the last value set to 1 by this fragment? That way the second stage will only start flagging multiples from the point where the first stage ended.

Mooshua: Interesting. When I first wrote this I tried this exact technique out and it ran slower than baseline, but just now I've gone in and re-done it it's running *faster*, by a fair margin. I've pushed a commit that has this now--Thanks for the feedback 😄

rbergen: To be honest, the change you made is not exactly what I meant, but great that you found another optimization that did work. I was just thinking out loud, anyway; it took me a little while to figure out what your algorithm was doing, and once I did I couldn't help mulling over it for a bit. Chances are you already did try what I suggested and found it slowed things down instead of speeding them up, as you said.

Anyway, I'll proceed with CI and merging this in its current state then.

Mooshua: OH, I see what you mean now. I’m very sorry it took me three swings at that :(. I implemented this and it interestingly was well within standard deviation (as in, it was both slower and faster than baseline). 
I didnt look too much into it, but I suspect this is due to LuaJIT having to compile three traces for this: One to start the loop, one to mark the non-primes, and one to end the loop—each coming with its overhead of guards. I’ll have to take some time to look into it though and confirm it is the case. Very good eye though!


Thanks 😄

PR #870, opened 2022-12-02
Adding Terra programming language to the race

Enter1he:
## Description
A solution for counting number of primes up to 1 million. Made with Terra programming language

## Contributing requirements


rbergen:
@Enter1he Thank you for submitting this. To allow inclusion of the solution in the drag race, please:

- Have your solution output a result line in drag race format, as specified in [the Output section of the contributing guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md#output).
- Add a Dockerfile to build and run your solution, in accordance with [the relevant section of the contributing guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md#dockerfile).

Enter1he:
Added Dockerfile and hopefully fixed the output. Should be correct now

PrimeTerra/solution_1/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM ubuntu:20.04
+
+WORKDIR ./
+
+RUN wget https://github.com/terralang/terra/releases/download/release-1.0.6/terra-Linux-x86_64-6184586.tar.xz &&
+    mkdir Terra &&
+    tar -C "Terra" -xvf terra-Linux-x86_64-6184586.tar.xz --strip-components 1 && 
+    PATH=~/Terra/bin:$PATH &&
+    terra Primes.t
```
rbergen: This Dockerfile is incomplete and doesn't work. Please make sure you have tested both the `docker build` and `docker run` with your Dockerfile before (re)submitting the file as finished.

PrimeTerra/solution_1/Primes.t
```diff
@@ -0,0 +1,155 @@
...
+        ct = c.clock() - start
+        pass = pass + 1
+        if ct/1000 >= dur then
+           
+            break;
+        end
+    end
+    c.printf("Computing primes to 1000000 on 1 thread for 5 seconds.\n\
+ Passes: %d, Time: %f, Avg: %f, Limit: %d, Count: %d, Valid: %d\n", pass, ct*0.001, 0.0, N, p.res, 64)
+    c.printf("Enter1he;%d;%f;64;algorithm=other,faithful=no,bits=1", pass, ct*0.001)
```
rbergen: This line of output reports:
- the thread count as 64. I believe that should be 1.
- the number of bits per prime as 1. I believe that should be 64.

rbergen (commented):
Thank you for the changes. However, I believe a few more will be necessary.

Enter1he:
Well, I changed Dockerfile and output. However, I can't test it on pure Linux machine, so test results are from Windows. Dockerfile were tested on wsl and it works fine, but I need additional tests on Linux to actually get the speed of my program.

rbergen:
I've tested this version on my machine. The Dockerfile now works, but can be condensed quite a lot. Before we get to that though, there is a more pressing issue: the solution itself doesn't work. This is the output I get when I run the solution using Docker:

```text
Computing primes to 1000000 on 1 thread for 5 seconds.

 Passes: 7, Time: 5.784000, Avg: 0.826286, Limit: 1000000, Count: 0, Valid: 0
Enter1he;7;5.784000;1;algorithm=other,faithful=no,bits=64
```

The issues are:

- The runtime claims to be close to 6 seconds, but actually clearly isn't. The solution runs _much_ shorter than that.
- The reported prime count is 0.
- The output itself indicates that the result isn't valid.

Enter1he:
Yep, final fix - the problem was clock function. Because originally written on Windows - clock was a real big mistake to use(WIndows handles it different) Now even in wsl everything works fine.

rbergen:
It seems your Dockerfile no longer works. I believe this is because you changed the name of the executable file from "Sieve" to "Sieve.exe", and did not update your Dockerfile accordingly.

Disregarding that for a second, I took a closer look at your program's logic and I'm sorry to say we cannot accept this solution in its current form. What the program seems to do is run a sieve once (I say "a" sieve because I have difficulty identifying the algorithm as the Sieve of Eratosthenes). It then reports:

- How many cycles in the main logic loop it took to find the primes as the number of passes
- The time it took to complete the single sieve

That's too far of a deviation from the contributing guidelines to be admissible as a solution for this project.

Some of the basic rules that apply to solutions in this repo are:

- They must use the Sieve of Eratosthenes as the sieving algorithm
- A "pass" is the completion of a full sieve. In other words, the identification of all prime numbers up to 1000000 is **one** pass
- The program must execute complete passes, until at least 5 seconds have passed. What must be reported is the number of _full passes_ the program completed in the "just over" 5 seconds, and the exact time it took to complete them.

If you would still like to submit a solution in the Terra language, I encourage you to read the [contributing guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md) in full. Then, please make sure your solution complies with the guidelines and works correctly in a Linux-based Docker container before requesting a(nother) review.

Enter1he:
I still need some help to classify my new solution. Can it be classify as faithfull? And is my bit array can be classified as is_prime array or I should specifically make a boolean one? I don't see any of those in PrimeCPP solution_1 on which I'm usually trying to compare mine.

rbergen:
It looks to me like your solution now creates and destructs the Primes struct with every pass, which is an important criterion for a faithful solution. With that in place, I do think your solution can be considered faithful. 
Your bit array can indeed be considered a suitable is_primes array as it is, so returning it will suffice to meet that requirement.

PrimeTerra/solution_1/Primes.t
```diff
@@ -0,0 +1,165 @@
...
+            i = check - 1
+            if self:get_subOne(i) == 1 then
+                for n = check+check, N+1, check do
+                    self:set_subOne(n - 1)
+                    i = i + 1
+                end
+            end
+            check = check + 1
+        end
+        self:count()
```
rbergen: I'm not sure I understand how this code actually returns `self.arr`?

PrimeTerra/solution_1/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN wget https://github.com/terralang/terra/releases/download/release-1.0.6/terra-Linux-x86_64-6184586.tar.xz -P /tmp/
+
+RUN tar -xvf /tmp/terra-Linux-x86_64-6184586.tar.xz -C /tmp/
+
+RUN adduser --disabled-password --gecos '' docker
+RUN adduser docker sudo
+
+
+RUN apt-get update
+RUN sudo apt-get install -y build-essential llvm-11-dev libclang-11-dev clang-11 libedit-dev libncurses5-dev zlib1g-dev libpfm4-dev
```
rbergen: Is there a reason not to include these packages in the first `apt-get install`? It would save another `apt-get update` and `apt-get install`.
Also, the use of sudo is not necessary because the `USER` statement only takes effect from the point it is used. In fact, I don't think the installation of sudo is needed at all if `USER` is used just before `CMD`, as it is now.

PrimeTerra/solution_1/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+
+
+COPY *.t ./
+RUN sudo /tmp/terra-Linux-x86_64-6184586/bin/terra Primes.t
```
rbergen: Same comment as before.

PrimeTerra/solution_1/Dockerfile
rbergen: Our Dockerfile linter will object to the use of multiple `RUN` statements in a row in an effort to reduce the number of image layers. Please consolidate the sequence of `RUN` statements into one:

```Dockerfile
 RUN apt-get update && apt-get install -y wget tar xz-utils sudo \
  && rm -rf /var/lib/apt/lists/* \
  && wget https://github.com/terralang/terra/releases/download/release-1.0.6/terra-Linux-x86_64-6184586.tar.xz -P /tmp/ \
  && tar -xvf /tmp/terra-Linux-x86_64-6184586.tar.xz -C /tmp/ \
  && ...
```

etc.

PrimeTerra/solution_1/Primes.t
Enter1he: It doesn't. Not yet, apparently. Should it return it after running a sieve?

Enter1he: Because I'm actually printing results in function draw() And it can be run in printResults() if boolean value true. I'm just following CPP solution here

PrimeTerra/solution_1/Dockerfile
Enter1he: Now fixed. Originally didn't know how to write it either way.

PrimeTerra/solution_1/Primes.t
```diff
@@ -0,0 +1,170 @@
...
+macro Primes:printResults(printNums, pass, ct)
+    return quote
+        if printNums then
+            self:draw()
+        end
+        self:count()
+        var ti = ct*clocks
+        c.printf("Computing primes to 1000000 on 1 thread for 5 seconds.\n\
+ Passes: %d, Time: %f, Avg: %f, Limit: %d, Count: %d, Valid: %d\n", pass, ti, ti/pass, N, self.res, self:validate())
+        c.printf("Enter1he;%d;%f;1;algorithm=base,faithful=yes,bits=1\n", pass, ti)
```
rbergen: This solution can't be marked as faithful. The reason is that it uses a C function (i.e. an external dependency) for memory allocation, which is part of the benchmarked code and as such instrumental to the calculation of the sieve.

rbergen (commented):
Please address the error found by the Dockerfile linter.

Enter1he:
I added workdir to dockerfile.

rbergen:
Thank you, I've taken the liberty of making it an absolute directory (`/opt/app` instead of `opt/app`). CI is now running.

rbergen:
LGTM. As this is a new solution (in a new language) I will request a review from the second maintainer as well.

rbergen (approved):
LGTM. As this is a new solution (in a new language) I will request a review from the second maintainer as well.

marghidanu (approved):

PR #871, opened 2022-12-03
NVIDIA CUDA solution

rbergen:
## Description
This is a solution based on the use of NVIDIA CUDA, meaning it requires the use of a CUDA-capable NVIDIA GPU.

Note that:

- I considered making this a C++ solution, but chose otherwise because "CUDA C++" includes some specific extensions that are critical to make the language work. This is reflected by the solution requiring a CUDA-specific compiler (NVCC) to build.
- A Dockerfile has not been provided. This is because the solution requires an NVIDIA GPU to be present at runtime (which is not a given), and because making the solution work requires some setup-specific steps/modifications.

## Contributing requirements

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "PrimeCUDA/solution_1/CUDASieve"]
+	path = PrimeCUDA/solution_1/CUDASieve
+	url = https://github.com/curtisseizert/CUDASieve.git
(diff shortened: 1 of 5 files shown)


marghidanu (approved):

davepl (commented):
Nice work!  And we should send Curtis a gift basket or something :-)

PR #872, opened 2022-12-07
 [PrimeRuby] solution_2 using numo gem near 10x speedup

darnellbrawner:
## Description
Using numo gem ruby's flavor of numpy.
Solution uses python's numpy method.
## Contributing requirements

### DIFF
--- a/PrimeRuby/solution_2/Dockerfile
+++ b/PrimeRuby/solution_2/Dockerfile
@@ -1,6 +1,6 @@
 FROM ruby:3.1-alpine3.14
 
-RUN apk add build-base
+RUN apk add --no-cache build-base
 RUN gem install numo-narray
 
 WORKDIR /opt/app
--- a/PrimeRuby/solution_2/Dockerfile
+++ b/PrimeRuby/solution_2/Dockerfile
@@ -1,0 +1,9 @@
+FROM ruby:3.1-alpine3.14
+
+RUN apk add build-base
+RUN gem install numo-narray
+
+WORKDIR /opt/app
+COPY *.rb .
+
+ENTRYPOINT [ "ruby", "prime.rb" ]
--- a/PrimeRuby/solution_2/Dockerfile
+++ b/PrimeRuby/solution_2/Dockerfile
@@ -1,7 +1,6 @@
 FROM ruby:3.1-alpine3.14
 
-RUN apk add --no-cache build-base
-RUN gem install numo-narray
+RUN apk add --no-cache build-base && gem install numo-narray:0.9.2.1
 
 WORKDIR /opt/app
 COPY *.rb .
--- a/PrimeRuby/solution_2/README.md
+++ b/PrimeRuby/solution_2/README.md
@@ -1,0 +1,24 @@
+# Ruby solution by rbergen
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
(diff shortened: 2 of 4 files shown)


PrimeRuby/solution_2/prime.rb
```diff
@@ -0,0 +1,57 @@
...
+loop do
+  sieve = PrimeSieve.new(sieve_size)
+  sieve.run_sieve
+  pass_count += 1
+  break if (Time.now.to_f - start_time) > 5.0
+end
+duration = (Time.now.to_f - start_time).round(3)
+
+puts "WARNING: result is incorrect! #{reference_results[sieve_size]} != #{sieve.count_primes}" unless reference_results[sieve_size] == sieve.count_primes
+puts "rbergen;#{pass_count};#{duration};1;algorithm=base,faithful=yes"
```
rbergen: This output line needs to be changed. The label should feature your name/GitHub username instead of mine (I really can't take any credit for this solution), and the `faithful` tag should be set to `no`.

PrimeRuby/solution_2/README.md
```diff
@@ -0,0 +1,28 @@
+# Ruby solution by darnellbrawner
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-no-yellowgreen)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-8-yellowgreen)
```
rbergen: I believe the bit count should be "unknown" instead of 8. I've seen in the Numo source code that the underlying data type for the "bits" (`BIT_DIGIT`) in the `Bit` array is C's `unsigned int`. The bit size of that data type can vary, and will definitely be larger than 8 on any platform these solutions will run on.

PrimeRuby/solution_2/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+This is an implementation in Ruby using Numo gem ruby's version of numpy
```
rbergen: I accept this can easily be perceived as a bit pedantic, but for clarity I would write this sentence as "This is an implementation in Ruby using the Numo gem, Ruby's version of Numpy."

PrimeRuby/solution_2/README.md
```diff
@@ -0,0 +1,28 @@
...
+### Docker
+A Dockerfile has been provided.
+
+# Results on my machine
+
+ - Macbook Pro M1
+
+## Output
+```
+rbergen;1550;5.028;1;algorithm=base,faithful=yes
```
rbergen: This sample output line needs to be changed in accordance with the comment I make about `prime.rb`.

rbergen (commented):
Hi @darnellbrawner, thank you for submitting this. I think there's a few things that need to be addressed before we can move forward to merging this.

PrimeRuby/solution_2/README.md
darnellbrawner: Thank you again for volunteering your time on this fun project. I'll try to do a multi threaded pony next since that should be its stronger showing.

PrimeRuby/solution_2/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM ruby:3.1-alpine3.14
+
+RUN apk add build-base
```
marghidanu: Please review the warnings below and fix them, just to ensure we follow the best practices.

rbergen (approved):
LGTM.

> Thank you again for volunteering your time on this fun project. 

You're welcome!

marghidanu (approved):
LGTM 2

PR #873, opened 2022-12-13
Multithreaded Ruby Solution using Ractor 

darnellbrawner:
## Description
Takes advantage of multi-core systems using Ractor ruby's actor model framework.
https://ruby-doc.org/core-3.0.0/Ractor.html

## Contributing requirements

### DIFF
--- a/PrimeRuby/solution_2/.gitattributes
+++ b/PrimeRuby/solution_2/.gitattributes
@@ -1,0 +1,1 @@
+*.rb binary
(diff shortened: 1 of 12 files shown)


rbergen:
Hi @darnellbrawner, I see you're now contributing the multithreaded implementation(s) you announced earlier. You have now structured these additions as two fully independent solutions. This means that for each of your three implementations a separate Docker image is built and run when the benchmark is executed, which is not very efficient. 

May I ask you to structure your PR to add these new implementations to your existing `PrimeRuby/solution_2` solution instead? That solution can then run all three implementations in sequence, in one `docker run`. It also means we can merge without requiring a review from the second maintainer.

PR #874, opened 2022-12-13
Added MultiThreaded support using Ractor

darnellbrawner:
## Description
numo multithreaded is set at only 2 Ractors do to a flaw in numo and total number of cpu's minus 1 for none numo solution.
## Contributing requirements


PrimeRuby/solution_2/README.md
```diff
@@ -2,10 +2,13 @@
 
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Faithfulness](https://img.shields.io/badge/Faithful-no-yellowgreen)
-![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Parallelism](https://img.shields.io/badge/Parallel-yes-green)
```
rbergen, line 6: As your solution includes both a single-threaded and multi-threaded implementations, please include both the "yes" and "no" Parallel badges.

PrimeRuby/solution_2/prime.rb
```diff
@@ -1,37 +1,89 @@
...
   end
 
   def count_primes
-    @bits.count_1
+    (@prime_array.class == Numo::Bit) ? @prime_array.count_1 : @prime_array.count(true)
   end
+
 end
+
+def run_sieve(sieve_size, prime_array)
```
rbergen: Is this function used anywhere?

PrimeRuby/solution_2/prime.rb
```diff
@@ -40,18 +92,65 @@ def count_primes
                       1_000_000 => 78_498,
                       10_000_000 => 664_579,
                       100_000_000 => 5_761_455 }.freeze
+SIEVE_SIZE = 1_000_000
```
rbergen: This seems to be a duplicate of the `sieve_size` variable that was already present?

rbergen (commented):
The next two questions are just out of curiosity, so not review findings as such.

PrimeRuby/solution_2/prime.rb
darnellbrawner: good catch I was testing performance  with calling method directly vs creating a instance of class and calling method. I'll remove it.

PrimeRuby/solution_2/prime.rb
darnellbrawner: The is related to Ractor's not being allowed to access variables outside of their block except for things like constants or classified shared variables.

darnellbrawner: never mind good catch that was self over when I was using a stand alone method call right in inside the Ractor block so it needed to be a constant vs passing in instance

PrimeRuby/solution_2/prime.rb
```diff
@@ -46,12 +73,58 @@ def count_primes
...
+
+r = nil
+begin
+  r = Ractor.select(*workers)
+  pass_count += 1
+end until (Time.now.to_f - start_time) > 5.0
+duration = (Time.now.to_f - start_time).round(3)
+
+puts 'WARNING: result is incorrect!' unless reference_results[SIEVE_SIZE] == r[1].count(true)
+puts "darnellbrawner-MultiThreaded;#{pass_count};#{duration};1;algorithm=base,faithful=no"
```
rbergen: I'm sorry for only noticing this just now, but this output is incorrect for a multithreaded solution. The "1" needs to be replaced by the number of threads your solution ran on. This obviously also applies to the other multithreaded implementation's output, and I'd like to ask you to update the example output in your README.md as well.

darnellbrawner: Fixed man its a bummer Numo has a multithreaded flaw I really wanted to bump past python :-( .

PrimeRuby/solution_2/prime.rb
```diff
@@ -46,12 +73,58 @@ def count_primes
...
   pass_count += 1
   break if (Time.now.to_f - start_time) > 5.0
 end
 duration = (Time.now.to_f - start_time).round(3)
 
 puts "WARNING: result is incorrect! #{reference_results[sieve_size]} != #{sieve.count_primes}" unless reference_results[sieve_size] == sieve.count_primes
-puts "darnellbrawner;#{pass_count};#{duration};1;algorithm=base,faithful=no"
+puts "darnellbrawner-Numo;#{pass_count};#{duration};1;algorithm=base,faithful=no"
+
+cores = Concurrent.processor_count - 1
```
rbergen: Just out of curiosity, why do you not utilize all available cores?

PrimeRuby/solution_2/prime.rb
```diff
@@ -46,12 +73,58 @@ def count_primes
...
+
+r = nil
+begin
+  r = Ractor.select(*workers)
+  pass_count += 1
+end until (Time.now.to_f - start_time) > 5.0
+duration = (Time.now.to_f - start_time).round(3)
+
+puts 'WARNING: result is incorrect!' unless reference_results[sieve_size] == r[1].count(true)
+puts "darnellbrawner-MultiThreaded;#{pass_count};#{duration};#{cores};algorithm=base,faithful=no"
```
rbergen: I see you mark this implementation as not faithful. Considering the fact that no external dependencies are used for parts of the implementation that are instrumental to the calculation of the sieve, I'd agree with the multi-threaded implementation that does not use Numo being marked as faithful.

PrimeRuby/solution_2/prime.rb
darnellbrawner: I'm new to concurrent program since there hasn't been a way around GVL in ruby  until last year with Ractor. Since the main ruby program takes up its own thread so a thread per core was the thinking.   Have you run it on your AMD Threadripper machine?

rbergen: I think the main Ruby program doesn't do much of anything while it waits for the threads that do the work to finish, which means it's not actively using any of the cores. Generally, you can even create more logical threads than the CPU has available, they will just be subject to scheduling (which does decrease their use significantly). 
All other multithreaded solutions I have seen in this repo use all available cores for the actual sieving, but you are free to deviate if you want.

I'd love to run this on my Threadripper machine, if I had one. It's somewhere on my list of things to buy if I win the lottery. :) As it stands, @davepl is the owner of the Threadripper machine, and your solution will be run on that the first time it runs a benchmark after this PR is merged.

darnellbrawner: When I was testing u could add a thread and see the performance increase as expected but at the limit of threads.count == cores.count I didn't notice much of a change.  I changed it to one to one. I googled around to see if there was any free to demo high thread count boxes u could run a program on as a fun test (no luck). Not sure why also in docker mode the multithreaded solution is only using 5 threads vs 10.

PR #875, opened 2022-12-26
Add second CUDA solution

rbergen:
## Description
This is a second CUDA solution, written from scratch (as in, not depending on CUDASieve) based on an idea from @davepl.. 

## Contributing requirements


PrimeCUDA/solution_2/primes.cu
```diff
@@ -0,0 +1,233 @@
...
+    uint64_t passes = 0;
+
+    auto startTime = steady_clock::now();
+    duration<double, std::micro> runTime;
+
+    Sieve *sieve = nullptr;
+
+    do
+    {
+        if (sieve)
```
davepl: It's more clear with it, but you don't need this test because delete nullptr is just fine, so you can always just delete the sieve if you like.

PrimeCUDA/solution_2/primes.cu
```diff
@@ -0,0 +1,233 @@
...
+            sieveSize,
+            primeCount,
+            countValidated);
+
+    printf("rbergen_faithful_cuda;%zu;%f;1;algorithm=base,faithful=yes,bits=1\n", passes, duration);
+}
+
+int main(int argc, char *argv[])
+{
+    uint64_t sieveSize = determineSieveSize(argc, argv);
```
davepl: Make me const, though the compiler likely does anyway.  I just marked this example, there are several, but it's a style issue you want to or not!

PrimeCUDA/solution_2/primes.cu
```diff
@@ -0,0 +1,233 @@
...
+#include <cstring>
+#include <cmath>
+#include <chrono>
+#include <map>
+#include <cuda_runtime.h>
+
+using namespace std::chrono;
+
+#define DEFAULT_SIEVE_SIZE 1'000'000
+#define THREAD_COUNT 128
```
davepl: Why 128 and not 64 or 256?  How do we pick a thread count?

PrimeCUDA/solution_2/primes.cu
```diff
@@ -0,0 +1,233 @@
...
+        cudaDeviceSynchronize();
+    }
+
+    ~Sieve() 
+    {
+        cudaFree(device_sieve_buffer);
+        free(host_sieve_buffer);
+    }
+
+    uint32_t *run()
```
davepl: I'm sure you know what you're doing, but why halfsize?  Don't you have to eliminate the multiples in the top half too?  What am I forgetting about the algorithm here?

PrimeCUDA/solution_2/primes.cu
```diff
@@ -0,0 +1,233 @@
...
+                for (index = (factor * factor) >> 1; index <= lastMultipleIndex; index += factor)
+                    host_sieve_buffer[WORD_INDEX(index)] &= ~(((uint32_t)1) << BIT_INDEX(index));
+            }
+        }
+
+        // Copy the first (square root of sieve size) buffer bytes to the device
+        cudaMemcpy(device_sieve_buffer, host_sieve_buffer, (size_sqrt >> 4) + 1, cudaMemcpyHostToDevice);
+        // Allocate device buffer for the list of primes and copy the prime list to it
+        uint32_t *devicePrimeList;
+        cudaMalloc(&devicePrimeList, primeCount * sizeof(uint32_t));
```
davepl: Check for cudaSuccess?

PrimeCUDA/solution_2/primes.cu
rbergen: Good point. I'll choose simplification over algorithmic clarity and change this. It's not a key part of the core algorithm anyway.

PrimeCUDA/solution_2/primes.cu
rbergen: No, I agree with you. It's not just a style issue in my view: a variable that should not be changed should be "protected" from mistakes by making it const. I will change this throughout.

PrimeCUDA/solution_2/primes.cu
rbergen: I tested with 256 and that didn't make any difference in overall speed on my own machine, even for larger sieves (up to 10 billion). 64 seems a bit low for sieves of the larger size. But yes, I did say "seems". If anyone knows of a more empirical way to decide what the best number of threads for this solution is (and maybe even derive it from the sieve size) then I'd love to learn about it.

PrimeCUDA/solution_2/primes.cu
rbergen: The code this comment is linked to is the `free`ing of the host sieve buffer, but I guess you're inquiring about the existence of the `half_size` member variable and its use as the basis for "everything memory". 
This is because we only care about odd prime candidates, as we know no even number besides 2 can be a prime and are allowed to use that knowledge as per this repo's rules. Therefore, every size and actual index throughout the algorithm is halved. In many places this is done by a "sneaky bit shift" to the right, but here I've actually given it a name.

PrimeCUDA/solution_2/primes.cu
rbergen: I should. Across the board. As I should check the return value of the normal `malloc` as well. However, writing this solution in a "production-ready" fashion will really bloat the code and "distract" from what this solution is trying to demonstrate. If it's okay with you, I'll accept that this program will die an ungraceful death if it's run with a sieve size so big that memory allocation fails. :)

PrimeCUDA/solution_2/primes.cu
rbergen: With the block method added, 256 threads is indeed faster than 128. I have thus made 256 the set number.

davepl:
Looks great!  It might be nice to comment each function like markmultiples to just summarize what they do for the casual reader, but otherwise I couldn't spot anything!

PR #880, opened 2023-01-03
Initial Algol 68g implementation

rzuckerm:
## Description
Initial Algol 68g implementation.

## Contributing requirements


PrimeAlgol68g/solution_1/README.md
```diff
@@ -0,0 +1,34 @@
+# Algol 68g solution by rzuckerm
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-no-yellowgreen)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-1-green)
```
rbergen: Are you sure that a member of an Algol 68g BOOL array indeed occupies one bit of memory space? Note that the bit count should reflect the number of bits actually _used in memory_, not how many are _needed from a binary/logical perspective_.

PrimeAlgol68g/solution_1/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+Faithfulness is *no* due to not being able to run a sieve size greater than 100 million.
```
rbergen: That is actually not the reason to mark this solution as not faithful, as the sieve size used in the benchmarks is 1 million. However, there is another reason why it is the correct qualification. The contributing guidelines state that a faithful solution "_...uses a class to encapsulate the sieve, or a (closest) equivalent feature in your language if a class construct is not available. This class must contain the full state of the sieve. Each iteration should re-create a new instance of this class from scratch._"

That's not the case with this solution. Embedded in the rule I just quoted is the requirement that the memory buffer used for the sieve must be allocated and released for every iteration, which is also not happening. 

So, the "not faithful" qualification is appropriate, but for a different reason than the README now states.

PrimeAlgol68g/solution_1/run.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+docker run --rm -it primes_algol68g:latest
```
rbergen: Do note that the benchmark runs the containers without the `-it` options. Please make sure your Dockerized solution also works without them.

PrimeAlgol68g/solution_1/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+if [ -n "$(docker images -q primes_algol68g 2>/dev/null)" ]
+then
+    docker rmi primes_algol68g:latest
+fi
```
rbergen: Is there a specific reason to remove any existing image before building?

rbergen (commented):
@rzuckerm Thank you for submitting this. I do have a few comments and questions.

PrimeAlgol68g/solution_1/README.md
rzuckerm: I'm looking into this. The documentation available for Algol 68 is rather sparse since this language isn't widely used anymore.

PrimeAlgol68g/solution_1/README.md
rzuckerm: I'm having a hard time finding out any information about how Algol knows when the free memory. That's the reason why I allocated the memory only once. Also, Algol doesn't support classes. However, it does support the equivalent of a C struct. I tried that, but the performance was severely degraded. I'll look into that some more.

PrimeAlgol68g/solution_1/run.sh
rzuckerm: I'll remove the `-it`. I believe it should still work.

PrimeAlgol68g/solution_1/README.md
rbergen: I'm well aware Algol 68 was more widely used "some time ago", which is actually also why I like that you're submitting a solution for it. There is the option of marking the number of bits as "unknown" if it is impossible to find an unequivocal answer to this. It would definitely not be the first time that happens.

PrimeAlgol68g/solution_1/build.sh
rzuckerm: I wanted to have a clean environment every time I ran build. but I can remove that.

PrimeAlgol68g/solution_1/README.md
rbergen: I understand the point about the classes, but it's been well documented that the "class or equivalent" rule still applies; we acknowledge that this means that a "faithful" solution cannot be written in certain languages. 

Concerning the degraded performance, and this may sound a bit odd: that is basically the point, to an extent. In many languages and platforms, memory allocation, initialization, deallocation and object construction take a meaningful amount of time. That's why we've concluded that they need to be included for a "fair comparison" between languages and implementations.

PrimeAlgol68g/solution_1/run.sh
rbergen: Thank you. I don't expect any issues myself. I can remember only one solution where things broke without `-it`. That was [COMAL/solution_2](https://github.com/PlummersSoftwareLLC/Primes/tree/drag-race/PrimeCOMAL/solution_2), I needed to apply some interesting trickery in the build and run scripts using `timeout` and `socat` to get that one to work. Which is to say that if you do run into problems, it's likely we'll be able to work around those.

PrimeAlgol68g/solution_1/build.sh
rbergen: No, it's fine in principle. It's just that our automated build and benchmark setups don't do this, so your solution can't rely on this step being performed, to make it work.

rzuckerm:
@rbergen I think I addressed all your comments. Please approve the workflow so that I can see if anything else needs to be changed. Thanks!

rbergen:
> @rbergen I think I addressed all your comments. Please approve the workflow so that I can see if anything else needs to be changed. Thanks!

~~I don't think I agree, for the following reasons:~~

- ~~You've now marked your solution as faithful, which I explained it (currently) isn't. As it stands, this will need to be reverted.~~
- ~~I don't think we've come to a conclusion about the appropriate number of bits per prime yet (1, another number, or unknown).~~

My apologies, I overlooked some of the changes you made when I wrote the previous text. My actual response is that I don't think we're there yet, because a CHAR definitely does not have a bit size of 1, but at least 8.

A question I have is if you've definitely established that a "delete" of an instantiated struct is not something you can trigger?

I'd (still) really like to have a "semifinal" version of the solution (i.e. one without open topics) in place before I trigger CI. Otherwise, it will have to run after every change that is made, even if it only concerns documentation.

rzuckerm:
> My apologies, I overlooked some of the changes you made when I wrote the previous text. My actual response is that I don't think we're there yet, because a CHAR definitely does not have a bit size of 1, but at least 8.

I change the bit size to 8

> A question I have is if you've definitely established that a "delete" of an instantiated struct is not something you can trigger?

Algol automatically implements garbage collection, so there is no way to directly trigger a delete. However, garbage collection can be forced with the `sweep heap` function. I have added this. I also added some heap debugging to show how many bytes were recovered due to garbage collection in the main loop. This is currently disabled, but I ran it once with it enabled, and here's what is showed:

```
                                  +0
                            +4000664
                            +4000664
                            +4000664
                            +4000664
                            +4000664
...
```

rbergen:
I really appreciate you triggering the garbage collection manually, but it's not necessary to take it to that level. We don't demand this of any other managed memory language/platform either. At the same time, I think your debug run actually shows that the number of bits per prime is not 8, but 32; the `sweep heap` function seems to clear a little over 4 million bytes for 1 million prime candidates. Which makes sense if the CHAR type actually represents UTF-32 characters.

I'd like to propose the following:

- Switch the prime candidate type back from CHAR to BOOL.
- Use the `sweep heap`/heap debugging combination to determine the size of the BOOL type. Base the bit count in README and program output on that.
- Remove the `sweep heap` and anything related to heap debugging from the source code.
- Commit one more time with these changes in place.

After these changes have been made, I'm happy to trigger CI and move forward towards merging this.

rzuckerm:
I changed back to `BOOL`, and the heap didn't change by much:
```
4000896
```

Since I only store odd values, there are only 500000 bins. Based on this, it looks like it might be 64-bits, which seems really wasteful. :frowning_face:

rzuckerm:
I'm just going to use `unknown` and be done with it. I suspect, the value changes based on machine word size, which is 64 for my case.

rbergen:
It would be wasteful. It could just be that under the hood, every integer type is implemented as a "long int". I have seen that before. You may also be right that it is machine dependent.
I'm happy to accept your findings as support for a prime candidate bit size of 64, but you can also choose to stick to "unknown" as you seem to prefer. Do note that in the benchmark output line that the program outputs, the "bits" flag then needs to be removed. Its absence in the output is the indication that the actual bit size is unknown.

rbergen:
I've taken the liberty of removing the flag from the benchmark output myself. I hope that's ok. It does come with the benefit that CI is triggered immediately.

rzuckerm:
Looks like the CI worked. Thanks for your help @rbergen !

rbergen (approved):
LGTM. As this is a new solution (and in a new language) I will ask for a second review from the other maintainer.

marghidanu (approved):
LGTM

PR #881, opened 2023-01-05
Fix some incorrect assignments with BOOL version and add BITS version

rzuckerm:
## Description
In the original implementation that uses `BOOL`, there were some `=` operators that should have been `:=`. An equivalent implementation uses the `BITS` data type has been added for comparison. This is 10x slower, but I thought it would be interesting to have. If this should be made a separate solution, please let me know.

## Contributing requirements

### DIFF
--- a/PrimeAlgol68g/solution_1/Dockerfile
+++ b/PrimeAlgol68g/solution_1/Dockerfile
@@ -7,8 +7,8 @@ RUN apt-get update -qq \
 
 WORKDIR /opt/app
 COPY primes*.a68 run_primes.sh ./
-RUN a68g -O3 --compile primes.a68
-RUN a68g -O3 --compile primes_bit.a68
+RUN a68g -O3 --compile primes.a68 && \
+    a68g -O3 --compile primes_bit.a68
 
 ENTRYPOINT [ "./run_primes.sh" ]
 
--- a/PrimeAlgol68g/solution_1/Dockerfile
+++ b/PrimeAlgol68g/solution_1/Dockerfile
@@ -6,8 +6,9 @@ RUN apt-get update -qq \
     && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /opt/app
-COPY primes.a68 .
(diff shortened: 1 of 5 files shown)


rbergen:
Really cool to have a base algorithm, faithful, 1 bit solution in this language! That means Algol will now be one of the languages with a solution in the "core set"; that being the ones that match the characteristics of Dave's original 3 implementations that started this whole project.

rbergen:
Hadolint found one issue in your Dockerfile (consecutive RUNs due to the added build for the BITS version), as indicated in the changed files. If you can please fix this then I think we're good to go,

rzuckerm:
> Hadolint found one issue in your Dockerfile (consecutive RUNs due to the added build for the BITS version), as indicated in the changed files.

A script that a developer can run locally to do this linting would be nice to have. The `docker run` command is rather cumbersome.

rbergen:
Personally I think it's a copy/paste from the contributing guidelines, with two pretty straightforward edits. But nothing is stopping you from opening a PR that adds a script like the one you suggest in the `tools` directory, that uses the `docker run` in a parameterized fashion. :)
I would not like to extend the list of software prerequisites for this, or (worse) force a local install of Hadolint.

rzuckerm:
> Personally I think it's a copy/paste from the contributing guidelines, with two pretty straightforward edits. But nothing is stopping you from opening a PR that adds a script like the one you suggest in the `tools` directory, that uses the `docker run` in a parameterized fashion. :)
> I would not like to extend the list of software prerequisites for this, or (worse) force a local install of Hadolint.

The intent is to just add a simple bash script that you can give two params: language and solution number. I'll do a PR for this later.

rbergen (approved):
LGTM

rbergen:
> The intent is to just add a simple bash script that you can give two params: language and solution number. I'll do a PR for this later.

I understand. If you do put a PR together, please include an updated version of CONTRIBUTING.md that replaces the `docker run` with the use of your script.

In the mean time, I will merge this one.

PR #882, opened 2023-01-06
Improve speed of Algol bits by about 1.6

rzuckerm:
## Description
Clearing the bits in a loop improved the speed from 14 to 23 passes (about 1.6 times faster).

## Contributing requirements


rbergen:
Ha, that approach looks familiar. I recently applied a version of it in my second CUDA solution. :)

I'll approve and merge if CI passes.

rbergen (approved):
LGTM

PR #883, opened 2023-01-07
Add script to run hadolint

rzuckerm:
## Description
Add script to run hadolint

## Contributing requirements


rbergen (commented):
Thank you for submitting this.

I think it would be good to explain that the "language" needs to precisely match the spelling (as it stands, including case) of the respective "PrimeXxx" directory. Personally, I think it would be even better if the script was written such that the language name match was not case-sensitive. Meaning that "cpp" should work as well as "CPP", to match the tool's "usage" instruction's example.

rbergen (approved):
LGTM

PR #885, opened 2023-01-15
Add Euphoria solution

rzuckerm:
## Description
Implementation in [Euphoria](https://en.wikipedia.org/wiki/Euphoria_(programming_language))

## Contributing requirements

### DIFF
--- a/PrimeEuphoria/solution_1/primes.ex
+++ b/PrimeEuphoria/solution_1/primes.ex
@@ -1,0 +1,129 @@
+include std/io.e
+without type_check
+
+enum SIEVE_SIZE, SIEVE
+constant TRUE = 1
+constant FALSE = 0
+
+function run_sieve(integer sieve_size)
+    integer num_bits = floor((sieve_size - 1) / 2)
+    sequence sieve = repeat(TRUE, num_bits)
+    integer q = floor(sqrt(sieve_size) / 2)
+    for bit = 1 to q
+    do
+        if sieve[bit] = TRUE
+        then
+            for k = 2 * bit * (bit + 1) to num_bits by 2 * bit + 1
+            do
+                sieve[k] = FALSE
+            end for
+        end if
+    end for
+
+    return {sieve_size, sieve}
+end function
+
+function count_primes(sequence this)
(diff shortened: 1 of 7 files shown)


PrimeEuphoria/solution_1/primes.ex
```diff
@@ -0,0 +1,129 @@
...
+    do
+        if this[SIEVE][bit] = TRUE
+        then
+            count += 1
+            if show_results = TRUE
+            then
+                printf(STDOUT, "%d, ", 2 * bit + 1)
+            end if
+        end if
+    end for
```
rbergen: I'm wondering if it would make sense to use the count_primes function to count the primes, instead of duplicating its code here? (The question also applies to primes_bit.ex, but I'll only ask it here.)

rbergen (commented):
Thanks for submitting this. I genuinely always like it when a solution is submitted in a language I've never heard of. I do have one question that applies to both implementations.

PrimeEuphoria/solution_1/primes.ex
rzuckerm: I could do that, but this is the same approach as the original C++ implementation [countPrimes](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/PrimeCPP/solution_1/PrimeCPP.cpp#L145-L152) and [printResults](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/PrimeCPP/solution_1/PrimeCPP.cpp#L117-L126). It's also the same approach that I took for the Algol implementation.

rzuckerm: @rbergen Never mind, I made the suggested change. I can make the same change to the Algol implementation if you'd like

rbergen: Thank you. Concerning the Algol solution, I'll leave it up to you if you want to implement the same change. As I said before on another topic, the Algol solution is already in the codebase as it is.

rbergen (approved):
LGTM

marghidanu (approved):
LGTM 2

PR #887, opened 2023-01-17
Add Phix solution

rzuckerm:
## Description
Add Phix solution. [Phix](http://phix.x10.mx/) is based on [Euphoria](https://en.wikipedia.org/wiki/Euphoria_(programming_language)), so this code is almost identical to the Euphoria solution. However, it does perform differently. See the README in this PR for further details. Also, the installation process in the `Dockerfile` is really bizarre since this language isn't available on docker hub or as any type of Linux package.

## Contributing requirements

### DIFF
--- a/PrimePhix/solution_1/Dockerfile
+++ b/PrimePhix/solution_1/Dockerfile
@@ -21,6 +21,6 @@ RUN apt-get update -qq \
     && echo "" | p
 
 WORKDIR /opt/app
-COPY primes.ex ./
+COPY primes*.ex run_primes.sh ./
 
-ENTRYPOINT ["p", "primes.ex"]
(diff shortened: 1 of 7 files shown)


rbergen:
Interesting! I see your point about the Dockerfile. We do have some other jewels in the repository, although mostly indirectly. That is to say that we create separate Dockerfile repos for the more complex/time consuming Docker base image builds, and post the resulting images on Docker Hub. I think the [Dockerfile for Chapel](https://github.com/PlummersSoftwareLLC/Chapel-docker/blob/master/Dockerfile) is my favorite of the ones I created.

rbergen (approved):
LGTM

marghidanu (approved):
LGTM 2

PR #894, opened 2023-02-19
Remove non-Bash external dependencies

dotysan:
- sleep and date aren't part of bash
 - also more portable by removing non-standard nanoseconds
 - still uses clock in microseconds

## Description
This is now pure bash. It might improve performance slightly by reducing syscalls to `sleep` and `date`.

Also only GNU `date` supports `%N` (nanoseconds), so it would have broken on BSD or other platforms that only support `strftim(3)` proper. The only drawback is internal `printf %(datefmt)T` supports only standard `strftim(3)` so we must use microseconds instead of nano.

## Contributing requirements

### DIFF
--- a/PrimeBash/solution_1/PrimeBash.sh
+++ b/PrimeBash/solution_1/PrimeBash.sh
@@ -90,20 +90,16 @@ function runSieve {
 }
 
 function printResults {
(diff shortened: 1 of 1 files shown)


rbergen:
Closing this as #895 includes this PR's changes

PR #896, opened 2023-02-22
Use built-in prime_list function in Euphoria

rzuckerm:
## Description
I've implemented an Euphoria solution using the built-in [prime_list function](https://openeuphoria.org/docs/std_primes.html#_4387_prime_list). It's pretty fast!

## Contributing requirements

### DIFF
--- a/PrimeEuphoria/solution_2/Dockerfile
+++ b/PrimeEuphoria/solution_2/Dockerfile
@@ -1,0 +1,11 @@
+FROM openeuphoria/euphoria:4.1.0
+
+RUN apt-get update -qq \
+    && apt-get install -y gcc \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /opt/app
+COPY primes.ex ./
+RUN euc -con -extra-lflags '-no-pie' primes.ex
+ENTRYPOINT ["./primes"]
--- a/PrimeEuphoria/solution_2/README.md
+++ b/PrimeEuphoria/solution_2/README.md
@@ -1,12 +1,15 @@
 # Euphoria solution by rzuckerm
 
-![Algorithm](https://img.shields.io/badge/Algorithm-unknown-yellowgreen)
+![Algorithm](https://img.shields.io/badge/Algorithm-other-yellowgreen)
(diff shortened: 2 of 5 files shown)


rbergen:
Very interesting to see an implementation based on a prime calculation facility provided by the language's core library. I think that's actually a first for this project. 

I will need a little more time to be able to do a full review, but I did notice you labelling the algorithm as `unknown` in the benchmark output. The contributing guidelines for this project state that if the algorithm is not a confirmed case of `base` or `wheel`, the (only) other option is `other`. So I'd like to kindly ask you to already make that change.

rbergen:
Haha, snap. Thanks!

rzuckerm:
@rbergen I'm taking the PR back to draft mode. After analyzing the `prime_list` function, I see that it is using a global variable to store the answer. On the first call, it builds the answer. On subsequent calls it just returns the same answer. I'm going to see if there is a way to reset the global variable each time.

rzuckerm:
@rbergen I'm closing this PR since there is no way to reset the global variable. The variable is private to `primes.e`. The only alternative would be to copy the function from the Euphoria library, and I didn't want to do that because I don't know if that will violate their licensing model. I may use it as a basis for my own trial division algorithm.

rbergen:
@rzuckerm Thanks for your updates, and the thorough investigation into the inner workings of the `prime_list` function. As I said, I would have loved to include a solution that's based on a core library prime number identification routine, but I understand your conclusion concerning it.

That said, I took a look at the Euphoria license, and it's basically the MIT license. That's a very permissive one, which does allow you to use, modify and redistribute (parts of) the Euphoria source code, if you want to. I'm just noting this in case you still consider adding a trial division-based solution, later on.

rzuckerm:
> That said, I took a look at the Euphoria license, and it's basically the MIT license. That's a very permissive one, which does allow you to use, modify and redistribute (parts of) the Euphoria source code, if you want to. I'm just noting this in case you still consider adding a trial division-based solution, later on.

Thanks @rbergen . I hadn't had time to research the licensing. I'm familiar with the MIT license. I'll do another pass at this using their code, and I might make some tweaks to it. For instance, there's no reason to yield to other tasks, and there's no reason to put a time limit on how long the primes calculation takes. Other than that, I'll try to stay true to the original implementation

PR #899, opened 2023-03-06
Improve phix dockerfile

rzuckerm:
## Description
In Phix 1.0.1, the `p` executable and the compiled output of `p` had to run through `/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2` in order to work. In 1.0.2, this is no longer necessary.

## Contributing requirements


rbergen (approved):
LGTM

PR #904, opened 2023-04-08
update C# solutions to dotnet 7

qrzychu:
## Description
dotnet 7 made big improvements to performance, so just changing the target framework makes 10-30% difference on my machine

### DIFF
--- a/PrimeCSharp/solution_1/Dockerfile
+++ b/PrimeCSharp/solution_1/Dockerfile
@@ -1,6 +1,6 @@
 #See https://aka.ms/containerfastmode to understand how Visual Studio uses this Dockerfile to build your images for faster debugging.
 
-FROM mcr.microsoft.com/dotnet/sdk:6.0-focal AS build
+FROM mcr.microsoft.com/dotnet/sdk:7.0 AS build
 WORKDIR /source
 
 # copy csproj and restore dependencies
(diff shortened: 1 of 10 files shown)


rbergen:
@qrzychu To be honest, I'm curious how you tested this in the Docker context. When I try to `docker build` your updated solution_1, this is what I get:

```text
~/source/Primes/PrimeCSharp/solution_1$ docker build -t csharp1-pr904 .
Sending build context to Docker daemon  311.8kB
Step 1/11 : FROM mcr.microsoft.com/dotnet/sdk:7.0-focal AS build
manifest for mcr.microsoft.com/dotnet/sdk:7.0-focal not found: manifest unknown: manifest tagged by "7.0-focal" is not found
```

qrzychu:
@rbergen sorry for that, fixed

rbergen:
Thanks for the new commit, that did indeed fix the Docker builds.

I've tested the old and new solutions, and I'd say .NET 7 yields a meaningful improvement to the performance of most implementations. The exceptions are the "stride" implementations in solution 4, where I would call the difference statistically insignificant.

All in all, personally I'm fine with merging this.

@Kinematics @davepl @tannergooding @ItalyToast, do you have any objections to upgrading your respective solution to .NET 7?

tannergooding:
No concerns on my end. Updating to the latest versions will generally provide some amount of performance wins and will keep the projects on supported versions of .NET

Kinematics:
Checking it over, it looks like a nice improvement on almost all methods, and no regressions. I don't mind it being updated.

ItalyToast:
I dont mind it being updated.

rbergen:
@tannergooding @Kinematics @ItalyToast Thank you for your response. I'm comfortable saying that Dave also won't mind, so I'll proceed with merging.

PR #905, opened 2023-04-08
Rust: added non-HT run

mike-barber:
## Description

This is a small addition: A multithreaded run using physical cores in addition to the existing run with available hyperthreads. This is relevant for an apples-to-apples comparison of per-thread throughput.

## Contributing requirements

### DIFF
--- a/PrimeRust/solution_1/prime-sieve-rust/src/main.rs
+++ b/PrimeRust/solution_1/prime-sieve-rust/src/main.rs
@@ -774,7 +774,11 @@ fn main() {
 
     let thread_options = match opt.threads {
         Some(t) => vec![t],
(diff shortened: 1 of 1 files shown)


Freaky:
It would probably be wise to deduplicate the vec so platforms without HT don't end up running the same test again.

mike-barber:
> It would probably be wise to deduplicate the vec so platforms without HT don't end up running the same test again.

Thanks @Freaky - I forgot about that.

rbergen (approved):
it's nice to see another PR from one of the earliest contributors to the project! LGTM!

mike-barber:
Thanks @rbergen! It's been ages indeed. Good to hear from you too!

PR #906, opened 2023-04-08
Initial commit for BitmaskSieve

ciaccia:
Another Java implementation with some further optimizations.


ciaccia:
Will recreated with some fixes

PR #908, opened 2023-04-09
Elixir: Improve performance of solution 2 using the Tasks API

DeemoONeill:
## Description
This improves on solution 2 and spawns as many tasks as there are schedulers to make the most of the available hardware which should see a nice improvement on the hardware that these tests run on.

I'm happy to remake this pull request as a new solution if that is more appropriate.

## Contributing requirements


rbergen:
@DeemoONeill Thanks for submitting this. This PR changes one of the key characteristics of the solution (parallelism), which can't be considered a straightforward optimization of the existing solution. I think there are two things you could do:

- Add the parallelized version as a second implementation to the existing solution
- Create a new solution based on the one this PR aims to change

Generally speaking, we prefer multiple implementations in one solution over individual solutions because it saves Docker resources (images, build time, etc,.), but I think a new solution would not be out of order per se.

DeemoONeill:
Thanks for the feedback, I'll add the parallelised version as a second run in main then.

First it will run the existing solution and then the _parr version

rbergen:
That sounds like a great approach to me. Thanks!

DeemoONeill:
Ok, main now does the sequential run, writes to stdout, then does the parr run and writes to stdout.

This keeps the original version and adds in a multiple core version.

rbergen (approved):
LGTM!

PrimeElixir/solution_2/lib/prime_sieve.ex
```diff
@@ -197,15 +197,46 @@ defmodule PrimeSieve do
...
+      |> Enum.map(fn _ ->
+        Task.async(fn -> start(sieve_size, start_time) end)
+      end)
+      |> Task.await_many(7000)
+      |> Enum.sum()
+
+    duration = Time.diff(Time.utc_now(), start_time, :microsecond)
+
+    IO.puts(
+      "thomas9911_parr;#{pass_count};#{Float.round(duration / 1_000_000, 3)};1;algorithm=base,faithful=yes,bits=1"
```
jesperes, line 236: This emits the wrong number of threads, should be something like
```elixir
    IO.puts(
      "thomas9911_parr;#{pass_count};#{Float.round(duration / 1_000_000, 3)};#{System.schedulers()};algorithm=base,faithful=yes,bits=1"
```

rbergen: That's a good catch, thanks for flagging it. As I missed this in my review, I'll open a PR to fix this.

DeemoONeill: Apologies, I completely missed that

PR #910, opened 2023-04-10
Squirrel Solution

tmob03:
## Description
A primes solution in Squirrel, based on the python_1 solution.

Squirrel is meant to be embedded and as such doesn't have a good compiler. After searching for a while I found the Going Nuts Compiler. After looking at the [licensing](https://forum.goingnuts.net/license-comparison.html ) though, you need to pay for cloud computing status, thus I did not try and learn Docker to upload it.
## Contributing requirements


rbergen:
@tmob03 Thanks for submitting this. I've reviewed the Gonuts license myself, and my conclusion is that Gonuts can be used in the context of this project. The project itself is clearly non-commercial (and can even be called educational to some extent).   Concerning the "cloud capabilities", the website includes [a page with this text](https://forum.goingnuts.net/gonuts-usage.html#:~:text=module%20their%20own.-,Cloud%20capabilites,-Gonuts%20cloud%20capabilities):

> Gonuts cloud capabilities usually utilize an online login, however some users may find it useful to have that login secured by a gonuts license and hence locked to their executable or amalgamation.
> **At the time of this writing, cloud services are non-existing and in planning stage only. Check the goingnuts website for more information.**

The use of Gonuts within the context of your solution does not require cloud capabilities, and couldn't even use them (yet) if it wanted to.
While looking into the licensing aspect, I also noticed that there is a Gonuts executable available for AMD64 Linux as well. 

All this being the case, could you please put a Dockerfile together to compile and run your solution?

tmob03:
Oh, neat. I’ll see what I can do.

tmob03:
Well, I can't seem to get docker to run on my PC. Not sure how to proceed if I can't even get it to run on my PC.

rbergen:
> Well, I can't seem to get docker to run on my PC. Not sure how to proceed if I can't even get it to run on my PC.

You could ask for help from someone who does get Docker to run. :)

I could try and put a Dockerfile together for you, if you add me as a collaborator on your fork of this repository. It will probably be the weekend that I get around to it, then.

marghidanu:
You can use a VM to run Docker in it if your PC doesn't allow it.

tmob03:
@rbergen I got docker to work on my PC! I set up a dockerfile, but it seems the linux_amd64 compiler has a bug, and runs out of memory allocating a 1 million size array. Is it good to use the windows exe instead?

rbergen:
The main reason for semi-requiring a Linux-based Docker image is that it allows solutions to be included in the daily benchmark runs. This being a drag-race, we like to do that if in any way possible. With a Windows executable, it just isn't.

If the issue you're running into is caused by a bug in Gonuts, then maybe it could be beneficial for you to create a post in the Going Nuts forums? It would be interesting to see if others can replicate the behaviour, and I'd imagine Going Nuts would also be interested to learn about a bug in their compiler.

tmob03:
I have done that now, but I don't expect a reply, the website looks pretty dead to me. I made the first thread other than a spambot since its creation in 2017.

rbergen:
Ok, I'll try to confirm the bug this weekend. If I can then I'm happy to proceed towards merging without a Dockerfile.

tmob03:
Well, I massively overcomplicated this. There's a Squirrel package for linux I never realized since I'm on Windows. :P Free performance boost with the official interpreter, though!

PrimeSquirrel/solution_1/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ubuntu:22.04
+
+RUN apt-get update -qq \
+	&& apt-get install -y squirrel3 \
+	&& apt-get clean \
+	&& mkdir -p usr/local/squirrel
+	
+
+COPY ["prime.nut", "run.sh", "usr/local/squirrel"]
```
rbergen: Hadolint's complaint was that the target ("/usr/local/squirrel") has to _end_ with a / when there are multiple files to copy. The last argument to COPY should be "/usr/local/squirrel/", with slashes at both the beginning and end.

PrimeSquirrel/solution_1/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+WORKDIR usr/local/squirrel
```
rbergen: The WORKDIR has to be absolute. so starting at root. The command should thus be: `WORKDIR /usr/local/squirrel` (with a slash at the beginning).

PrimeSquirrel/solution_1/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ubuntu:22.04
+
+RUN apt-get update -qq \
+	&& apt-get install -y squirrel3 \
+	&& apt-get clean \
+	&& mkdir -p usr/local/squirrel
```
rbergen: I'd recommend making the directory absolute here as well, to avoid confusion. So: `&& mkdir -p /usr/local/squirrel`

rbergen (commented):
I noticed Hadolint found some issues with your Dockerfile, and I see you've made some changes to address these. I may be able to give a few pointers to assist, see the following comments.

tmob03:
Thanks!

PrimeSquirrel/solution_1/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ubuntu:22.04
+
+RUN apt-get update -qq \
+	&& apt-get install -y squirrel3 \
+	&& apt-get clean \
```
rbergen, line 5: Thank you for a) putting a Dockerfile together and b) addressing these findings. You can address the Hadolint finding in the last CI run by adding the following line after `&& apt-get clean \`: `&& rm -rf /var/lib/apt/lists/* \`

rbergen (approved):
LGTM! As it concerns a new solution in a new language, I'll ask a review from @marghidanu as well.

marghidanu (approved):

PR #912, opened 2023-04-12
Inlined bit operations and moved sieve definition inside loop

rzuckerm:
## Description
For the both implementations, I moved the sieve definition inside the main prime loop to ensure that it keeps
getting reallocation every iteration.

For the packed bit implementation (`primes_bit.ex`), I inlined the bit operations. This improved the speed by a factor of about 2.1x.

## Contributing requirements

### DIFF
--- a/PrimeEuphoria/solution_1/README.md
+++ b/PrimeEuphoria/solution_1/README.md
@@ -7,7 +7,8 @@
 ![Bit count](https://img.shields.io/badge/Bits-1-green)
 
 * `primes.ex` uses `integer` for each sieve item
-* `primes_bit.ex` packs 64 sieve items in an `atom`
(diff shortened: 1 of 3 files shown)


rbergen:
@rzuckerm Thanks for submitting this. The changes themselves look good to me, but I did notice something else that I must have missed when I reviewed the version of the solution that's in the current codebase: you use the same label ("rzuckerm") in the drag-race output line for both implementations included in this solution. Could you please label each of them differently? It makes it easier to distinguish the implementations in the output of the automated benchmark runs (like shown in PrimeView, for example).

A very simple solution would be to change the one for the bit-level implementation to "rzuckerm-bit", but any other approach that leads to different labels would also be fine. I apologize for missing this before.

rzuckerm:
@rbergen Done

rbergen:
Thank you!

rbergen (approved):
LGTM!

PR #916, opened 2023-04-15
C# solution 1 maintenance updates

Kinematics:
## Description
Update nuget packages and performance results as a result of upgrading from .NET 6 to .NET 7.

## Contributing requirements


rbergen (approved):
@Kinematics Thanks for submitting this. LGTM

PR #917, opened 2023-04-15
Use Phix docker image

rzuckerm:
## Description
Since the process for creating a docker image for Phix is so convoluted, I decided to make my own. If you're curious, here's the [source](https://github.com/rzuckerm/phix-docker-images) for that image.

## Contributing requirements


rbergen (approved):
LGTM

PR #918, opened 2023-04-25
Fixed bug in output

kostyakozko:
## Description
Fixed output bug in CPP solution 3

## Contributing requirements


rbergen:
@kostyakozko Thank you for proposing a fix to a perceived bug. However, the current calculation of "average" is correct. 
This is because the number in that particular line of output is the average time spent per pass, not the average number of passes per second.

You can refer to @davepl's original C++ solutions (1 and 2) for comparison; they also calculate the average in the exact same way.

kostyakozko:
Thank you, I've missed the point

PR #924, opened 2023-05-10
Haskell Solution2 Maximum Threads

GordonBGood:
## Description

When the maximum available (logical) threads is less than four, use this number of threads rather than four; this gives a better benchmark score for CPU's such as some Intel Celeron's with only two cores (and no Hyper-Threading - HT), although it still wouldn't quite be fair as to passes per thread for some CPU's with only two cores but with HT or Simultaneous Multi Threading (ofwhich none are currently tested)...

## Contributing requirements


rbergen (approved):
LGTM!

PR #926, opened 2023-05-13
Rust: adjust number of threads

mike-barber:
## Description

The number of physical cores is not reliably detected on many of the benchmark machines, and it's always a fun problem with containers and virtualization. This little PR makes some small adjustments to the runs to deal with this sensibly:
- changed physical cores count to simple (logical_cores / 2), similar to other solutions
- added run on 4 threads (if possible) for comparability with @GordonBGood updates
- removed some of the older "striped" solutions from default run 
  - less noise in reports from the earlier algorithms
  - less time for the overall run

## Contributing requirements

### DIFF
--- a/PrimeRust/solution_1/prime-sieve-rust/src/main.rs
+++ b/PrimeRust/solution_1/prime-sieve-rust/src/main.rs
@@ -860,7 +860,8 @@ fn main() {
             );
         }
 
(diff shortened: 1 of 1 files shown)


rbergen:
@mike-barber Thank you for submitting this update!

PR #933, opened 2023-08-03
Make bit counting more clever

rbergen:
## Description

This updates PrimeCUDA/solution_2 to use Brian Kernighan's algorithm for counting the number of set bits in an integer.

## Contributing requirements


marghidanu (approved):
LGTM

PR #939, opened 2023-09-09
Primes in Microsoft Basic v2

rzuckerm:
## Description
This is an improvement on the Microsoft Basic v2 solution by davepl. In addition, it contains a 1-bit version which can handle sieve size up to 100000

## Contributing requirements


rbergen:
@rzuckerm Thanks for this submission. However, as it is in part an improvement on an existing solution, I think we should indeed structure it as an iteration on PrimeBASIC/solution_4. That is to say: apply the algorithm improvements to the implementation that's already there, and add the 1-bit implementation to it. For one, Dave has said multiple times himself that he prefers improvements to existing solutions over the creation of new, deduced ones - his own are no exception.

Of course, it's perfectly fine to take credit for the improvement/addition in the solution's README.md and implementation output. As always, I'm happy to help with the phrasing of things if needed.

Please do retain the possibility to run the solution without Docker that PrimeBASIC/solution_4 now offers.

rzuckerm:
@rbergen How would I annotate the original solution? Would I retain the `davepl`? For the output, would I list the original result, or would I list the results that I got for the old implementation vs. the new?

rbergen:
In the 1-bit implementation you can obviously just tag yourself. In the one you improve I'd go for `davepl-rzuckerm` - in the end that implementation is effectively co-authored. 

The output should reflect the output of whatever is the solution in its current state. In other words, that should be replaced by the output of the solution _after_ you've modified it.

PrimeBASIC/solution_4/README.md
```diff
@@ -1,21 +1,39 @@
-# Microsoft BASIC solution by davepl
+# Microsoft BASIC solution by davepl and rzuckerm
 
 ![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
 ![Faithfulness](https://img.shields.io/badge/Faithful-no-yellowgreen)
 ![Parallelism](https://img.shields.io/badge/Parallel-no-green)
 ![Bit count](https://img.shields.io/badge/Bits-unknown-yellowgreen)
+![Bit count](https://img.shields.io/badge/Bits-1-yellowgreen)
```
rbergen: The "Bits: 1" badge can be made green.

```suggestion
![Bit count](https://img.shields.io/badge/Bits-1-green)
```

PrimeBASIC/solution_4/README.md
```diff
@@ -33,12 +51,16 @@ Execute the following commands from the implementation directory, in a bash shel
 A Dockerfile has been provided. With Docker installed on a supported operating system, it can be used with the following commands:
 
 ```bash
-docker build -t primes-basic-4 .
-docker run primes-basic-4
+./build-docker.sh
+./run-docker.sh
```
rbergen: I think we should keep the old Docker build and run instructions. They are one-lines themselves anyway, and the use of bash scripts here limits the execution of the solution with Docker to *NIX systems with bash installed. Using the solution on other systems is exactly the scenario that Docker enables (at least theoretically.)

```suggestion
docker build -t primes-basic-4 .
docker run primes-basic-4
```

rbergen (commented):
Thanks for the work on merging your solution with the existing one. 

I have two small comments, see below.

rbergen (approved):
LGTM! Thanks again for submitting this!

PR #944, opened 2023-11-07
CPP with inlined amd64 assembler

hildenborg:
## Description
Multithreaded CPP with inlined amd64 assembler.
I chosed to put it in it's own folder as it is neither fully C++ or fully assembler.

## Contributing requirements


rbergen:
@hildenborg As you describe it this is a mixed language solution, for which the PrimeMixed directory already exists. This solution would then become the one in directory PrimeMixed/solution_5.

hildenborg:
I didn't see the mixed language category. Moved into the correct place now.

rbergen (approved):
> I didn't see the mixed language category. Moved into the correct place now.

Not a problem, one of the reasons I'm here. :) 

I took the liberty of setting the execute bit on run.sh, without which the solution wouldn't run. With that, it LGTM.

As this is a new solution in a new mix of languages, I will ask for a second review from @marghidanu also.

rbergen:
Merging based on "he who is silent agrees".

PR #949, opened 2023-12-28
Primes in Pyret

rzuckerm:
## Description
Implementation of Primes in Pyret.

## Contributing requirements

### DIFF
--- a/PrimePyret/solution_1/Dockerfile
+++ b/PrimePyret/solution_1/Dockerfile
@@ -1,0 +1,6 @@
+FROM rzuckerm/pyret:0.0.27-2
+
+WORKDIR /opt/app
+COPY primes*.arr ./
+RUN pyret -qkc primes.arr
+ENTRYPOINT ["node", "primes.jarr"]
--- a/PrimePyret/solution_1/Dockerfile
+++ b/PrimePyret/solution_1/Dockerfile
@@ -1,4 +1,4 @@
-FROM rzuckerm/pyret:0.0.27-2
+FROM rzuckerm/pyret:0.0.27-3
 
 WORKDIR /opt/app
 COPY primes*.arr ./
--- a/PrimePyret/solution_1/.gitignore
+++ b/PrimePyret/solution_1/.gitignore
@@ -1,0 +1,2 @@
+*.jarr
+.pyret
--- a/PrimePyret/solution_1/README.md
+++ b/PrimePyret/solution_1/README.md
@@ -30,7 +30,7 @@ where `<options>` are optional command-line arguments:
 
 ## Output
 
-On an 12th Gen Intel(R) Core(TM) i7-12850HX 2.10 GHz with 32 GB of memory on a Windows 10
+On a 12th Gen Intel(R) Core(TM) i7-12850HX 2.10 GHz with 32 GB of memory on a Windows 10
 laptop running a Ubuntu 22.04 VM in VirtualBox 7.0.6:
 
 ```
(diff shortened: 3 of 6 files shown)


rbergen:
@rzuckerm Thank you for submitting this. I always enjoy adding a "new" language to the collection.

After taking a first look at your solution and the Pyret language, I do think some changes will have to be made to be able to mark the solution as "faithful". As per the contributing guidelines, a "faithful" solution must

> [use] a class to encapsulate the sieve, or a (closest) equivalent feature in your language if a class construct is not available. This class must contain the full state of the sieve. Each iteration should re-create a new instance of this class from scratch.

In my understanding of Pyret, it is possible to implement a class-like construct using data declarations, which includes the possibility to incorporate methods that operate on the contents of the actual data type.  
I'm therefore of the opinion that a "faithful" Pyret solution should be based on the data declaration construct. 

Would it be possible for you to restructure the solution accordingly?

rzuckerm:
@rbergen I'll try. The pyret documentation is not great. I looked over the `data` declaration and couldn't quite understand how to use it.

rzuckerm:
@rbergen I think I have what you requested. The closest thing to a class in Pyret is a [data declaration](https://pyret.org/docs/latest/s_declarations.html#%28part._s~3adata-decl%29).

rbergen:
@rzuckerm Thank you for taking the time and trouble to adapt the solution to use data declarations. I'll test and review a little later today.

PrimePyret/solution_1/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM rzuckerm/pyret:0.0.27-2
```
rbergen: I noticed that the Pyret image you uploaded to Docker Hub is linux/amd64 only; as in, it's not a multi-arch image that includes the linux/arm64 architecture. We prefer Docker base images that support both architectures, which can be built using the information available here: <https://docs.docker.com/build/building/multi-platform/>.

If that's not possible for any reason, then I think this solution should include an `arch-amd64` flag file to indicate it is strictly AMD64 only.

rbergen (commented):
@rzuckerm One comment relating to the Docker base image from me, otherwise LGTM.

I'll push an update to PrimeView to add Pyret to its language map as well.

PrimePyret/solution_1/Dockerfile
rzuckerm: @rbergen I added the arch-amd64 flag

rbergen:
@rzuckerm I may not have explained sufficiently clearly that the use of the `arch-*` flag file really is a last-resort measure that should only be used if creating a multi-architecture Docker image is not possible.

I've just built a combined linux/amd64 and linux/arm64 pyret image off of your pyret-docker-image repo myself, and pushed it to primeimages/pyret:0.0.27-2 (available on Docker Hub here: <https://hub.docker.com/r/primeimages/pyret>). I then changed the FROM line in the Dockerfile to pull that image instead of yours, and tested the solution on both architectures. That worked, which means it is possible to build and run your Pyret image and hence this solution on arm64 as well.

As I see it we can now do either of two things:

- You build and push a multi-arch Pyret Docker image yourself, or
- We update the solution to pull the primeimages/pyret image instead.

Obviously, in both cases the arch-amd64 flag file should once again be removed.

I'll leave the decision on which approach to take to you.

rzuckerm:
@rbergen I changed my image to also support linux/arm64

rbergen:
Great, thanks! Your ARM64 image also works on my ARM64 device. I'll delete the primeimages Pyret Docker image, as that is now redundant. This being a new solution in a new language, I'll request a second review from @marghidanu and give him a few days to give it, if he wants to.

rbergen (approved):

marghidanu (approved):
LGTM

PR #961, opened 2024-02-17
:heavy_plus_sign: Add Assembly solution 4.

DanielAtCosmicDNA:
## Description
This PR adds an optimized assembly solution to showcase the concept in the ongoing race.

## Contributing requirements


rbergen:
@DanielAtCosmicDNA Thank you for your submission. What exactly have you changed in the assembly code coming out of the compiler? Or is the code in PrimeAssembly.s literally the output of `clang++ $CXX_ARGS -S -masm=intel PrimeCPP_CONSTEXPR.cpp -o PrimeAssembly.s`?

DanielAtCosmicDNA:
> @DanielAtCosmicDNA Thank you for your submission. What exactly have you changed in the assembly code coming out of the compiler? Or is the code in PrimeAssembly.s literally the output of `clang++ $CXX_ARGS -S -masm=intel PrimeCPP_CONSTEXPR.cpp -o PrimeAssembly.s`?

PrimeAssembly.s is literally the output of that line, except for the fact that I removed the neutral line `.file	"PrimeCPP_CONSTEXPR.cpp"` for testing. The only difference I can think of now in terms of performance is that this solution runs just after a light build script that links the assembly. That might put this solution at a disadvantage as the processor clocks would not have already been accelerated due to the CPU intensive processing that took place in the CPP solution 3.

rbergen:
Thanks for clarifying that. In that case I don't think we can add this as a new solution, as it's effectively an intermediate step in the compilation of another one - the actual solution is still the one in PrimeCPP/solution_3. Reasoning from a different angle: if we were to make this an Assembly solution in itself, we could do the same for every other compiled-language solution, and then add a range of bytecode solutions for certain other ones (like Java and .NET) on top of that.

Instead of adding this as a separate solution, I would be happy to merge a PR that updates the README.md of PrimeCPP/solution_3 to mention that it's possible to inspect the Assembly code for that solution by running clang++ in the way you now described.

DanielAtCosmicDNA:
Alright, makes sense.

PR #962, opened 2024-02-18
Updated OCaml solutions to fix issue (#958)

### LINKED ISSUE #958: OCaml solutions don't build with current version of OCaml
While updating a number of Docker images published on Docker Hub in the context of this project, I found that the OCaml solutions don't build with a current version of OCaml (5.1, to be precise). Two reasons I've been able to identify but also noting that there may be more:

- `corebuild` is no longer available. After reading some documentation it seems to me that `dune` is now "the" build system for OCaml, but I may be wrong.
- `Core.Unix` has been deprecated.

Personally, I am definitely not sufficiently knowledgeable about OCaml to make the modifications that would be required to make the solutions build again.

Without wanting to imply any obligation to look into this by the solutions' author @gkpotter, I'm calling out to him specifically in case he would be willing and able to do so.

gkpotter:
## Description
Updated OCaml solutions to use Dune and the latest version of OCaml to fix #958.

## Contributing requirements


rbergen:
@gkpotter Thank you so much! I'm currently running CI on this, which I already know will yield some complaints from Hadolint, our Dockerfile linter. Please let me know if you're ok with addressing those, or if you'd like me to take care of that for you - which will require me to update your main branch.

Correction: CI actually fails because the base images you refer to are untagged - I forgot Hadolint marks this as an error instead of a warning. The "latest" version of the primeimages/ocaml image is 5.1, for the alpine image that would be 3.19.

rbergen (approved):
LGTM! Thanks again @gkpotter!

gkpotter:
Glad I could help keep this project up to date!

PR #965, opened 2024-02-25
Move Java/solution_1 to adoptopenjdk/openjdk base image

rbergen:
## Description

This makes Java/solution_1 use an AdoptOpenJDK-provided, Alpine-based, OpenJDK image as the its base image, because the currently used "official" OpenJDK image has been deprecated since July 2022.

## Contributing requirements


rbergen:
Merging based on lack of reviewer feedback.

PR #966, opened 2024-05-09
Adding second solution for PrimeTerra

Enter1he:
## Description
Adding second solution for Terra. Now using metaprogramming abilities of the language.

## Contributing requirements


rbergen:
@Enter1he Thank you for your submission. The existing Terra solution is also created by you and has a Dockerfile that is effectively identical to that of the second solution you now propose to add. Therefore, please structure this second approach as a second implementation of the existing solution, instead of a new solution. You can do that by updating the existing solution so that it runs both the existing and the new Terra programs.

Also, and more on a side note, there is a merge conflict between the existing solution's README and this PR. Please correct that as well.

PR #970, opened 2024-07-31
Add MACRO-11, apply updates

rbergen:
## Description

This:

- Adds a solution in MACRO-11, the macro assembly language for the PDP-11 series of computers
- Bumps base image versions for Prolog/solution_1 and BrainF*ck/solution_1, because the former base images ran out of maintenance and caused the solutions' builds to fail. In the latter case, the Dockerfile was also changed to switch to distro-provided versions of clang and cmake, as they are now of sufficiently recent versions to meet the solution's requirements. Both solutions have been confirmed to build and run with changes applied.
- A number of actions in the GitHub CI workflow have been bumped to newer versions. This is triggered by deprecation of older action versions due to end of maintenance on older versions of Node.js.

## Contributing requirements


davepl (approved):
The CI is voodoo but the code looks right!

PR #972, opened 2024-08-06
Use latest Julia version to represent current performance

christiangnrd:
## Description
This PR updates the version of Julia used in all 4 solutions from 1.6 to 1.10. 1.10 will be the next long-term support release once 1.11 releases in a few weeks, and while I don't think there's a big change in performance, it is more representative of current Julia performance.

@dcbi, @epithet, @louie-github, @GordonBGood

### DIFF
--- a/PrimeJulia/solution_1/Dockerfile
+++ b/PrimeJulia/solution_1/Dockerfile
@@ -1,4 +1,4 @@
-FROM julia:1.6-alpine3.13
+FROM julia:1.10-alpine
 
 WORKDIR /opt/app
 
--- a/PrimeJulia/solution_1/Dockerfile
+++ b/PrimeJulia/solution_1/Dockerfile
@@ -1,4 +1,4 @@
-FROM julia:1.10-alpine
+FROM julia:1-alpine
 
 WORKDIR /opt/app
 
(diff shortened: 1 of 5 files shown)


rbergen:
@christiangnrd Thanks for opening this! It's nice to have someone (else) concern themselves with some of the "regular maintenance" a project like this requires. :)

I'll give the people you tagged a week or so to respond before moving towards merging.

christiangnrd:
I tried to make the least amount of changes to ensure this gets merged, but I personally think we should be using the latest stable version, as recommended on the [downloads](https://julialang.org/downloads/#current_stable_release) page. That would mean using the `1`/`1-alpine` tags instead of `1.10`/`1.10-alpine`.

Curious about your opinion.

rbergen:
In principle everything is good to merge as long as an actual version tag is used (i.e. anything but "latest"). There are no set rules for how _specific_ that version tag needs to be.

However, the more flexibility you allow in the tag's exact version interpretation, the bigger the chance that some future specific version within the range you specify will break (a) build(s) at some point. And if that happens in a way that makes CI fail I'll do what it takes and what I think is best to fix that, right then and there. And that may not be what you (or a solution contributor) prefers.

In that non-negotiable context, it is up to you. :)

christiangnrd:
That sounds reasonable. The second sentence in the linked downloads pages mentions "Great care is taken not to break compatibility with older Julia versions, so older code should continue to work with the latest stable Julia release.", so I think it's worth the risk. 

I pushed a change to use the latest stable version. If any of the original authors objects I'll change it back for their solution.

rbergen:
In the absence of feedback from solution authors and considering CI passes without any hiccups, I'm merging this.

PR #977, opened 2024-09-13
Add timeout to benchmark solution runs

rbergen:
## Description

I've found that on certain systems/in certain situations, benchmark runs can get stuck for reasons that are unclear. This adds a timeout for each of the solution runs kicked off during the benchmark.

The default timeout is 10 minutes, this can be changed by a new make variable that's also been added to the documentation.

## Contributing requirements


rbergen:
Merging after out-of-band review by @davepl.

PR #979, opened 2024-10-19
Switch to uint8_t, some optimizations

davepl:
## Description
Replace this text with your description.

## Contributing requirements


PrimeCPP/solution_2/PrimeCPP_PAR.cpp
```diff
@@ -24,80 +24,66 @@ using namespace std::chrono;
 const uint64_t DEFAULT_UPPER_LIMIT = 10'000'000LLU;
 
 class BitArray {
-    uint32_t *array;
+    uint8_t *array;
     size_t arrSize;
```
rbergen: Is it necessary to declare `arrSize` as a member variable? It's private so only visible to the `BitArray` member functions, and I don't think it's used anywhere outside the constructor.

PrimeCPP/solution_2/PrimeCPP_PAR.cpp
```diff
@@ -24,80 +24,66 @@ using namespace std::chrono;
...
 public:
-    explicit BitArray(size_t size) : arrSize(size) 
+    explicit BitArray(size_t size) : logicalSize(size)
     {
-        array = new uint32_t[arraySize(size)];
-        std::memset(array, 0xFF, (size >> 3) + ((size & 7) > 0));
+        arrSize = (size + 1) / 2; // Only store bits for odd numbers
+        array = new uint8_t[arraySize(arrSize)];
+        // Bits are left at zero default, so no need to initialize them
+        // std::memset(array, 0x00, arraySize(arrSize));
```
rbergen: Are you absolutely sure? When I was learning C++, it was hammered into me by my teacher that with C and C++, one has to assume garbage when allocating memory via `new`.

I actually looked up the documentation in the C++ reference, and this is what it says about default initialization, which is what's done when `new T` is used ("if T is an array type, every element of the array is default-initialized", as per [Default-initialization](https://en.cppreference.com/w/cpp/language/default_initialization)):

> When storage for an object with automatic **or dynamic storage duration** is obtained, the object has an indeterminate value.
> 
> If no initialization is performed for an object, that object retains an indeterminate value until that value is replaced.

It may work by accident in your office, but it looks to me we can't rely on that happening elsewhere. :)

rbergen (commented):
I always like improvements to one of the original solutions! :)

I have two comments, the second one (array initialization) is the one I think actually matters.

davepl:
Fixed!

rbergen (approved):
Thanks, LGTM!

PR #985, opened 2024-11-06
CPP/solution_2: add mask implementation

davepl:
This splits solution_2 into two distinct solutions:
solution_2, which now uses bit masking and direct array access
solutiion_2a, which uses half as much memory, no direct array access, and is faster

Why keep solution_2 at all?  Because one day I hope to learn SIMD and will need direct array access, so want to keep the basic methodology around


PrimeCPP/solution_2/PrimeCPP_PAR.cpp
```diff
@@ -61,7 +61,55 @@ class BitArray {
         array[index(n)] |= (uint8_t(1) << (n % 8));
     }
 
-    constexpr size_t size() const 
+    static constexpr uint32_t buildSkipMask(size_t skip, size_t offset)
```
davepl: Is there a backwards-Kernigan way to do this?   Seems like someone has probably thought about how to create a mask before, and there's likely a better way.  Like maybe as bits are moved to the mask, you clear them from i and stop when i is zero?

rbergen (approved):
LGTM!

PR #987, opened 2024-11-07
Revert "Walk prebuilt mask through memory for small steps"

rbergen:
Reverts PlummersSoftwareLLC/Primes#984


rbergen:
Reverting because #984 merged a change that was intended to be part of a new implementation, as provided in #985.

Issue #954, opened 2024-01-21
Request: more esoteric languages


rbergen:
I'd like to put a request out to potential contributors to add solutions for more esoteric languages. We have some in the collection, but many more exist. At some point, I'd love to have a solution for each of the languages listed on the [Wikipedia page discussing the concept of esoteric programming language](https://en.wikipedia.org/wiki/Esoteric_programming_language), at least.

cmester0:
I have a solution in a higher level language compiled to Piet (https://github.com/cmester0/Piet/blob/main/Rust/tests/advc/primes.advc), I can generate the Piet solution if you would like it.

rbergen:
@cmester0 I'd love a proper solution in Piet! Ideally, we should be able to "compile" the Piet solution to what actually gets executed.

Just out of curiosity and chancing a blatant exposure of complete ignorance on my part, but what's the higher-level language in the file you referred to?

cmester0:
I have made a compiler stack to code Piet solutions easier. So this "high level language" is a simple language with function calls, local and global variables, and a simple standard library. So I am not coding the solution directly in Piet, but have some levels of abstraction on top of Piet. The compiled version is here: https://github.com/cmester0/Piet/blob/main/examples/primes.png

rbergen:
I understand. In that case it wasn't strange I didn't recognise the language.

I'd love to add a Piet solution to the collection. To make it a "proper" one, we'd need:

- The source code from which the Piet solution can be generated.
- The compiler that turns it into a Piet solution.
- A Piet interpreter that can execute the Piet solution.
- A Dockerfile that ties everything together and executes the Piet solution in such a way that it runs a sieve with an upper limit of 1,000,000 and prints the number of primes that were identified.

I've found a Piet interpreter here: https://github.com/your-diary/piet_programming_language. I'm also happy to help with the composition of the Dockerfile, if needed.

Kexisx:
I wanted to ask for a bit of clarification on where a language stands if there is no practical way to complete even a single full pass within the 5-second benchmark window.

This has come up while looking at some of the more esoteric languages. In some cases it is possible to write a faithful/base sieve in the language, but the interpreter/runtime is slow enough that a complete pass to 1,000,000 is not realistically achievable inside the timed window.

For example, with LOLCODE I currently get:

cwager_lolcode;1;17.121261;1;algorithm=base,faithful=yes,bits=1

[...]

PR #989, opened 2024-12-15
Add Primes in Whitespace

rzuckerm:
## Description
Prime Sieve in Whitespace. This contribution was inspired by #954 

## Contributing requirements

### DIFF
--- a/PrimeWhitespace/solution_1/Dockerfile
+++ b/PrimeWhitespace/solution_1/Dockerfile
@@ -1,0 +1,19 @@
+FROM ubuntu:24.04
+RUN apt-get update && \
+    apt-get install -y git make python3 ghc && \
+    mkdir -p /opt && \
+    cd /opt && \
+    git clone --depth 1 https://github.com/TryItOnline/WSpace && \
+    cd WSpace && \
+    make && \
+    cp wspace /usr/bin/whitespace && \
+    cd / && \
+    apt-get remove -y git make && \
+    apt-get autoremove -y && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* /opt/WSpace
+
+WORKDIR /opt/app
+COPY *.ws *.sh *.py ./
+
+ENTRYPOINT ["./run_primes.sh"]
--- a/PrimeWhitespace/solution_1/Dockerfile
+++ b/PrimeWhitespace/solution_1/Dockerfile
@@ -4,17 +4,9 @@ WORKDIR /tmp
 RUN apt-get update && \
     apt-get install -y ghc git make python3 && \
     git clone --depth 1 https://github.com/TryItOnline/WSpace && \
-    apt-get remove -y git && \
-    apt-get autoremove -y && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
-
-WORKDIR /tmp/WSpace
-RUN apt-get update && \
-    apt-get install -y make && \
-    make && \
-    cp wspace /usr/bin/whitespace && \
-    apt-get remove -y make && \
+    make -C WSpace && \
+    cp WSpace/wspace /usr/bin/whitespace && \
+    apt-get remove -y git make && \
     apt-get autoremove -y && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/* /tmp/WSpace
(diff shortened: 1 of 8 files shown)


rbergen:
@rzuckerm I'll have to do my proper review a little later, but I just want to immediately say that I love that you made the effort to submit this.

As documented in issue #954 I opened almost a year ago, I really like having more esoteric languages in. The dedication you've shown here by first writing your own assembler for it is something I truly admire.

After a first quick look, I can already say that you correctly marked your solution as non-faithful, but not because of the timing or sieve size aspects. The reason for marking it as such would be that the sieve is not held in any class-like structure whatsoever - that is the main criterion for a faithful implementation. 

As I said, I will do a proper review soon (probably tomorrow) and let you know my further thoughts then.

Thanks again,
Rutger

davepl:
Very cool that we’re still seeing submissions of new languages!  Thanks for taking the time for all of the work it must have been to even get to the point of a working Whitespace solution!

Cheers,
Dave

rzuckerm:
Whitespace does not have anything comparable to class, nor does Unicat, and I don't remember a class-like structure being mentioned in that review.

rbergen:
That's a fair comment. At the same time, I would have brought it up if you had pushed for a "faithful" classification, but the topic never came up (and I did read back the PR comments before I wrote that). 

I figured you marked that solution as not faithful because of this reason, but it's true I didn't check.

PrimeWhitespace/solution_1/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+WORKDIR /tmp/WSpace
+RUN apt-get update && \
+    apt-get install -y make && \
+    make && \
+    cp wspace /usr/bin/whitespace && \
+    apt-get remove -y make && \
+    apt-get autoremove -y && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* /tmp/WSpace
```
rbergen: I think this can be combined/compacted as follows:

```suggestion
RUN apt-get update && \
    apt-get install -y ghc git make python3 && \
    git clone --depth 1 https://github.com/TryItOnline/WSpace && \
    make -C WSpace && \
    cp WSpace/wspace /usr/bin/whitespace && \
    apt-get remove -y git make && \
    apt-get autoremove -y && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/* /tmp/WSpace
```

rbergen (commented):
As I said, I love this kind of contribution to the project. It doesn't _fully_ address #954 in the sense that I'd love to see even more esoteric language solutions, but this is definitely a very nice addition!

I have one suggestion concerning the Dockerfile, see my comment below. Please take a look at it to see if you want to adopt it. I'm happy to approve if you don't, but @marghidanu - who I will also ask for a review - may have a stronger opinion about Dockerfile inefficiencies. :)

While we work towards finalizing and merging this PR, I will update PrimeView to include Whitespace in its collection of known languages.

PrimeWhitespace/solution_1/Dockerfile
rzuckerm: I forgot about `make -C`. The only reason why I did it the way I did was due to a `hadolint` warning about using `WORKDIR` instead of the `cd`s that I was doing. This looks closer to what I originally had

rzuckerm:
> It doesn't fully address https://github.com/PlummersSoftwareLLC/Primes/issues/954 in the sense that I'd love to see even more esoteric language solutions, but this is definitely a very nice addition!

Yeah, I wasn't intending it to fully address the issue, but I figured I'd mention that issue so that the issue would have a record of this PR. I changed the wording from "addresses" to "inspired by" :smile_cat:

rzuckerm:
@rbergen I've made the requested `Dockerfile` change. Is there anything else I need to do?

rbergen:
@rzuckerm Nope. During work hours I'm a CTO, and that means that at this time of year I'm pretty caught up in things like year plans, annual reviews, etc. Which means Primes PR reviews can take a little more time than usual.

I'll request a review from @marghidanu as soon as I've had another look at the changes.

rbergen:
LGTM. Requesting review from @marghidanu now.

marghidanu (approved):
LGTM

PR #991, opened 2024-12-21
Faster PHP solution.

andrewthecodertx:
## Description
PHP solution that is about 5x faster than solution1

## Contributing requirements

### DIFF
--- a/PrimePHP/solution_2/Dockerfile
+++ b/PrimePHP/solution_2/Dockerfile
@@ -1,0 +1,9 @@
+FROM php:8.3-cli
+
+RUN mv "$PHP_INI_DIR/php.ini-production" "$PHP_INI_DIR/php.ini" && \
+    printf "zend_extension=opcache.so\nopcache.enable=1\nopcache.enable_cli=1\nopcache.jit_buffer_size=100M\nopcache.jit=tracing\nmemory_limit=-1\n" >> "$PHP_INI_DIR/php.ini"
+
+WORKDIR /opt/app
+COPY *.php .
+
+ENTRYPOINT [ "php", "prime_sieve.php" ]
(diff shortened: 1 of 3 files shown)


rbergen:
@andrewthecodertx Thank you for your submission. I've taken a look at it, and there are some things that I think need to be addressed before I can consider this PR further:

- This is the main one: the results of the solution are invalid, because the sieve size is wrong. You set the sieve size to 100,000. Because of that it yields a prime count of 9,592, which is not the correct value for a sieve size of 1,000,000. In fact, the solution's own output clearly states it isn't:

  ```log
  Passes: 6592, Time: 5000ms, Avg: 0ms, Limit: 1000000, Count: 9592, Valid: False
  
  andrewthecoder;6592;5.000142;1;algorithm=base,faithful=yes,bits=8
  ```

- The rules for a faithful solution indicate that: 
  
  > The sieve size and corresponding prime candidate memory buffer (or language equivalent) are set/allocated dynamically at runtime. The size of the memory buffer must correspond to the size of the sieve.
  
  In your code, the sieve size is effectively fixed to 100,000 because the sieve class constructor does not accept a parameter to set it. That would make this solution unfaithful.

- This solution has the same basic characteristics (algorithm, faithfulness, number of bits) as solution_1. As per the contribution guidelines, that usually means that if a speed improvement can be implemented, a PR to improve the existing solution should be opened instead of proposing to create a new solution. In this particular case I've compared the code of solution_1 to the solution you propose to add, and they are really very similar. Therefore, I would certainly not accept this as a new solution.

If you correct the first two issues and then still conclude there is a meaningful speed improvement compared to solution_1, then I suggest you modify this PR to augment solution_1 with the actual algorithmic differences that make that speed improvement happen (thereby addressing the third point).

andrewthecodertx:
you are exactly right. not sure how I missed that. thanks for the response.

PR #994, opened 2025-03-08
Add Primes in TI-Basic

rzuckerm:
## Description
This uses a python-based TI-84 Basic simulator. The docker image can be found [here](https://github.com/rzuckerm/ti-basic-docker-image/blob/main/Dockerfile).

## Contributing requirements

### DIFF
--- a/PrimeBASIC/solution_5/README.md
+++ b/PrimeBASIC/solution_5/README.md
@@ -1,0 +1,73 @@
+# TI-84 Basic Solution by rzuckerm
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-no-yellowgreen)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-unknown-yellowgreen)
+![Deviation](https://img.shields.io/badge/Deviation-sievesize-blue)
+
+This version is for the TI-84 graphing calculator. It uses a 
+[python-based simulator](https://github.com/rzuckerm/pitypas) that is a fork
+of [this repo](https://github.com/lunixbochs/pitybas). This program
+on a real TI-84 would most likely not handle a sieve size much larger than
+a few thousand due to memory constraints.
+
+Note that TI-Basic has no classes or even subroutines, so the code in just
+one big function. Also, there are no bitwise operations.
+
+In this implementation, each array element indicates whether the value is
+prime (0) or composite (1). The array only holds odd values from 3 to the
+sieve size. Also, the TI-84 only has a clock that measures hours, minutes,
+and seconds, where all values are an integer. Therefore, the duration is
+truncated to a one-second increment.
+
+## Run instructions
+
+Build the docker image with this:
+
+```bash
+./build.sh
+```
+
+You should only need to do this once. Run the docker image:
+
+```bash
+./run.sh
+```
+
+`run.sh` has three optional command-line arguments:
+
+- 1: Sieve size (default=100000)
+- 2: Time limit in sec (default=5)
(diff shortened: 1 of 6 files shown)


rbergen:
@rzuckerm Thanks for this submission. As you know I like adding solutions in less common languages, and I think Ti-Basic falls in that category. It also brings back memories of my own programmable calculator days, now between 25 and 30 years ago.

I've looked at the code, and I basically have one question: why not do a benchmark run with a sieve size of 1,000,000? What's the reason for the size deviation?

PrimeBASIC/solution_5/README.md
```diff
@@ -0,0 +1,73 @@
+# TI-84 Basic Solution by rzuckerm
+
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-no-yellowgreen)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-unknown-yellowgreen)
+![Deviation](https://img.shields.io/badge/Deviation-sievesize-blue)
+
+This version is for the TI-84 graphing calculator. It uses a 
+[python-based simulator](https://github.com/rzuckerm/pitypas) that is a fork
```
rbergen: I think there is a typo in the URL.

```suggestion
[python-based simulator](https://github.com/rzuckerm/pitybas) that is a fork
```

rzuckerm:
> I've looked at the code, and I basically have one question: why not do a benchmark run with a sieve size of 1,000,000? What's the reason for the size deviation?

The size deviation is that, at least on my PC, it takes over 20 sec to do a sieve size of 1,000,000. If that is OK, I can remove it. A similar reason was used for my Whitespace submission (which took several minutes to complete with that sieve size).

rbergen:
I expected that was the reason. 20-ish seconds is fine, though. The "soft limit" is about a minute per solution, which that is comfortably under. 

I would therefore prefer a benchmark run with the regular sieve size of 1,000,000.

rzuckerm:
> I expected that was the reason. 20-ish seconds is fine, though. The "soft limit" is about a minute per solution, which that is comfortably under.
> 
> I would therefore prefer a benchmark run with the regular sieve size of 1,000,000.

Done

rbergen:
Thanks for your changes. 

I'm afraid we will have to still mark the solution as unfaithful, though. Not because it's your fault (in fact, there's nothing you can do about it), but a faithful solution **must** use _some_ form of sieve encapsulation, including creation and tear-down of whatever the incarnation of sieve encapsulation is. 

Quoting from the contributing guidelines:

> - It uses a class to encapsulate the sieve, or a (closest) equivalent feature in your language if a class construct is not available. This class must contain the full state of the sieve. Each iteration should re-create a new instance of this class from scratch.

Again, I acknowledge that there is no way to do this in Ti-Basic, but that just means the language cannot be used to write a faithful solution by the standing definition. This restriction applies to a number of other languages as well.

rzuckerm:
The `ClrList l1` instruction effectively does the deallocation of the sieve, while `B->dim(l1)` does the allocation, so in that respect, it is faithful. However, the language lacks any type of data structure and the ability to call functions, so I can see why that would qualify it as unfaithful. In general, the older BASICs lack the ability of data structures.

Since this is being classified as unfaithful, what should I list as the reason?

rbergen:
You can list as the reason that Ti-Basic lacks the ability to encapsulate the full sieve as required for a faithful implementation. And that as such, it's an aspect of the language, not the solution in itself.

rzuckerm:
OK, I made the solution unfaithful again and added the reason why. What do I put in the "Deviations" field?

rbergen:
@rzuckerm I'm sorry, I missed that last question. Nothing, because there is no deviation any more. Which is what I saw in your README, hence my decision to merge.

PR #995, opened 2025-04-11
Refactor of Prime Extend algorithm in C

rogiervandam:
## Description
Refactored the code and improved speed by >10%

## Contributing requirements


rbergen:
That's quite the refactor! 

I'm just noting for the record that I did not performance-check this implementation in comparison to its previous version, because this is an update to the solution made by its original author.

PR #996, opened 2025-04-21
CFML

Willeye-uk:
## Description
This is my CFML addition

## Contributing requirements


rbergen:
@Willeye-uk Nice, thanks for submitting this! 

Very early in my career (read: almost a quarter of a century ago) I spent some time developing a lightweight e-commerce solution that combined NetObjects Fusion for the web storefront design and ColdFusion for the back-end logic. So, it's nice to have a CFML solution in!

However, I think we need to change the Docker setup on this. To be able to include the solution in the daily benchmark runs and as per the contribution guidelines, the container needs to start, run the sieve, output the results to stdout, and terminate.

I think this can quite easily be achieved by having the Docker image run a script that uses curl (or something similar) to pull the output from the Lucee server running within the container, and then convert the HTML to plaintext. 

With this setup, the container also doesn't have to expose its port, which in turn might make one drop the Docker Compose file and go with a straight docker build/docker run approach. CI and the benchmark logic will use that approach anyway (which means it needs to work regardless), but of course it's up to you to decide what you want to do within the context of your solution.

Is this something you could implement?

PrimeCFML/solution_1/PrimeSieve.cfc
```diff
@@ -0,0 +1,66 @@
...
+		
+		return cnt;
+	}
+
+	function validateResults() {
+		return structKeyExists(VALIDATION_DATA, this.n) and VALIDATION_DATA[this.n] eq count();
+	}
+
+	function runBitSet() {
+		var halfLimit = int(this.n \ 2);
```
rbergen: I think we already have this as `this.half_n`, so there is no need to recalculate it?

PrimeCFML/solution_1/PrimeSieve.cfc
```diff
@@ -0,0 +1,66 @@
...
+					sieve[i] = 0;
+				}
+			}
+		}
+	}
+
+	function printResults(duration, passes) {
+		var count = this.count();
+		var label = "willeyuk";
+		var bits =  "1";
```
rbergen: I don't believe this is correct. A CFML `number` translates to a Java `double`, which is 64 bits in length.

If you do want a 1-bit implementation, you could use a Java `BitSet` for that. ColdFusion being based on Java since 2002 and Java interop having been an explicitly supported and encouraged part of CFML since then - particularly for memory-efficient use cases like this -, I'm happy to allow a `CreateObject` invocation to create one.

(In fact, you can also use the `new java.util.BitSet` syntax considering you're targeting Lucee, but that would obviously break the solution if one were to run it on Adobe CF.)

PrimeCFML/solution_1/index.cfm
```diff
@@ -0,0 +1,22 @@
...
+sieve = new PrimeSieve(limit);
+passes = 0;
+
+// Timed run
+start = getTickCount();
+do {
+		sieve.runBitSet();
+		passes++;
+
+} while (getTickCount() - start < runTime);
```
rbergen: As per the contribution guidelines, the creation and initialization of the sieve needs to take place within the timed loop.

rbergen:
Concerning the Dockerfile stuff, ChatGPT put this together for me: [Dockerstuff.zip](https://github.com/user-attachments/files/19831051/Dockerstuff.zip)

I'm sharing it for convenience, but haven't tested it myself. Which means it may not work, but it doesn't look like it will burn down your house either.

Willeye-uk:
I will implement all your suggestions and i will create a solution 2 thats uses java

rbergen:
I appreciate that, thanks in advance! 

About the second solution: if you can, please combine both implementations in one solution - this is widely done throughout the project. It will save the build and warm-up time and additional storage of a second Docker image for those people who execute the benchmark runs.

PrimeCFML/solution_1/index.cfm
```diff
@@ -0,0 +1,21 @@
...
+passes = 0;
+
+// Timed run
+start = getTickCount();
+sieve = new PrimeSieve(limit);
+do {
+		sieve.runBitSet();
+		passes++;
+
+} while (getTickCount() - start < runTime);
```
rbergen: This may now start to sound like nit-picking, but the sieve creation/initialization really needs to be _inside_ the do/while loop. The point is that each run starts with a completely new and freshly initialized sieve.

Willeye-uk: I have fixed this now

Willeye-uk:
I tried a few other solutions nothing was faster so i have reverted to lucee class

PrimeCFML/solution_1/index.cfm
Willeye-uk: Fixed

Willeye-uk:
So the latest commit, works when container starts Tomcat, it waits the runs the page and outputs to console results

PrimeCFML/solution_1/README.md
```diff
@@ -0,0 +1,16 @@
+# cfml (Lucee) solution by willeyeuk
+
```
rbergen, line 2: Suggested addition to add badges to the README:

```suggestion

![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
![Parallelism](https://img.shields.io/badge/Parallel-no-green)
![Bit count](https://img.shields.io/badge/Bits-64-yellowgreen)

```

PrimeCFML/solution_1/README.md
```diff
@@ -0,0 +1,16 @@
...
+```shell
+Run container
+```
+
+## Output
+
+```shell
+Passes: 20, Time: 5.042, Avg: 0.2521, Limit: 1000000, Count: 78498, Valid: true
+willeyuk;20;5.042;1;algorithm=base,faithful=yes,bits=1
```
rbergen: Suggested update to reflect the actual number of bits used:

```suggestion
willeyuk;20;5.042;1;algorithm=base,faithful=yes,bits=64
```

rbergen (commented):
Thanks for these updates. I made two concrete suggestions for changes to the README, please refer to the comments.

I think a 1-bit implementation will also not be faster than this one due to the bit shifting and masking that's required, but such an implementation would qualify for the single-threaded leaderboard.

Willeye-uk:
Added new version as used in my current work which gives
Passes: 18, Time: 5.133, Avg: 0.285166666667, Limit: 1000000, Count: 78498, Valid: true
willeyeuk;18;5.133;1;algorithm=base,faithful=yes,bits=64
Passes: 6856, Time: 5, Avg: 0.000729288215, Limit: 1000000, Count: 78498, Valid: true
willeyeuk;6856;5;1;algorithm=base,faithful=yes,bits=1
true

Willeye-uk:
Would it be worth creating a wasm version?

rbergen:
> Would it be worth creating a wasm version?

Assuming you mean "wasm" as in "WebAssembly", then in the context of this solution I don't see how that would work, as there is no logic executed within the browser. 

In a more general sense, I'd love a WebAssembly solution, but it would basically have to be coded by hand. Similar to Java's bytecode or .NET's CIL, WebAssembly is usually a compilation target, not a language that people code in. Which is probably also why neither bytecode nor CIL are represented in the is project's list of languages either.

PrimeCFML/solution_1/Dockerfile
rbergen, line 1: Stating first I haven't tested it, but I think we still have a problem with this Dockerfile. I mean that in the sense that the container keeps running after the solution has been executed and its output has been written to stdout. To be included in the benchmark runs, the container **must** terminate after solution execution has ended.

I believe the only way to do this is specifying a CMD in the Dockerfile that runs a script that starts Tomcat/Lucee, waits for it to be up, does the curl, prints its output, shuts Tomcat down again, and ends. That script can itself use the script that the Tomcat base image normally uses to start Tomcat, which I believe to be catalina.sh.

I'll have time to look into this in more detail (including working on an updated Dockerfile), this weekend.

If you want to _also_ include a Dockerfile that keeps the container running you can do that, but it can't be called "Dockerfile", as that's the one CI and the benchmark runs use.

Willeye-uk: I will fix dockerfile

Willeye-uk: Fixed shutdown issue

rbergen:
@Willeye-uk, I've spent some more time reviewing your solution, and I think we developed some static on the line, as well as have slightly different ideas on how to isolate core logic from execution context.

In short:

- Your second implementation is effectively a pure Java implementation, invoked from CFML. That's not something we can do in a solution marked as "CFML"; it would become a mixed-language solution instead. I mentioned before that I was okay with a CFML implementation creating and using a java.util.BitSet within the CFML code (for reasons explained at the time), but that's a different thing. 
- You solved the shutdown of the container by invoking the Tomcat control script from within index.cfm. That approach tightly integrates the actual code with the benchmark execution context, and is something I think we should not do. People might want to run the solution in their own (possibly non-Dockerized) context, and then this approach will either error out, or shutdown a Lucee instance they really need to keep running.

To short-circuit this, I've created an updated version of the solution in the [`cfml1-update`](https://github.com/rbergen/Primes/tree/cfml1-update) branch in my fork. I'll open a PR on your `drag-race` branch from it, so you can see the changes I think need to be made and adopt them if you want. I've tested it, and it works on my machine.

Some other changes I've made:

- The code is now compatible with both Lucee and ColdFusion - as I said, I used to program in ColdFusion back in the day, so I like the idea of the solution running in that environment as well.
- The benchmark run script that's used in the Dockerfile now polls Tomcat to see if Lucee is up.
- I pipe the index.cfm output through html2text to remove the HTML tags; this is necessary for the benchmark result parsing to work.

Have a look, and let me know what you think.

Willeye-uk:
Happy with your changes, i added the java class as that is the route we are taking at work to port applications. I used this as a demo.

rbergen:
I see. I now understand that on the Java front I connected some dots that I shouldn't have.

Am I correct in concluding you're okay with submitting the solution as it now stands? In that case I will move on to request a second review on this; we always do this for a new solution in a new language.

Willeye-uk:
All happy

rbergen:
Alright, great! Tagging @marghidanu for the second review.

marghidanu (approved):
LGTM

rbergen (approved):
...same here, as indicated.

PR #997, opened 2025-04-26
BoxLang Addittion

Willeye-uk:
Description
This is my BoxLang  https://boxlang.io/ addition

Contributing requirements
 I read the contribution guidelines in CONTRIBUTING.md.
 I placed my solution in the correct solution folder.
 I added a README.md with the right badge(s).
 I added a Dockerfile that builds and runs my solution.
 I selected drag-race as the target branch.
 All code herein is licensed compatible with BSD-3.


Willeye-uk:
This was supposed to only be my BoxLang addition, Not used to PR's over forks

rbergen:
This is because you continued working on the drag-race branch in your fork that was created when you created your fork. It's why I create a feature branch in my fork (which you can do straight off the main tree) just before I start working on a new solution or feature.

In any case, in terms of commits it doesn't matter, as I always squash commits on merge. It just means GitHub will be inclined to credit me as a co-author of that squashed commit, which I will try to prevent from happening.

PrimeBoxLang/solution_1/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM ortussolutions/boxlang:cli-snapshot
```
rbergen: Please don't use snapshot releases. In principle, we require solutions to use specific versions, but I see on Docker Hub that none are provided for boxlang. In that case, the still more stable "cli" is preferred.

Willeye-uk: Ok will fix

Willeye-uk:
All fixed

PrimeBoxLang/solution_1/sieve.bx
```diff
@@ -0,0 +1,61 @@
...
+                cnt++;
+            }
+        }
+
+        return cnt;
+    }
+
+    function printResults(duration, passes) {
+        var count = this.countPrimes();
+        var label = "willeyeuk-BoxLang";
```
rbergen, line 56: Just noting this for the record: you don't have to add the language to a label to distinguish it from other implementations in other languages. Labels are always interpreted within the context of the language it concerns (BoxLang in this case, CFML for your the solution you recently submitted).

I'll leave it to you if you want to change the label or not, it's not a requirement to do so.

Willeye-uk: I will leave as is, but noted. thanks

rbergen (approved):
LGTM

@Willeye-uk I made some changes in the context of this PR to make CI on one of the V solutions pass; this is not your fault, it's just how V(lang) works from time to time. 

As with the CFML solution, I will ask for a second review from @marghidanu, this also being a new solution in a new language.

rbergen:
Merging on the premise of approval by absence of objection.

marghidanu:
I'm a bit late to the party, but LGTM 👍

PR #998, opened 2025-08-30
Add Primes in M4

rzuckerm:
## Description
Implementation of prime sieve in M4.

## Contributing requirements

The base docker image is available [here](https://hub.docker.com/r/rzuckerm/m4/tags). Here is the [Dockerfile](https://github.com/rzuckerm/m4-docker-image/blob/main/Dockerfile).

### DIFF
--- a/PrimeM4/solution_1/Dockerfile
+++ b/PrimeM4/solution_1/Dockerfile
@@ -1,0 +1,6 @@
+FROM rzuckerm/m4:1.4.19-3
+
+WORKDIR /opt/app
+COPY primes.m4 run_primes.sh ./
+
+ENTRYPOINT ["./run_primes.sh"]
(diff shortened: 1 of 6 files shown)


rbergen (approved):
@rzuckerm Thanks for submitting this. I've said it before and I'll say it again: I love solutions in languages that are not intuitively suitable for solving the challenge this project poses, like this one.

LGTM, but as it's a new solution in a new language, I'll ask for a review from @marghidanu as well.

rzuckerm:
I realized that I was making the time calculation more complicated than it needed to be. M4 truncates any number, no matter how large, to 32 bits:
```m4
eval(`12345678901234567890')
-350287150
eval(`123456789012345678901234567890')
1312754386
```

Also, `start > end` doesn't need to be mitigated. For example:
```m4
define(`start', `0x7fffffff')
define(`end', `0x80000000')
eval(start)
2147483647
eval(end)
-2147483648
eval(end-start)
1
```

rbergen:
Makes sense. Still LGTM.

marghidanu:
LGTM for me too. Let's roll it.

PR #999, opened 2025-08-30
Suppress all checks for calculation

Evan-LucasCurrie:
Ada does runtimes checks; using suppress will improve performance by disabling them all. Link to documentation: https://docs.adacore.com/live/wave/gnat-ccg/html/gnatccg_ug/gnat_ccg/gnat_ccg.html#:~:text=3.3.%20Enabling/Disabling%20Runtime%20Checks,-%EF%83%81

Using godbolt, we see a reduction of a few thousand lines of assembly.

### DIFF
--- a/PrimeAda/solution_1/main.adb
+++ b/PrimeAda/solution_1/main.adb
@@ -1,3 +1,4 @@
+pragma Suppress (All_Checks);
 with Ada.Containers.Ordered_Maps;
 with Ada.Text_IO;   use Ada.Text_IO;
 with Ada.Numerics.Long_Long_Elementary_Functions;


rbergen:
As this is a solution submitted by @BoopBeepBoopBeep, I will allow them some time to indicate if they agree with this change to their solution.

Evan-LucasCurrie:
This was an oversight on my behalf but when I add all compiler flags I see a reduction. However, when I single out -gnatp there is no change I figured it was a quirk of ada and the compiler. This might be worth investigating but I am going to close this PR, as I think the issue might be my machine. 

This is a prime example of works on my machine as I didn't test this with docker itself, thats on me.

PR #1015, opened 2025-09-07
Note that CUDA/solution_1 only runs on AMD64

rbergen:
## Description

CUDASieve contains embedded AMD64 Assembly, which makes it only builds on that architecture. Hence, the same applies to CUDA/solution_1; this updates the README for that solution to mention this.

## Contributing requirements


rbergen:
Merging based on out-of-band approval.

PR #1018, opened 2025-09-07
Mojo/solution_1 Dockerfile: Clean up after apt-get install

rbergen:
## Description

We overlooked that the apt-get install in the Dockerfile for Mojo/solution_1 was not followed by the appropriate clean-up. This fixes that.

## Contributing requirements


rbergen:
Merging based on out-of-band approval.

PR #1023, opened 2025-10-12
add CITATION.cff and LICENSE files

jonaspleyer:
## Description
I am currently writing a research article where I would like to mention the results of this study. Therefore I would like to cite this work. Thus I am adding the `CITATION.cff` file.

Furthermore, I noticed that the project does not seem to have a `LICENSE` file although all submissions are required to be compatible with the [BSD 3-Clause](https://opensource.org/license/BSD-3-clause) license. Within the license file, one typically states the year and the copyright owners. How we might fill the latter is debatable. I view my PR as a suggestion and of course hold no authority over this. Please let me know what you think.

## Contributing requirements

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,11 @@
+cff-version: 1.2.0
+message: "If you use this software, please cite it as below."
+authors:
+  - family-names: van Bergen
+    given-names: Rutger
+  - family-names: Marghidanu
+    given-names: Tudor
+  - family-names: Plummer
+    given-names: David W
+title: "Primes | A Software Drag Race"
+url: https://github.com/PlummersSoftwareLLC/


rbergen:
@jonaspleyer First off, thank you for making these suggestions, as well as shaping them as a PR.

I think labelling this project as a "study" is very flattering and maybe giving it too much credit. As Dave has stated multiple times, and quite recently in a Lex Fridman podcast, this basically started as a code share of a Dave's Garage episode that then got a bit out of hand. 

That said, it is true that as maintainers we have made a sincere effort to make the collection of solutions suitable for a "fair" comparison, and facilitate that comparison being executed and the result being reviewable on a daily basis.

Concerning the contents of the PR in terms of files, I'm happy to accept the addition of the CITATION.cff file. The LICENSE file I will not.

The reason for the latter decision, which we've actually made some years ago, is that the requirement is that any code that is submitted is _compatible_ with BSD-3 in the sense that it is _at least as permissive_ as that license. The license that applies to an individual solution does not have to be BSD-3 itself, and some contributors have indeed chosen to license their solutions differently. Also, including a LICENSE file at the project level that mentions David, Tudor and myself as the copyright holders implies or can easily be seen to claim that the copyright of all the solutions in this project lies with us. That's not the case, per se.

All that said, I would love to take a look at your research article, once you've written it and provided you're willing to share it.

jonaspleyer:
Thanks for the quick feedback. I think that there are very good arguments that the results of this project can be considered scientifically valuable. My article is about the usage of Rust in the context of scientific computing and as such this project provides a good window to showcase that Rust's performance matches that of C or C++ (which are the languages that I am most concerned about). I think that this point is made clear when inspecting the [results](https://plummerssoftwarellc.github.io/PrimeView/). Once the article is available, I can link it here again.

Concerning the License file: I understand your motives and have thus removed the commit which introduced it.

PR #1024, opened 2025-10-25
Vectordave - Add vectorization on x86/ARM

davepl:
This adds vectorization approaches for SSE2, AVX, AVX512, and NEON.  It's worth about 30% perf on the Mac and ubdellamd.

It also enables __restrict__ in the hopes it keeps the compiler assured that sieve memory is not being aliased.

### DIFF
--- a/PrimeCPP/solution_5/PrimeCPP_array.cpp
+++ b/PrimeCPP/solution_5/PrimeCPP_array.cpp
@@ -20,6 +20,27 @@
 #include <memory>
 #include <cstdlib>
 
+#if defined(_WIN32)
+#  include <malloc.h>
+#endif
+
+#if defined(__aarch64__) || defined(_M_ARM64)
+#  include <arm_neon.h>
+#  define PRIMECPP_VECTOR_NEON 1
+#elif defined(__AVX512F__)
+#  include <immintrin.h>
+#  define PRIMECPP_VECTOR_AVX512 1
+#  define PRIMECPP_VECTOR_AVX2 1
+#  define PRIMECPP_VECTOR_SSE2 1
+#elif defined(__AVX2__)
+#  include <immintrin.h>
+#  define PRIMECPP_VECTOR_AVX2 1
+#  define PRIMECPP_VECTOR_SSE2 1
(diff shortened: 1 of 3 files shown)


PrimeCPP/solution_5/PrimeCPP_array.cpp
```diff
@@ -256,7 +317,176 @@ class BitArray
...
+        }
+
+        if (cycleLen > 0)
+        {
+#if defined(PRIMECPP_VECTOR_AVX512)
+            while (wordIndex + cycleLen <= fullWordCount)
+            {
+                size_t idx = 0;
+                while (idx + 8 <= cycleLen)
+                {
```
rbergen, line 363: This question applies to all main and nested `while` loops in this section of code: is there a particular reason for using `while` instead of `for`?

I know this is dangerously close to a coding style comment and I'm the first to say we shouldn't spend too much time on those, certainly in this project. However, I think combining the loop condition and the increment expression (which is neatly at the end of every loop) in one line clarifies the "cycle logic" in this particular case.

davepl:
Good point - the structure required it the way I was doing the preprocessor stuff originally, but now it looks silly.  Any cases that have init, condition, AND increment should be a for !

Thanks
Dave

davepl:
So I went through the code and did an AI refactor to for(), and it’s uglier :-). Most of the fors() can’t have the indexer in score because it’s used in multiple loops.

You wind up with a lot of this, and I actually prefer the while()!

- Dave

￼

rbergen:
The screen snippet doesn't come through when you email respond to GitHub comments, but I imagine it's an example of a for loop with an empty first clause (`for (; wordIndex + cycleLen <= fullWordCount; wordIdex += cycleLen)` or `for(; idx + 8 <= cycleLen; idx += 8)`). 

If that's right then I was aware of this, but I still think the looping logic itself is more clearly expressed in a `for` context. However, the fact I'm one of the maintainers of this project doesn't change it's _your_ solution, so if you prefer the `while`, then that's what we'll merge.

davepl:
I prefer the current while structure!  So, LGTM*

- Dave

* let’s get this merged!

Issue #1027, opened 2025-12-13
Mojo/solution_1 fails to build


rbergen:
Tagging @Evan-LucasCurrie as the author of the solution.

Mojo/solution_1 started generating errors during the docker build step. I've included the output of the build step in question below:

```log
#13 [6/6] RUN mojo build primesieve.mojo -o primesieve
#13 0.415 /app/primesieve.mojo:64:32: warning: deprecated implicit conversion from 'UInt' to 'Int'
#13 0.415         self.array.clear(index // 2)
#13 0.415                          ~~~~~~^~~~
#13 0.415 /app/primesieve.mojo:64:32: note: call 'Int(...)' explicitly
#13 0.415         self.array.clear(index // 2)
#13 0.415                          ~~~~~~^~~~
#13 0.415                          Int(      )
#13 0.415 /app/primesieve.mojo:1:1: note: '@implicit' constructor 'Int.__init__' declared here
#13 0.415 import time
#13 0.415 ^
#13 0.448 /app/primesieve.mojo:80:31: warning: deprecated implicit conversion from 'Int' to 'UInt'
#13 0.448                 self.clearBit(num)
#13 0.448                               ^~~
#13 0.448 /app/primesieve.mojo:80:31: note: call 'UInt(...)' explicitly
#13 0.448                 self.clearBit(num)
#13 0.448                               ^~~
#13 0.448                               UInt( )
#13 0.448 /app/primesieve.mojo:1:1: note: '@implicit' constructor 'UInt.__init__' declared here
#13 0.448 import time
#13 0.448 ^
#13 0.467 /app/primesieve.mojo:153:29: error: invalid redefinition of 'sieve_size'
#13 0.467     fn __call__(self: Self, sieve_size: Int) -> Self:
#13 0.467                             ^
#13 0.467 /app/primesieve.mojo:144:30: note: previous definition here
#13 0.467 struct prime_sieve_1bit_meta[sieve_size: Int](Runnable):
#13 0.467                              ^
#13 0.469 /app/primesieve.mojo:162:32: warning: deprecated implicit conversion from 'UInt' to 'Int'
#13 0.469         self.array.clear(index // 2)
#13 0.469                          ~~~~~~^~~~
#13 0.469 /app/primesieve.mojo:162:32: note: call 'Int(...)' explicitly
#13 0.469         self.array.clear(index // 2)
#13 0.469                          ~~~~~~^~~~
#13 0.469                          Int(      )
[...]

PR #1028, opened 2025-12-13
Fix build errors in existing solutions

rbergen:
## Description

This PR addresses build errors in 4 existing solutions:

- Assembly/solution_1, Umple/solution_1 and Yoix/solution_1 started failing because the deprecated openjdk Docker Hub images that the solutions' Dockerfiles derived from are no longer available. I have switched these solutions to corresponding eclipse-temurin images instead.
- Mojo/solution_1 fails because the solution's code is no longer compatible with the version of Mojo that the Dockerfile installs. I have disabled the build and hence benchmark runs for this solution until a fix is provided; I have opened issue #1027 to request one.

## Contributing requirements

### DIFF
--- a/PrimeAssembly/solution_3/Dockerfile
+++ b/PrimeAssembly/solution_3/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:22-slim
+FROM eclipse-temurin:22-jdk-noble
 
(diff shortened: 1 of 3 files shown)


marghidanu (approved):
LGTM

PR #1029, opened 2025-12-20
Rewrite Batch Worker: Correctness Fix + Major Performance Improvements Using A[n] Loop Implementation

ch2laughlin:
## Summary
This PR replaces the original Batch worker with a corrected and significantly faster implementation based on A[n] loop logic and an optimized Newton integer square root. The new worker is faithful, produces correct prime lists, and dramatically improves performance across all sieve sizes.

The original worker (using goto labels and a broken sqrt) is preserved as worker_orig.bat for historical reference and comparison.  A new worker that uses a bit packing agorithm is also added and results available for others to study or improve on.

### ✔ Correctness
- Replaces the broken square‑root logic with a fast, accurate Newton integer sqrt.
- Removes label‑based control flow in the hot path.
- Produces correct prime lists for all tested sieve sizes.

### ✔ Performance
The new worker uses:
- `for /L` loops instead of `goto`/labels  
- short variable names (`A[n]`) to reduce environment‑variable overhead  
- incremental variable creation (instead of pre‑allocating)  
- fewer interpreter‑level operations per composite  

## Performance Comparison

| Sieve Size | Original  | New A[n] loop Worker | Speedup |
|-----------:|------------------------:|----------------:|--------:|
| 1k         | 1.52 s                 | 0.12 s          | 12.7×   |
| 5k         | 8.74 s                 | 0.81 s          | 10.8×   |
| 10k        | 18.74 s                | 1.69 s          | 11.1×   |
| 50k        | 154.39 s               | 28.41 s         | 5.4×    |
| 100k       | 631.49 s               | 106.26 s        | 5.9×    |
| 500k       | —                      | 2,276.11 s      | —       |
| 1M         | 37,915.85 s            | 9,937.02 s      | 3.82×   |

## Changes
* worker.bat - new A[n] loop implementation (default)
* worker_orig.bat - preserved original implementation
* worker_packed.bat - packed/bitmap version
* main.bat - updated framework with improved timing accuracy n timing precision
* readme.md - updated documentation and performance data.  Added details about the 3 available worker algorithms.
* timekeeper.bat - no longer needed timing synchronization script

## Contributing requirements

This PR is structured to be squash‑merged so that only the final combined change lands in upstream history


PrimeBatch/solution_1/README.md
```diff
@@ -1,13 +1,69 @@
-# Batch solution by Sxxov
+# Batch solution by Sxxov & Ch2Laughlin
 
-![Algorithm](https://img.shields.io/badge/Algorithm-base-green)![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)![Parallelism](https://img.shields.io/badge/Parallel-no-green)![Bit count](https://img.shields.io/badge/Bits-unknown-yellowgreen)
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
+![Parallelism](https://img.shields.io/badge/Parallel-no-green)
+![Bit count](https://img.shields.io/badge/Bits-32-yellowgreen)
```
rbergen: I believe this to be incorrect. In my understanding:

- for worker.bat, the bit count per prime candidate is unknown, because whether bits are used or not depends on if the candidate is marked as a composite or not. I.e.: the total number of bits used for the full sieve depends on the number of composites found, which makes the number bits per prime candidate undetermined from an algorithm perspective.
- for worker_pack.bat, the bit count per prime candidate is 1.

Hence, the bit count badges should be set as such:

```suggestion
![Bit count](https://img.shields.io/badge/Bits-unknown-yellowgreen)
![Bit count](https://img.shields.io/badge/Bits-1-green)
```

PrimeBatch/solution_1/README.md
```diff
@@ -1,13 +1,69 @@
-# Batch solution by Sxxov
+# Batch solution by Sxxov & Ch2Laughlin
 
-![Algorithm](https://img.shields.io/badge/Algorithm-base-green)![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)![Parallelism](https://img.shields.io/badge/Parallel-no-green)![Bit count](https://img.shields.io/badge/Bits-unknown-yellowgreen)
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
```
rbergen, line 4: I'm aware that what follows also applies to the original solution. I'd argue that if we're improving the solution, we should do the same with its faithfulness classification.

The solution is unfaithful because two of the conditions for faithful solutions are not met:

- For both worker.bat and worker_pack.bat: _It uses a class to encapsulate the sieve, or a (closest) equivalent feature in your language if a class construct is not available. This class must contain the full state of the sieve. Each iteration should re-create a new instance of this class from scratch._
  **Reasoning**: there is no class-like encapsulation for either implementation. I note for completeness that the absence of such a capability in the language is not a reason to deviate from the condition.
- For worker.bat: _The sieve size and corresponding prime candidate memory buffer (or language equivalent) are set/allocated dynamically at runtime. The size of the memory buffer must correspond to the size of the sieve._
  **Reasoning**: There is no allocation of a memory buffer, but a variable per identified composite number. Hence, the amount of memory used corresponds to the number of composites within the sieve, not the size of the sieve itself.

Hence, the faithfulness badge should be set as follows:

```suggestion
![Faithfulness](https://img.shields.io/badge/Faithful-no-yellowgreen)
```

PrimeBatch/solution_1/main.bat
```diff
@@ -7,156 +7,188 @@ setlocal enableDelayedExpansion
...
+
+    set "csDisplayFr=!%3.cs!"
+    if !csDisplayFr! lss 10 set csDisplayFr=0!csDisplayFr!
+
+	echo framework: !%3.hours!:!%3.mins!:!%3.secs!.!csDisplayFr! (!secsPartFr!.!csPartFr!s total)
 	echo.
 	echo ---
-	echo batch;%2;!%1.totalSecs!.!%1.ms!;%options.workers%;algorithm=base,faithful=yes,bits=unknown
-	(exit /b)
+	echo batch;%2;!secsPart!.!csPart!;%options.workers%;algorithm=base,faithful=yes,bits=32
```
rbergen: ~~In line with earlier comments, the flag section should reflect that the solution is not faithful and the bit count is either unknown (absent) or 1, depending on what implementation is being used.~~

PrimeBatch/solution_1/worker_pack.bat
```diff
@@ -0,0 +1,138 @@
...
+			)
+		)
+	)
+	goto :eof
+
+:markComposite (int: num)
+    set /a "varindex=%1/30"
+    set /a "boffset=%1%%30"
+    set /a "mask=1<<boffset"
+    set /a "C[%varindex%]=!C[%varindex%]!|mask"
```
rbergen, line 97: Just noting for the record that further packing is possible by dividing num by 2; only odd numbers will ever be marked as composites, so it's not necessary to reserve bits in the sieve space for even numbers. Of course, other functions (`isPrime` for one) should be changed accordingly.

rbergen (changes requested):
Thank you for submitting this. I have a few comments, mainly about the characteristic qualifications.

PrimeBatch/solution_1/README.md
ch2laughlin: I settled on 32bit based on the size of a candidate.  Not the size of a prime, which is an optimized case in this environment.  As N get large the # of primes diminishes and the approximation PrimeSize = bits_used_for_all_numbers / sieve_size rapidly approaches the size of a candidate - which is 32bit.

in worker.bat composites are set with /a which makes them a 32bit number, the basis of all integer math in cmd.  If a composite is not set (the number is prime) its actually zero bits - as the env var is not set (its not a zero, its not equal to null, its an undefined var).  However, all integer math is 32bit in cmd - and all the sets are of form /a so its safe to say that all candidates are 32bit integers by defn of the cmd language.  For proof, you could add a setting of all candidates at the start of the algorithm to be = 0, then allow the current algorithm to run unchanged except use the prime_test as candiate=0 instead of 'is defined'.  The algorithm will work the same, but much slower since candidate=0 is internally both a 'is defined' and then comparison of value.   However this approach does not help answer the bit size question - you either end up with 0bits cause the primes themselves are undefined or 32bits cause by env. definition all math is 32bit.

Now, if you wish to argue that about what the size of an undefined variable memory size is in CMD, its 0.  It takes up no space.  That is not, an accurate assessment of the bit count used for a given sieve size.  I would propose the bit count is asking the size of any given candidate *or* prime.  The closest would be 32bit - an integer number when performing math via set /a and we use the smallest value possible in this case.

You could also argue, what is the size if we only consider the candidates as strings - the default env var storage.  Well, that is either 8bit or 16bit depending on if your using an older ANSI cmd version or a newer Unicode enabled version.  And given that there is a null at the end and candidates are a single '1' when set you end up with 16bit for the older ansi vers and 32bit for the widely available cmd versions found in Win 8/10/11 etc.  So back to 32bits as the approximation.

As best I can tell - 32bit is the best answer giving an accurate sense of how much space we allocate per candidate when the algorithm is run.

I hope this clarifies.

PrimeBatch/solution_1/README.md
ch2laughlin: 1. worker.bat _is_ the class equivalent feature.  it recreates from scratch the sieve state from scratch.

2. The amount of memory allocated corresponds to the amount needed to hold the sieve size and corresponding prime candidates.  Your stating that languages with non-uniform arrays are not faithful, I don't see that the best interpretation of "the memory buffer must correspond to the size of the sieve".  If an algorithm wants to mark only Candidates in a hash or in a map the size of the dictionary in both cases is the memory buffer.  Worker.bat is allocating dynamically during runtime a memory buffer that holds all composites (and by negation the primes).  That memory buffer is unique and corresponds to the sieve size. 

While in C you may want to use an array which means both candidates and primes are going to take up space - there are plenty of languages that support non-uniform arrays.  Then I may define just the primes or just the candidates or both.  Its also worth noting that usually in such cases you need the size of the array/mapping _and_ the mapping/array.  Worker.bat fits well into this situation containing both the size of the sieve _and_ the composites are all defined.  The primes are all undefined in the same memory allocation.  This is by defn a full working memory set that corresponds to the size of the sieve.

It is possibly helpful to understand that inside CMD while executing a batch file all variables are consolidated into a single buffer - especially if set local is true and we later need to 'pop' all the variables defined by the script and revert back to a previous set.

PrimeBatch/solution_1/worker_pack.bat
ch2laughlin: noted.  the algorithm already skips evens and the further packing does not make it a better choice as the slowness of the extra bitflip code is what kill the perf.  Others can explore further, this was good enough for the lunch crowd which was curious about trying to reduce the impact of scanning the larger env var space as the candidates grew.  The multiple sets can also be packed together into a single set cmd with the ',' - but small diffs like that won't win the day against the loop implementation in worker.bat

PrimeBatch/solution_1/README.md
rbergen: It really is about the actual number of _memory bits used per prime candidate_, as has been well-documented in many conversations I've had about this topic over the past ~5 years. You're arguing to take the _intended number of bits per marked composite_ as the number, and that is simply not the same thing.

Aside from the fact that the number of bits per prime candidate varies because of the algorithm (i.e., some candidates taking 0), the actual memory used per marked composite is more than 32 bits. That is indeed the bit width used for the calculations, but the results are stored in CMD strings. 

On top of that, you accurately bring up the other point of variables living in environment space, which makes the overall actual memory being used per marked composite even more unclear.

Hence, the number of memory bits used per prime candidate is unknown for the unpacked algorithm.

For the packed implementation, I'm happy to take 1 as the bit count, because the algorithm clearly allocates the entire sieve space in one go, and then modifies one bit per prime candidate - composite or not. As has also been documented in earlier discussions, we can then take the string + env space memory usage as "consistent overhead". If you want to argue with that you can; the alternative is to mark that implementation's bit count as unknown also, because of said overhead.

PrimeBatch/solution_1/README.md
rbergen: 1. I was going to argue that worker.bat is the implementation of a code _structure_ capability, not an actual language feature; with the latter I mean a feature of the language as specified in its documentation. Even though I still stand by the latter point, I can see that turning into a semantical conversation, so I'm willing to accept your definition, on one condition: at least the sieve size and the noValidate option need to be passed as parameters to the workers when invoked. With that in place, I think it's fair to say that the workers are self-contained logic units, with the runtime configurability that classes typically offer. 
   To make it more clear that you consider the workers the class equivalent, it may also be helpful to rename worker.bat to sieve.bat, and the packed version accordingly.

2. I've already responded to this in my previous comment. As I've said there, I've had these conversations many, many times, so I am comfortable with my interpretation of the rules being the right one. You can still disagree with it but it does provide consistency across solutions, which matters in itself in a project like this.

ch2laughlin: the sieve size and noValidate options are passed to the workers for all 3 worker implementations.  This is done using the 'options.<field>' set of variables and can be set on the command line when calling the main framework. If they are not passed then the worker loads default values for each.

Would naming the worker bat files as sieve.bat sieve_packed.bat etc be closer to the common structure?  Its easy to rename a few files to make the structure clearer.

rbergen: I'm aware of the `options.` set of variables. My point was that an explicit pass of the parameters at the time of invocation of a worker more closely matches the semantics of a class instance being constructed, and would therefore much better support the plea to consider the solution faithful.

In C++ parlance, it's using constructor parameters (command-line args when invoking the worker) vs. using global variables (`options.` variables). And this has been a topic in earlier conversations about the faithfulness of solutions, too.

And yes, the renames you refer to is exactly what I intended to suggest.

PrimeBatch/solution_1/main.bat
rbergen: Updated version of the previous comment: in line with earlier comments, the flag section should reflect that the bit count is either unknown (absent) or 1, depending on what implementation is being used.

PrimeBatch/solution_1/README.md
ch2laughlin: I will make the requested changes.  Later I'd be interested in reading the background discussion on why its this way.

rbergen: I appreciate that, thank you in advance. 

The background discussion is not one discussion, but a collection of dozens if not hundreds of exchanges about the application of the rules - rules which themselves evolved to their current phrasing and application in the early phases of this project.

Bullet-point with what I think is a responsible amount of simplification:

- Dave's original implementations that kicked off this project (the implementations discussed in the YouTube video the README links to) were all in object-oriented languages. That made object-orientation a de facto premise for the implementations in other languages that were then added by others. I guess an underlying reasoning was that there is a cost to structuring things in a certain way, and that a fair comparison in the context of the leaderboard requires this to be applied to all languages on it.
- The fair comparison thing was also the foundation for other rules, including the determination of the way the "bits per prime" are counted. In actual fact, from a leaderboard perspective the definition of "one bit per prime" is most relevant, as only those implementations qualify for it. To avoid the situation where confusion or ambiguity about the way the bit count per prime is established for implementations using more than one bit permeates the clarity of the definition of one bit per prime, we decided to be equally strict about all definitions. 
- In multiple conversations about the interpretation of the rules, the objection was raised that the rules' are somewhat arbitrary. I'd say that is true insofar it concerns their _composition_ (one can come up with alternatives for each one of them), but a genuine effort has been made to make their _application_ consistent.
- The "faithful" characteristic turned out to be an unfortunate choice from a semantic perspective, because of the negative connotations of the term in other contexts. We stuck with that one mainly for purposes of "backwards compatibility".
- As already mentioned, the phrasing (first) and application (second) of the rules have developed throughout the first 6 to 12 months of this project's existence. My best guess is that things were stable by the time we had our first ~50 programming languages in.

I hope this fills in the background of the conversations we've had in the context of this PR, at least to some extent.

rbergen:
@ch2laughlin Do you still intend to follow-up our earlier discussions with the changes you indicated you would make?

rbergen (changes requested):
Requested changes have already been discussed.

ch2laughlin:
> @ch2laughlin Do you still intend to follow-up our earlier discussions with the changes you indicated you would make?

Yes, my plan is to make the requested markup changes, even though I disagree with them, so the goodness in the work can at least be submitted.  Tagging can be easily updated later...

Thanks for your descriptive comments during the process.

PrimeBatch/solution_1/README.md
ch2laughlin: The invocation of the worker bat in the PR explicitly passes the size of the array and other parameters.  These are passed as command line arguments to the worker.bat which then interprets them.  I beleive this is what your were asking for and can confirm it is done so in this PR - whereas in the previous work it was a bit half n half (some passed, some global)

main.bat LOC: 127 

		start "" /affinity !mask! cmd /c "^"!workBat!^" /id:!expectedPasses! /noValidate:%options.noValidate% /title:%WORKER_TITLE% /sieveSize:%options.sieveSize% /resultsDirectory:^^^"%resultsDirectory%^^^"" 

this becomes ...in essense a list of constructor parameters
     cmd /c worker.bat /id:8 /novalidate:false /title:something /sieveSize:100 ... 

Hopefully this at least addresses that concern.  I agree that in the past some of these were global variables - I changed that to make the current solution better and make it easier to add multiple new solutions (like the packed worker version).

PrimeBatch/solution_1/README.md
ch2laughlin: I have made the requested tagging changes for bits used in the latest PR update.

PrimeBatch/solution_1/main.bat
ch2laughlin: this is done / addressed.  resolving here as well.

PrimeBatch/solution_1/README.md
ch2laughlin: I have updated the faithful tagging per request in the lastest update.

PrimeBatch/solution_1/main.bat
rbergen: It hasn't. The main.bat script output (i.e. this very line of code) still claims bits=32.

As the actual value of bits varies per implementation, as per the comment I'm going to be making on README.md in this review round, it's probably just best to leave both unspecified, as long as you use one "echo" for both implementation:

```suggestion
	echo batch;%2;!secsPart!.!csPart!;%options.workers%;algorithm=base,faithful=yes
```

PrimeBatch/solution_1/README.md
```diff
@@ -1,13 +1,79 @@
-# Batch solution by Sxxov
+# Batch solution by Sxxov & Ch2Laughlin
 
-![Algorithm](https://img.shields.io/badge/Algorithm-base-green)![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)![Parallelism](https://img.shields.io/badge/Parallel-no-green)![Bit count](https://img.shields.io/badge/Bits-unknown-yellowgreen)
+![Algorithm](https://img.shields.io/badge/Algorithm-base-green)
+![Faithfulness](https://img.shields.io/badge/Faithful-no-yellowgreen)
```
rbergen: The faithfulness change is not necessary, as I'll explain below.


```suggestion
![Faithfulness](https://img.shields.io/badge/Faithful-yes-green)
```

PrimeBatch/solution_1/README.md
```diff
@@ -1,13 +1,79 @@
...
+- Uses an optimized Newton integer square root (no loops required).
+- Builds environment variables incrementally as composites are found.
+- Produces correct prime lists.
+
+#### Faithfulness
+This algorithm is not considered faithful by the drag-race maintainers at this time for 2 reasons.
+1. 'The size of the memory buffer must correspond to the size of the sieve' not the # of found candidates.  This algo is using env. variables like a
+sparse array that grows as the prime candidates are found.  That does not meet the defn of a faithful implementation.
+2. The worker.bat needs to take as input parameters the size of the sieve similar to how class constructors work - not depend on global like options that
+are used now.
```
rbergen: This is no longer true, as I tried to explain in earlier comments. I am happy to consider the solution **faithful**, as long as the **bit count** is correctly specified:

- for worker.bat: unknown
- for worker_pack.bat: 1

In more detail:

- objection 1 only applies to _one_ implementation, being the worker.bat one. It does not apply to the packed version, where a buffer is reserved that matches the sieve space size. Furthermore, I've never raised that point as a _faithfulness_ issue, only as an issue with the reported _bit count_.
- objection 2 no longer applies. You've convinced me that the invocation of the worker scripts is in fact sufficiently class-like.

So: 

- for **worker.bat**: the solution is **faithful**, bit count is **unknown**
- for **worker_pack.bat**: the solution is **faithful**, bit count is **1**

rbergen (commented):
I think we've developed a little bit of static on the line, leading to you "downgrading" the labelling of your solution beyond what I asked for, in the end. 

I've explained this in the individual comments below. To speed up the merging of your changes, I am willing to both make the change that is still necessary (in main.bat) and undo the changes that are no longer necessary (in README.md), **if you'll allow me to apply them to your drag-race branch (i.e. the head branch of this PR).**

Please let me know if you're okay with this.

PrimeBatch/solution_1/README.md
ch2laughlin: I am ok with this.  I've attempted to polish & incorporate your feedback into the readme.  Thanks for your patience.

PrimeBatch/solution_1/README.md
ch2laughlin: I'm ok with this.

ch2laughlin:
I've accepted your changes and then did some additional polishing on the readme markdown file.  Hopefully, we have things squared away this time.

rbergen (approved):
LGTM! Thanks for your patience and persistence. I believe we now have a significant improvement of the solution in place.

rbergen:
CI fails because of a failing build in a different solution, because your drag-race branch is out of sync with the main tree - these errors were fixed 2 months ago. Please sync your branch to fix this.

rbergen:
I've performed the sync myself and found there are new compile-time errors in Mojo/solution_1. I've reopened #1035 in relation to this.

As CI now passes, I will merge this. Thanks again for your patience.

PR #1031, opened 2025-12-25
Improve Dockerfile layer caching for PrimeMojo solution_1

relmer:
## Description
This PR improves rebuild times for the Mojo solution’s Docker image by restructuring layers so that the expensive dependency install is cacheable.

What changed

- Install `modular` (via pip) before copying the frequently-changing source.
- Copy only the needed Mojo source into the image (instead of copying the whole directory).
- Fix a minor comment typo.

Why

Previously, edits to the Mojo source invalidated the COPY layer and forced Docker to re-run `pip install modular` on every rebuild. With this ordering, most edits only rerun mojo build, which is much faster than re-installing the toolchain.

Verification

`docker build -t mojosieve .`
`docker run --rm -it mojosieve`

## Contributing requirements

### DIFF
--- a/PrimeMojo/solution_1/Dockerfile
+++ b/PrimeMojo/solution_1/Dockerfile
@@ -13,8 +13,6 @@ RUN apt-get update && apt-get install -y \
 
 WORKDIR /app
 
-COPY . .
-
 RUN python3 -m venv /opt/venv \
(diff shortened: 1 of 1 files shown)


rbergen (approved):
LGTM

PR #1032, opened 2025-12-25
Fix Odin solution_1 build to support current Odin (dev-2025-12)

### LINKED ISSUE #956: Odin/solution_1 doesn't compile with the current (dev-2024-02) version of Odin
While updating a number of Docker images published on Docker Hub in the context of this project, I found that [the Odin solution](https://github.com/PlummersSoftwareLLC/Primes/tree/drag-race/PrimeOdin/solution_1) does not compile on the current version of Odin. 

Without wanting to imply an obligation on the original contributors of the solution, I'm calling out to @omgitsmoe, @gingerBill and @Kelimion in case they would be willing and able to look into this.

relmer:
## Description

Hope it's ok I'm updating this code--seems like it may be abandoned given the issue is 2 years old.  Fixes Odin solution_1 to build and run on the current Odin dev snapshot and updates the Docker build to use a current Odin release tarball.

What changed

Updated the code for compatibility with the current language/toolchain expectations (build flags, import constraints, type/size fixes) per Kelimion's comments (not tagging him here since he seemed surprised to be referenced in the issue).
- Replaced the previous thread synchronization approach with `sync.Wait_Group`.
- Fixed a threaded-mode crash by avoiding passing a procedure value through thread poly-data; split workers by sieve type instead.
- Switched sieve allocations to `context.allocator` to avoid relying on thread-unsafe temporary allocation patterns.
- Implemented Linux core-count via glibc `get_nprocs()` (and kept the Windows implementation separate).
- Updated the Docker setup to download a modern Odin release and build with current odin build flags.

How to verify

`docker build -t odin-primes .`
`docker run --rm odin-primes`

Fixes #956.

## Contributing requirements

### DIFF
--- a/PrimeOdin/solution_1/Dockerfile
+++ b/PrimeOdin/solution_1/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:24.04 AS odin
+FROM ubuntu:24.04 AS build
 
 ARG ODIN_VERSION=dev-2025-12
 ARG ODIN_ARCHIVE=odin-linux-amd64-${ODIN_VERSION}.tar.gz
@@ -6,20 +6,14 @@ ARG ODIN_URL=https://github.com/odin-lang/Odin/releases/download/${ODIN_VERSION}
 
 RUN apt-get update \
 	&& apt-get install -y --no-install-recommends ca-certificates curl clang lld \
-	&& rm -rf /var/lib/apt/lists/*
-
-RUN mkdir -p /opt/odin \
+	&& rm -rf /var/lib/apt/lists/* \
+	&& mkdir -p /opt/odin \
 	&& curl -fsSL "$ODIN_URL" -o /tmp/odin.tar.gz \
 	&& tar -xzf /tmp/odin.tar.gz -C /opt/odin --strip-components=1 \
 	&& rm /tmp/odin.tar.gz
 
 ENV ODIN_ROOT=/opt/odin
 ENV PATH="/opt/odin:${PATH}"
-
-RUN odin version
-
-
-FROM odin AS build
 
 WORKDIR /opt/app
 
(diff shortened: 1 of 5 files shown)


PrimeOdin/solution_1/Dockerfile
```diff
@@ -1,14 +1,36 @@
...
+	&& tar -xzf /tmp/odin.tar.gz -C /opt/odin --strip-components=1 \
+	&& rm /tmp/odin.tar.gz
+
+ENV ODIN_ROOT=/opt/odin
+ENV PATH="/opt/odin:${PATH}"
+
+RUN odin version
+
+
+FROM odin AS build
```
rbergen: I think this part of the Dockerfile can be simplified as follows:

- the separate creation of the `odin` image can be skipped; the first image can be marked as the `build` image, as in the original Dockerfile.
- The two `RUN` statements between the `ARG` and `ENV` blocks can be combined. In fact, the current setup with the two subsequent `RUN` statements will trigger Hadolint when CI is executed.
- The `RUN odin version` statement can be removed.

rbergen (changes requested):
@relmer Thank you for submitting this.

As for updating this solution while it was submitted by others: I think #956 has been open long enough for the original authors to provide their updates.

As for your changes, I think:

- the README needs an update to the run command included there to align it with the changes in the arguments passed to `odin build` in the Dockerfile.
- the Dockerfile can be simplified and requires a change to pass Hadolint validation. Please refer to the specific comment concerning the Dockerfile for more details.

PrimeOdin/solution_1/Dockerfile
relmer: Thanks, that is better.

rbergen (approved):
@relmer Thank you for the follow-up. LGTM now!

PR #1039, opened 2026-02-21
Add Prime-Sniper V5 (Faithful JS ArrayBuffer implementation by Helron…

Helron1977:
…1977)

## Description
Adding `Prime-Sniper V5`, a high-throughput faithful implementation in JavaScript (NodeJS).
To maximize V8's JIT compilation and CPU branch prediction while remaining strictly algorithmic ("Faithful" class):
- Uses an external `ArrayBuffer` linked to an `Uint32Array` to bypass NodeJS Garbage Collection latency.
- Aggressive algorithmic Loop Unrolling (unrolled by 8) for the impact loop.
- Exclusive use of bitwise operations (`>>>` and `&`) to eliminate Division and Moduluar instructions completely.
## Contributing requirements

### DIFF
--- a/PrimeJavaScript/solution_1/PrimeJavaScript_rogier_optimized.js
+++ b/PrimeJavaScript/solution_1/PrimeJavaScript_rogier_optimized.js
@@ -1,0 +1,175 @@
+"use strict";
+const { performance } = require('perf_hooks');
+
+const NOW_UNITS_PER_SECOND = 1000;
+const WORD_SIZE = 32;
+
+let config = { sieveSize: 1000000, timeLimitSeconds: 5, runtime: 'node' };
+
+const sieveBits = config.sieveSize >>> 1;
+const wordArraySize = 1 + (sieveBits >>> 5);
+const GLOBAL_POOL = new Int32Array(wordArraySize);
+
+function runSieveInline() {
+    const arr = GLOBAL_POOL;
+    arr.fill(0);
+
+    let factor = 1;
+    let blocksize_bits = 1;
+    let range = 3;
+    const q = Math.ceil(Math.sqrt(sieveBits));
+
+    while (factor <= q) {
+        const step = factor * 2 + 1;
+        const start = factor * step + factor;
+
+        if (range < sieveBits) {
+            let dest_stop = blocksize_bits * step * 2;
+            if (dest_stop > sieveBits) dest_stop = sieveBits;
+
+            const source_start = blocksize_bits;
+            const dest_start = blocksize_bits * 2;
+            const size = dest_start - source_start;
+            let copy_start = dest_start;
+
+            if (size < 64) {
+                let copy_max = 64 + source_start;
+                for (let index = 0; index < size; index++) {
+                    if (arr[(source_start + index) >>> 5] & (1 << ((source_start + index) & 31))) {
+                        let copy_index = dest_start + index;
+                        while (copy_index < copy_max) {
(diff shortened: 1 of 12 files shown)


rbergen (commented):
@Helron1977 Thank you for submitting this.

I have the following comments:

- As per the [contributing guidelines](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md), the directory for your solution should be solution_2:
  _In the drag-race branch, see what the highest numbered solution is for the language you chose, and place your solution in the following folder: Prime<Language>/solution\_<highest+1>/ where "highest" is the number of the highest numbered solution you found._
- The solution is not faithful, for a number of reasons:
  - The sieve reuses a global memory buffer, instead of a memory buffer that is embedded in the class. The key part of the relevant faithfulness rule is in bold:
    _It uses a class to encapsulate the sieve, or a (closest) equivalent feature in your language if a class construct is not available. **This class must contain the full state of the sieve.** Each iteration should re-create a new instance of this class from scratch._
  - The sieve buffer is not dynamically allocated based on the sieve size, but a fixed-size buffer that is statically allocated:
    _The sieve size and corresponding prime candidate memory buffer (or language equivalent) are set/allocated dynamically at runtime. The size of the memory buffer must correspond to the size of the sieve._ 

If you still want to submit your solution then you will have to rename the solution folder, and either:

- Change the faithfulness indication to "no"
- Address the deviations from a faithful implementation that I mentioned

Helron1977:
Thank you for the detailed review and clarification on the rules, it's highly appreciated!

I will comply with the guidelines and go with the Faithful route. I will rename the folder to solution_2 and move the ArrayBuffer allocation inside the class constructor so it is dynamically re-created at each pass.

However, as a side note from an engineering perspective, I find the strict constraint of constantly destroying and recreating memory buffers a bit limiting for evaluating true algorithmic performance, especially in garbage-collected environments like NodeJS. Also, while fixing the benchmark size to 1,000,000 allows for fast drag-racing, it somewhat masks a major flaw in many current top-tier "faithful" implementations: they simply crash with Out-Of-Memory exceptions if asked to sieve up to 5 billion, whereas the segmented/memory-pooled approach of Prime-Sniper was specifically designed to handle infinite ranges without crashing the RAM.

But rules are rules for a fair drag-race! I will push the compliance updates to this PR shortly. Thanks again for the great repository and the fun challenge!

rbergen:
I know you've already indicated you will comply with the rules, but just for the record: I actually believe the buffer (re)allocation requirement is a fair part of the rules. It's not realistic to have an algorithm run without doing the setup required to be able to do so.

I see we now have 4 PRs open from your hand, which seem to all contain the same files, at least in name. Please:

- Create **one** PR with **only** the _changes_ you propose to the existing JavaScript solution (solution_1), if you still aim to make any. Do not add additional implementations to this PR, or the solution_1 directory.
- Create **one other** PR with the implementations you'd like to add. Put all of those in the **solution_2** directory, i.e. make it a PR that aims to add one multi-implementation solution. I'm specifically mentioning the solution_2 directory because the directory your PRs are currently adding  is still called solution_helron (you can check the "Files changed" tab on this PR to verify). Again, make sure that no new implementation files are added to the solution_1 directory, and that no changes are made to existing solution_1 files, in this PR.

I will review once we have (at most) two PRs structured this way.

Helron1977:
After reading carefully the rules . this one has to be closed
Pb allocation dynamique
Pb global var
Pb object encapsulé. class propre

PR #1045, opened 2026-02-25
solution 2

Helron1977:
Algorithm
Faithfulness
Parallelism
Bit count

## Description

High-performance "God Mode" implementation using a static mask for the first 11 primes and a highly optimized bit-sieve for the rest.I tryed to follow each rule of the challenge. I tryed to make it as fast as possible. I hope it will be accepted, it was pretty hard for me to learn and follow the rules and especially   the publish rules. I withdrawed the previous fauthfull submission contest because it was not able to overtake the records.

## Run instructions

```bash
node PrimeJavaScript_sniper.js
```

## Output

```log
helron-sniper;23536;5.0001;1;algorithm=base,faithful=no,bits=1
```

### DIFF
--- a/PrimeJavaScript/solution_2/Dockerfile
+++ b/PrimeJavaScript/solution_2/Dockerfile
@@ -1,0 +1,5 @@
+FROM node:18-alpine
+WORKDIR /app
+COPY . .
+RUN chmod +x run.sh
+ENTRYPOINT ["./run.sh"]
--- a/PrimeJavaScript/solution_2/PrimeJavaScript_sniper.js
+++ b/PrimeJavaScript/solution_2/PrimeJavaScript_sniper.js
@@ -1,0 +1,125 @@
+"use strict";
+const { performance } = require('perf_hooks');
+const NOW_UNITS_PER_SECOND = 1000;
+let config = { sieveSize: 1000000, timeLimitSeconds: 5 };
+
+/**
+ * GOD MODE V8 - STATIC MASK (17k+ Passes)
+ * - Crible binaire (1 bit par nombre impair).
+ * - Pré-calcul de TOUS les multiples des primes jusqu'à 37 dans un masque statique.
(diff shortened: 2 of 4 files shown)


Helron1977:
Y a t il à nouveau un soucis de nomenclature ou de repertoire ?

rbergen:
C'est dommage, mais je ne parle pas Français.

With which I mean to say: please post your comments in English if you'd like a response.
As for the review, I will have time for it this weekend.

rbergen:
Actually, I had some time now.

Your implementation uses prior knowledge of a set of numbers being primes, which means your approach does not cover the entire sieve space using the Sieve of Eratosthenes algorithm. That in turn means you cannot mark the solution as following the base algorithm.

The applicable quote from the contributing guidelines, under the [Base algorithm](https://github.com/PlummersSoftwareLLC/Primes/blob/drag-race/CONTRIBUTING.md#base-algorithm) header:

> * When seeking factors (the first operation), the algorithm sequentially **checks all odd numbers, starting at 3**.

Your algorithm is wheel-like, but also not _exactly_ the traditional wheel factorisation algorithm either. Because of this, the most appropriate way to mark the algorithm of your solution would be as "other". Please apply this change to both the solution benchmark format output and the README.

Helron1977:
I have updated the algorithm classification to 'other' as requested. Thanks!

rbergen (approved):
The solution LGTM at a code level, in the sense that I'll allow it as a single demonstration of a somewhat outlandish implementation that skirts the rules. 

Part of the justification for this is that the solution does not implement the base algorithm and is unfaithful, which means it's disqualified from the actual leaderboard and doesn't impact that.

The previous is also a statement I will do this once for this type of implementation; that being where an apparently random number of primes at the beginning of the sieve space have been precomputed.

Considering the nature of the solution, I will test it locally before merging. As I will be travelling for most of this week, it may take another few days before I get round to that.

PR #1046, opened 2026-03-10
Add PrimePython solution_4: numpy + numba JIT sieve

TylerDOC1776:
Uses numpy bool_ array with Numba @njit to compile the sieve loop to native machine code. Warm-up call absorbs JIT compilation cost before the timed benchmark starts.

## Description
Replace this text with your description.

## Contributing requirements


rbergen:
@TylerDOC1776 Thank you for submitting this. Considering this is effectively a relatively small extension to solution_3 (I mean that in the sense it uses the exact same logic and numpy, and has the exact same solution characteristics), could you please structure it as a second implementation of solution_3 instead? It will save everyone who builds and runs the benchmark suite another Docker image and container.

Of course, you can still credit yourself in the implementation output and the README, in a similar way you have now.

TylerDOC1776:
Updated as requested — restructured solution_4 as a second implementation within solution_3 to avoid an extra Docker image. The Numba JIT sieve (PrimePY_numba.py) now shares the same container as the numpy implementation, with a run.sh entrypoint that runs both in sequence. solution_4 has been removed.

rbergen (approved):
LGTM!

PR #1047, opened 2026-03-10
Submit/helron v3 final

Helron1977:
Description
This implementation is a Faithful Sieve of Eratosthenes using an optimized bit-storage and a wheel factorization technique.

Faithful Approach: The algorithm follows the classic Sieve of Eratosthenes logic. It uses 1-bit per number (ignoring evens) and calculates all prime multiples within the sieve range.

Optimization (Wheel Factorization): I implemented a wheel for the first primes (3, 5, 7, 11, 13). Instead of inefficiently re-calculating these high-frequency patterns in the main loop, I use a periodic bit-pattern replication.

Zero Pre-computation: No results are pre-stored. The "Super-Wheel" pattern (15,015 words) is a purely mathematical representation of the 3-13 prime cycles. It only clears multiples that the sieve would have cleared anyway, but does so with optimized memory alignment.

Performance: Each pass includes full object instantiation and memory allocation, "Drag Race" conform.

Passes: ~12,000 - 13,000 (Hardware dependent)

Status: faithful 

### DIFF
--- a/PrimeJavaScript/solution_3/Dockerfile
+++ b/PrimeJavaScript/solution_3/Dockerfile
@@ -1,0 +1,5 @@
+FROM node:18-alpine
+WORKDIR /app
+COPY . .
+RUN chmod +x run.sh
+ENTRYPOINT ["./run.sh"]
--- a/PrimeJavaScript/solution_3/PrimeJavaScript_extreme_sieve.js
+++ b/PrimeJavaScript/solution_3/PrimeJavaScript_extreme_sieve.js
@@ -1,0 +1,143 @@
+"use strict";
+
+const { performance } = require('perf_hooks');
+
+const SIEVE_SIZE = 1000000;
+const LIMIT_BITS = SIEVE_SIZE >>> 1;
+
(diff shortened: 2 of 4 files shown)


rbergen:
@Helron1977 Please roll this up into solution_2 as a second implementation. You can do this after I merge #1045, if you don't want to interfere with that process.

rbergen:
I've merged #1045, so you can proceed with making this a second implementation in solution_2.

Helron1977:
i was hoping this one would be base faithfull because i didn't use pre-knowned primes, and the solution 1 was a mem copy.

rbergen:
It doesn't matter if it is or not for my request. It's possible to have multiple implementations with mixed characteristics in one solution, as other solutions show. 

The reason I'm requesting you to combine your various implementations in one solution is that every solution generates a Docker image and container when CI or a benchmark is run, so it's more economic from a resource perspective to combine multiple implementations in one solution.

You can add the different characteristics for the various implementations in the badges at the top of the README, and each implementation can output its own correct characteristics in the benchmark format output. In the benchmark results lists, each implementation will still show up independently.

PR #1048, opened 2026-03-14
Submit/helron consolidated

Helron1977:
# JavaScript solution 2 by Helron1977

This solution contains two different implementations of the Prime Sieve merging my solutions under solution_2 as requested.
Ty

### DIFF
--- a/PrimeJavaScript/solution_2/PrimeJavaScript_extreme_sieve.js
+++ b/PrimeJavaScript/solution_2/PrimeJavaScript_extreme_sieve.js
@@ -1,0 +1,140 @@
+"use strict";
+
+const { performance } = require('perf_hooks');
+
+const SIEVE_SIZE = 1000000;
+const LIMIT_BITS = SIEVE_SIZE >>> 1;
+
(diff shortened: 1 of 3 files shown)


PrimeJavaScript/solution_2/PrimeJavaScript_extreme_sieve.js
```diff
@@ -0,0 +1,140 @@
...
+        passes++;
+    }
+
+    const duration = (performance.now() - timeStart) / 1000;
+    console.log(`helron-extreme;${passes};${duration};1;algorithm=other,faithful=yes,bits=1`);
+
+    const verify = new PrimeSieve(sieveSize);
+    verify.runSieve();
+    if (verify.getPrimeCount() !== 78498) {
+        console.error(`Validation failed: expected 78498, got ${verify.getPrimeCount()}`);
```
rbergen: Please restructure this to only output a benchmark result line if the sieve result count is correct. 

This is not technically an implementation submission requirement, but you did use that approach in the other implementation and it is much preferred.

rbergen (approved):
Thank you, LGTM.

PR #1050, opened 2026-03-22
Primes in ALGOL 60 and directory rename

rzuckerm:
## Description
This PR contains the following:
- Rename existing `PrimesAlgol68g` directory to `PrimesALGOL` (BTW,  [ALGOL](https://en.wikipedia.org/wiki/ALGOL) is the correct spelling)
- Change the output for existing ALGOL 68 samples to indicate `algol68`
- Add ALGOL 60 implementation using the [GNU MARST 2.8 ALGOL to C translater](https://ftp.gnu.org/gnu/marst/marst-2.8.tar.gz)

Here is the [rzuckerm/algol68 docker image](https://github.com/rzuckerm/algol60-docker-image) repo. The [image](https://hub.docker.com/r/rzuckerm/algol60/tags) supports amd64 and arm64

## Contributing requirements


rbergen (approved):
LGTM!

I'm noting that the PrimeAlgol68g directory rename will require an update to the PrimeView language map. I am currently in the process of upgrading that project to .NET 10, and will either make the lang map change when that upgrade completes if it does soon enough (read: today). If it doesn't then I'll implement the lang map change on the current .NET 8 codebase. 

Until the lang map change is in place, solution links in PrimeView benchmark reports will not work for the new ALGOL language directory.

PR #1053, opened 2026-03-28
Use dynamic memory allocation for ALGOL 60 implementation

rzuckerm:
## Description
It turns out that ALGOL 60 **does** support dynamic memory allocation. This can be done inside a `begin/end` construct. I changed the code to do this inside a `for` loop that keeps track of the number of passes. Also, it turns out that arrays should be passed by name, not value. Removing `value sieve` from each procedure gave a significant speed improvement (2 to 5x). This allows a sieve size up to 1 billion to be supported.

## Contributing requirements


rzuckerm:
Sorry for all the churn. I'm finished now

rbergen (approved):
> Sorry for all the churn. I'm finished now

No worries, I've seen - and done - much worse.

Nice update to your solution, LGTM, thanks for submitting it!

PR #1054, opened 2026-04-07
C64 solution

RazCamelot:
## Description
I'm hereby submitting my C64 solution to the Prime Sieve challenge. It's a full solution (bitfield of 500,000 bits / $f424 bytes, sqroot calculation) and it comes with build-in validation of the primes + build in timer (auto-calibrated for both NTSC and PAL), as well as a menu for starting options.

Target system is (evidently) C64 and target CPU is the MOS 6510 (all variants) - the non C64 specific code itself is compatible with all 6502 variants, except 65c02. I have added two lines with benchmark results: for 1) NTSC and 2) PAL.

I have also included a very comprehensive README.md that explains my implementation in details.

Regarding Docker: As this solution (incl validation + timing) runs entirely on the C64, I have opted for the setup also used for solution_3 (Ben Eater's breadboard computer): Include instructions on assembly (also in my case a single line) and execution, rather that setting up a Dockerfile. Let me know, if this is an acceptable approach - if not, I'll make adjustments.

-Rasmus Wernersson, rasmus@wernersson.dk 

## Contributing requirements


rbergen (changes requested):
Thank you for submitting this! My hat's off to you for this very sophisticated C64 solution, and that is coming from the guy who wrote all other three 6502 assembly solutions that we already have. 

I do have a few requests:

- Please move the images to a location outside of the repo, and remove the .prg file altogether. As a rule, we don't allow any binaries in this repo, and program binaries need to be compiled (or in this case, assembled) from the submitted source code in all cases. 
- In part to compensate for the .prg file going, please provide step-by-step instructions for turning the .asm file into a .prg using the assember you used.
- You refer to solution_3 as a motivation for not including a Dockerfile, but I'm guessing you did so knowing that solution_1 and solution_2 do include ones that also use VICE. In fact, the VICE base image that Dockerfile relies on was created specifically for this project (by yours truly, as it happens). I would therefore like to urge you to indeed provide a Dockerfile - you can use one of the two mentioned as a source of inspiration. 
- The build-no file is only necessariy if a Dockerfile _is_ present. So if you do provide a Dockerfile for this solution it should indeed be kept, but can and should be removed otherwise.

RazCamelot:
Thanks for the kind reply. I'll of cause be happy to implement the suggested changes.
@rbergen: I'll send an email with an additional question about the Docker setup (just to make sure I get it as right as possible).
Cheers, Rasmus

rbergen:
@RazCamelot I opened RazCamelot#2 to make the solution work in Docker in the head branch of this PR. I'm happy to continue the path towards merging this PR after the other one has been merged.

rbergen (commented):
Now that we're this close to merging this, I've taken a closer look at the exact algorithm implementation. I believe that the lazy mode initialization calculates a bitmap for primes 3, 5 and 7, and projects that bitmap on the entire sieve space. This is also the way the sieve is run when the solution is operated in batch/benchmark mode.

As per the base algorithm rules and the way they have been applied throughout this project, it _is_ allowed to compute a multiples marking bitmask for a particular identified prime number, _but only for one prime number at a time_. (In fact, it so happens that this has been a topic of conversation in the context of the other PR that is open when I write this, which also happens to be an assembly solution. Coincidence can work in funny ways.)  
Creating such a multiple masking bitmap for a collection of primes basically makes it a wheel algorithnm, as also described in the contributing guidelines.

I think the most appropriate thing to do here is to include both the algorithm=base and algorithm=wheel tags to the solution, make the full clear algorithm the one to use in batch mode, and link the two algorithmic approaches to the respective algorithm labels in the README.  
IF you choose to stick to the wheel algorithm in benchmark mode, then I think the output should reflect that the wheel algorithm was used to compose it. Knowing you only have about two handfuls of bytes left available, it is okay with me to tag that output addition on to the output in the run.sh script, or one of the supporting Python scripts.

In case you interpret this as a take-down of your approach: it isn't. I actually **love** the fact there are two different implementations in this solution, whithin the very limited memory space this platform provides. I'm just duty-bound to ensure that the solution's bookkeeping is in order, in the context of the many solutions we have and the - sometimes very animated - conversations that have taken place about these very topics over the past 5 years.

RazCamelot:
Thanks for the swift reply - I’m happy to have the algorithm variant run in batchmode be the “vanilla” one. This will be super easy to implement. 
I did originally make the option for running either version to address exactly the situation, that the one with the lazy bitmap init might be taking it too far =)

I’ll do the updates and submit the changes.

Cheers
-Raz

RazCamelot:
Ok - think I got it now =)
I have just pushed a few updates.

Cheers
-Raz

------------------------
Rasmus Wernersson
Maglebjergvej 7
3200 Helsinge
------------------------

Prime6502Assembly/solution_4/README.md
```diff
@@ -0,0 +1,348 @@
...
+
+Finally, there is a Dockerfile for wrapping up the entire process and automating the build, running + checking of results (thanks to Rutger van Bergen for help with this).
+
+## Results
+
+```
+RazCamelot-c64PAL;1;21.2;1;algorithm=base,faithful=no,bits=1
+RazCamelot-c64NTSC;1;20.6;1;algorithm=base,faithful=no,bits=1
+RazCamelot-c64PAL;1;14.2;1;algorithm=wheel,faithful=no,bits=1
+RazCamelot-c64NTSC;1;13.8;1;algorithm=wheel,faithful=no,bits=1
```
rbergen: I think I'm happy to let this go, but there isn't _actually_ a way to make the code (.asm, .py and .sh combined) output this, right?  
I'm asking because for other solutions this is the actual output format they generate.

RazCamelot:
Hmm - I think I could make it work with some extra wrapping: 
1) NTSC vs. PAL is easy enough since that is a parameter to Vice. 
2) Running both algos in batchmode would require a bit more assembly time magic + generating two distinct binaries.
3) The output from the current evaluation script can quite easily be formatted to the specifications.

If I dig into this, would you prefer build.sh to generate two binaries, or handle the differential builds in the wrapper for running all four variations?

Cheers
-Raz

rbergen:
I appreciate your willingness to put in the extra effort for this. 

As to your question: I don't _really_ have a preference. If I'm honest I mainly care about the Dockerfile scenario working, exactly how you get there is less relevant to me personally. I'm duty-bound to add to that that at a conceptual level "the Docker way" would be doing all the building in the build phase, i.e. build.sh in the context of this solution.

My earlier suggestions regarding restructuring the shell and Python scripts related to code and config duplication, which made the review process more difficult. In this case, I am happy to roll with whatever decision you make.

rbergen:
@RazCamelot Just checking in to ask if you still intend to make the changes you mentioned earlier?

RazCamelot:
Yes, indeed - life got in the way this week. I expect to be able to wrap it up the coming week.

Cheers
-Raz

rbergen:
That's what I thought, but I thought I'd check. 

No pressure, I just really want to get this one in. :)

RazCamelot:
Hi Rutger

Finally found a bit of quiet time to look at it. 
I still think it should be possible to do, but it will require more changes to the build and run scripts than I anticipated.

Build:
I would like to build different versions of the binary - a clunky solution would be to build and rename, but the correct way to do it would be to encode it directly in the SEGMENT output definition (together with additional assembly time flags). I’ll need to do some research here - the issue is that a name is always hardcoded in this section:

.file [name="%o.prg", segments="Upstart,Code,Data,StatusLine,CodeInit,DataReloc,DataShowOnce"]

Running / outputting should be a bit more straightforward - I plan to add an extra option to analyze_memdump.py with a formatting string.
I think I need to get the workflow to work outside the Docker setup while testing, and then I can eventually backport it to the repo.

Q: Would it be possible to accept the PR as is (as you eluded to earlier)? Then I can work on a more permanent solution in the background without holding up things.

Cheers
-Raz

------------------------
Rasmus Wernersson
Maglebjergvej 7
3200 Helsinge
------------------------

rbergen:
You definitely can and I would still appreciate your efforts greatly, even if they are implemented later. My only request at this time is to then change your README to display what the actual output of the Dockerfile run is, instead of what it will/may be further down the line.

RazCamelot:
Great - thanks for the swift response. I have updated the README accordingly.

Thanks - and have a great weekend =)
-Raz

------------------------
Rasmus Wernersson
Maglebjergvej 7
3200 Helsinge
------------------------

rbergen (approved):
LGTM! Thank you very much for submitting this!

RazCamelot:
Great =) Thanks for all the help along the way.

Cheers
-Raz

------------------------
Rasmus Wernersson
Maglebjergvej 7
3200 Helsinge
------------------------

PR #1056, opened 2026-04-16
PrimeAssembly: add cwager_x64ff_mt multithreaded amd64 solution

Kexisx:
Adds `cwager_x64ff_mt`, a multithreaded amd64 NASM implementation.

Why a new solution:
- existing amd64 assembly solutions in `PrimeAssembly` are single-threaded
- this implementation materially changes the parallelism characteristic by using more than one thread

Characteristics:
- algorithm=base
- faithful=yes (worker state carries runtime sieve metadata and each pass recreates a fresh sieve buffer)
- bits=1
- multithreaded

Implementation notes:
- each worker owns runtime sieve metadata and sieve state in a worker object
- each timed pass recreates a fresh sieve buffer from scratch
- the sieve work itself remains the base odd-factor marking algorithm

Validation:
- verified local build/run
- verified Docker build/run


rbergen (approved):
LGTM. Thanks for your submission!

Kexisx:
Thank you for taking the time to review and approve this.

Just out of curiosity, will this now be submitted to the runs that show on the Reports page? If so, how long does that typically take to appear?

rbergen:
If the Docker image indeed works, your solution's results will indeed be included. This should start happing the next time the various benchmark runners execute a daily benchmark run. By looking at the timestamps of earlier runs, you can tell when that is.

One of the runners is an ARM device (a Raspberry Pi 4), on which your solution will of course not be executed.

PR #1058, opened 2026-04-26
Add x86-64 NASM multithreaded solutions with extreme optimized variant

Kexisx:
This PR adds a new multithreaded x86-64 NASM variant:

* `cwager_x64ff_mt_extreme.asm`

## Summary

This builds on the existing `cwager_x64ff_mt.asm` implementation by introducing a more aggressively optimised marking strategy while remaining within the base Sieve of Eratosthenes rules.

Both implementations:

* use a packed 1-bit representation for odd candidates
* discover all sieving factors dynamically from the runtime sieve state
* allocate and initialise sieve memory per pass
* validate against the expected 78,498 primes at 1,000,000
* output results in the standard drag-race format

## Extreme Variant

The new `cwager_x64ff_mt_extreme.asm` variant introduces:

* specialised dense and sparse projection paths for runtime-discovered factors
* optimised handling for smaller factors
* more efficient marking logic to reduce instruction count in hot paths

All optimisations are applied only after a factor is identified from the sieve state. No precomputed primes, wheel factorisation, or hardcoded prime tables are used.

## Classification

* `algorithm=base`
* `faithful=yes`
* `bits=1`

The implementation remains faithful as:

* factors are discovered by scanning the sieve state at runtime
* each pass recreates the sieve from scratch
* sieve memory is dynamically allocated per pass

## Build and Run

```bash
./build.sh
./run.sh
```

Run individually:

```bash
./cwager_x64ff_mt
./cwager_x64ff_mt_extreme
```

## Docker

Docker configuration has been updated to build and run both variants for comparison:

```bash
docker build -t cwager-x64ff-mt .
docker run --rm cwager-x64ff-mt
```

## Notes

Targets x86-64 Linux and links against pthread and libc.

The extreme variant focuses on improving projection efficiency while preserving base-algorithm semantics.


PrimeAssembly/solution_4/cwager_x64ff_mt_extreme.asm
```diff
@@ -0,0 +1,19043 @@
...
+    dq run_sieve_dense_121
+    dq run_sieve_dense_123
+    dq run_sieve_dense_125
+    dq run_sieve_dense_127
+    dq run_sieve_dense_129
+
+dense_p3_masks dq 0xdb6db6db6db6db6d, 0x6db6db6db6db6db6, 0xb6db6db6db6db6db
+dense_p5_masks dq 0xbdef7bdef7bdef7b, 0x7bdef7bdef7bdef7, 0xf7bdef7bdef7bdef, 0xef7bdef7bdef7bde, 0xdef7bdef7bdef7bd
+dense_p7_masks dq 0xf7efdfbf7efdfbf7, 0xfbf7efdfbf7efdfb, 0xfdfbf7efdfbf7efd, 0x7efdfbf7efdfbf7e, 0xbf7efdfbf7efdfbf, 0xdfbf7efdfbf7efdf, 0xefdfbf7efdfbf7ef
+dense_p11_masks dq 0xeffdffbff7feffdf, 0xbff7feffdffbff7f, 0xffdffbff7feffdff, 0xff7feffdffbff7fe, 0xfdffbff7feffdffb, 0xf7feffdffbff7fef, 0xdffbff7feffdffbf, 0x7feffdffbff7feff, 0xffbff7feffdffbff, 0xfeffdffbff7feffd, 0xfbff7feffdffbff7
```
rbergen: The way these masks are organized - and quite possibly used, but I'll get to that if I'm right when I reach the relevant code - does imply prior knowledge what numbers are primes, from an overall algorithm perspective. I would not object to a variation where masks are set up for all odd numbers to a point that makes sense from the perspective of the optimization approach, and the code picks the mask for whatever prime number it actually identified - and thus ends up not using the other masks.

PrimeAssembly/solution_4/cwager_x64ff_mt_extreme.asm
```diff
@@ -0,0 +1,19043 @@
...
+align 16
+run_sieve_outer:
+    cmp r11d, r12d
+    jae run_sieve_done
+
+    test r13, r10
+    jz run_sieve_next
+
+    ; This path is entered only after p has been discovered from the sieve
+    ; bitset at runtime. It does not encode prior-prime knowledge.
```
rbergen: But it does (seem to do so) in the sense there are code paths that only exist for numbers that "happen to be" primes. As implied before, I'd like either:

- an algorithmic explanation that makes this specific selection of code paths make sense without their primeness being a factor.
- an implementation that includes code paths, that are considered in terms of cmp tests, for odd numbers that are not primes but are within the range of the algorithm's general approach. Again, this is necesssary for a fair claim that we don't use prior knowledge of numbers being primes even if we "secretly" actually do know those code paths won't ever be used.

rbergen (commented):
Thank you for submitting this addition to your solution; it's been a while since anyone aimed to add a performance-focused implementation. I do have a couple of comments about apparent reliance on prior knowledge regarding numbers being primes in the second loop.

You can respond either with an explanation why my interpretation is incorrect, or an implementation that doesn't lend itself for that interpretation.

Kexisx:
Thanks, that’s fair.

I’ve pushed a revised version that removes the prime-shaped paths entirely. The old p3/p5/p7/p11 mask tables and direct handlers are gone.

The dense path now uses a contiguous odd-skip dispatch table covering every odd skip from 3 through 129, including composite values. The marking path is only selected after the candidate factor has been discovered from the sieve state at runtime, so the dispatch decision is now based on skip range rather than prior knowledge of which values are prime.

I’ve also cleaned up the sparse path naming so those labels describe modulo-16 residue classes rather than looking like selected skip values.

rbergen (commented):
After reviewing your code again, I had a nagging discomfort with the fully precomputed multiple masking bitmaps because of the "multiples are marked individually" stipulation in the definition of the base algorithm.

I've done some proper digging into the existing solution set, and based on previously submitted `algorithm=base` solutions, I'm going to have to decide as follows: it _is_ allowed to use computed bitmasks for multiples marking, as long as:

- the individual bits in the masks are identifiable in the code, either as literals or through bit shifts, and
- the actual composition of the masks is done in code.

The `algorithm=base` solution that I can find that is closest to what you are submitting is PrimeV/solution_2 (more specifically the functions in the dense_bitset array), and I remember having a discussion with the author of that solution about this very point.

What's relevant to note specifically is that I wrote "in code" under the second bullet. The patterns that are composed in the dense_bitset functions in the V solution _may_ be optimized into flat constants by the V compiler - that is allowed, as long as that optimization is an embedded capability of the compiler in question (i.e. not applied through separate tooling).

This means that the code of your second implementation would have to be restructured to match the bullets mentioned above to be able to be marked as `algorithm=base`.

Kexisx:
Thanks, that makes sense, and I agree with the distinction you are drawing here.

I have pushed another revision to address this. The extreme variant no longer uses the fully precomputed dense multiple bitmap constants.

The dense helpers are now generated from NASM macros that emit individual `btr` operations for the relevant bit positions. So the submitted source now shows the multiple-marking pattern being composed in code from identifiable individual bits, rather than storing the finished pattern as opaque `dq` bitmap data.

The dense dispatch is still selected only after the candidate factor has been discovered from the runtime sieve state, and the table still covers the full odd skip range from 3 through 129, including composites.

Current Docker output after the change is:

cwager_x64ff_mt_extreme;326674;5.000;16;algorithm=base,faithful=yes,bits=1
cwager_x64ff_mt;132908;5.000;16;algorithm=base,faithful=yes,bits=1

Hopefully this now matches the rule you described: individual bits identifiable in the code, and the actual composition/marking expressed in the implementation rather than through precomputed bitmap constants.

PrimeAssembly/solution_4/Dockerfile
```diff
@@ -1,11 +1,33 @@
-FROM alpine:3.13
+# ---------- 1. Build stage ----------
+FROM debian:bookworm AS build
```
rbergen, line 2: I was wondering what the reason is for switching from Alpine to Debian? (For clarity: this is not a review comment as such, just a question out of interest.)

PrimeAssembly/solution_4/README.md
```diff
@@ -1,62 +1,134 @@
 # cwager_x64ff_mt
 
+# cwager x86-64 NASM solutions
```
rbergen: You have two H1 headers now - I'm guessing the intention was to replace the upper with the lower?

PrimeAssembly/solution_4/README.md
```diff
@@ -1,62 +1,134 @@
...
+The extreme variant no longer uses fully precomputed flat dense bitmap constants.
 
-Example output from this machine:
+Instead, the dense helper code is generated from NASM macros that emit identifiable individual `btr` operations for the relevant bit positions.
 
-```text
-cwager_x64ff_mt;128068;5.001;16;algorithm=base,faithful=yes,bits=1
-cwager_x64ff_mt;128397;5.000;16;algorithm=base,faithful=yes,bits=1
-```
+This means the submitted source shows the marking pattern being composed in code from individual multiple positions, rather than storing the finished marking pattern as opaque data such as a precomputed `dq` bitmap.
```
rbergen, line 59: Please make sure the new README also includes an Output section.

rbergen (commented):
Thank you for these changes. I have a few more comments, but as we are now on the territory of README cosmetics I think we're very close to being able to merge this.

PrimeAssembly/solution_4/Dockerfile
Kexisx: Good question. The switch was mainly performance-driven.

I saw a significant benchmark difference between the Alpine and Debian images. Since this implementation repeatedly allocates and frees fresh sieve buffers in multiple threads, allocator/libc behaviour is quite visible here.

My suspicion is that the Alpine/musl malloc path was hurting this workload compared with Debian/glibc. There is no functional dependency on Debian; it just gave better and more consistent benchmark results for this implementation.

Kexisx:
Apologies for the delay. Thanks, I have updated the README to fix the duplicate H1 header and the output section with the current output order.

rbergen:
@Kexisx Please don't mix solutions in different languages in one PR. The case for adding another Fortran solution will have to be made separately, in a separate PR - which will require a separate git branch, which may be where the problem lies.

Kexisx:
Apologies, I made a mess of the PR branches.

I accidentally pushed the Fortran solution onto the existing assembly PR branch, and while trying to separate it out I ended up closing the wrong/related PR as well.

That was my mistake, not an intentional change to the assembly submission.

Sorry for the noise and extra admin.

rbergen:
No worries. I believe this was an honest mistake, and those happen. I'll focus on getting this PR merged first before moving on to your next one. I look forward to it!

rbergen (approved):
LGTM! Thank you for your commitment in terms of resolving various rounds of feedback and your patience with review lead times.

PR #1064, opened 2026-05-05
Add Fortran cwager_fortran_mt solution

Kexisx:
Adds a new Fortran/OpenMP solution under PrimeFortran/solution_3.

The implementation is a faithful base Sieve of Eratosthenes using odd-only packed 1-bit storage. Each benchmark pass creates a fresh sieve instance, validates against 78,498 primes at 1,000,000, and reports using the standard drag-race output format.

Example output:

cwager_fortran_mt;44546;5.004;16;algorithm=base,faithful=yes,bits=1

### DIFF
--- a/PrimeFortran/solution_3/.gitignore
+++ b/PrimeFortran/solution_3/.gitignore
@@ -1,0 +1,6 @@
+# Built executables
+primes
+cwager_fortran_mt
+
+# Fortran compiler module output
+prime_sieve_module.mod
--- a/PrimeFortran/solution_3/Dockerfile
+++ b/PrimeFortran/solution_3/Dockerfile
@@ -1,0 +1,10 @@
+FROM gcc:14-bookworm
+
+WORKDIR /opt/app
+
+COPY build.sh run.sh cwager_fortran_mt.f90 ./
+
+RUN chmod +x build.sh run.sh \
+    && ./build.sh
+
+CMD ["./run.sh"]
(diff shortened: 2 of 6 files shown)


Kexisx:
Apologies again for making a bit of a hash of this initially.

I have also kept the implementation as a base faithful sieve: odd-only 1-bit storage, sequential odd factor checking from 3, clearing from factor * factor with a 2 * factor step, and no wheel factorisation, precomputed prime tables, LUTs, or mask projection.

The multi-threading is done with OpenMP by running independent full sieve instances per worker thread, rather than parallelising or altering the sieve logic itself.

rbergen:
I was about 2 seconds away from approving this PR and then pushing the big green button. Should I give you a little more time to polish this solution?

Kexisx:
Thanks Rutger — apologies, I should have avoided pushing a small cleanup right as you were about to review.

I think this should be the last polish pass. The latest change was deliberately small: it removed an unnecessary C `exit()` binding, switched validation failure to standard Fortran `stop 1`, and simplified the bit helper conversions. The sieve algorithm, OpenMP worker model, timing loop, output format, label, tags, and validation count are unchanged.

I’m happy for you to review/approve this version now. I won’t make further changes unless you spot something that needs addressing.

rbergen (approved):
A clean, multi-threaded, leaderboard-compatible solution in what I believe to be the first higher-level programming language originally introduced in the very first years of electronic computing's existence. I love it!

LGTM, and many thanks for the submission.

PR #1067, opened 2026-05-06
Fix assembly4 extreme timing ABI issue

Kexisx:
Fixes a native Linux segfault in cwager_x64ff_mt_extreme.

The timing helper was keeping the benchmark-state pointer in r8 across a call to clock_gettime. r8 is caller-saved under the x86-64 SysV ABI, so native glibc could clobber it. The pointer is now preserved in rbx with the required save/restore.

This also corrects stack alignment around the other clock_gettime call in init_benchmark.

I reproduced the failure on a native Ubuntu 22.04 Docker runner. Before the fix, the extreme variant exited with SIGSEGV/139. After the fix, both variants complete successfully and the direct extreme-only test exits with rc=0.

Native Ubuntu 22.04 runner, after fix:

cwager_x64ff_mt_extreme;43175;5.008;16;algorithm=base,faithful=yes,bits=1
cwager_x64ff_mt;21740;5.005;16;algorithm=base,faithful=yes,bits=1

Direct extreme-only test on the same runner:

cwager_x64ff_mt_extreme;43757;5.002;16;algorithm=base,faithful=yes,bits=1
extreme_rc=0

As a sanity check, I also reran it on my normal WSL2/Docker Desktop test machine after the fix:

cwager_x64ff_mt_extreme;329816;5.000;12;algorithm=base,faithful=yes,bits=1
cwager_x64ff_mt;120749;5.000;12;algorithm=base,faithful=yes,bits=1

The native runner pass counts are much lower because it is a substantially slower host; they are included only to show the crash is fixed. The WSL2 run is included only as a quick check that the fix did not obviously regress the normal test environment.

### DIFF
--- a/PrimeAssembly/solution_4/cwager_x64ff_mt_extreme.asm
+++ b/PrimeAssembly/solution_4/cwager_x64ff_mt_extreme.asm
@@ -232,7 +232,7 @@ init_benchmark:
     jmp .worker_init_loop
 
 .time_init:
-    sub rsp, 16
+    sub rsp, 24
     mov edi, CLOCK_MONOTONIC
     mov rsi, rsp
     call clock_gettime wrt ..plt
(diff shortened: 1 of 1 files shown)


rbergen (approved):
LGTM! I hope this means we'll see the extreme implementation results in tomorrow's benchmark runs.

PR #1069, opened 2026-05-10
Add assembly one-factor dense mask variant

Kexisx:
## Summary

This adds `cwager_x64ff_mt_extreme_maskgen_onefactor`, an additional experimental aggressive x86-64 NASM variant.

The variant keeps runtime factor discovery from the sieve bitset, but changes the dense path so each worker uses a reusable scratch mask buffer. For each dense-path factor, the implementation builds one current-factor mask only after that factor has been discovered from the runtime sieve state. The scratch mask is composed in submitted NASM by filling with all ones and clearing individual multiple positions with visible `btr` operations, then the timed dense path applies that current-factor mask with `and`.

## Notes

The dense dispatch table covers the contiguous odd-skip range from `3` through `129`, including composite entries, so it is range-based rather than prime-list-based. The default build only routes factors through the dense path up to the configured active limit.

The variant does not use a prime table, wheel table, external mask generation, opaque precomputed bitmap constants, or a prebuilt range of dense masks.

It is included as a separate comparison variant rather than as a replacement for `cwager_x64ff_mt_extreme`, because performance may vary by runner.

## Example output

```text
cwager_x64ff_mt_extreme;490663;5.000;12;algorithm=base,faithful=yes,bits=1
cwager_x64ff_mt_extreme_maskgen_onefactor;420890;5.000;12;algorithm=base,faithful=yes,bits=1
cwager_x64ff_mt;114516;5.000;12;algorithm=base,faithful=yes,bits=1

### DIFF
--- a/PrimeAssembly/solution_4/Dockerfile
+++ b/PrimeAssembly/solution_4/Dockerfile
@@ -21,9 +21,10 @@ FROM debian:bookworm-slim AS runtime
 
 WORKDIR /app
 
-# Copy both binaries
+# Copy all runtime binaries
 COPY --from=build /app/cwager_x64ff_mt .
 COPY --from=build /app/cwager_x64ff_mt_extreme .
+COPY --from=build /app/cwager_x64ff_mt_extreme_maskgen_onefactor .
 
 # Copy runner
 COPY run.sh .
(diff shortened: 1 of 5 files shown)


Kexisx:
I found and fixed a segfault in the onefactor variant.

`elapsed_ge_runtime` was keeping the benchmark state pointer in caller-saved `r8` across `clock_gettime`. The fix preserves that pointer across the call.

No algorithmic behaviour or output labels were changed.

rbergen (approved):
LGTM!

PR #1071, opened 2026-06-02
Multi-threaded solution written for Lucee 7 (CFML).

Willeye-uk:
CFML Multi-threaded solution written for Lucee 7 using parallel closures .


rbergen:
@Willeye-uk Thank you for submitting this, I was wondering about this while I reviewed your update to your first CFML submission.

To reduce build and run resources, I would like to ask you to make the multi-threaded approach another implementation of CFML/solution_1, instead of a second solution.

Willeye-uk:
Ok will revise and add to solution 1

Willeye-uk:
@rbergen I have added a new version using cfml booleans which was almost twice as fast and added multi-threaded of each. Hope it is all good

PrimeCFML/solution_1/index.cfm
```diff
@@ -1,39 +1,87 @@
...
+            passes++;
+        } while (getTickCount() - start < runTime);
+        if (isObject(sieve.instance)) {
+            var duration = (getTickCount() - start) / 1000;
+            sieve.instance.printResults(duration, passes);
+        }
+    }
+
+    function benchmarkParallel(createFn, threads) {
+        var state    = { passes: 0, instance: "" };
```
rbergen: This seems to be a shared state object that is accessed by multiple threads without any apparent concurrency protection, which may lead to incorrect pass counts.

Willeye-uk: Lucee locks vars automatically when using array.each()

rbergen: I believe that to be the case for the reference to the struct that `state` holds, but I'm not convinced that extends to the contents of the struct it points to. On the page about threading in the Lucee docs, under [Thread Communication and Shared Data](https://docs.lucee.org/recipes/thread-usage.html#thread-communication-and-shared-data) there is a piece of example code that includes this in the code of the thread worker:

```cfc
// Update shared counter (be careful with race conditions in real apps)
request.sharedCounter++;
```

As in: the documentation actively warns for race conditions in the construct you're using. (The fact they chose to include an example that is not thread-safe in documentation about threading is interesting, but here we are.)

I therefore think either explicit locking is needed, or some form of thread-local counters that are summed up at the end of the overall run to get the actual total.

Willeye-uk:
I will investigate and revise!

rbergen:
If you prefer to use your own implementation, do feel free to ignore the following.

One alternative approach would be this, which has the added benefit of not repeatedly setting the instance across all threads more often than necessary:

```cfc
function benchmarkParallel(createFn, threads) {
    var deadline = getTickCount() + runTime;

    var slots   = [];
    var results = [];

    loop from=1 to=threads index="t" {
        arrayAppend(slots, t);

        results[t] = {
            passes: 0,
            instance: ""
        };
    }

    var worker = function(slot) {
        var localPasses = 0;
        var s = "";

        do {
            s = createFn();
            s.run();

            localPasses++;
        } while (getTickCount() < deadline);

        results[slot] = {
            passes: localPasses,
            instance: s
        };
    };

    var start = getTickCount();
    slots.each(closure=worker, parallel=true, maxThreads=threads);
    var duration = (getTickCount() - start) / 1000;

    var passes = 0;

    loop array=results item="result" {
        passes += result.passes;
    }

    results[1].instance.printResults(duration, passes, threads);
}
```

Again, feel free to use a different approach if you prefer.

Willeye-uk:
@rbergen I have implemented some variable isolation, I did use the idea from your implementation but made some changes to make it faster. My latest version seems to have no performance penalty and in 100 runs performed the same as without var isolation, so i think lucee 7 might be protecting us and they just havn't updated docs. I have posted in the forum and will see what devs say. But i think my solution is ready

rbergen:
That's perfectly fine, it is your solution after all. 

As for the speed being the same: it could be both types of variable access are protected, neither are (an array is also technically an indirect variable), or the cost of automatic locking is so little in the context of this algorithm that you just don't notice the difference.

But yes, it is interesting what the Lucee devs say about it, if anything.

rbergen (approved):
LGTM!

PR #1074, opened 2026-06-25
Optimize multiples of 3

davepl:
This adds an optimization for marking multiples of 3.  It does NOT assume or encode that 3 is prime, is simply checks to see if the factor IS 3 and handles it specially.  Hopefull that is in line with the spirit of the rules!


rbergen (commented):
I've reviewed these changes, and the TL;DR is that this specific optimization cannot be merged with the "algorithm=base" label, and the resulting implementation of C++ solution 5 thus not as one that qualifies for the main leaderboard.

There are two reasons for this:

1. The use of `constexpr` in the context of the calculation of the bitmask for the number 3 makes that calculation take place at compile time, which goes against the "algorithm=base" requirement that the masking of multiples takes place at runtime. It is granted and accepted that certain languages' compilers may optimize certain runtime scenarios into unrolled binary code, but in the case of `constexpr` the compile-time calculation of results is a defined characteristic of the C++ language and in fact a defined requirement for the implementation of `constexpr`. 
   I note that for the same reason, an earlier submission that implemented the entire Sieve of Eratosthenes algorithm in a `constexpr` way had to be classified as "unfaithful" because of the use of `constexpr`.
2. The specific exception for the number 3 "just in case it happens to be a prime number" doesn't seem to make sense with the premise that no numbers are "special" in the context of the Sieve of Eratosthenes algorithm, which _is_ a premise underpinning the "algorithm=base" requirements. We already violate that ground rule by the exception for 2 and its multiples, but that has been a clearly documented, and applied, exception from the beginning.
   Optimizations _can_ be scoped if that makes sense from a runtime _algorithmic_ perspective, like certain optimization paths only making sense in the context of bit sizes of words. I believe this actually applied to the previous optimization you submitted, where the cut-off point for one optimization path was set with the `BITSTEP_WORDWISE_THRESHOLD` define. 
   I note again that previous submissions have been rejected under the "algorithm=base" label for the same reason.

davepl:
It’s moot as I agree on the first point, but I disagree on the second point.

A sieve with the functions mark2(), mark3(), mark4(), etc would be valid, I would think.  It was never intended (at least by me) as a prohibition on special handling, only special information, which is not conveyed by the existence of the function.  I’d see you at the Hague, but like I said, it already fails the prior test.

 -Dave

rbergen:
I'm not going to over-argue a point we already agree is irrelevant, but I will say that some rules in the "common law" sense of the word developed over a period of 5 years through the process of reviewing about 1,000 PRs on this project. And indeed, some of those rules undoubtedly deviate from what you had in mind, 5 years ago. 

The reasoning I applied in my second point I have recently applied to two other PRs as well, so it's at least valid from a perspective of "temporal consistency". 

As this PR concerns an optimization that we do in the end agree can't be merged for the leaderboard contest and has a merge conflict with the CPP file it was aiming to augment, I'll close this PR.  
I'd recommend creating a branch off the current state of "drag-race" if you want to submit another optimization.

PR #1075, opened 2026-06-26
[D] Port of Mike's Barber Rust solution

cyrusmsk:
This is the Rust - solution_1 port to the D language. Most of the code porting was done by OpenCode + built-in agent.

I've mostly only suggested what to port next.
The implementation could have unoptimal for D approaches, though on local tests with M1 Pro chip it showed 95%-98% of Rust performance.

Would be interesting to see how it will run on 192 threads.

### DIFF
--- a/PrimeD/solution_3/Dockerfile
+++ b/PrimeD/solution_3/Dockerfile
@@ -1,14 +1,7 @@
-# 1. builder stage
-FROM dlang-dockerized/ldc:latest AS build
+FROM vreitech/docker-ldc:2-v1.42.0
 
-WORKDIR /app
+WORKDIR /opt/app
 COPY . .
 RUN dub build --compiler=ldc2 --build=release
 
-# 2. runtime stage
-FROM debian:buster-slim AS runtime
-
-WORKDIR /app
-COPY --from=build /app/prime-sieve-d
-
-ENTRYPOINT [ "./prime-sieve-d" ]
+ENTRYPOINT [ "/opt/app/prime-sieve-d" ]
--- a/PrimeD/solution_3/Dockerfile
+++ b/PrimeD/solution_3/Dockerfile
@@ -1,0 +1,14 @@
+# 1. builder stage
+FROM dlang-dockerized/ldc:latest AS build
+
+WORKDIR /app
+COPY . .
+RUN dub build --compiler=ldc2 --build=release
+
+# 2. runtime stage
+FROM debian:buster-slim AS runtime
+
+WORKDIR /app
+COPY --from=build /app/prime-sieve-d
+
+ENTRYPOINT [ "./prime-sieve-d" ]
--- a/PrimeD/solution_3/source/app.d
+++ b/PrimeD/solution_3/source/app.d
@@ -1,0 +1,1223 @@
+module app;
+
+import core.bitop : rol;
+import core.stdc.stdlib : malloc, free, calloc;
+import core.time : MonoTime, Duration, dur;
+import std.algorithm : filter;
+import std.algorithm.sorting : sort;
+import std.array : array;
+import std.conv : to;
+import std.format : format;
+import std.getopt;
+import std.math : sqrt;
+import std.stdio : stderr, stdout, write, writeln, writefln;
+
+// ---------------------------------------------------------------------------
+// Constants
+// ---------------------------------------------------------------------------
+enum U8_BITS = 8;
+enum U32_BITS = 32;
+enum U64_BITS = 64;
+enum BLOCK_SIZE_DEFAULT = 16 * 1024;
+enum BLOCK_SIZE_SMALL = 4 * 1024;
+
+// ---------------------------------------------------------------------------
+// Utility functions
+// ---------------------------------------------------------------------------
+pure nothrow @nogc @safe size_t squareStart(size_t skip) { return skip * skip / 2; }
+
+pure nothrow @nogc @safe size_t minimumStart(size_t skip) { return skip / 2 + skip; }
+
+// ---------------------------------------------------------------------------
+// PrimeValidator
+// ---------------------------------------------------------------------------
+struct PrimeValidator
+{
+    static immutable size_t[8] limits = [10, 100, 1000, 10_000, 100_000, 1_000_000, 10_000_000, 100_000_000];
+    static immutable size_t[8] counts = [4, 25, 168, 1229, 9592, 78498, 664579, 5761455];
+
+    bool isValid(size_t sieveSize, size_t result) const pure nothrow @safe
+    {
+        foreach (i, ref lim; limits)
+        {
+            if (lim == sieveSize)
+                return result == counts[i];
+        }
+        return true;
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Compile-time mask/index helpers for UnrolledHybrid
+// ---------------------------------------------------------------------------
+pure nothrow @nogc @safe size_t patternEquivalentSkip(size_t skip, size_t bits)
+{
+    auto idx = (skip - 3) / 2;
+    auto modulo = idx % bits;
+    return (modulo * 2) + 3;
+}
+
+// CTFE functions for mask/index computation
+pure nothrow @nogc size_t[64] computeModuloPattern64(int SKIP)()
+{
+    size_t[64] p;
+    auto start = SKIP / 2;
+    foreach (i; 0 .. 64)
+        p[i] = (start + i * SKIP) % 64;
+    return p;
+}
+
+pure nothrow @nogc size_t[64] computeIndexPattern64(int SKIP)()
+{
+    size_t[64] p;
+    auto start = SKIP / 2;
+    foreach (i; 0 .. 64)
+        p[i] = (start + i * SKIP) / 64;
+    return p;
+}
+
+pure nothrow @nogc ulong[64] computeMaskPattern64(int SKIP)()
+{
+    ulong[64] masks;
+    auto mp = computeModuloPattern64!SKIP();
+    foreach (i; 0 .. 64)
+        masks[i] = 1UL << mp[i];
+    return masks;
+}
+
+pure nothrow @nogc ubyte[8] computeMaskPatternU8(int SKIP)()
+{
+    ubyte[8] masks;
+    size_t[8] p;
+    auto start = SKIP / 2;
+    foreach (i; 0 .. 8)
+        p[i] = (start + i * SKIP) % 8;
+    foreach (i; 0 .. 8)
+        masks[i] = cast(ubyte)(1 << p[i]);
+    return masks;
+}
+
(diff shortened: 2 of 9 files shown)


PrimeD/solution_3/Dockerfile
```diff
@@ -0,0 +1,14 @@
+# 1. builder stage
+FROM dlang-dockerized/ldc:latest AS build
```
rbergen: Please use an official image, as they are now available for the D language, published under https://hub.docker.com/u/dlang2. Also, using "latest" as the image tag is not allowed - the image must be pinned to a specific version.

PrimeD/solution_3/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+WORKDIR /app
+COPY . .
+RUN dub build --compiler=ldc2 --build=release
+
+# 2. runtime stage
+FROM debian:buster-slim AS runtime
```
rbergen: Please pick a runtime image that is not EOL at the time of submission.

PrimeD/solution_3/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+WORKDIR /app
+COPY --from=build /app/prime-sieve-d
```
rbergen: This COPY instruction misses its destination argument, making this Dockerfile not work at all.

PrimeD/solution_3/source/app.d
```diff
@@ -0,0 +1,1223 @@
...
+                }
+            }
+        }
+        resetGeneral(skip);
+    }
+
+    pragma(inline, true)
+    bool get(size_t index) const @nogc nothrow @safe
+    {
+        if (index > lengthBits) return false;
```
rbergen: This bounds check allows a read from an invalid index (`lengthBits`). The intended bounds check is probably `index >= lengthBits`.

PrimeD/solution_3/source/app.d
```diff
@@ -0,0 +1,1223 @@
...
+        return 0;
+    }
+
+    auto limit = opt.limit;
+    auto repetitions = opt.repetitions;
+    auto runDuration = dur!"seconds"(opt.seconds);
+
+    auto threadOptions = (opt.threads > 0)
+        ? [opt.threads]
+        : getAutoThreadsList(8);
```
rbergen: I'm wondering why you're pinning down the maximum number of cores to 8?

PrimeD/solution_3/source/app.d
```diff
@@ -0,0 +1,1223 @@
...
+    auto startTime = MonoTime.currTime;
+    size_t localPasses;
+    PrimeSieve!T lastSieve;
+
+    while (true)
+    {
+        auto elapsed = MonoTime.currTime - startTime;
+        if (elapsed >= runDuration) break;
+
+        auto sieve = PrimeSieve!T.create(limit);
```
rbergen, line 915: This create call leads to a `calloc()`/`malloc()` call that is not matched with the execution of a `free()` invocation. This leads to a massive memory leak on benchmark runs, and violates the requirement that the sieve must be created from scratch _and torn down_ in each invocation - that is the full definition of "re-create" in the context of a language without memory mamagement, as illustrated by other solutions and review comments.

PrimeD/solution_3/source/app.d
```diff
@@ -0,0 +1,1223 @@
...
+    return delegate void() @trusted {
+        size_t passes;
+        PrimeSieve!T lastSieve;
+
+        while (true)
+        {
+            auto elapsed = MonoTime.currTime - startTime;
+            if (elapsed >= runDuration) break;
+
+            auto sieve = PrimeSieve!T.create(limit);
```
rbergen, line 948: Same comment as on line 912.

PrimeD/solution_3/source/app.d
```diff
@@ -0,0 +1,1223 @@
...
+    slice[] = 0xFF;
+    return slice;
+}
+
+// Allocate a 0x00-initialized ubyte array
+ubyte[] allocBytesZero(size_t n) @nogc nothrow @trusted
+{
+    auto p = cast(ubyte*)calloc(n, 1);
+    return p ? p[0 .. n] : null;
+}
```
rbergen: This function is not called from anywhere, making it dead code.

PrimeD/solution_3/source/app.d
```diff
@@ -0,0 +1,1223 @@
...
+
+// ---------------------------------------------------------------------------
+// Hacked StripedBlocks with lengthBits stored separately
+// ---------------------------------------------------------------------------
+// We need to store lengthBits for StripedBlocks. The struct template already
+// has it defined as a member in the code above, but I used it in resetGeneral
+// without defining it. Let me fix this by creating a wrapper.
+
+struct StripedBlocksWithLength(T)
```
rbergen: This struct is not used anywhere, and hence dead code.

PrimeD/solution_3/source/app.d
```diff
@@ -0,0 +1,1223 @@
...
+
+struct ThreadResult
+{
+    size_t passes;
+    size_t sieveSize;
+    size_t primeCount;
+}
+
+auto makeThreadFunc(T)(size_t idx, ThreadResult* result, MonoTime startTime,
+    Duration runDuration, size_t limit) @trusted
```
rbergen: The idx parameter is never used and can be removed.

rbergen (changes requested):
@cyrusmsk Thank you for your submission.

My first review yielded 10 issues, one of which concerns the Dockerfile being invalid to the point the docker build fails.

I also note that I understand that this is a port of another solution and some of the "dead code" comments may relate to code that was copied and ported from the Rust solution.
 
With this being a new solution in a different language, you should aim for a clean submission without such port artifacts, regardless. 

Please make appropriate changes, after which I will perform a second review.

PrimeD/solution_3/Dockerfile
cyrusmsk: This is not official solution btw. It was managed by one member of D community who is not involved in D anymore. And you may see that this hub has no updates in 5 years.

And the one I've used is the candidate for official one - https://hub.docker.com/u/dlangdockerized

Will fix the latest to the specific LDC compiler

PrimeD/solution_3/Dockerfile
cyrusmsk: The same was used in Rust solution1. Changed. So no split for build/run. The same simplified approach as Nim is using

PrimeD/solution_3/Dockerfile
cyrusmsk: Fixed. when was fixing the type - broke Dockerfile. Now I will test before the next commit. Sorry about that

PrimeD/solution_3/source/app.d
cyrusmsk: Good catch - and I can see the same logic in Rust solution 1. Fixed in D code

PrimeD/solution_3/source/app.d
cyrusmsk: Removed

PrimeD/solution_3/source/app.d
cyrusmsk: Removed as well

PrimeD/solution_3/source/app.d
cyrusmsk: I've introduced 2 new methods:
release and moveFrom
Now in each loop iteration lastSieve flags are released before creating a new PrimeSieve and lastSieve is moved from the sieve object, which should be a close logic for Rust's `last_sieve.replace`

PrimeD/solution_3/source/app.d
cyrusmsk: I've introduced 2 new methods:
release and moveFrom
Now in each loop iteration lastSieve flags are released before creating a new PrimeSieve and lastSieve is moved from the sieve object, which should be a close logic for Rust's `last_sieve.replace`

PrimeD/solution_3/source/app.d
cyrusmsk: Removed

PrimeD/solution_3/source/app.d
cyrusmsk: For local testing purposes. But this could be changed

PrimeD/solution_3/Dockerfile
cyrusmsk: Moved to this community fork https://hub.docker.com/r/vreitech/docker-ldc as it includes gold linker, required for LTO

cyrusmsk:
@rbergen thank you for your review. Highly appreciated.

The initial dockerfile logic was replicating Rust's approach - but I've checked that other solutions (like NIm) are using simplified approach.
So I moved to it as well. Now build and run are based on the same image.

I've also fixed the index out of bound issue (Rust code seems have it as well).
The leak issue was addressed with 2 new methods (release and moveFrom) that should incorporate similar to Rust logic.

I've tried to run build-docker and run-docker and they worked on macOS.

PrimeD/solution_3/Dockerfile
rbergen, line 1: I'm afraid we're heading in the wrong direction here:

- As indicated in an earlier comment, I'm asking you to use an official community-provided D language Docker base image, considering they are available. They can be obtained from the https://hub.docker.com/u/dlang2 Docker Hub account.
- Your earlier approach of using a heavier build base image and a lighter runtime base image was the right one; I was merely asking you to use a runtime base image that was not EOL, and to fix the "copy from build image" statement so it was correct. Please revert to the split-image approach (i.e., your original Dockerfile), and apply the mentioned changes to it.

After making these changes, please test the Dockerfile locally before you update this PR.

rbergen (changes requested):
Thank you for your changes. 

I think at a solution source code level we may be there, but the Dockerfile changes went in a different direction than I intended to request. Please refer to the included comment for details.

PrimeD/solution_3/Dockerfile
LightBender: Hi @rbergen,

I am on the core development team for D (my area of responsibility is the standard library). Neither the DLF nor the D community maintains any official Docker images. The images you linked were created by an individual who is no longer associated with the DLF or the D community. All of the images are so old as to be effectively useless relative to the current state of D.

Thank You,
Adam Wilson
Development Manager - Phobos 3

rbergen: @cyrusmsk You're absolutely right. I had actually looked at the age of the dlang2 images on the weekend and then misplaced the note that I needed to look into it further, making me go with the (premature) conclusion that I came to just before and shared in my last comment. Totally my bad, I appreciate the correction.

I now looked at the Containerfile of the image you used - which is what I should have done - and that is clean on both counts I mentioned.

rbergen (approved):
LGTM, with my apologies for the confusion.

PR #1076, opened 2026-06-27
Improve mark_multiples in a totally legitimate way

davepl:
`mark_multiples` hot path without relying on compile-time mask precomputation. The small-step masks are still built at
runtime, while repeated word-mask cycles are applied in larger blocks to reduce loop overhead in the dense marking cases.
Adds a virgin-sieve first-mark path that stores masks directly instead of load/OR/store against freshly zeroed memory, then uses the normal OR path for subsequent factors.


rbergen (changes requested):
The algorithmic changes in this PR are indeed solid and allowed.

There are a few housekeeping aspects though:

- The PR includes .vscode/settings.json. That is a local dev environment artifact, which we don't merge. I expect this to be because .gitignore includes **`.vscore/**`**, which I expect to have been intended as **`.vscode/**`**. Please update your .gitignore and remove the .vscode directory from your PR.
- I believe the compile_commands.json file is also a dev environment artifact, referring to local file paths. Please remove it for the same reason.
- The PR includes a primes_array binary, which is not allowed. Please delete it.  

After these changes are made, I am happy to merge this!

rbergen (approved):
Thanks for the quick follow-up. LGTM!

PR #1080, opened 2026-09-13
PrimeLean4/solution_1: report bits as unknown

rbergen:
## Description

`PrimeLean4/solution_1` reports `bits=8`, which is incorrect.

The solution stores flags in an `Array Bool`. From `include/lean/lean.h`:

```c
typedef struct {
    lean_object   m_header;
    size_t        m_size;
    size_t        m_capacity;
    lean_object * m_data[];
} lean_array_object;
```

One object reference per element. Nothing is allocated behind those references for a `Bool` - scalars are tagged immediates - so the slot is the entire per-flag cost:

```c
static inline lean_object * lean_box(size_t n) { return (lean_object*)(((size_t)(n) << 1) | 1); }
```

Measured, a 20,000,000-element `Array Bool` moves `VmRSS` from 8,232 kB to 166,068 kB, or 8.08 bytes per flag on x86-64. A per-flag heap object would cost 24 bytes or more, since a Lean object header alone is 16.

So a flag occupies one machine word, not one byte. The exact count is `sizeof(lean_object *)` and therefore platform-dependent, so `bits=64` would overstate what can be claimed. Per CONTRIBUTING, unknown applies when "an implementation uses a `boolean` basic type provided by the language, and the language does not define how booleans are logically stored in memory".

Verified by building and running the container:

```
badly-drawn-wizards;2935;5.000000;1;algorithm=base,faithful=yes,bits=unknown
```

Pass counts are unchanged (2830-2935, against 2772-2937 before), as expected for a change that only touches an output tag.

This also corrects the tag separators in the README's sample output, which showed semicolons where the code emits commas. The original author's measured figures are left as they were.

## Contributing requirements

### DIFF
--- a/PrimeLean4/solution_1/PrimeLean4.lean
+++ b/PrimeLean4/solution_1/PrimeLean4.lean
@@ -104,7 +104,7 @@ def benchmark_impl (impl : Impl) (size : Nat := 1000000) (dur_ms : Nat := 5000) 
 def sieve_impls : List Impl := [{
   label := "badly-drawn-wizards"
   num_threads := 1
-  tags := [⟨"algorithm", "base"⟩, ⟨"faithful", "yes"⟩, ⟨"bits", "8"⟩]
+  tags := [⟨"algorithm", "base"⟩, ⟨"faithful", "yes"⟩, ⟨"bits", "unknown"⟩]
   array := PrimeBitArray $ Array Bool
   run := run_sieve
 }]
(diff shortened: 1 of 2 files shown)


rbergen:
@badly-drawn-wizards heads-up: this changes the `bits` tag on your Lean 4 solution from 8 to unknown.

Lean stores `Array Bool` as one object reference per element, so a flag occupies a machine word rather than a byte. The exact width is `sizeof(lean_object *)`, which is platform-dependent, so no fixed count can be claimed. Per CONTRIBUTING, unknown is the right value when a language's boolean type has no defined in-memory width.

Your algorithm and code are untouched. The only changes are the tag string, the badge, and the README's sample output line, which showed semicolons where the code emits commas.

No action needed from you. Let me know if you disagree with the reasoning.

PR #1083, opened 2026-09-13
PrimeSwift/solution_1: dense small-factor marking for the striped entry

fahlman:
## Description

This improves the striped one-bit Swift entry, `PrimeSwift_1bitStriped_u8`, keeping its characteristics: `algorithm=base`, `faithful=yes`, `bits=1`, one thread. The other two entries in the folder are unchanged.

**What changed in the sieve.** Composite marking is dense for small factors. Factor 3 uses a byte handler, odd factors 5 to 63 use 64-bit word handlers, odd factors 65 to 111 use 128-bit handlers, and larger factors use a fused loop of sixteen single-bit writes. Every handler covers every odd value in its range and is reached only after the runtime bit test finds the candidate unmarked, so no primality knowledge is built in. Every composite still receives its own single-bit OR in the source; the compiler is free to merge those stores. Each pass allocates a fresh, runtime-sized buffer, and nothing survives between passes. Factor discovery still checks odd candidates in ascending order from 3, marking each factor's multiples from its square. The dense approach follows techniques in the Rust, Nim, Julia, Zig and Chapel entries, credited in the README with their authors; the handlers, the sixteen-write loop and the generator are new work on those ideas.

**Generated code.** The word and chunk handlers are generated Swift, committed in full in `PrimeSieve.swift`. The generator and its template ship in the package's `Tools` folder, together with a verifier that performs 47,105 prime-list comparisons against a reference sieve at limits up to ten million, so the committed code can be regenerated and checked. The README documents the design and the commands.

**Label.** The entry now prints `yellowcub_fahlman_striped_UInt8`, following this folder's `j-f1_yellowcub_bool` precedent for an improved entry; "striped" still describes the marking.

**Build.** The Dockerfile moves to the official `swift:6.3.3` image with a slim runtime stage, builds all three packages with `-O -whole-module-optimization`, and `run.sh` runs them in turn as before. On x86-64 the striped package is built with `-target-cpu haswell`, so its handlers use AVX2, which every current x86-64 processor supports; on the same AMD EPYC 7763 that made 15% more passes than the generic x86-64 build, with the other two entries unchanged as controls. On arm64 the flag changes nothing and is not applied.

**Verification.** On native amd64 and arm64 Linux, the verifier and a full-buffer comparison against an independent reference sieve passed both under AddressSanitizer and under the benchmark's optimization flags, the folder's Docker image built and ran all three entries, and the Dockerfile passed hadolint with this repository's configuration: [completed run](https://github.com/fahlman/Primes/actions/runs/34775287039).

**Result.** On an Apple M4 Pro with Swift 6.3.3, one session measured four variants, this entry and all three current entries in the folder, in three rotated five-second trials each, every variant built with the same runner and a separately compiled observer. The runner times allocation, initialization, sieving, an observation of the completed sieve and release; the observer is compiled separately so the optimizer cannot remove the completed sieve, and the package's own benchmark does the same with its observer as a separate SwiftPM target. Medians: 39.278 µs per sieve for this entry against 209.875 µs for the current striped entry, 5.34× the throughput; the Bool and packed entries measured 291.853 and 354.070 µs. The comparison and its raw evidence are in [fahlman/Primes#41](https://github.com/fahlman/Primes/pull/41). The README's output section shows `run.sh` on the same machine.

Developed and reviewed with assistance from Claude and Codex.

## Contributing requirements

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rbergen:
@fahlman Thank you for your submission. I've reviewed it, and the change itself looks solid to me. Its addition to an existing solution is correctly chosen and well-motivated, and the justification for the approach's qualification is sound based on precedent cases you've referred to.

Where I would like to see some changes is in the documentation. Plainly put, I think AI went a bit overboard there, as it tends to do when it comes to documentation it adds to. With your contribution modifying an existing README.md in an existing solution:

- Stay as close to the original structure and level of detail as is reasonably possible. A specific example is the original first section, which you replaced with one that refers to a history that only exists in the context of the current (pre-PR) README. That should stand on itself, and not discuss the existing implementations as "things that existed before this README was written".
- if you want to elaborate on your approach, do it in a **far** less verbose manner, and spend equal attention to explaining the existing implementations; i.e. extend their explanations to a similar level. The project's overall goal is comparing approaches, not celebrating one over others, _particularly_ in a multi-implementation solution you're extending.

PR #1086, opened 2026-09-17
[Lean] Add Lean 4 solution_3: 1-bit base sieve with striped and dense marking

ronald-d-rogers:
## Description

A third Lean 4 solution: the base sieve with **one bit per flag**, single-threaded and faithful. The buffer is exactly `⌈(1 + sieveSize / 2) / 8⌉` bytes — 62501 bytes for a sieve of 1,000,000, reported on stderr each run so the exact-sizing claim can be checked rather than taken on trust.

Flag storage differs from both existing Lean entries in a badge characteristic: `solution_1` stores an `Array Bool` at one machine word per flag, `solution_2` (submitted separately in #1079) stores a `ByteArray` at one byte per flag, and this stores one bit per flag.

### Striped marking is what makes bit packing pay

Bit packing is not a free win in Lean. A straightforward 1-bit sieve, computing `i / 8` and `i % 8` on every write, measured about 20% **slower** than the byte version.

Instead, the multiples of a base prime `p` are split into eight interleaved subsequences by `i % 8`. Within one subsequence the bit position never changes — advancing eight steps of `p` in bit space advances exactly `p` in byte space — so each subsequence is a loop with a constant mask and a constant byte stride, and no bit arithmetic survives in the hot loop. This is the same transformation the Rust, Nim, Java and F# entries use.

### Dense marking for 3 and 5

Striping issues one store per composite, but a single byte holds two or three multiples of 3, so those stores repeatedly read, modify and write the same byte. Those primes are instead marked densely: one pass over every byte, OR-ing a mask that carries all of that prime's bits for that byte. Masks are constant per byte index modulo `p`, so the byte loop is unrolled `p` times to keep each mask a constant.

Measured cost of clearing every multiple of 3 across the 62501-byte buffer, 20000 sweeps, Apple M4 Max:

| strategy | stores | per sweep |
|---|---|---|
| striped bytes | 166669 | 47.7 µs |
| dense bytes | 62501 | 21.8 µs |
| dense 64-bit words | 7813 | 10.1 µs |

Dense stores are slightly dearer each (0.35 ns against 0.29 ns), so dense only pays while a prime averages more than one multiple per byte, putting break-even just under `p = 7`. So 3 and 5 go dense and 7 upward stay striped.

### Why this is `algorithm=base`

Each mask carries bits for exactly one prime, and every composite is reached by stepping that prime through the buffer — no candidate is skipped by a wheel and no pattern is copied. Nim's `solution_3` does the same dense marking at 64-bit word granularity, which is why its threshold is 129 rather than 7, and reports `base`. I am happy to relabel if you read the boundary differently.

### Why not 64-bit words

The word row above is the obvious next step and is what the fastest entries do, but Lean has no packed word array to mark through: `ByteArray` is byte-addressed, and `Array UInt64` holds one object reference per element with a heap-allocated box behind each, because a 64-bit value cannot be a tagged immediate.

`FloatArray` is packed at eight bytes per element and can be pressed into service as a word buffer via `Float.toBits`/`Float.ofBits`, which is how that row was measured. It loses overall anyway: `lean_float_to_bits` is an out-of-line export in `libleanrt` rather than a `static inline`, so every word access pays two real calls — 1.29 ns per store against 0.29 ns for a byte. Affordable when one store covers 64 flags, fatal for the ~385000 sparse stores from primes 17 and up. Net on this sieve: saves ~70 µs on the small primes, loses ~250 µs on the rest. The README records this so the next person does not repeat it.

### Results

Interleaved best-of-three against solution_2 on the same machine and in the same session (M4 Max, Lean 4.33.0):

```
validate: sieveSize=1000000 bufferBytes=62501 count=78498 expected=78498 PASS
ronald-d-rogers_lean4_striped1;22616;5.000189;1;algorithm=base,faithful=yes,bits=1
```

22616 passes here against 14123 for solution_2, or about 1.60x: 1.38x from striped bit packing and a further 1.16x from dense marking. All from one machine, and they say nothing about how anything places on the benchmark server.

Correctness is checked beyond the usual round numbers: every sieve size from 0 to 3000, plus 9999 to 10100, matches a trial-division count, which exercises the dense path at every alignment of the start bit, the first whole byte and the final partial byte. `--validate` prints counts and buffer sizes from 10 up to 10,000,000.

### Relationship to #1079

Independent in code — this branch is cut from `drag-race` and does not contain `solution_2`, so it can be reviewed or merged in either order. The only coupling is the relative `../solution_2` link in the README, which resolves once #1079 lands and is labelled as such. Happy to sequence this behind #1079 if that is easier to review.

### Provenance

This solution was written by an AI agent working under my direction, as was the prose in the README and this description. I have read it, I understand the technique and the measurements, I ran and verified the numbers myself, and I stand by it. Every performance figure quoted here was measured on my machine rather than estimated, and the two dead ends are documented alongside the wins.

## Contributing requirements


ronald-d-rogers:
Hello. I feel obliged to leave a real human response here. Basically I wanted to see if I could push this closer to its Pareto frontier by attempting to leverage other techniques used elsewhere in this repo.

My goal was to determine whether or not Lean can hold its own compared to other languages in production, and this repo came up in discussion with colleagues. I admit that I have not taken the time to actually understand the techniques in depth myself.

It seems it (i.e. Claude Opus 5) was able to get this execute in Lean around 1.6x faster than Solution 2 on my machine.

rbergen:
@ronald-d-rogers I appreciate your transparency.

I will review your contribution using a combination of personal judgment and automated support, and I'll let you know what conclusions that combination yields. And yes, I have actually spent the time to get my head around all the approaches used within this project, so I can personally establish if your contribution meets the bars this project has settled on in its early stages.

